PR #46, opened 2015-05-05
Added manifest-src directive to request rel. table

marcoscaceres:
Defined here: https://w3c.github.io/webappsec/specs/content-security-policy/#manifest_src

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -589,7 +589,7 @@ objects.
   <tr>
    <td>"<code title>manifest</code>"
    <td>"<code title>none</code>"
-   <td>&mdash;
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Wait, it wasn't merged. It's just a PR.

annevk:
I see, but we ended up with a tweet because it was a commit on whatwg/fetch. I guess that should only follow master. Sorry.

domenic:
Yeah, I had to bug GitHub to add an option to the tweeter thingy so that it only follows master. Probably worth enabling on all whatwg repos now that it exists.

annevk:
I've added "master" to Filter in most places now.

PR #122, opened 2015-09-09
Fix "nosniff?" link in order to avoid anolis bug.

yutakahirano:
Anolis cannot treat definitions including "?" correctly in some environments. This change
replaces "...nosniff?" definition with "...nosniff" in order to avoid the bug. 

See https://github.com/whatwg/xref/pull/7 for details.


annevk:
I ended up changing your PR slightly to also add an explicit ID. Otherwise the Anolis bug would surface again due to the special handling around `<dfn>` inside `<hn>` elements and we would have broken incoming URLs.

Issue #118, opened 2015-09-07
Add Response.body


yutakahirano:
Merge https://github.com/yutakahirano/fetch-with-streams/ to the spec.

See also: https://github.com/yutakahirano/fetch-with-streams/issues/25

annevk:
Thank you @yutakahirano. Real glad this is progressing.

PR #128, opened 2015-09-18
Introduce Response.body. (#118)

yutakahirano:
This change introduces `body` property of Response class. Note that we don't
introduce `Request.body` at this moment and hence some properties and algorithms
defined in the Body mixin are moved to Request and Response.

`bodyUsed` property is now based on `IsDisturbed` defined in the streams spec.
Request.bodyUsed emulates the behavior without using streams.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-5-october-2015">Living Standard — Last Updated 5 October 2015</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-6-october-2015">Living Standard — Last Updated 6 October 2015</h2>
 
 <dl>
  <dt>Participate:
@@ -1167,30 +1167,29 @@ clearly stipulates that <a href="#concept-connection" title="concept-connection"
 
 <h4 id="readablestream"><span class="secno">3.6.1 </span>ReadableStream</h4>
 
-<p>A <dfn id="concept-readablestream" title="concept-ReadableStream">ReadableStream</dfn> represents a
+<p>A <dfn id="concept-readablestream" title="concept-ReadableStream">ReadableStream</dfn> object represents a
 <a class="external" data-anolis-spec="streams" href="https://streams.spec.whatwg.org/#rs-class" title="readablestream">stream of data</a>. In this section, we
 define common operations for ReadableStream. <a href="#refsSTREAMS">[STREAMS]</a>
 
-<p>To <dfn id="concept-enqueue-readablestream" title="concept-enqueue-ReadableStream">enqueue</dfn> a byte sequence
-(<var title="">chunk</var>) into a <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a>
-<var title="">stream</var>, run the these steps:
+<p>To <dfn id="concept-enqueue-readablestream" title="concept-enqueue-ReadableStream">enqueue</dfn> <var>chunk</var> into a
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var>, run these steps:
 
 <ol>
- <li><p>Call <a href="https://streams.spec.whatwg.org/#enqueue-in-readable-stream"><code class="external" data-anolis-spec="streams">EnqueueInReadableStream</code></a>(<var title="">stream</var>,
- <var title="">chunk</var>). Rethrow any exceptions.
+ <li><p>Call <a href="https://streams.spec.whatwg.org/#enqueue-in-readable-stream"><code class="external" data-anolis-spec="streams">EnqueueInReadableStream</code></a>(<var>stream</var>,
+ <var>chunk</var>). Rethrow any exceptions.
 </ol>
 
 <p>To <dfn id="concept-close-readablestream" title="concept-close-ReadableStream">close</dfn> a
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var>, run these steps:
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var>, run these steps:
+
 <ol>
- <li><p>Call <a href="https://streams.spec.whatwg.org/#close-readable-stream"><code class="external" data-anolis-spec="streams">CloseReadableStream</code></a>(<var title="">stream</var>).
+ <li><p>Call <a href="https://streams.spec.whatwg.org/#close-readable-stream"><code class="external" data-anolis-spec="streams">CloseReadableStream</code></a>(<var>stream</var>).
  Rethrow any exceptions.
 </ol>
 
 <p>To <dfn id="concept-construct-readablestream" title="concept-construct-ReadableStream">construct</dfn> a
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> with given <var title="">strategy</var>,
-<var title="">pull</var> action and <var title="">cancel</var> action, all of which are optional, run
-these steps:
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object with given <var>strategy</var>,
+<var>pull</var> action and <var>cancel</var> action, all of which are optional, run these steps:
 
 <ol>
  <li><p>Let <var title="">init</var> be a new object.
@@ -1205,20 +1204,20 @@ these steps:
  <var title="">strategy</var> is given.
 
  <li><p>Set <var title="">init</var> be the result of calling the initial value of
- <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> as constructor with
+ <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> as constructor with
  <var title="">init</var> as an argument. Rethrow any exceptions.
 
  <li><p>Return <var title="">stream</var>.
 </ol>
 
 <p>To <dfn id="concept-construct-fixed-readablestream" title="concept-construct-fixed-ReadableStream">construct</dfn> a fixed
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> with given <var title="">chunks</var>,
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object with given <var>chunks</var>,
 run these steps:
 
 <ol>
  <li><p>Let <var title="">stream</var> be the result of
  <a href="#concept-construct-readablestream" title="concept-construct-ReadableStream">constructing</a> a
- <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a>. Rethrow any exceptions.
+ <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object. Rethrow any exceptions.
 
  <li><p>For each <var title="">chunk</var> in <var title="">chunks</var>,
  <a href="#concept-enqueue-readablestream" title="concept-enqueue-ReadableStream">enqueue</a> <var title="">chunk</var> into
@@ -1230,8 +1229,25 @@ run these steps:
  <li>Return <var title="">stream</var>.
 </ol>
 
+<p>To <dfn id="concept-construct-readablestream-with-byte-stream" title="concept-construct-ReadableStream-with-byte-stream">construct</dfn> a
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object with given a byte stream
+<var>byteStream</var>, run these steps:
+
+<p class="XXX">This operation is a workaround and will be deleted when we stop using byte streams. 
+
+<ol>
+ <li><p>Let <var>stream</var> be a <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object
+ such that consuming it will result in chunks each of which is a <code title="">Uint8Array</code>
+ object wrapping an <code title="">ArrayBuffer</code> and the result of concatenating them all will be
+ equivalent to the byte sequence which would be read from <var>byteStream</var>. Rethrow any
+ exceptions.
+
+ <li><p>Return <var>stream</var>.
+</ol>
+
 <p>To <dfn id="concept-get-reader" title="concept-get-reader">get</dfn> a reader from a
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var>, run these steps:
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var>, run these steps:
+
 <ol>
  <li><p>Let <var title="">reader</var> be the result of calling
  <a href="https://streams.spec.whatwg.org/#acquire-readable-stream-reader"><code class="external" data-anolis-spec="streams">AcquireReadableStreamReader</code></a>(<var title="">stream</var>).
@@ -1241,7 +1257,7 @@ run these steps:
 </ol>
 
 <p>To <dfn id="concept-read-all-bytes-from-readablestream" title="concept-read-all-bytes-from-ReadableStream">read</dfn> all bytes from a
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> with <var title="">reader</var>, run these
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object with <var>reader</var>, run these
 steps:
 
 <ol>
@@ -1256,7 +1272,7 @@ steps:
  <ul>
   <li><p>When <var title="">read</var> is fulfilled with an object whose <code title="">done</code>
   property is false and whose <code title="">value</code> property is a
-  <code title="">Uint8Array</code>, append the <code title="">value</code> property to
+  <code title="">Uint8Array</code> object, append the <code title="">value</code> property to
   <var title="">bytes</var> and run the above step again.
 
   <li><p>When <var title="">read</var> is fulfilled with an object whose <code title="">done</code>
@@ -1276,7 +1292,7 @@ steps:
 to read cannot be observed. Implementations could use more direct mechanism if convenient.
 
 <p>To <dfn id="concept-tee-readablestream" title="concept-tee-ReadableStream">tee</dfn> a
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var>, run these steps:
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var>, run these steps:
 
 <ol>
  <li><p>Return the result of calling
@@ -1285,30 +1301,29 @@ to read cannot be observed. Implementations could use more direct mechanism if c
 </ol>
 
 <p>An <dfn id="concept-empty-readablestream" title="concept-empty-ReadableStream">empty</dfn>
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> is the result of
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object is the result of
 <a href="#concept-construct-fixed-readablestream" title="concept-construct-fixed-ReadableStream">constructing</a> a fixed
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> with an empty array.
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object with an empty list.
 
 <p class="note no-backref">Constructing an <a href="#concept-empty-readablestream" title="concept-empty-ReadableStream">empty</a>
-<a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> will not throw an exception.
+<a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object will not throw an exception.
 
-<p>A <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var> is said to be
-<dfn id="concept-readablestream-readable" title="concept-ReadableStream-readable">readable</dfn> if <var title="">stream</var>@[[state]] is
+<p>A <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var> is said to be
+<dfn id="concept-readablestream-readable" title="concept-ReadableStream-readable">readable</dfn> if <var>stream</var>@[[state]] is
 "readable".
 
-<p>A <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var> is said to be
-<dfn id="concept-readablestream-closed" title="concept-ReadableStream-closed">closed</dfn> if <var title="">stream</var>@[[state]] is
-"closed".
+<p>A <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var> is said to be
+<dfn id="concept-readablestream-closed" title="concept-ReadableStream-closed">closed</dfn> if <var>stream</var>@[[state]] is "closed".
 
-<p>A <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var> is said to be
-<dfn id="concept-readablestream-errored" title="concept-ReadableStream-errored">errored</dfn> if <var title="">stream</var>@[[state]] is
+<p>A <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var> is said to be
+<dfn id="concept-readablestream-errored" title="concept-ReadableStream-errored">errored</dfn> if <var>stream</var>@[[state]] is
 "errored".
 
-<p>A <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var> is said to be
+<p>A <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var> is said to be
 <dfn id="concept-readablestream-locked" title="concept-ReadableStream-locked">locked</dfn> if the result of calling
-<a href="https://streams.spec.whatwg.org/#is-readable-stream-locked"><code class="external" data-anolis-spec="streams">IsReadableStreamLocked</code></a>(<var title="">stream</var>) is true.
+<a href="https://streams.spec.whatwg.org/#is-readable-stream-locked"><code class="external" data-anolis-spec="streams">IsReadableStreamLocked</code></a>(<var>stream</var>) is true.
 
-<p>A <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var> is said to
+<p>A <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var> is said to
 <dfn id="concept-readablestream-need-more-data" title="concept-ReadableStream-need-more-data">need more data</dfn> if the following conditions
 hold:
 
@@ -1320,9 +1335,9 @@ hold:
  positive.
 </ul>
 
-<p>A <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> <var title="">stream</var> is said to be
+<p>A <a href="#concept-readablestream"><code title="concept-ReadableStream">ReadableStream</code></a> object <var>stream</var> is said to be
 <dfn id="concept-readablestream-disturbed" title="concept-ReadableStream-disturbed">disturbed</dfn> if the result of calling
-<a href="https://streams.spec.whatwg.org/#is-readable-stream-disturbed"><code class="external" data-anolis-spec="streams">IsReadableStreamDisturbed</code></a>(<var title="">stream</var>) is true.
+<a href="https://streams.spec.whatwg.org/#is-readable-stream-disturbed"><code class="external" data-anolis-spec="streams">IsReadableStreamDisturbed</code></a>(<var>stream</var>) is true.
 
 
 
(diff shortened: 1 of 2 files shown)


yutakahirano:
html diff: http://services.w3.org/htmldiff?doc1=https%3A%2F%2Ffetch.spec.whatwg.org%2F&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fyutakahirano%2Ffetch%2Fresponse-body%2FOverview.html#dom-response

annevk:
What is the reason for not introducing `Request.prototype.body`?

Overview.html
```diff
@@ -4080,11 +4131,14 @@ <h3 id="response-class"><span class="secno">6.4 </span>Response class</h3>
...
-   <li><p>Let <var title="">stream</var> and <var title="">Content-Type</var> be the result of
+   <li><p>Let <var title="">bytes</var> and <var title="">Content-Type</var> be the result of
    <a href="#concept-bodyinit-extract" title="concept-BodyInit-extract">extracting</a> <var title="">body</var>.
 
-   <li><p>Set <var title="">r</var>'s <a href="#concept-response-response" title="concept-Response-response">response</a>'s
-   <a href="#concept-response-body" title="concept-response-body">body</a> to <var title="">stream</var>.
+   <li><p>Set <var title="">r</var>'s
+   <a href="#concept-response-readable-stream" title="concept-Response-readable-stream">readable stream</a> to the result of
+   <a href="#concept-construct-fixed-readablestream" title="concept-construct-fixed-ReadableStream">constructing</a> a fixed
+   <a href="#concept-readablestream" title="concept-ReadableStream">ReadableStream</a> with an array consisting of a
```
domenic: The input to the constructing a readable stream algorithm isn't a JS array, but instead an abstract list type. (In ES it would be a List, but most other specs including fetch are a bit less formal.) I'd say either "with a Uint8Array..." or "with a list whose single element is a Uint8Array..."

Overview.src.html
```diff
@@ -3965,9 +3962,63 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+<p class = "note no-backref">The type of the associated
+<span title=concept-Response-readable-stream>readable stream</span> is defined as
+<span title=concept-ReadableStream>ReadableStream</span>, but it is currently discussed. Depending
+on the discussion, the type might be changed to <code>ReadableByteStream</code>. Hence the type name
+is not strictly specified at this moment, though the behavior is specified because
+<code>ReadableByteStream</code> has all methods and properties which
+<span title=concept-ReadableStream>ReadableStream</span> has.
+
+<p>A <code>Response</code> object also has an associated
```
domenic: A better phrasing would be "A Response object is said to be locked if its associated readable stream..." Similarly for disturbed.

Although maybe these concepts should apply to abstract "responses" and not to Response objects? I am not sure.

Overview.html
```diff
@@ -3415,119 +3415,84 @@ <h3 id="body-mixin"><span class="secno">6.2 </span>Body mixin</h3>
...
 <p>The <dfn id="dom-body-bodyused" title="dom-Body-bodyUsed"><code>bodyUsed</code></dfn> attribute's getter must
-return true if the <a href="#concept-body-used-flag" title="concept-Body-used-flag">used flag</a> is set, and false
-otherwise.
+return true if <a href="#concept-body-disturbed" title="concept-Body-disturbed">disturbed</a>, and false otherwise.
 
 <p>Objects implementing the <a href="#body"><code>Body</code></a> mixin also have an associated
-<dfn id="concept-body-consume-body" title="concept-Body-consume-body">consume body</dfn> algorithm, which given a
-<var title="">type</var>, runs these steps:
+<dfn id="concept-body-package-data" title="concept-Body-package-data">package data</dfn> algorithm, which given a
+<var title="">bytes</var>, a <var title="">type</var> and a <var title="">MIME type</var>, runs these steps:
```
domenic: given bytes, not given a bytes

Overview.src.html
domenic: Sorry, disturbed is a bit different. It is mostly fine as you defined it, except I think it should probably apply to interface types, not to individual objects. But I guess the spec as it is currently defines them on individual objects, so no need to change that in your patch.

I think locked would be clearer as I suggested though, unless I am missing something.

Overview.src.html
```diff
@@ -3965,9 +3962,63 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ <li><p>If <var title>stream</var> is null, set <var title>stream</var> to an
+ <span title=concept-empty-ReadableStream>empty ReadableStream</span>.
+
+ <li><p>Let <var title>reader</var> be the result of <span title=concept-get-reader>getting</span>
+ a reader from <var title>stream</var>. If that threw an exception, return a new promise rejected
+ with that exception.
+
+ <li><p>Let <var title>p</var> be a new promise.
+
+ <li><p>Let <var title>promise</var> be the result of
```
domenic: I think this section would be clearer if instead of p and promise, you just said "Let promise be the result of reading all bytes from stream with reader. Return the result of transforming promise by a fulfillment handler that ...". It translates more naturally to idiomatic promise code that way too.

Overview.src.html
```diff
@@ -3965,9 +3962,63 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
-<p>A <code>Response</code> object's <span title=concept-Body-body>body</span> is its
-<span title=concept-Response-response>response</span>'s
-<span title=concept-response-body>body</span>.
+<p>A <code>Response</code> object also has an associated
+<dfn title=concept-Response-readable-stream>readable stream</dfn> of type
+<span title=concept-ReadableStream>ReadableStream</span> which is initially null.
+
+<p class = "note no-backref">The type of the associated
+<span title=concept-Response-readable-stream>readable stream</span> is defined as
+<span title=concept-ReadableStream>ReadableStream</span>, but it is currently discussed. Depending
```
domenic: insert "being" between currently and discussed.

yutakahirano:
> What is the reason for not introducing Request.prototype.body?

Hmm, I made that decision ago, and at that time it looked difficult to reading body stream bytes when fetching. I reconsidered it and it might be possible by changing the fech() function as following.

from:

 - Fetch _request_

to:

 - Let _reader_ be the result of getting a reader from _request_.body. If that throws an exception, reject _p_ with that exception and abort these steps.
 - Let _promise_ be the result of reading all data from _request_.body with _reader_.
  - When _promise_ is fulfilled with _bytes_, set _request_'s body to _bytes_ and Fetch _request_.
  - When _promise_ is rejected with an error, reject _p_ with that error.

What do you think about it?

annevk:
That does not seem ideal since it doesn't stream.

yutakahirano:
> That does not seem ideal since it doesn't stream.

I agree, and addressing the problem requires much effort, because it needs a fetch algorithm modification. There is a [ticket](https://github.com/yutakahirano/fetch-with-streams/issues/41) for the problem and I marked it as "not included in the first release".

annevk:
Okay, fair enough I guess.

yutakahirano:
Thanks, then I will include Request.body in this change (I'm on vacation and will be back on Sep 28).

annevk:
Sorry, I think you misunderstood. I agree with you that it's better to wait we have actual streaming sorted.

yutakahirano:
Oops, sorry for my misunderstanding. Let me confirm, Is it OK to 

1. Land this change (with addressing reviewers' comments) and
1. Address https://github.com/yutakahirano/fetch-with-streams/issues/41 after that ?

annevk:
Yes.

yutakahirano:
Thank you!

Overview.src.html
```diff
@@ -3355,119 +3355,84 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 <!-- https://lists.w3.org/Archives/Public/public-whatwg-archive/2014Jun/thread.html#msg72 -->
 
-<p>Objects implementing the <code>Body</code> mixin gain an associated
-<dfn title=concept-Body-body>body</dfn> (a byte stream), a
-<dfn title=concept-Body-used-flag>used flag</dfn> (initially unset), and a
+<p>Objects implementing the <code>Body</code> mixin gain a
 <dfn title=concept-Body-MIME-type>MIME type</dfn> (initially the empty byte sequence).
 
+<p>Objects implementing the <code>Body</code> mixin must define an associated
+<dfn title=concept-Body-disturbed>disturbed</dfn> predicate and a
```
annevk: What is a predicate? Can we turn this into a flag?

Overview.src.html
```diff
@@ -3355,119 +3355,84 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <p>Otherwise, <span data-anolis-spec=webidl>throw</span> a<code title>TypeError</code>.
 
-     <dt><i title>text</i>
-     <dd><p>Resolve <var title>p</var> with the result of running
-     <span data-anolis-spec=encoding>utf-8 decode</span> on <var title>bytes</var>.
-    </dl>
-  </ol>
+  <dt><i title>JSON</i>
+  <dd>
+   <p>Let <var title>JSON</var> be the result of invoking the initial value of the
```
annevk: Shouldn't you return the result?

Overview.src.html
```diff
@@ -3557,9 +3522,40 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+<ol>
+ <li><p>If this <code>Request</code> is <span title=concept-Body-disturbed>disturbed</span>, return
+ a new promise rejected with a <code>TypeError</code>.
+
+ <li><p>Set <span title=concept-Request-disturbed-flag>disturbed flag</span>.
+
+ <li><p>Let <var title>p</var> be a new promise.
+
+ <li><p>Run these substeps <span data-anolis-spec=html>in parallel</span>:
```
annevk: Newline after `<li>` since it has more than one child.

Overview.src.html
```diff
@@ -4167,40 +4235,90 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <code title>TypeError</code> and abort these steps.
 
-   <li><p>Otherwise, resolve <var title>p</var> with a new <code>Response</code> object
-   associated with <var title>response</var> and a new <code>Headers</code> object whose
+   <li><p>Let <var title>res</var> be a new <code>Response</code> object associated with
+   <var title>response</var> and a new <code>Headers</code> object whose
    <span title=concept-Headers-guard>guard</span> is "<code title>immutable</code>".
+
+   <li><p>If <var title>response</var>'s status is a <span>null body status</span>, resolve
+   <var title>p</var> with <var title>res</var> and abort these steps.
```
annevk: I don't think this is needed here. This is already enforced within Fetch, no?

Overview.src.html
```diff
@@ -4167,40 +4235,90 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <p>To <span>process response body</span> for <var title>response</var>, run these substeps:
+  <ol>
+   <li><p>If <var title>stream</var> is null, abort these steps.
+
+   <li><p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code> whose
+   contents are <var title>response</var>'s <span title=concept-response-body>body</span> into
+   <var title>stream</var>.
 
-  <p>To <span>process response body</span> for <var title>response</var>, do nothing.
+   <li><p>Clear out <var title>response</var>'s <span title=concept-response-body>body</span>.
```
annevk: I don't like this. We should define this more naturally. concept-response-body should be a stream that simply pipes into this one or some such.

annevk:
Also, there's no need to use `<var title>` anymore, just use `<var>` for new text (and when changing existing text).

Overview.src.html
domenic: A predicate is a function that returns a boolean. It seems like a better model than a flag in this case, since you want it to be re-computed every time it is consulted.

annevk: Hmm mkay.

Overview.src.html
yutakahirano: Hmm, that is [another TODO] (https://github.com/yutakahirano/fetch-with-streams/issues/40). In my plan I would do the refactoring after landing this change. Replacing &lt;concept-response-body&gt; will impact other specs in non-trivial ways (https://xhr.spec.whatwg.org/#dom-xmlhttprequest-send for example), do you want to do it here?

Overview.html
yutakahirano: Done.

Overview.src.html
yutakahirano: Fixed the definition of "locked".

Overview.html
yutakahirano: Done.

Overview.src.html

Overview.src.html

Overview.src.html
yutakahirano: Thanks, done.

Overview.src.html
yutakahirano: Done. Is the indentation correct?

Overview.src.html
yutakahirano: This is related to the below comment. If &lt;concept-response-body&gt; is a ReadableStream, we don't need this item. Otherwise, we need this early-exit to make `res` have null body.

yutakahirano:
> Also, there's no need to use `<var title>` anymore, just use `<var>` for new text (and when changing existing text).

Fixed added / modified ones. How about &lt;code title&gt; and &lt;code title=""&gt;?

annevk:
`<code title>` is still needed as `<code>` still has linking semantics.

Overview.src.html
```diff
@@ -3557,9 +3520,41 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ a new promise rejected with a <code>TypeError</code>.
+
+ <li><p>Set <span title=concept-Request-disturbed-flag>disturbed flag</span>.
+
+ <li><p>Let <var>p</var> be a new promise.
+
+ <li>
+  <p>Run these substeps <span data-anolis-spec=html>in parallel</span>:
+  <ol>
+   <li><p>Let <var>bytes</var> be <span title=concept-Request-request>request</span>'s
```
annevk: Shouldn't this do a bit more? E.g., read/consume the body somehow since it's a stream?

Overview.src.html
```diff
@@ -3979,9 +3975,52 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 is itself associated with <span title=concept-Response-response>response</span>'s
 <span title=concept-response-header-list>header list</span>.
 
-<p>A <code>Response</code> object's <span title=concept-Body-body>body</span> is its
-<span title=concept-Response-response>response</span>'s
-<span title=concept-response-body>body</span>.
+<p>A <code>Response</code> object also has an associated
+<dfn title=concept-Response-readable-stream>readable stream</dfn> of type
```
annevk: I think I would prefer "(null or a ReadableStream object)" as done elsewhere. Or perhaps "which is null or a ReadableStream object". And then the initial value separately.

Overview.src.html
```diff
@@ -3979,9 +3975,52 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+<span title=concept-ReadableStream>ReadableStream</span> which is initially null.
+
+<p class = "note no-backref">The type of the associated
+<span title=concept-Response-readable-stream>readable stream</span> is defined as
+<span title=concept-ReadableStream>ReadableStream</span>, but it is currently being discussed.
+Depending on the discussion, the type might be changed to <code>ReadableByteStream</code>. Hence the
+type name is not strictly specified at this moment, though the behavior is specified because
+<code>ReadableByteStream</code> has all methods and properties which
+<span title=concept-ReadableStream>ReadableStream</span> has.
```
annevk: I think it would be clearer to have something like `<p class=XXX><a href={linktoissue}>This might become a <code>ReadableByteStream</code></a>.`

Overview.src.html
```diff
@@ -3979,9 +3975,52 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ with that exception.
+
+ <li><p>Let <var>promise</var> be the result of
+ <span title=concept-read-all-bytes-from-ReadableStream>reading</span> all bytes from
+ <var>stream</var> with <var>reader</var>.
+
+ <li><p>Return the result of transforming <var>promise</var> by a fulfillment handler that
+ returns the result of the <span title=concept-body-package-data>package data</span> algorithm
+ with its first argument, <var>type</var> and <span title=concept-body-mime-type>MIME type</span>.
+</ol>
```
annevk: I don't really understand what transforming a promise means. @domenic?

domenic: http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting

Overview.src.html
```diff
@@ -4178,44 +4233,92 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
-  <p>To <span>process response</span> for <var title>response</var>, run these substeps:
+  <p>To <span>process response</span> for <var>response</var>, run these substeps:
 
   <ol>
-   <li><p>If <var title>response</var>'s <span title=concept-response-type>type</span> is
-   "<code title>error</code>", reject <var title>p</var> with a
-   <code title>TypeError</code>.
+   <li><p>If <var>response</var>'s <span title=concept-response-type>type</span> is
+   "<code title>error</code>", reject <var>p</var> with a <code title>TypeError</code> and abort
```
annevk: abort -> terminate.

Overview.src.html
```diff
@@ -4178,44 +4233,92 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   these steps.
 
-   <li><p>Otherwise, resolve <var title>p</var> with a new <code>Response</code> object
-   associated with <var title>response</var> and a new <code>Headers</code> object whose
+   <li><p>Let <var>res</var> be a new <code>Response</code> object associated with
+   <var>response</var> and a new <code>Headers</code> object whose
    <span title=concept-Headers-guard>guard</span> is "<code title>immutable</code>".
+
+   <li><p>If <var>response</var>'s status is a <span>null body status</span>, resolve <var>p</var>
```
annevk: xref status. Having said that, I don't think we need this step at all. Why do you think we need it?

Overview.src.html
```diff
@@ -4178,44 +4233,92 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   with <var>res</var> and abort these steps.
+
+   <li><p>Let <var>pull</var> be an action that <span title=concept-fetch-resume>resumes</span> the
+   ongoing fetch if it is <span title=concept-fetch-suspend>suspended</span>.
+
+   <li><p>Let <var>cancel</var> be an action that
+   <span title=concept-fetch-terminate>terminates</span> the ongoing fetch with reason
+   <i title>end-user abort</i>.
+
+   <li><p>Let <var>strategy</var> be an object. The user agent may choose any object.
```
annevk: Newline after `<li>` since it has two children.

Overview.src.html
```diff
@@ -4178,44 +4233,92 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <p class="note no-backref"><var>strategy</var> is used to control the queuing strategy of
+   <var>stream</var> constructed below.
+
+   <li><p>Set <var>stream</var> to the result of
```
annevk: New line after `<li>` since it has two children. Same below.

Overview.src.html
```diff
@@ -4178,44 +4233,92 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <p>To <span>process response body</span> for <var>response</var>, run these substeps:
+  <ol>
+   <li><p>If <var>stream</var> is null, abort these steps.
+
+   <li><p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code> whose
+   contents are <var>response</var>'s <span title=concept-response-body>body</span> into
+   <var>stream</var>.
 
-  <p>To <span>process response body</span> for <var title>response</var>, do nothing.
+   <li><p>Clear out <var>response</var>'s <span title=concept-response-body>body</span>.
```
annevk: I don't understand this. "body" is itself a stream of sorts. Why would it be cleared?

Overview.src.html
annevk: Hmm, we really need IDL to define this terminology.

Overview.src.html
yutakahirano: Maybe I don't understand what "byte stream" is. What operations does it have?

"Let _bytes_ be the result of consuming request's body as a byte sequence" makes sense?

annevk: "Let bytes be the result of reading from stream until it returns end-of-stream." is what I use now, mostly based on https://encoding.spec.whatwg.org/#concept-stream and waiting for the Streams standard to become ready.

yutakahirano: Fixed.

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html
yutakahirano: Without the early-exit, `res` has an empty associated readable stream instead of null. I stopped using null body status here and check `response`'s body's nullity instead.

Overview.src.html

Overview.src.html

Overview.src.html
yutakahirano: I thought it might be unclear if reading is an idempotent operation or a destructive operation. If it's clear that reading from a byte stream is destructive, we can remove the sentence. Done.

Overview.src.html
```diff
@@ -3355,119 +3355,82 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
-
-       <li><p>If that fails for some reason, reject <var title>p</var> with a
-       <code title>TypeError</code>.
-
-       <li><p>Otherwise, resolve <var title>p</var> with a new
-       <code data-anolis-spec=xhr>FormData</code> object, appending each entry, resulting
-       from the parsing operation, to
-       <span data-anolis-spec=xhr title=concept-FormData-entry>entries</span>.
-      </ol>
+ <li><p>Switch on <var>type</var>:
```
annevk: This seems to be the only step in this list? I think would be better to avoid the `<ol>` then and say something along the lines of "switch on type and run the associated steps" in the previous paragraph.

Overview.src.html
```diff
@@ -3355,119 +3355,82 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+  <dt><i title>Blob</i>
+  <dd><p>Return a <code data-anolis-spec=fileapi>Blob</code> whose contents are
+  <var>bytes</var> and <code data-anolis-spec=fileapi title=dom-Blob-type>type</code>
+  is <var>MIME type</var>.
+
+  <dt><i title>FormData</i>
+  <dd>
+   <p>If <var>MIME type</var> (ignoring parameters) is `<code>multipart/form-data</code>`,
+   run these subsubsteps:
```
annevk: These are now just substeps.

Overview.src.html
```diff
@@ -3557,9 +3520,43 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+   <li><p>If <span title=concept-Request-request>request</span>'s
+   <span title=concept-request-body>body</span> is not null, set <var>bytes</var> to the result of
+   reading from <span title=concept-Request-request>request</span>'s
+   <span title=concept-request-body>body</span> until it returns end-of-stream.
+
+   <li><p>Resolve <var>p</var> with the result of running the
+   <span title=concept-body-package-data>package data</span> algorithm with <var>bytes</var>,
+   <var>type</var> and <span title=concept-Body-MIME-type>MIME type</span>. If that threw an
+  exception, reject <var>p</var> with that exception.
```
annevk: Indentation for "exception" is wrong.

Overview.src.html
```diff
@@ -3979,9 +3977,51 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 is itself associated with <span title=concept-Response-response>response</span>'s
 <span title=concept-response-header-list>header list</span>.
 
-<p>A <code>Response</code> object's <span title=concept-Body-body>body</span> is its
-<span title=concept-Response-response>response</span>'s
-<span title=concept-response-body>body</span>.
+<p>A <code>Response</code> object also has an associated
+<dfn title=concept-Response-readable-stream>readable stream</dfn> which is null or a
+<span title=concept-ReadableStream>ReadableStream</span>. Its initial value is null.
```
annevk: ReadableStream is an object, right? Should we not use `<code>` for it and follow it with the word object? E.g., `<code title=...>ReadableStream</code> object`?

Overview.src.html
```diff
@@ -3979,9 +3977,51 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+<p class="XXX">This might become a <code>ReadableByteStream</code>, depending on the
+<a href="https://github.com/whatwg/streams/issues/379">the discussion</a>.
+
+<p class="note no-backref">The type might be changed, but the behavior is specified because
+<code>ReadableByteStream</code> has all methods and properties which
+<span title=concept-ReadableStream>ReadableStream</span> has.
```
annevk: This can be folded into the XXX.

Overview.src.html
annevk: Can you add `<!-- XXX IDL really needs to define "transforming". For now it is defined in http://www.w3.org/2001/tag/doc/promises-guide -->` (wrapped if it exceeds the line length, aligned as per other comments in the document) please?

Overview.src.html
```diff
@@ -4034,11 +4074,14 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <li><p>Let <var>bytes</var> and <var>Content-Type</var> be the result of
+   <span title=concept-BodyInit-extract>extracting</span> <var>body</var>.
 
-   <li><p>Set <var title>r</var>'s <span title=concept-Response-response>response</span>'s
-   <span title=concept-response-body>body</span> to <var title>stream</var>.
+   <li><p>Set <var>r</var>'s
+   <span title=concept-Response-readable-stream>readable stream</span> to the result of
+   <span title=concept-construct-fixed-ReadableStream>constructing</span> a fixed
+   <span title=concept-ReadableStream>ReadableStream</span> with a list whose single element is a
+   <code>Uint8Array</code> whose contents are <var>bytes</var>. Rethrow any exceptions.
```
annevk: We should probably also say object after `Uint8Array`. Also, instead of "whose contents are" I think "whose buffer is" is more accurate. Do we want to spell out that the buffer is a new `ArrayBuffer`? I think we might. It seems you are already handling that throwing.

annevk: Also, do we really want to require this consuming to happen during the constructor? It seems to throw away some of the benefits of a stream being asynchronous.

Overview.src.html
```diff
@@ -4178,44 +4234,93 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   these steps.
 
-   <li><p>Otherwise, resolve <var title>p</var> with a new <code>Response</code> object
-   associated with <var title>response</var> and a new <code>Headers</code> object whose
+   <li><p>Let <var>res</var> be a new <code>Response</code> object associated with
+   <var>response</var> and a new <code>Headers</code> object whose
    <span title=concept-Headers-guard>guard</span> is "<code title>immutable</code>".
+
+   <li><p>If <var>response</var>'s <span title=concept-response-body>body</span> is null, resolve
+   <var>p</var> with <var>res</var> and terminate these steps.
```
annevk: Then p never gets returned.

Overview.src.html
```diff
@@ -4178,44 +4234,93 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+    <p>Let <var>strategy</var> be an object. The user agent may choose any object.
+    <p class="note no-backref"><var>strategy</var> is used to control the queuing strategy of
+    <var>stream</var> constructed below.
+
+   <li>
+    <p>Set <var>stream</var> to the result of
+    <span title=concept-construct-ReadableStream>constructing</span> a
+    <span title=concept-ReadableStream>ReadableStream</span> with <var>strategy</var>,
+    <var>pull</var> and <var>cancel</var>. If that threw an exception, run the following
+    substeps:
```
annevk: Newline.

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html
yutakahirano: Hm, I added "constructing a ReadableStream with a byte stream" operation and use it here.

Overview.src.html
yutakahirano: Here is in the "process response for response" handler, so I thought the termination was scoped in it (as "return" in the "process request body" handler in the XHR spec).

Overview.src.html

Overview.src.html
annevk: That could work. But then use the "substep" (or subsubstep depending on the level of nesting) wording.

Overview.src.html
```diff
@@ -1073,7 +1073,7 @@ <h4 id=terminology-headers>Headers</h4>
 define common operations for ReadableStream. <span data-anolis-ref>STREAMS</span>
 
 <p>To <dfn title=concept-enqueue-ReadableStream>enqueue</dfn> a byte sequence
-(<var title>chunk</var>) into a <span title=concept-ReadableStream>ReadableStream</span>
+(<var title>chunk</var>) into a <code title=concept-ReadableStream>ReadableStream</code>
```
annevk: `ReadableStream` object* (throughout)

Overview.src.html
```diff
@@ -1131,8 +1131,39 @@ <h4 id=terminology-headers>Headers</h4>
  <li>Return <var title>stream</var>.
 </ol>
 
+<p>To <dfn title=concept-construct-ReadableStream-with-byte-stream>construct</dfn> a
+<code title=concept-ReadableStream>ReadableStream</code> with given a byte stream
+<var>bstream</var>, run these steps:
```
annevk: `<var>byteStream</var>`*

Overview.src.html
```diff
@@ -1131,8 +1131,39 @@ <h4 id=terminology-headers>Headers</h4>
...
+ <li>
+  <p>Let <var>pull</var> be an action that runs these substeps:
+
+  <ol>
+    <li><p>Let <var>bytes</var> be a byte sequence which is 
+    <span data-anolis-spec=encoding title=concept-stream-read>read</span> from <var>bstream</var>.
+
+    <li><p>If <span data-anolis-spec=encoding title=concept-stream-read>read</span> returns
+    <span data-anolis-spec=encoding>end-of-stream</span>,
+    <span title=concept-close-ReadableStream>close</span> <var>stream</var> defined below.
```
annevk: This reads rather weird. Perhaps introduce a flag before these substeps, set it here, and then after these substeps and after `<var>stream</var>` is allocated, use it to close it if the flag is set?

Overview.src.html
```diff
@@ -3355,120 +3386,81 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 <p>The <dfn title=dom-Body-bodyUsed><code>bodyUsed</code></dfn> attribute's getter must
-return true if the <span title=concept-Body-used-flag>used flag</span> is set, and false
-otherwise.
+return true if <span title="concept-Body-disturbed">disturbed</span>, and false otherwise.
 
 <p>Objects implementing the <code>Body</code> mixin also have an associated
-<dfn title=concept-Body-consume-body>consume body</dfn> algorithm, which given a
-<var title>type</var>, runs these steps:
+<dfn title=concept-Body-package-data>package data</dfn> algorithm, which given
+<var>bytes</var>, a <var>type</var> and a <var>MIME type</var>, switches on <var>type</var> and runs
```
annevk: comma after `<var>type</var>`.

Overview.src.html
```diff
@@ -4034,11 +4103,14 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
     <p class="note no-backref"><code title>101</code> is included in
     <span>null body status</span> due to its use elsewhere. It does not affect this step.
 
-   <li><p>Let <var title>stream</var> and <var title>Content-Type</var> be the result of
-   <span title=concept-BodyInit-extract>extracting</span> <var title>body</var>.
+   <li><p>Let <var>bstream</var> and <var>Content-Type</var> be the result of
```
annevk: byteStream

Overview.src.html
yutakahirano: _pull_ is called multiple times and end-of-stream should be checked every time.

annevk: Oh I see, but then `pull` is presumably passed some arguments? That would make this more clear.

yutakahirano: In fact, byte stream lacks some signals required by ReadableStream such as "process response body" and "process response end-of-file". I'm afraid we cannot specify this operation perfectly. Is it possible to state something like

To __construct__ a ReadableStream object from a byte stream _byteStream_, return a ReadableStream object so that consuming the ReadableStream object will result in chunks each of which is a Uint8Array object wrapping an ArrayBuffer and concatenate them all will result in a byte sequence that is equivalent to the byte sequence which would be read from _byteStream_.

?

annevk: I guess so. Is there a long term plan to improve this?

Overview.src.html
```diff
@@ -1106,20 +1106,20 @@ <h4 id=terminology-headers>Headers</h4>
...
- <span title=concept-ReadableStream>ReadableStream</span> as constructor with
+ <code title=concept-ReadableStream>ReadableStream</code> as constructor with
  <var title>init</var> as an argument. Rethrow any exceptions.
 
  <li><p>Return <var title>stream</var>.
 </ol>
 
 <p>To <dfn title=concept-construct-fixed-ReadableStream>construct</dfn> a fixed
-<span title=concept-ReadableStream>ReadableStream</span> with given <var title>chunks</var>,
+<code title=concept-ReadableStream>ReadableStream</code> with given <var title>chunks</var>,
```
annevk: object*

Overview.src.html
yutakahirano: I will use ReadableStream instead of byte stream in request and response.

For extracting Strings, ArrayBuffers and FormData objects, they can be converted to byte sequences and thus easily enqueued to a ReadableStream object.

For extracting Blobs, it would be possible to define an operation returning a ReadableStream object consisting of the blob contents (At least, as [blob] (https://w3c.github.io/FileAPI/#blob) refers to a byte sequence, enqueuing it to the ReadableStream object is possible. Allowing the UA to split the sequence to chunks would be beneficial, though).

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html
annevk: Yeah only using `ReadableStream` makes sense. That or an equivalent abstract concept.

Overview.src.html
```diff
@@ -4178,44 +4248,96 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+    <ol>
+     <li><p>Reject <var>p</var> with that exception.
+
+     <li><p><span title=concept-fetch-terminate>Terminate</span> the ongoing fetch with reason
+     <i title>fatal</i>.
+    </ol>
+
+   <li>
+    <p>Otherwise, run the following substeps:
```
annevk: This "Otherwise" is a tad confusing since the previous step doesn't start with an "If". Since the previous step ends up aborting the substeps, do we need an otherwise and substeps here?

Overview.src.html
```diff
@@ -4178,44 +4248,96 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
-   <li><p>If <var title>response</var>'s <span title=concept-response-type>type</span> is
-   "<code title>error</code>", reject <var title>p</var> with a
-   <code title>TypeError</code>.
+   <li><p>If <var>response</var>'s <span title=concept-response-type>type</span> is
+   "<code title>error</code>", reject <var>p</var> with a <code title>TypeError</code> and terminate
+   these substeps.
 
-   <li><p>Otherwise, resolve <var title>p</var> with a new <code>Response</code> object
-   associated with <var title>response</var> and a new <code>Headers</code> object whose
+   <li><p>Let <var>res</var> be a new <code>Response</code> object associated with
```
annevk: Perhaps we should call this variable `responseObject` or some such? Or maybe not rename the internal response from `r` to `response`...

Overview.src.html
```diff
@@ -3979,9 +3990,52 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
-<p>A <code>Response</code> object's <span title=concept-Body-body>body</span> is its
-<span title=concept-Response-response>response</span>'s
-<span title=concept-response-body>body</span>.
+<p>A <code>Response</code> object also has an associated
+<dfn title=concept-Response-readable-stream>readable stream</dfn> which is null or a
+<code title=concept-ReadableStream>ReadableStream</code> object. Its initial value is null.
+
+<p class="XXX">This might become a <code>ReadableByteStream</code> object, depending on the
+<a href="https://github.com/whatwg/streams/issues/379">the discussion</a>. The type might be
```
annevk: "the the discussion" is wrong. I think this entire thing can be clearer as:

`<a>This might become a ReadableByteStream object</a>. While the type might change, the behavior specified will be equivalent since the hypothetical ReadableByteStream object will have the same members as the ReadableStream object has today.`

yutakahirano: Oops, sorry. Fixed.

Overview.src.html
yutakahirano: Terminating a fetch operation and terminating substeps are different things. We should state both in this case. Fixed.

Overview.src.html
yutakahirano: I don't like `r` because it can be read as either one of request, response, Request and Response. I also want to use consistent names in this function - If I use `request` for a request, I wanna use `response` for a response and vice versa.

`responseObject` sounds good. Fixed.

Overview.src.html
```diff
@@ -3355,120 +3370,81 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
-<var title>type</var>, runs these steps:
+<dfn title=concept-Body-package-data>package data</dfn> algorithm, which given
+<var>bytes</var>, a <var>type</var> and a <var>MIME type</var>, switches on <var>type</var>, and
+runs the associated steps:
 
-<ol>
- <li><p>Let <var title>p</var> be a new promise.
+<dl class=switch>
+ <dt><i title>ArrayBuffer</i>
+ <dd><p>Return an <code>ArrayBuffer</code> whose contents are <var>bytes</var>.
```
annevk: Do you want to fix #134 here as well? It's okay if you don't, but with your new setup it seems fairly easy to just add the rethrow sentence.

annevk:
Apart from that this looks good. Do you want to squash the commits and rebase or should I?

Overview.src.html
yutakahirano: Are you OK with https://github.com/yutakahirano/fetch/commit/bf979b16cf394dc0b84ef2716dbce4b264800d8a?

annevk: Yes.

yutakahirano:
Thank you very much for the review!
I squashed and rebased the commits.

yutakahirano:
Addressing #118

annevk:
Thank you for the great patch!

PR #131, opened 2015-09-24
Reference ABNF spec

sideshowbarker:


Overview.src.html
```diff
@@ -395,6 +395,7 @@ <h4 id=terminology-headers>Headers</h4>
   <p>If the ABNF for <var title>name</var> allows a single
   <span title=concept-header>header</span> and <var title>headers</var> contains more than
   one, return failure.
+  <span data-anolis-ref=ABNF>ABNF</span>
```
annevk: Just `<span data-anolis-ref>ABNF</span>` is sufficient. However, more in style of this specification would be to just add ABNF to the list of specifications referenced in terminology. I think that might be better than reference it a ton of times, since ABNF is fairly well established.

sideshowbarker: OK, will change it to do it that way

PR #138, opened 2015-10-07
Add an example using Response body

yutakahirano:
Addressing #118.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -3049,7 +3049,7 @@ Entries may be removed before that moment arrives.
 <p class=no-backref>The <code title=dom-global-fetch>fetch()</code> method is relatively
 low-level API for <span title=concept-fetch>fetching</span> resources. It covers slightly
 more ground than <code data-anolis-spec=xhr>XMLHttpRequest</code>, although it is
-currently lacking when it comes to reporting progression.
+currently lacking when it comes to request progression (not response progression).
 
 <div class="example no-backref">
  <p>The <code title=dom-global-fetch>fetch()</code> method makes it quite straightforward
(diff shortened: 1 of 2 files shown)


annevk:
Can you also update the initial paragraph of https://fetch.spec.whatwg.org/#fetch-api to say "lacking when it comes to request progression (not response progression)"?

annevk:
@domenic could you have a look at the example just in case?

domenic:
I really preferred the example before https://github.com/yutakahirano/fetch-with-streams/pull/58. It was more idiomatic promise usage and much clearer. The version in this PR is an example of the [explicit promise construction anti-pattern](http://stackoverflow.com/q/23803743/3191).

But given that naive implementation of the promise spec currently causes the more idiomatic example to retain some memory for each chunk, maybe the proposed example is better since it avoids those issues in current implementations. I am a bit torn. Since @yutakahirano thinks this proposed version is better than I am OK with it.

Overview.src.html
```diff
@@ -3082,6 +3082,31 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  var total = 0
+  return new Promise((resolve, reject) => {
+    function pump() {
+      reader.read().then(({done, value}) => {
+        if (done) {
+          resolve()
+          return
+        }
+        total += value.byteLength
+        log("received " + value.byteLength + " bytes (" + total + " bytes in total).")
```
domenic: Optional ES6 goodness: ``log(`received ${value.byteLength} bytes (${total} bytes in total)`)``.

yutakahirano: Thanks, done.

yutakahirano:
> Can you also update the initial paragraph of https://fetch.spec.whatwg.org/#fetch-api to say "lacking when it comes to request progression (not response progression)"?

Done.

Regarding the example, code, I agree that the previous one is more elegant and it's unfortunate that existing implementations cannot execute it efficiently. That said, we received a report related with this problem once, so encouraging the pattern is dangerous at least for now, I think.

annevk:
Rebase? LGTM.

yutakahirano:
Rebased.

PR #140, opened 2015-10-13
Update Fetch with shiny, new CSP hooks

mikewest:
Content Security Policy has defined a set of integrations with Fetch
that make a little more sense than the hand-wavey "Whenever the
user agent requests a resource ..." descriptions in CSP1 and CSP2.
Finally.

This patch updates Fetch's placeholders to link to the new algorithms,
and adds a "policy list" attribute to response objects in order to
support those algorithms' requirements.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -908,6 +908,12 @@ suites, certificates for an "Internal Name", or certificates with an overly long
 period. How exactly a user agent can use "<code title>deprecated</code>" is not defined by
 this specification.
 
+<p>A <span title=concept-response>response</span> has an associated
+<dfn title=concept-response-csp-list>CSP list</dfn>, which is a list of
+<a href=https://w3c.github.io/webappsec-csp/#policy>Content Security Policy objects</a>
+for the <span title=concept-response>response</span>. The list is empty unless otherwise
+specified. <span data-anolis-ref>CSP</span>
+
 <hr>
 
 <p>A <span title=concept-response>response</span> whose
@@ -1663,7 +1669,7 @@ redirects.
  <li><p>If
  <a href="https://w3c.github.io/webappsec/specs/mixedcontent/#should-block-fetch">should fetching <var title>request</var> be blocked as mixed content</a>
  or
- <span class=XXX>should fetching <var title>request</var> be blocked as content security</span>
+ <a href="https://w3c.github.io/webappsec-csp/#should-block-request">should fetching <var>request</var> be blocked by Content Security Policy</a>
  returns <b title>blocked</b>, set <var title>response</var> to a
  <span title=concept-network-error>network error</span>.
  <span data-anolis-ref>MIX</span>
@@ -1796,7 +1802,7 @@ redirects.
 
  <li><p>If
  <a href=https://w3c.github.io/webappsec/specs/mixedcontent/#should-block-response>should <var>internalResponse</var> to <var>request</var> be blocked as mixed content</a>,
- <span class=XXX>should <var>internalResponse</var> to <var title>request</var> be blocked as content security</span>,
+ <a href=https://w3c.github.io/webappsec-csp/#should-block-response>should <var>internalResponse</var> to <var>request</var> be blocked by Content Security Policy</a>,
  or
  <span title="should response to request be blocked due to nosniff">should <var>internalResponse</var> to <var>request</var> be blocked due to nosniff</span>
  returns <b title>blocked</b>, set <var title>response</var> to a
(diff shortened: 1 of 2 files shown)


mikewest:
WDYT, @annevk? I think this deals with the things you noted in your initial pass.

Overview.src.html
```diff
@@ -908,6 +908,13 @@ <h4 id=terminology-headers>Headers</h4>
 period. How exactly a user agent can use "<code title>deprecated</code>" is not defined by
 this specification.
 
+<p>A <span title=concept-response>response</span> has an associated
+<dfn title=concept-response-policy-list>policy list</dfn>, which is a list of
+<a href=https://w3c.github.io/webappsec-csp/#policy>Content Security Policy objects</a>
+delivered in the <span title=concept-response>response</span>'s
+<span title=concept-response-header-list>header list</span>. The list is empty unless
+otherwise specified.
```
annevk: Perhaps "CSP list" would be better? I suspect this might not be the last policy. Also, any particular reason the source is mentioned?

Overview.src.html
```diff
@@ -2087,6 +2094,11 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
      "<code title>manual</code>" and <var title>response</var>'s
      <span title=concept-response-type>type</span> is "<code title>opaqueredirect</code>".
     </ul>
+
+   <li>
+     <p>Execute <a href=https://w3c.github.io/webappsec-csp/set-response-policy-list>Set
```
annevk: "Execute Set" has incorrect casing.

Overview.src.html
```diff
@@ -2087,6 +2094,11 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+     <var title>response</var>'s policy list</a> on <var title>response</var>.
+     <span data-anolis-ref>CSP</span>
```
annevk: Should this clause not apply to opaque responses? Why not here, but it's fine for the next one?

mikewest: Sorry, I don't follow the question.

This bit takes care of responses pushed in from Service Workers. The change in HTTP-network fetch takes care of requests pushed in from the network (either inside or outside of a Service Worker). Where else do you think we ought to be setting the CSP list?

annevk: You can't get to the header list of an opaque response (it will always be empty).

I guess as long as opaque responses always come from the network this might work (since the internal response will have its policy list set at that point), but it's not clear to me they always will. But maybe if they don't come from the network CSP doesn't matter...

On top of that, should we avoid setting a policy list here if the response already has one?

mikewest: > You can't get to the header list of an opaque response (it will always be empty).

I see. As you note, responses which came from the network will be covered by the HTTP-network fetch hook, so the responses we need to deal with are opaque responses which are synthetically constructed by the SW.

Would it be appropriate for this hook to address the `internal response`'s header list? That seems like the most straightforward way to handle things.

> On top of that, should we avoid setting a policy list here if the response already has one?

Probably. I'm not sure whether that's the right thing to do, though. I think I'm going to want to read the policy information as early as possible in fetch when grabbing things from the network (for `Set-Cookie` if no other reason). It's not clear that I can avoid setting the policy twice if I want to do that, and handle synthetic responses coming from the service worker.

I think what might work is for the `set the CSP list` algorithm to clear the response's CSP list, and overwrite it with whatever the response's header list (or it's internal response's header list) contains. WDYT?

mikewest: I've taken a stab at this in the current Fetch patch, and poked at the CSP side in https://github.com/w3c/webappsec-csp/commit/0545ae08d6c74788f006e07b2b991bbf925c9691. WDYT?

annevk:
I guess this looks reasonable. I suppose we can tweak it if it doesn't work out. Hopefully implementations will report bugs...

s/`<var title>`/`<var>`/ for the changed lines.

mikewest:
Thanks. I hope it's the right thing to do, but if not, it's still a better explanation than today's text. :)

bb10b20 drops `title` and squashes.

Overview.src.html
```diff
@@ -1796,7 +1803,7 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 
  <li><p>If
  <a href=https://w3c.github.io/webappsec/specs/mixedcontent/#should-block-response>should <var>internalResponse</var> to <var>request</var> be blocked as mixed content</a>,
- <span class=XXX>should <var>internalResponse</var> to <var title>request</var> be blocked as content security</span>,
+ <a href=https://w3c.github.io/webappsec-csp/#should-block-response>should <var>internalResponse</var> to <var title>request</var> be blocked by Content Security Policy</span>,
```
annevk: Needs to be closed with `</a>`. Also `<var title>` -> `<var>`.

Overview.src.html
```diff
@@ -1663,7 +1670,7 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <li><p>If
  <a href="https://w3c.github.io/webappsec/specs/mixedcontent/#should-block-fetch">should fetching <var title>request</var> be blocked as mixed content</a>
  or
- <span class=XXX>should fetching <var title>request</var> be blocked as content security</span>
+ <a href="https://w3c.github.io/webappsec-csp/#should-block-request">should fetching <var title>request</var> be blocked by Content Security Policy</a>
```
annevk: `<var title>` -> `<var>`.

mikewest:
Fixed, ans squished. I also rebuilt the HTML, in the hopes that that makes your job easier. :)

annevk:
Thank you!

Issue #38, opened 2015-04-05
Cache state: partial content


annevk:
@mayhemer it's not entirely clear to me how to deal with partial content. Does us handling that mean we also pay special attention to a 206 response? What's the logic here, if we notice we have a partial cache entry, we set `If-Range`, then if there's a 206 response, we combine some stuff and hand out the combined response as a 200 to content? Does this only happen for media elements basically?

mayhemer:
Not sure what is the question here.  Is the "cache mode" from this bug title the cache mode on fetch?

annevk:
Sorry, I messed up the title, this is about a related feature: https://fetch.spec.whatwg.org/#concept-response-cache-state And also how we should deal with partial content cache entries overall within Fetch.

mayhemer:
> it's not entirely clear to me how to deal with partial content. Does us handling that mean we also pay special attention to a 206 response? What's the logic here, if we notice we have a partial cache entry, we set If-Range, then if there's a 206 response, we combine some stuff and hand out the combined response as a 200 to content? 

Exactly.  We always hand out a 200 (on a new 200 server response -> cache state "none", or a 206 response from the server that we, as you say, combine -> cache state "partial" or could well be "validated" since 206 is mostly the same as 304.  But depends.  I kinda don't see a use case right now for the "partial" state, tho)

> Does this only happen for media elements basically?

No, this can happen for anything, large HTML, CSS whatever.  The response just needs to fulfill some conditions to be "resumable".  I don't think you need to be concerned about it tho.

wanderview:
Some IRC discussion of what we do in gecko's http cache:

12:13 PM <valentin> mayhemer: bkelly asked: do you know if our http cache does anything special for range requests?  are they automatically combined together or anything?
12:14 PM <mayhemer> we only support caching partial content in range <1-N> where N is anything between 1 and content-length
12:14 PM <mayhemer> when we have sych a partial entry (for N < content-length)
12:14 PM <mayhemer> we do an If-Range request and merge, invisibly to the consumer of nsIChannel
12:15 PM <mayhemer> (we merge on 206)
12:15 PM <mayhemer> (we refetch on 200)
12:15 PM <mayhemer> (we report any server error otherwise)
12:15 PM <mayhemer> bkelly: ^^
12:16 PM <mayhemer> note that not all content is resumable so in few cases we throw any partial cached data as soon as the interruption occurs
12:17 PM <bkelly> mayhemer: thanks!
[...]

PR #141, opened 2015-10-20
Update headers upon a 304; fixes #97

### LINKED ISSUE #97: 206/304 handling broken?
@mnot, at some point you mentioned the handling of 304 was broken. Presumably that also applies to 206, which is not really handled at all at the moment. Could you elaborate on that or point to where we discussed this?

mnot:


annevk:
I don't think this was the main problem. The problem was that 304/206 were handled at an earlier stage than the current draft suggests. And this cannot fix that issue as it does not handle 206. Or am I missing something?

mnot:
I agree that 206 needs to be handled. Not sure what you mean by "earlier stage" though -- the underlying issue AFAIK was that response headers weren't updated by the 304. Please explain?

mnot:
WRT 206 - a decision similar to the one made for validation needs to be made -- what to do when there's a Range header on the incoming request. Is it worth creating a whole new mode for "don't combine partial responses with what you have cached, pass them through"?

annevk:
So reading that thread linked from the issue. One of the problems is that where 304 is handled now in Fetch, CORS has already happened. So the 304 would be rejected if it lacked CORS headers. Which per that thread is wrong. We should first get the response from the cache, update things all around, and then do a CORS check.

mnot:
Ah, right. Will dig more.

mnot:
I think I need to move the 304 case (and presumably 206 when it materialises) into HTTP-network-or-cache fetch. Make sense?

annevk:
Yeah, I think that's what we concluded.

mnot:
I feel like 

> If request's cache mode is "default" and request's header list contains a header named `If-Modified-Since`, `If-None-Match`, `If-Unmodified-Since`, `If-Match`, or `If-Range`, set request's cache mode to "no-store".

should go in there too. Any objection?

annevk:
Sounds good.

mnot:
OK, take a look RE: 304. 

What do you want to do with 206?

annevk:
It seems the `<!-- XXX xref "HTTP cache" -->` comments can be removed. And for 206 I guess we want something similar. https://github.com/whatwg/fetch/issues/38#issuecomment-125277789 has some comments on how Gecko only supports a single partial entry, but I think the specification could support multiple if that doesn't complicate things too much.

mnot:
I'd like to do 206 as a separate discussion, because it could be pretty intrusive. Talking to @jakearchibald  last night, it's not clear how capable we want to make it. 

Can we push that out of #97 into a separate issue?

annevk:
Separating that out seems fine. Comments still need to be removed, maybe?

Overview.src.html
```diff
@@ -2481,6 +2434,18 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <var title>httpRequest</var>'s
  <span title=concept-request-header-list>header list</span>.
 
+ <li><p>If <var title>httpRequest</var>'s
```
annevk: s/`<var title>`/`<var>`/ for changed `<var>`s.

Overview.src.html
```diff
@@ -2602,6 +2567,52 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
    <var title>httpRequest</var> with the <i title>credentials flag</i> set if set.
   </ol>
 
+ <li>
+   <p>If <var title>response</var>'s <span title=status>status</span> is <code>304</code>, run
```
annevk: Single-space indentation.

Overview.src.html
```diff
@@ -2602,6 +2567,52 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+      <span title=concept-response-status>status</span> which is most likely <code title>200</code>
+      now.
+
+      <p>Set <var title>response</var>'s
+      <span title=concept-response-cache-state>cache state</span> to
+      "<code title>validated</code>".
+    </ol>
+  </ol>
+
+
```
annevk: One newline less here.

Overview.src.html
mnot: what is a changed <var>?

annevk: Basically any that is on a green line in this diff view thingie.

annevk:
I ended up having to fix the hyperlinks a bit since you had two closing tags for them. But other than that this was mostly fine. Thank you!

PR #146, opened 2015-10-28
Use the URL from the response, if it has one

annevk:
This mostly affects the way stylesheets work in service workers. See
https://github.com/slightlyoff/ServiceWorker/issues/757 for details.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -1897,10 +1897,6 @@ redirects.
   <p class="note">This operates on <var title="">response</var> as this algorithm is not
   supposed to observe <var>internalResponse</var>.
 
- <li><p>Set <var title="">internalResponse</var>'s
- <a href="#concept-response-url-list" title="concept-response-url-list">url list</a> to a copy of
- <var title="">request</var>'s <a href="#concept-request-url-list" title="concept-request-url-list">url list</a>.
-
  <li>
   <p>If <var title="">request</var>'s <a href="#synchronous-flag">synchronous flag</a> is set,
   <a href="#wait-for-a-response" title="wait for a response">wait for <var>internalResponse</var></a>, and then
@@ -2114,6 +2110,8 @@ indicates an attempt to authenticate.
 <ol>
  <li><p>Let <var title="">response</var> be null.
 
+ <li><p>Let <var>actualResponse</var> be null.
+
  <li>
   <p>If <var title="">request</var>'s <a href="#skip-service-worker-flag">skip-service-worker flag</a> is unset and
   either <var title="">request</var>'s <a href="#concept-request-client" title="concept-request-client">client</a> is
@@ -2130,6 +2128,10 @@ indicates an attempt to authenticate.
     <a href="#refsSW">[SW]</a>
 
     <p><a href="#read-a-request" title="Read a request">Read <var>request</var></a>.
+
+   <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
+   <a href="#concept-filtered-response" title="concept-filtered-response">filtered response</a>, and to <var>response</var>'s
+   <a href="#concept-internal-response" title="concept-internal-response">internal response</a> otherwise.
 
    <li>
     <p>If one of the following conditions is true, return a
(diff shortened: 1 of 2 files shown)


annevk:
@wanderview, I'd appreciate review.

wanderview:
I don't actually see where Response's url list is set with the URL from a network load.  Is it implied in step 4 of HTTP-network fetch?

> Let response be the result of making an HTTP request over connection using request, following the relevant requirements from HTTP, and waiting until all the headers are transmitted or fetch is being terminated with reason reason. If fetch is being terminated, set response's termination reason to reason.

Since we're making this logic dependent on response url list being set, we should be explicit about where it is it I think.

annevk:
Okay, so the way to solve this would indeed be to set response's url list to a copy of request's url list in HTTP-network fetch. That way only synthetic responses end up with an empty url list in main fetch.

annevk:
@wanderview should be fixed now.

I guess to make it even better I could create some sharing of code between service worker responses and network responses. Not sure where to place that though. Ideas?

wanderview:
LGTM, although it would be nice if github provided a way to see the final document as a rendered html page I can view in the browser.  Reading HTML diffs makes me a bit cross-eyed.

wanderview:
Gecko bug: https://github.com/whatwg/fetch/pull/146

wanderview:
Gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1222008

jakearchibald:
Blink bug https://code.google.com/p/chromium/issues/detail?id=553535

wanderview:
> Blink bug https://code.google.com/p/chromium/issues/detail?id=553535

@mattto @horo-t Has this been implemented in chrome?  This issue still says "assigned", but this glitch test suggests the final Response.url is being synthesized across to the intercepted request:

https://sw-passthrough-redirect.glitch.me/

In firefox I currently get:

```
==> SW fetch() got Response url:https://sw-passthrough-redirect.glitch.me/final.txt redirected:true
==> window fetch() got Response url:https://sw-passthrough-redirect.glitch.me/redirect.txt redirected:false
```

In chrome I get:

```
==> SW fetch() got Response url:https://sw-passthrough-redirect.glitch.me/final.txt redirected:true
==> window fetch() got Response url:https://sw-passthrough-redirect.glitch.me/final.txt redirected:true
```

Note that the window fetch() is seeing "final.txt".

Is this a fetch() only thing right now in chrome, or have you implemented it for all intercepted network requests?

I'm happy if this is implemented in chrome, because I'd like to fix:

https://bugzilla.mozilla.org/show_bug.cgi?id=1222008

wanderview:
I am able to test edge 16 service workers as well, and they also show the final URL.

wanderview:
@jakearchibald pointed me at this which suggests chrome is not using the synthetic `Response.url` for the CSS case:

  https://sw-passthrough-redirect.glitch.me/

So it seems chrome is sometimes using synthetic `Response.url` and sometimes not.

I can't test edge on the CSS case yet because its hitting some error even without a SW.

jakearchibald:
It seems that Edge doesn't update the base URL on redirected with CSS. URLs remain relative to the request URL, not the response URL.

annevk:
I'm happy for you to explore this further here, but please do escalate eventually to either an open (new) issue or web-platform-tests for better visibility.

wanderview:
> I'm happy for you to explore this further here, but please do escalate eventually to either an open (new) issue or web-platform-tests for better visibility.

We're currently implementing this spec change and will fix any WPT tests that need fixing as part of that effort.  I keep asking here in case someone has an objection.

mattto:
Sorry not ignoring this, I just don't recall what we did off-hand and have been meaning to investigate commits/bugs. @horo-t do you remember what we've changed?

wanderview:
No problem.  I was just trying to understand the current situation a bit better in case there were lurking problems.  Tom is making good progress on implementing it in firefox.

wanderview:
Note, we may have a run into a thorny spec question with this change here:

https://bugzilla.mozilla.org/show_bug.cgi?id=1222008#c38

Consider:

```
var xmlDoc = document.implementation.createDocument(null, null, null);
xmlDoc.load('load_cross_origin_xml_document_cors.xml');
// SW intercepts and passes a cross-origin CORS response to respondWith()
```

In this case the outer network request is set up as same-origin (at least we make that restriction for XMLDocument in gecko).  What should we do in the case where the service worker synthesizes a cross-origin CORS Response?  Should we expose the cross-origin URL for the DocumentURI?  That seems somewhat surprising and potentially breaking.

This problem exists for other same-origin requests as well such as Worker() top level scripts, same-origin fetch(), etc.

How would people feel if we rejected the outer network request if you synthesized a cross-origin Response onto a same-origin Request?

Issue #61, opened 2015-06-08
Proposed backward-incompatible change: requiring cloning all requests/responses


domenic:
In https://github.com/yutakahirano/fetch-with-streams/issues/37 @yutakahirano, @wanderview, @tyoshino and I have been going in circles with regard to a pretty big mismatch between how streams work and how Request/Response currently work. We have a proposal that would make everything a lot simpler, but it does require a backward-incompatible change. Hoping to get opinions on whether it's a feasible change. /cc @slightlyoff @jakearchibald 

Basically, the premise is that it's extremely surprising to developers that

```js
var req = new Request(url, { method: "POST" });
fetch(req);
fetch(req);
cache.add(req);
```

works, but

```js
var req = new Request(url, { method: "POST", body: "foo" });
fetch(req);
fetch(req);
cache.add(req);
```

fails on the second fetch, since the body is drained. That is, the latter case requires cloning before use, whereas the former does not.

We propose that:

- Passing a request to fetch or cache.add, or calling .text() etc., sets req.bodyUsed to true.
- If any of those methods gets a request with req.bodyUsed as true, they throw.

In my opinion this helps the story a lot. Right now the rules for when you need to clone are pretty darn arbitrary, and in fact can be a refactoring hazard, as shown above. It also solves a lot of headaches around streams + fetch integration---if you want to dive into the linked thread, you'll see that we're having a really hard time rationalizing bodyUsed and related behavior in terms of streams.

However, there might be people out there already using multi-fetch/cache on empty-body requests/responses. So this might be a non-starter. I suggested adding use counters, but that might take a while. What are others thoughts? I am hopeful that this is a rare pattern, but I might be wrong...

wanderview:
I just want to note there are alternatives here.

For example, streams could track and expose a .drained state.  This is a boolean.  Its cheap.  As far as I can tell it has no practical downsides.  Its just not theoretically pure to the current vision of streams.

Obviously @domenic and I disagree here.  I just want to point it out as an option, though, before we decide to break backward compat.

domenic:
>  Its just not theoretically pure to the current vision of streams.

That's really overstating it. It mutates streams into a completely new data structure. You wouldn't ask someone to add a .drained to arrays, asking if they had started out nonempty but then someone .pop()ed them into emptiness.

wanderview:
> You wouldn't ask someone to add a .drained to arrays, asking if they had started out nonempty but then someone .pop()ed them into emptiness.

[...]

PR #149, opened 2015-10-29
Mark passed Request object as disturbed

yutakahirano:
Let `req` be a Request with a non-null body. When constructing a new Request
from it, it should be marked as disturbed so as not to confuse users. As a
Request with the null body cannot be disturbed, this change sets `req`'s body
to an empty byte stream.

This was discussed at #61. It is not yet decided whether we make it possible to disturb a Request with the null body or setting the body to a non-null stream in such cases, but they are indistinguishable until we expose Request.prototype.body. See https://github.com/yutakahirano/fetch-with-streams/issues/60.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-15-october-2015">Living Standard — Last Updated 15 October 2015</h2>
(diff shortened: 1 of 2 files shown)


yutakahirano:
The HTTP AUTH change looks caused by a mistake made in [the previous commit] (https://github.com/whatwg/fetch/commit/8ce550b53c53e2b6e8376373b94d66b2eeada8f8).

tyoshino:
> The HTTP AUTH change looks caused by a mistake made in the previous commit.

I agree that the main change of 8ce550b53c53e2b6e8376373b94d66b2eeada8f8 does nothing with RFC7235, but the entry should really be included in the reference as for example the `Authorization` header is built in the Fetch Standard, I think.

tyoshino:
Except for the reference change, 652da784588cd78068fec9519fa9e0d2d2bf9002 lgtm

annevk:
You might have to update whatwg/xref.

yutakahirano:
Oops, thank you for pointing it out. Fixed.

PR #173, opened 2015-12-10
fixed an outdated url in HTTP fetch spec

nikkisquared:
I noticed, in substep 5 of step 3 in the HTTP Fetch specification, that the "set response's CSP list" link is outdated, so I've changed it to what seems like the correct url.


annevk:
This link is wrong twice. Would you mind updating your commit?

PR #181, opened 2015-12-15
Fix Response body IDL

yutakahirano:
This change fixes the type of Response.prototype.body.
 - It should not be [SameObject] as clone() will change the attribute.
 - It should be nullable.


annevk:
Thank you!

PR #188, opened 2016-01-05
Update the "determine request's referrer" algorithm link

tyoshino:


annevk:
Thank you. I amended your commit to also update all the other URLs they broke.

PR #193, opened 2016-01-11
Fix #169: ban combination of "no-cors" and "same-origin"

### LINKED ISSUE #169: "no-cors" and credentials mode
Step 4.3 of HTTP fetch reads:

```
Let credentials flag be set if one of
 - request's credentials mode is "include"
 - request's credentials mode is "same-origin" and either the CORS flag is unset or
   response tainting is "opaque" 
```

That second bullet should read "or response tainting is NOT opaque".  The logic on the opaque check is reversed.

annevk:


annevk:
Upon further discussion this is not the approach we're going to take.

PR #194, opened 2016-01-11
Introduce HTTP-redirect fetch

annevk:
See https://github.com/whatwg/html/issues/461 for context. This allows
HTML to invoke a single algorithm to “handle redirects”, a concept it
mentions but does not actually define.

This changes the processing rules for redirect mode “manual”, but since
that is a rather new feature that should be fine.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-29-january-2016">Living Standard — Last Updated 29 January 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-5-february-2016">Living Standard — Last Updated 5 February 2016</h2>
 
 <dl>
  <dt>Participate:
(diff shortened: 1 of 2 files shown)


annevk:
@wanderview, @tyoshino, @jakearchibald, what do you think?

Overview.src.html
```diff
@@ -2369,6 +2293,86 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+ <li><p>If <var>location</var> is null, return <var>response</var>.
+
+ <li><p>If <var>location</var> is failure, return a
+ <span title=concept-network-error>network error</span>.
+ <!-- only Gecko does this; and even that is currently more complicated -->
+
+ <li><p>Let <var>locationURL</var> be the result of
+ <span data-anolis-spec=url title=concept-url-parser>parsing</span> <var>location</var> with
+ <var>request</var>'s <span title=concept-request-current-url>current url</span>.
```
jakearchibald: If we change the document's base URL to be based on the response url, I imagine we'd change this to resolve location against the the response URL too.

annevk: When the service worker creates a redirect, there is no response URL. If the redirect came from the server, this only matters if the service worker stored the opaque response and served it against a different request url later on. Or am I missing something? I suppose foreign fetch might make this more complicated.

In that case though, I guess you're right that you might want to use the URL of the response rather than that of the request, otherwise you could figure out things about the redirect. Intriguing. Probably a good follow up issue.

jakearchibald:
```js
event.respondWith(fetch("//static.example.com/foo/", { redirect: 'manual' }))
```

Where:

* The request URL is `//example.com/bar/`
* Request has a manual redirect mode
* `//static.example.com/foo/` redirects to `/bar/`

This would result in redirecting to `//example.com/bar/` rather than `//static.example.com/bar/` right?

Overview.html
```diff
@@ -2430,7 +2354,87 @@ <h3 id="http-fetch"><span class="secno">5.3 </span>HTTP fetch</h3>
...
+
+ <li><p>Let <var>location</var> be the result of <a href="#concept-header-parse" title="concept-header-parse">parsing</a>
+ `<code title="">Location</code>` in <var>actualResponse</var>'s
+ <a href="#concept-response-header-list" title="concept-response-header-list">header list</a>.
+
+ <li><p>If <var>location</var> is null, return <var>response</var>.
+
+ <li><p>If <var>location</var> is failure, return a
+ <a href="#concept-network-error" title="concept-network-error">network error</a>.
+ <!-- only Gecko does this; and even that is currently more complicated -->
```
jakearchibald: Interesting, what do the other browsers do?

Overview.src.html
annevk: Yeah, that seems like a security vulnerability. Since you can figure out where a cross-origin redirect goes to. So we need to do one of these:

* Resolve against the response URL (but this does not work as a generic solution, see above)
* Check that the request URL is equal to the response URL when response's type is "opaqueredirect".

Or is there a better solution that I'm missing? Either way, this seems like a separate issue from this rewrite, no?

Overview.html
annevk: They allow multiple `Location` headers, but only use the first one.

PR #200, opened 2016-01-21
Replace byte stream with ReadableStream.

yutakahirano:
This change replaces byte streams used in request and response with ReadableStream. This change
also unifies some properties and operations which was defined separately in Request and Response
classes.

The change is based on
https://github.com/yutakahirano/fetch-with-streams/blob/master/body-readable-stream.md.
Depending specs need to be changed as well.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -160,7 +160,6 @@ annotated as one of the following:
  <li><dfn>process request body</dfn>
  <li><dfn>process request end-of-file</dfn>
  <li><dfn>process response</dfn>
- <li><dfn>process response body</dfn>
  <li><dfn>process response end-of-file</dfn>
 </ul>
 
@@ -190,17 +189,18 @@ annotated as one of the following:
 
 <hr>
 
-<p>To <dfn>wait for a <var>response</var></dfn>, wait for either end-of-file to have been
-pushed to <var>response</var>'s <span title=concept-response-body>body</span> or for
-<var>response</var> to have a
-<span title=concept-response-termination-reason>termination reason</span>.
+<p>To <dfn>wait for a <var>response</var></dfn>, if <var>response</var>'s
+<span title=concept-response-body>body</span> is non-null, wait for <var>response</var>'s
+<span title=concept-response-body>body</span>'s <span title=concept-body-stream>stream</span> to be
+<span title=concept-ReadableStream-closed>closed</span> or
+<span title=concept-ReadableStream-errored>errored</span>.
 
 <p>To <dfn>read a <var>request</var></dfn>, if <var>request</var>'s
 <span title=concept-request-body>body</span> is non-null, whenever
 <var>request</var>'s <span title=concept-request-body>body</span> is read from (i.e.
 is transmitted or read by script), increase <var>request</var>'s
 <span title=concept-request-body>body</span>'s
-<span title=concept-body-transmitted>transmitted</span> with the amount of payload body
+<span title=concept-body-transmitted>transmitted bytes</span> with the amount of payload body
 bytes transmitted and then <span>queue a fetch task</span> on <var>request</var> to
 <span>process request body</span> for <var>request</var>.
 <!-- XXX xref "read", "payload body" -->
@@ -466,10 +466,37 @@ steps:
 
 <h4>Bodies</h4>
 
-<p>A <dfn title=concept-body>body</dfn> is a byte stream. It has an associated
-<dfn title=concept-body-transmitted>transmitted</dfn> which is an integer and initially 0,
-<dfn title=concept-body-length>length</dfn> which is an integer and initially 0, and
-<dfn title=concept-body-error-flag>error flag</dfn> which is initially unset.
+<p>A <dfn title=concept-body>body</dfn> consists of:
+
+<ul>
+ <li>
+  <p>A <dfn title=concept-body-stream>stream</dfn> (a
+  <code title=concept-ReadableStream>ReadableStream</code> object).
+
+  <p class="XXX"><a href="https://github.com/whatwg/streams/issues/379">This might become a
+  <code>ReadableByteStream</code> object</a>. While the type might change, the behavior specified
+  will be equivalent since the hypothetical <code title>ReadableByteStream</code> object will have
+  the same members as the <code title=concept-ReadableStream>ReadableStream</code> object has today.
+
+ <li><p>A <dfn title=concept-body-transmitted>transmitted bytes</dfn> (an integer), initially 0.
+
+ <li><p>A <dfn title=concept-body-total-bytes>total bytes</dfn> (an integer), initially 0.
+</ul>
+
+<p>To <dfn title=concept-body-clone>clone</dfn> a <span title=concept-body>body</span>
+<var>body</var>, run these steps:
+
+<ol>
+ <li><p>Let «<var>out1</var>, <var>out2</var>» be the result of
+ <span title=concept-tee-ReadableStream>teeing</span> <var>body</var>'s
+ <span title=concept-body-stream>stream</span>. Rethrow any exceptions.
+
+ <li><p>Set <var>body</var>'s <span title=concept-body-stream>stream</span> to <var>out1</var>.
+
+ <li><p>Return a <span title=concept-body>body</span> whose
+ <span title=concept-body-stream>stream</span> is <var>out2</var> and other members are copied from
+ <var>body</var>.
+</ol>
 
 <p>To <dfn>handle content codings</dfn> given <var>codings</var> and
 <var>bytes</var>, run these substeps:
@@ -902,6 +929,21 @@ whose <span title=concept-request-destination>destination</span> is neither
 <p class=note>See <span data-anolis-spec=sw>handle fetch</span> for usage of these terms.
 <span data-anolis-ref>SW</span>
 
+<p>To <dfn title=concept-request-clone>clone</dfn> a <span title=concept-request>request</span>
+<var>request</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>newRequest</var> be a copy of <var>request</var>, except for its
+ <span title=concept-request-body>body</span>.
+
+ <li><p>If <var>request</var>'s <span title=concept-request-body>body</span> is non-null, set
+ <var>newRequest</var>'s <span title=concept-request-body>body</span> to the result of
+ <span title=concept-body-clone>cloning</span> <var>request</var>'s
+ <span title=concept-request-body>body</span>. Rethrow any exceptions.
+
+ <li><p>Return <var>newRequest</var>.
+</ol>
+
 
 <h4>Responses</h4>
 
@@ -1065,6 +1107,26 @@ introducing new APIs, do not use the
 <span title=concept-internal-response>internal response</span> for internal specification
 algorithms as you will leak information.
 
+<p>To <dfn title=concept-response-clone>clone</dfn> a <span title=concept-response>response</span>
+<var>response</var>, run these steps:
+
+<ol>
+ <li><p>If <var>response</var> is a <span title=concept-filtered-response>filtered response</span>,
+ return a new identical filtered response whose <span title=concept-internal-response>internal
+ response</span> is a <span title=concept-response-clone>clone</span> of <var>response</var>'s
+ <span title=concept-internal-response>internal response</span>. Rethrow any exceptions.
+
+ <li><p>Let <var>newResponse</var> be a copy of <var>response</var>, except for its
+ <span title=concept-response-body>body</span>.
+
+ <li><p>If <var>response</var>'s <span title=concept-response-body>body</span> is non-null, set
+ <var>newResponse</var>'s <span title=concept-response-body>body</span> to the result of
+ <span title=concept-body-clone>cloning</span> <var>response</var>'s
+ <span title=concept-response-body>body</span>. Rethrow any exceptions.
+
+ <li><p>Return <var>newResponse</var>.
+</ol>
+
 
 <h3>Authentication entries</h3>
 
@@ -1192,6 +1254,15 @@ define common operations for ReadableStream. <span data-anolis-ref>STREAMS</span
  Rethrow any exceptions.
 </ol>
 
+<p>To <dfn title=concept-error-ReadableStream>error</dfn> a
+<code title=concept-ReadableStream>ReadableStream</code> object <var>stream</var> with given
+<var>reason</var>, run these steps:
+
+<ol>
+ <li><p>Call <code data-anolis-spec=streams>ErrorReadableStream</code>(<var>stream</var>,
+ <var>reason</var>). Rethrow any exceptions.
+</ol>
+
 <p>To <dfn title=concept-construct-ReadableStream>construct</dfn> a
 <code title=concept-ReadableStream>ReadableStream</code> object with given <var>strategy</var>,
 <var>pull</var> action and <var>cancel</var> action, all of which are optional, run these steps:
@@ -1234,23 +1305,7 @@ run these steps:
  <li>Return <var>stream</var>.
 </ol>
 
-<p>To <dfn title=concept-construct-ReadableStream-with-byte-stream>construct</dfn> a
-<code title=concept-ReadableStream>ReadableStream</code> object, given a byte stream
-<var>byteStream</var>, run these steps:
-
-<p class=XXX>This operation is a workaround and will be deleted when we stop using byte streams.
-
-<ol>
- <li><p>Let <var>stream</var> be a <code title=concept-ReadableStream>ReadableStream</code> object
- such that consuming it will result in chunks each of which is a <code title>Uint8Array</code>
- object wrapping an <code title>ArrayBuffer</code> and the result of concatenating them all will be
- equivalent to the byte sequence which would be read from <var>byteStream</var>. Rethrow any
- exceptions.
-
- <li><p>Return <var>stream</var>.
-</ol>
-
-<p>To <dfn title=concept-get-reader>get</dfn> a reader from a
+<p>To <dfn title=concept-get-reader>get a reader</dfn> from a
 <code title=concept-ReadableStream>ReadableStream</code> object <var>stream</var>, run these steps:
 
 <ol>
@@ -1261,7 +1316,7 @@ run these steps:
  <li><p>Return <var>reader</var>.
 </ol>
 
-<p>To <dfn title=concept-read-all-bytes-from-ReadableStream>read</dfn> all bytes from a
+<p>To <dfn title=concept-read-all-bytes-from-ReadableStream>read all bytes</dfn> from a
 <code title=concept-ReadableStream>ReadableStream</code> object with <var>reader</var>, run these
 steps:
 
@@ -1585,8 +1640,7 @@ response <span title=concept-header>header</span> can be used to require checkin
  <p>That is, it either returns a <span title=concept-response>response</span> if
  <span title=concept-request>request</span>'s <span>synchronous flag</span> is set, or it
  <span title="queue a fetch task">queues tasks</span> annotated
- <span>process response</span>, <span>process response body</span>, and
- <span>process response end-of-file</span> for the
+ <span>process response</span>, and <span>process response end-of-file</span> for the
  <span title=concept-response>response</span>.
 
  <p>To capture uploads, if <span title=concept-request>request</span>'s
@@ -1872,7 +1926,7 @@ redirects.
   `<code title>HEAD</code>` or `<code title>CONNECT</code>`, or <var>internalResponse</var>'s
   <span title=concept-response-status>status</span> is a <span>null body status</span>,
   set <var>internalResponse</var>'s <span title=concept-response-body>body</span> to
-  null and disregard any pushing toward it (if any).
+  null and disregard any enqueuing toward it (if any).
 
   <p class=note>This standardizes the error handling for servers that violate HTTP.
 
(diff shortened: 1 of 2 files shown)


Overview.src.html
```diff
@@ -442,10 +442,38 @@ <h4 id=terminology-headers>Headers</h4>
...
+<dfn title=concept-body-stream>stream</dfn> (a
+<code title=concept-ReadableStream>ReadableStream</code> object).
+
+<p class="XXX"><a href="https://github.com/whatwg/streams/issues/379">This might become a
+<code>ReadableByteStream</code> object</a>. While the type might change, the behavior specified will
+be equivalent since the hypothetical <code title>ReadableByteStream</code> object will have the same
+members as the <code title=concept-ReadableStream>ReadableStream</code> object has today.
+
+<p>A <span title=concept-body>body</span> has an associated
+<dfn title=concept-body-transmitted>transmitted</dfn> (an integer). Unless stated it is 0.
```
annevk: "Unless stated otherwise it is 0". Same below.

Overview.src.html
```diff
@@ -853,6 +881,21 @@ <h4 id=terminology-headers>Headers</h4>
...
 <span data-anolis-ref>SW</span>
 
+<p>To <dfn title=concept-request-clone>clone</dfn> a <span title=concept-request>request</span>
+<var>request</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>newRequest</var> be a copy of <var>request</var>, except for its
+ <span title=concept-request-body>body</span>.
+
+ <li><p>If <var>request</var>'s <span title=concept-request-body>body</span> is not null, set
```
annevk: I prefer "non-null" over "not null".

Overview.src.html
```diff
@@ -1874,39 +1929,11 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
-
-   <li><p><span>Queue a fetch-done task</span> using <var>request</var> and
-   <var>response</var>.
-  </ol>
-
- <li>
-  <p>Otherwise, if <var>internalResponse</var>'s
-  <span title=concept-response-body>body</span> is non-null, run these substeps:
-
+  <p>Run these substeps <span data-anolis-spec=html>in parallel</span>:
```
annevk: At this point we are already running in parallel so I'm not sure this is needed.

Overview.src.html
```diff
@@ -2663,6 +2689,29 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   <p><span title="Read a request">Read <var>request</var></span>.
 
  <li>
+  <p>Let <var>strategy</var> be an object. The user agent may choose any object.
+  <p class="note no-backref"><var>strategy</var> is used to control the queuing strategy of
```
annevk: Newline before the second `<p`.

Overview.src.html
```diff
@@ -2760,9 +2808,14 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
      <span title="handle content codings">handling content codings</span> given
      <var>codings</var> and <var>bytes</var>.
 
-     <li><p>Push <var>bytes</var> to <var>response</var>'s
-     <span title=concept-response-body>body</span>.
-     <!-- XXX xref payload body / streams -->
+     <li>
+      <p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code>
+      object wrapping an <code>ArrayBuffer</code> containing <var>bytes</var> to <var>stream</var>.
+      <p class="note no-backref">This operation will not throw an exception.
```
annevk: That newline again 😊

Overview.src.html
```diff
@@ -3458,15 +3535,22 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
-<p>Objects implementing the <code>Body</code> mixin gain a
-<dfn title=concept-Body-MIME-type>MIME type</dfn> (initially the empty byte sequence).
+<p>Objects implementing the <code>Body</code> mixin gain an associated
+<dfn title=concept-Body-body>body</dfn> (null or a <span title=concept-body>body</span>) and
+a <dfn title=concept-Body-MIME-type>MIME type</dfn> (initially the empty byte sequence).
+
+<p>An object implementing the <code>Body</code> mixin is said to be
+<dfn title=concept-Body-disturbed>disturbed</dfn> if <span title=concept-Body-body>body</span> is
+not null and its <span title=concept-body-stream>stream</span> is
```
annevk: non-null

annevk:
This looks really great @yutakahirano. Apart from the above nits, I think whenever we allocate a `Uint8Array` and corresponding buffer, we need to account for that throwing. That could be a follow up if you want since I know we don't do it now either.

I would also greatly appreciate it if @domenic and @wanderview could have a look at this PR (search for `Overview.src.html` to get the diff you want to read).

Overview.src.html

Overview.src.html
yutakahirano: Fixed.

Overview.src.html
yutakahirano: You're right. Fixed.

Overview.src.html

Overview.src.html

Overview.src.html

yutakahirano:
Regarding OOM handling, I would create a separate PR.

yutakahirano:
I squashed the commits.
Any other comments?

annevk:
@domenic said he would take a look, but I think he's been in meetings all week. I suspect he can review this on Monday. Would that be alright with you?

yutakahirano:
Yes, thanks!

Overview.src.html
```diff
@@ -442,10 +442,38 @@ <h4 id=terminology-headers>Headers</h4>
 
 <h4>Bodies</h4>
 
-<p>A <dfn title=concept-body>body</dfn> is a byte stream. It has an associated
-<dfn title=concept-body-transmitted>transmitted</dfn> which is an integer and initially 0,
-<dfn title=concept-body-length>length</dfn> which is an integer and initially 0, and
-<dfn title=concept-body-error-flag>error flag</dfn> which is initially unset.
+<p>A <dfn title=concept-body>body</dfn> consists of a
```
domenic: I think this would be slightly clearer like:

A **body** consists of:

- A **stream** (a ReadableStream object), initially null (?).
- A **transmitted** (an integer), initially 0.
- A **total bytes** (an integer), initially 0.

And maybe s/transmitted/transmitted bytes/g?

Overview.src.html
```diff
@@ -160,7 +160,6 @@ <h2 class="short">Conformance</h2>
  <li><dfn>process request body</dfn>
  <li><dfn>process request end-of-file</dfn>
  <li><dfn>process response</dfn>
- <li><dfn>process response body</dfn>
```
domenic: I checked and HTML does not use this, but maybe other people do? I guess it's probably OK to remove.

Overview.src.html
```diff
@@ -442,10 +442,38 @@ <h4 id=terminology-headers>Headers</h4>
...
+<p class="XXX"><a href="https://github.com/whatwg/streams/issues/379">This might become a
+<code>ReadableByteStream</code> object</a>. While the type might change, the behavior specified will
+be equivalent since the hypothetical <code title>ReadableByteStream</code> object will have the same
+members as the <code title=concept-ReadableStream>ReadableStream</code> object has today.
+
+<p>A <span title=concept-body>body</span> has an associated
+<dfn title=concept-body-transmitted>transmitted</dfn> (an integer). Unless stated otherwise it is 0.
+
+<p>A <span title=concept-body>body</span> has an associated
+<dfn title=concept-body-total-bytes>total bytes</dfn> (an integer). Unless stated otherwise it is 0.
```
domenic: maybe keep the concept-body-length ID to not break links? @annevk would need to make the call on that.

Overview.src.html
```diff
@@ -2663,6 +2685,30 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   <p><span title="Read a request">Read <var>request</var></span>.
 
  <li>
+  <p>Let <var>strategy</var> be an object. The user agent may choose any object.
```
domenic: Maybe in the future we should define "valid queuing strategy". Right now this could be taken to allow any object, including ones that cause the constructor to throw. For now it seems fine as-is.

domenic: Filed https://github.com/whatwg/streams/issues/427 to track this in the stream spec.

Overview.src.html
```diff
@@ -2663,6 +2685,30 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+  <p class="note no-backref"><var>strategy</var> is used to control the queuing strategy of
+  <var>stream</var> constructed below.
+
+ <li><p>Let <var>pull</var> be an action that <span title=concept-fetch-resume>resumes</span> the
```
domenic: I would do "function" instead of "action" here and below.

domenic: Oh, I see, the "function" part is taken care of by the "construct a ReadableStream" algorithm. Nevermind!

Overview.src.html
```diff
@@ -2760,9 +2805,15 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+      object wrapping an <code>ArrayBuffer</code> containing <var>bytes</var> to <var>stream</var>.
+
+      <p class="note no-backref">This operation will not throw an exception.
+
+     <li><p>If <var>stream</var> doesn't <span title=concept-ReadableStream-need-more-data>need more
+     data</span> and <var>request</var>'s <span>synchronous flag</span> is unset, ask the user agent
+     to <span title=concept-fetch-suspend>suspend</span> the ongoing fetch.
     </ol>
 
     <p class="note no-backref">This means that the `<code title>Content-Length</code>`
```
domenic: This note now looks a bit like it's referring to step 5... I am not sure there is a good fix.

I also don't really understand the note. Why is Content-Length "no longer" reliable for responses? Was it ever reliable?

Overview.src.html
annevk: It's fine to break that, it's relatively recent.

Overview.src.html
annevk: I suspect the idea is to revamp XMLHttpRequest to use streams directly rather than this indirection.

Overview.src.html
```diff
@@ -3391,21 +3456,23 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   <dl class=switch>
    <dt><code data-anolis-spec=fileapi>Blob</code>
    <dd>
-    <p>Push a copy of <var>object</var>'s contents to <var>stream</var>.
+    <p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code> object
```
domenic: This is normatively requiring a single chunk in the stream. Do we want to allow more flexibility perhaps? This concern applies to everything but BufferSource. Maybe @wanderview has opinions?

Overview.src.html
```diff
@@ -3433,12 +3501,19 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 
    <dt><code title>USVString</code>
    <dd>
-    <p>Push the result of running <span data-anolis-spec=encoding>utf-8 encode</span> on
-    <var>object</var> to <var>stream</var>.
+    <p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code> object
+    wrapping an <code>ArrayBuffer</code> containing the result of running
+    <span data-anolis-spec=encoding>utf-8 encode</span> on <var>object</var> to <var>stream</var>.
```
domenic: Can this ever fail?

annevk: Yeah, @yutakahirano said above he will create a new PR for handling `ArrayBuffer` OOM.

Overview.src.html
```diff
@@ -3535,6 +3617,35 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ <li><p>If this object is <span title=concept-Body-disturbed>disturbed</span>
+ or <span title=concept-Body-locked>locked</span>, return a new promise rejected with a
+ <code>TypeError</code>.
+
+ <li><p>Let <var>stream</var> be <span title=concept-Body-body>body</span>'s
+ <span title=concept-body-stream>stream</span> if <span title=concept-Body-body>body</span> is
+ non-null, or an <span title=concept-empty-ReadableStream>empty</span>
+ <code title=concept-ReadableStream>ReadableStream</code> object otherwise.
+
+ <li><p>Let <var>reader</var> be the result of <span title=concept-get-reader>getting</span>
```
domenic: Editorial nit: I would link "getting a reader" not "getting"

Overview.src.html
```diff
@@ -3535,6 +3617,35 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ a reader from <var>stream</var>. If that threw an exception, return a new promise rejected
+ with that exception.
+
+ <li><p>Let <var>promise</var> be the result of
+ <span title=concept-read-all-bytes-from-ReadableStream>reading</span> all bytes from
```
domenic: Similarly I would link "reading all bytes"

Overview.src.html
```diff
@@ -3535,6 +3617,35 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ <var>stream</var> with <var>reader</var>.
+
+ <li><p>Return the result of transforming <var>promise</var> by a fulfillment handler that
+ returns the result of the <span title=concept-body-package-data>package data</span> algorithm
+ with its first argument, <var>type</var> and <span title=concept-body-mime-type>MIME type</span>.
```
domenic: Maybe "and the body's MIME type"?

Overview.src.html
```diff
@@ -3944,11 +4021,22 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <code title=concept-ReadableStream>ReadableStream</code> object.
+
    <li><p>Set <var>input</var>'s <span title=concept-Request-request>request</span>'s
-   <span title=concept-request-body>body</span> to an empty byte stream.
+   <span title=concept-request-body>body</span> to a new <span title=concept-body>body</span> whose
+   <span title=concept-body-stream>stream</span> is <var>stream</var>.
 
-   <li><p>Set <var>input</var>'s
-   <span title=concept-Request-disturbed-flag>disturbed flag</span>.
+   <li>
```
domenic: This step confuses me. If input has a request body, then we override it with an empty stream, then immediately get a reader for the empty stream and read all the bytes in it, then never release the reader?

Shouldn't it instead be the case that we consume the request body, or transfer it, or tee it, or something? I forget what behavior we landed on, but this behavior seems wrong...

Overview.src.html
domenic: I meant the encoding actually.

Overview.src.html

Body's stream cannot be null and has no initial value. We always specify its stream explicitly when we create a body.

Overview.src.html

Overview.src.html

Overview.src.html
yutakahirano: Fixed as "this object's MIME type".

Overview.src.html
yutakahirano: The operations are just to make request disturbed and locked. See https://github.com/whatwg/fetch/commit/e7d718ac577b12f190e77370db42b68745d13eac and https://github.com/yutakahirano/fetch-with-streams/issues/60.

I think using pipeTo is the right way, but it's not distinguishable at this moment because the Request class doesn't have `body` property (yet).

Overview.src.html
yutakahirano: I think utf8-encoding never fails: it uses replacement characters when seeing an error.

Overview.src.html
yutakahirano: @annevk What do you think?

Overview.src.html
domenic: I still don't really understand. How do the bytes get transferred from the input's body to this new request's body?

yutakahirano:
```
2. Let temporaryBody be input's request's body if input is a Request object, and null otherwise.
...
33. Set r's request's body to temporaryBody.
```

domenic: Oh, wow, all the way up in step 2! Thanks, that makes sense. I would suggest some slight refactoring then to make this clearer:

- Move step 2 to right before step 31.
- Rename _temporaryBody_ to _rBody_.
- In step 35, rename _stream_ to _dummyStream_.
- Add a note to step 35. Something like:

   > These steps are meant to produce the observable equivalent of "piping" _input_'s [body]'s [stream] into _r_. That is, _r_ is left with a body stream that is disturbed, locked, and empty, while _r_'s body's stream is now equal to what used to be _input_'s.
   >
   > This step will likely change when [`Request`] gets a public `body` property, since we do not want the new [`Request`]'s `body` to be observably equal to _input_'s, and we do not want to observably change _input_'s `body property.

I agree this is hard to spec before we actually have identity transform streams or something similar in place.

Overview.src.html
annevk: That note is there for "handling content codings". It means to say that `Content-Length` is no longer reliable even if your server is configured correctly.

(The other note about not throwing an exception seems inaccurate given that `ArrayBuffer` can throw on allocation.)

Overview.src.html
annevk: Ah yes, that cannot fail. The utf-8 encoder accepts all Unicode scalar values. There are no errors. (Implementations might have error situations that in the specification are handled through the USVString annotation. Depends on how you implement things.)

Overview.src.html

I would use _inputBody_ instead of _rBodyStream_ to differentiate body and ReadableStream.

Overview.src.html
annevk: @yutakahirano is that the plan?

Overview.src.html
```diff
@@ -2760,9 +2804,15 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
      <var>codings</var> and <var>bytes</var>.
 
-     <li><p>Push <var>bytes</var> to <var>response</var>'s
-     <span title=concept-response-body>body</span>.
-     <!-- XXX xref payload body / streams -->
+     <li>
+      <p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code>
+      object wrapping an <code>ArrayBuffer</code> containing <var>bytes</var> to <var>stream</var>.
+
+      <p class="note no-backref">This operation will not throw an exception.
```
annevk: Since this is false, we should probably not add this note.

Overview.src.html
annevk: I think we should move this note up directly under the "handling content codings" step. That will make it a little clearer. And maybe reword the note to say "This makes the Content-Length header unreliable to the extent that it was reliable to begin with." to address the concern from @domenic.

Overview.src.html
yutakahirano: Yes.

Overview.src.html
yutakahirano: Fixed.

Overview.src.html
yutakahirano: Fixed.

Overview.src.html
annevk: Thinking about this we should probably allow more flexibility here. I suspect browsers want to stream reading from the disk. (Same for data URLs.)

What do you think @yutakahirano?

Overview.src.html
annevk: Just to be clear, are you planning on providing a PR for XMLHttpRequest?

yutakahirano: Yes, based on https://github.com/yutakahirano/fetch-with-streams/blob/master/body-readable-stream.md#xhr.

Overview.src.html
wanderview: Yes, I would say we want more flexibility to stream a file backed blob into memory.  Consider if someone is trying to stream a large zip off disk and pick out a single compressed item.  Requiring the entire zip to be in memory defeats much of the purpose of streams.

yutakahirano: 1. Let _buffers_ be a list of ArrayBuffer objects so that the concatenation of their contents equals to _object_'s contents.
2. For each _buffer_ in _buffers_:
 - Enqueue a Uint8Array object wrapping _buffer_.

What do you think about this?
Chrome loads a blob asynchronously: Is "in parallel" needed to allow such implementation?

wanderview: Can we say something like "convert blob to a stream and equivalent to pipeTo()"?

domenic: pipeTo what though? We got rid of the idea of a writable stream for requests.

I think the way to do this is to be vague. It also does need to be done in parallel. Hmm. Something like:

In parallel, enqueue a sequence of `Uint8Array` objects representing _object_'s contents, as they become available while reading the blob.

yutakahirano: Read _object_ in parallel:
 - Whenever one or more bytes are read, let _bytes_be the read and enqueue a Uint8Array object wrapping an ArrayBuffer containing _bytes_ to _stream_.
  - If an exception is thrown, error _stream_ and cancel reading.
 - When the reading is done, close _stream_.

"Read _object_" is not defined for FormData, but is it OK?

domenic: missing word "let bytes be the **bytes** read", I think. Maybe "If an exception is thrown" should be "If an exception is thrown creating the ArrayBuffer".

But overall that sounds good to me. It's pretty clear what the intent is.

annevk:
Wit those nits fixed I think this is ready to land. I'd appreciate if you could then fix XMLHttpRequest so that everything still makes some sense. If there's other things we'll just fix them as they come. Getting more eyes on all this Streams language is useful.

annevk:
@yutakahirano I'm happy to merge this by the way. If you want to do it, please see the README for instructions. You're probably already familiar with it from Streams, we basically prefer a clean history with single commits for a change.

annevk:
Unless someone tells me otherwise I plan to land this on Monday.

yutakahirano:
AI:
 - Fix XHR (@yutakahirano)
 - Fix xref (@yutakahirano)
 - Fix [Blob read operation](https://w3c.github.io/FileAPI/#readOperationSection)
 - Fix ServiceWorker
 - Allocation error handling (@yutakahirano)

yutakahirano:
Thank you very much, reviewers!

annevk:
What does AI mean? Something similar to TODO?

Thank your for working on this. It's greatly appreciated.

yutakahirano:
Yes, I meant TODO.

Issue #93, opened 2015-07-23
Redesign request's context and context frame type features


annevk:
One that returns one of "navigation", "worker", or "resource".

See https://github.com/slightlyoff/ServiceWorker/issues/673 for earlier discussion.

Tentative proposal is `contextSet` since I'd like to avoid both "group" and "type" in the name. Any better ideas?

wanderview:
Do we really need a new enum for this?  I think I would prefer something like:

  if (req.navigation)

It would also sidestep the hassle of naming a collection of contexts, etc.

annevk:
I think you're correct. `isNavigation` is fine. That way we could have `isClient` down the line too, or something like `isImage`. Without "is" seems problematic, since we might get `request.client` and who knows what other names we need there for variables.

annevk:
An alternative that @sicking suggested is to group all navigation-like contexts into a single one. Since there's not much point distinguishing a hyperlink from setting location.href.

I believe the main case is `<form>` for CSP, but I would need to check.

Also, if we go down that route https://code.google.com/p/chromium/issues/detail?id=455116 needs to be notified.

annevk:
So if we'd have "navigate", "internal" from https://fetch.spec.whatwg.org/#navigation-request-context would be restricted to non-navigation fetches. We'd need a "context element" to handle form/object/embed for CSP, but that does not need to be exposed.

I think @mikewest might slightly prefer this model too. Anyone else opinions?

ehsan:
I have been implementing the various RequestContext types in Gecko, and based on my experience, I think we need to pause and agree on what RequestContext is supposed to represent first.

Currently, RequestContext represents values on three axes:

* What a specific thing that is being loaded from the network is (for example: "font")
* How a specific thing is being loaded (for example: "hyperlink")
* Where (as in, what specific tag, for example) a request is coming from (for example: "frame" vs. "iframe")

This can lead to a number of difficult questions to answer.  For example: "How should we represent an audio file being downloaded from a <video> tag?", "How should we represent a hyperlink click that results in a download inside an iframe?" and so on.

As far as I can tell, there are two things that make these questions difficult to answer:

* It is not clear which one of these axes a specific RequestContext value appears on (for example, the question about loading audio inside a <video> tag.)
* It is not clear what specific value should be used in specific cases, since the respective specs usually are not based on top of Fetch.

I think it may make more sense to split the first two axes into two completely separate enums.  The third axis seems not too valuable to support, as far as I can tell.

[...]

PR #211, opened 2016-02-02
expand Request.destination with more granular values

igrigorik:
As per [discussion on IRC](http://logs.glob.uno/?c=freenode%23whatwg&s=2+Feb+2016&e=2+Feb+2016#c983589), I'm proposing that we break apart and expand the current "unknown" and "subresource" values for [#concept-request-destination](https://fetch.spec.whatwg.org/#concept-request-destination). This would allow us to address the Preload use case (see https://github.com/w3c/preload/issues/55) by remapping it to use `request.destination` instead of `request.initiator`, and I believe it's consistent with what we intended in https://github.com/whatwg/fetch/issues/93. 

/cc @sicking @ehsan @mikewest @yoavweiss

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -566,7 +566,6 @@ not have to explicitly set <span title=concept-request>request</span>'s
 <dfn title=concept-request-initiator>initiator</dfn>, which is one of
 the empty string,
 "<code title>download</code>",
-"<code title>fetch</code>",
 "<code title>imageset</code>",
 "<code title>manifest</code>", and
 "<code title>xslt</code>". Unless stated otherwise it is the empty string.
@@ -592,11 +591,19 @@ the empty string,
 <dfn title=concept-request-destination>destination</dfn>, which is one of
 the empty string,
 "<code title>document</code>",
+"<code title>embed</code>",
+"<code title>font</code>",
+"<code title>image</code>",
+"<code title>object</code>",
+"<code title>manifest</code>",
+"<code title>media</code>",
+"<code title>report</code>",
 "<code title>serviceworker</code>",
+"<code title>script</code>",
 "<code title>sharedworker</code>",
-"<code title>subresource</code>",
-"<code title>unknown</code>", and
-"<code title>worker</code>".
+"<code title>style</code>",
+"<code title>worker</code>", and
+"<code title>xslt</code>".
 
 <div class=note>
  <p>The following table illustrates the relationship between a
@@ -613,43 +620,52 @@ the empty string,
    <th>CSP directive
    <th>Features
   <tr>
-   <td rowspan=14>""
-   <td rowspan=4>""
-   <td>""
+   <td rowspan=16>""
+   <td rowspan=6>""
+   <td>"<code title>report</code>"
    <td rowspan=2>?
    <td>CSP, NEL reports.
   <tr>
    <td>"<code title>document</code>"
    <td>HTML's navigate algorithm.
   <tr>
-   <td>"<code title>subresource</code>"
+   <td>"<code title>document</code>"
+   <td><code title>child-src</code>
+   <td>HTML's <code>&lt;iframe></code> and <code>&lt;frame></code>
+  <tr>
+   <td>""
    <td><code title>connect-src</code>
    <td><code>navigator.sendBeacon()</code>, <code>EventSource</code>,
-   HTML's <code>ping=""</code>, <code>XMLHttpRequest</code>
+   HTML's <code>ping=""</code>, <code title=dom-global-fetch>fetch()</code>,
+   <code>XMLHttpRequest</code>, Cache API?
   <tr>
-   <td>"<code title>unknown</code>"
+   <td>"<code title>object</code>"
    <td><code title>object-src</code>
-   <td>HTML's <code>&lt;embed></code> and <code>&lt;object></code>
+   <td>HTML's <code>&lt;object></code>
+  <tr>
+   <td>"<code title>embed</code>"
+   <td><code title>object-src</code>
+   <td>HTML's <code>&lt;embed></code>
   <tr>
    <td>"<code title>audio</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>media</code>"
    <td><code title>media-src</code>
    <td>HTML's <code>&lt;audio></code>
   <tr>
    <td>"<code title>font</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>font</code>"
    <td><code title>font-src</code>
    <td>CSS' <code>@font-face</code>
   <tr>
    <td>"<code title>image</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>image</code>"
    <td><code title>img-src</code>
    <td>HTML's <code>&lt;img src></code>, <code title>/favicon.ico</code> resource,
    SVG's <code>&lt;image></code>, CSS' <code>background-image</code>, CSS'
    <code>cursor</code>, CSS' <code>list-style-image</code>, &hellip;
   <tr>
    <td rowspan=4>"<code title>script</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>script</code>"
    <td><code title>script-src</code>
    <td>HTML's <code>&lt;script></code>, <code>importScripts()</code>
   <tr>
@@ -666,17 +682,17 @@ the empty string,
    <td><code>Worker</code>
   <tr>
    <td>"<code title>style</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>style</code>"
    <td><code title>style-src</code>
    <td>HTML's <code>&lt;link rel=stylesheet></code>, CSS' <code>@import</code>
   <tr>
    <td>"<code title>track</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>media</code>"
    <td><code title>media-src</code>
    <td>HTML's <code>&lt;track></code>
   <tr>
    <td>"<code title>video</code>"
-   <td>"<code title>subresource</code>"
+   <td>"<code title>media</code>"
    <td><code title>media-src</code>
    <td>HTML's <code>&lt;video></code> element
   <tr>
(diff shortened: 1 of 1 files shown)


annevk:
@wanderview, @jungkees, @jakearchibald should probably review this from a service worker perspective, which uses "subresource" I think. Fetch could define a "is a subresource" handle of sorts if that's useful.

wanderview:
Looks reasonable to me, but I defer to @ehsan on this one.

yoavweiss:
Looks good from the `preload` perspective. Will ES6 modules be fetched using the same rules as subresource scripts? Or will they have a separate CSP policy, etc? If it's the latter case, we probably want to address that as well.

ehsan:
I think expanding "unknown" and "subresource" is fine, but I would like to see a more concrete description of what the new proposed values are.

Assuming that the suggestion is to add something like [this](https://docs.google.com/spreadsheets/d/1YXoso_Vm0tCuGjFxNmD70UlQd87PKIgGBM97DG9E2L0/edit#gid=0), here are some problems I can think of:

* For "audio" and "video", the distinction seems arbitrary given that you can load an audio file in a <video> element and vice versa (Firefox implements this, Chrome doesn't last I checked.)
* Using "script" for `importScripts()` but "worker" for top-level worker scripts seems inconsistent.  Can we use "worker" for the latter?
* Not sure what the rationale of using "script" for XSLT is...

(Note that I will be on vacation between Feb 10-20.)

igrigorik:
> For "audio" and "video", the distinction seems arbitrary given that you can load an audio file in a element and vice versa (Firefox implements this, Chrome doesn't last I checked.)

We can collapse `audio`, `video`, and `track` into a single entry: `media`. WDYT?

> Using "script" for importScripts() but "worker" for top-level worker scripts seems inconsistent. Can we use "worker" for the latter?

I think this continuation of https://github.com/w3c/webappsec-csp/issues/31. If that bug gets resolved in favor of applying child-src then yeah we can eliminate "worker" here as well.

> Not sure what the rationale of using "script" for XSLT is...

Don't have full history on this either. @annevk @mikewest?

yoavweiss:
> We can collapse audio, video, and track into a single entry: media. WDYT?

Inside Blink there's distinction between `audio` & `video` (both considered media resource types) and `track` which is considered `TextTrack`.

So merging `audio` and `video` into `media` would be easy and would make things simpler. Merging `track` into `media` is probably also possible.

annevk:
XSLT can execute scripts, that's probably why CSP uses `script-src`, but we could still have a distinct type here that ends up mapping to `script-src`.

igrigorik:
OK, updated the PR: merged audio+video+track under `media` destination.

@annevk by distinct type do you mean request.type, or destination? If its destination, "xslt"?

annevk:
I meant distinct destination enum value. Sorry for the confusion. "xslt" probably works. All depends on how granular we want to make this.

igrigorik:
@annevk sgtm, added. Anything else we need to address?

ehsan:
This looks good to me.  Thanks, @igrigorik!

Overview.src.html
```diff
@@ -592,11 +592,21 @@ <h4 id=terminology-headers>Headers</h4>
...
+"<code title>fetch</code>",
+"<code title>font</code>",
+"<code title>image</code>",
+"<code title>object</code>",
+"<code title>manifest</code>",
+"<code title>media</code>",
 "<code title>serviceworker</code>",
+"<code title>script</code>",
+"<code title>style</code>",
 "<code title>sharedworker</code>",
```
annevk: "sharedworker" would should appear before "style", if we want to keep this alphabetical.

Overview.src.html
```diff
@@ -592,11 +592,21 @@ <h4 id=terminology-headers>Headers</h4>
...
 "<code title>subresource</code>",
-"<code title>unknown</code>", and
-"<code title>worker</code>".
+"<code title>unknown</code>",
```
annevk: What is still "unknown" in this brave new world?

annevk:
I would love for @mikewest to review this. I think this should make parts of CSP and maybe Mixed Content simpler.

igrigorik:
@annevk updated, anything else? Sounds like the last bit is for @mikewest to review?

Overview.src.html
```diff
@@ -622,34 +631,42 @@ <h4 id=terminology-headers>Headers</h4>
    <td>"<code title>document</code>"
    <td>HTML's navigate algorithm.
   <tr>
+   <td>"<code title>document</code>"
+   <td><code title>child-src</code>
+   <td>HTML's <code>&lt;iframe></code> and <code>&lt;frame></code>
+  <tr>
    <td>"<code title>subresource</code>"
    <td><code title>connect-src</code>
    <td><code>navigator.sendBeacon()</code>, <code>EventSource</code>,
```
annevk: Does it still make sense to call this subresource?

Overview.src.html
```diff
@@ -688,25 +705,25 @@ <h4 id=terminology-headers>Headers</h4>
   <tr>
    <td>"<code title>fetch</code>"
    <td>""
-   <td>"<code title>subresource</code>"
+   <td>"<code title>fetch</code>"
    <td><code title>connect-src</code>
-   <td><code title=dom-global-fetch>fetch()</code>, Cache API?
+   <td><code title=dom-global-fetch>fetch()</code>, <code>XMLHttpRequest</code>, Cache API?
```
annevk: Why did XMLHttpRequest move here? And why is "fetch" a destination? Should we not group this with the other connect-src things?

Overview.src.html
```diff
@@ -889,11 +906,10 @@ <h4 id=terminology-headers>Headers</h4>
 <p>A <dfn>potential-navigation-or-subresource request</dfn> is a
 <span title=concept-request>request</span> whose
 <span title=concept-request-destination>destination</span> is
-"<code title>unknown</code>".
+"<code title>object</code>" or "<code title>embed</code>".
 
 <p>A <dfn>non-subresource request</dfn> is a <span title=concept-request>request</span>
-whose <span title=concept-request-destination>destination</span> is neither
-"<code title>subresource</code>" nor "<code title>unknown</code>".
+whose <span title=concept-request-destination>destination</span> is "<code title>document</code>", "<code title>serviceworker</code>", "<code title>sharedworker</code>", "<code title>worker</code>", or the empty string.
```
annevk: I use 100 column forced wrap (keeping inline elements on a single line). I can take care of that I suppose while merging, but you're welcome to do it too.

Overview.src.html
igrigorik: Ah, I think I moved it while originally shuffling the values, and in my head fetch() and XHR are logically grouped together. We _can_ merge this row with the other connect-src one, modulo one thing: this row has its initiator set to "fetch" whereas the other is the empty string.. do we have anything depending on this?

Overview.src.html
igrigorik: Perhaps not, do you have any suggestions?

Overview.src.html
annevk: The only thing that's different is that "fetch" does not get an Accept-Language header by default. We did this to give developers more control. Originally we also did that for Accept, but apparently that broke too many things.

annevk: If we give up on that they'd be equivalent. @domenic?

Overview.src.html
annevk: Either "fetch" (and fetch() can use that too) or the empty string, but if we do the empty string perhaps reports need to get their own destination and/or type?

igrigorik: FWIW, empty string would map cleanly for the use case in https://github.com/w3c/preload/issues/55 -- empty `as` is same as calling fetch(). How about "report" for the CSP/NEL?

Overview.src.html
igrigorik: Presumably developers can still override Accept-Language, correct? I don't have a strong preference either way, but collapsing these does make sense to me.

Overview.src.html
annevk: That sounds good. I'm having a hard time judging whether a report is a type or a destination, but I think this makes sense.

Overview.src.html
annevk: The main difference is that it would no longer be omitted by default for `fetch()`. And there'd be no way to omit it (we'd have to add something new for that). I think I'd be okay with that for this header, personally. Especially since it aligns a bunch of things.

igrigorik:
@annevk merged fetch rows, added "report" destination + few other updates. ptal.

igrigorik:
@annevk ack, updated. Hopefully we're getting close(r) :-)

annevk:
I think with that, and rebasing and squashing if you would, this can land. (Though I might review it one last time to be sure.) I'd also appreciate a clear summary of the changes in the body of the commit message. The title of this PR seems good enough for the commit title.

Thanks a lot for working on this by the way, I think it's a big improvement.

igrigorik:
@annevk updated and squashed. It should land cleanly.. I hope :-)

Overview.src.html
```diff
@@ -592,11 +591,19 @@ <h4 id=terminology-headers>Headers</h4>
...
 "<code title>document</code>",
+"<code title>embed</code>",
+"<code title>font</code>",
+"<code title>image</code>",
+"<code title>object</code>",
+"<code title>manifest</code>",
+"<code title>media</code>",
+"<code title>report</code>",
 "<code title>serviceworker</code>",
+"<code title>script</code>",
```
annevk: This needs to be before service worker. (I can fix this while landing, just making a note here.)

Overview.src.html
```diff
@@ -592,11 +591,19 @@ <h4 id=terminology-headers>Headers</h4>
...
 "<code title>sharedworker</code>",
-"<code title>subresource</code>",
-"<code title>unknown</code>", and
-"<code title>worker</code>".
+"<code title>style</code>",
+"<code title>worker</code>", and
+"<code title>xslt</code>".
```
annevk: We should add that the initial value is the empty string here.  (I can fix this while landing, just making a note here.)

Overview.src.html
```diff
@@ -613,43 +620,52 @@ <h4 id=terminology-headers>Headers</h4>
    <th>CSP directive
    <th>Features
   <tr>
-   <td rowspan=14>""
-   <td rowspan=4>""
-   <td>""
+   <td rowspan=16>""
```
annevk: This needs to be 15 I think. As far as I can tell we're only adding one `<tr>`. You probably forgot to update this after removing "unknown"".  (I can fix this while landing, just making a note here.)

Overview.src.html
```diff
@@ -613,43 +620,52 @@ <h4 id=terminology-headers>Headers</h4>
...
+   <td rowspan=6>""
```
annevk: This needs to be 5 I think.  (I can fix this while landing, just making a note here.)

Overview.src.html
```diff
@@ -592,11 +591,19 @@ <h4 id=terminology-headers>Headers</h4>
 <dfn title=concept-request-destination>destination</dfn>, which is one of
 the empty string,
 "<code title>document</code>",
+"<code title>embed</code>",
+"<code title>font</code>",
+"<code title>image</code>",
+"<code title>object</code>",
```
annevk: This was also wrong. Needs to be before "report". (I'm also updating the enum.)

annevk:
Committed as https://github.com/whatwg/fetch/commit/6b9cdaa9be7acd2b8074c74823981ef994d17857. Thank you @igrigorik! (And everyone else for reviewing!)

PR #216, opened 2016-02-16
Fixup the request destination table values 

yoavweiss:
The request destination table got broken in https://github.com/whatwg/fetch/commit/6b9cdaa9be7acd2b8074c74823981ef994d17857

This fixes that table.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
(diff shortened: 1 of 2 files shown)


annevk:
I think what should happen instead is that 15 should become 16 and 5 should become 6 for rowspan. @igrigorik had that correct after all and I'm not sure why I thought I knew better. 😟

annevk:
(Because "embed" is not a type and "video" is not an initiator.)

igrigorik:
Yep, looks like the embed and video rows need to be shifted right by 1 :)

annevk:
Can you change Overview.src.html too @yoavweiss? I'm happy to generate Overview.html from it, up to you.

yoavweiss:
@annevk - added the Overview.src.html file (and removed the Overview.html)

annevk:
Committed as https://github.com/whatwg/fetch/commit/b18594b356ebe2a87602a66eb77ad3d80f3250dd. Also added you to the acknowledgments section, since you weren't there yet surprisingly enough!

PR #220, opened 2016-02-22
Vary:-header refers to request headers, not response headers

mstock:
If I understand the [`Vary:`-header specification](http://tools.ietf.org/html/rfc7231#section-7.1.4) and [a blog post about "Caching with CORS"](https://www.fastly.com/blog/caching-cors) correctly, the header refers to request headers like `Accept-Encoding`, and not response headers. So the example should `Vary` on the `Origin` request header, not the `Access-Control-Allow-Origin` response header.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -4608,7 +4608,7 @@ however, it is perfectly fine to do so.
 <span data-anolis-ref>HTML</span>
 <span data-anolis-ref>HTTP</span>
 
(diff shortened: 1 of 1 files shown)


annevk:
Indeed, thank you. Would you please add your name to the acknowledgments section?

mstock:
Done :).

annevk:
Thanks again!

PR #221, opened 2016-02-22
Clarify the hooks into RFC6265

mikewest:
Fetch's current language around cookies is fairly vague. I think we can
do a little better, even without changing the Cookie specification.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-26-february-2016">Living Standard — Last Updated 26 February 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-29-february-2016">Living Standard — Last Updated 29 February 2016</h2>
 
 <dl>
  <dt>Participate:
@@ -2656,12 +2656,22 @@ The <i title="">credentials flag</i> is one too.
   <p>If <var>credentials flag</var> is set, run these substeps:
 
   <ol>
-   <li><p>Modify <var>httpRequest</var>'s
-   <a href="#concept-request-header-list" title="concept-request-header-list">header list</a>
-   per HTTP State Management Mechanism.
-   <a href="#refsCOOKIES">[COOKIES]</a>
-   <!-- XXX proper hook? -->
+   <li>
+    <p>If the user agent is not configured to block cookies for <var>httpRequest</var> (see
+    <a href="https://tools.ietf.org/html/rfc6265#section-7">section 7</a> of
+    <a href="#refsCOOKIES">[COOKIES]</a>), then run these substeps:
 
+    <ol>
+     <li><p>Let <var>cookies</var> be the result of running the "cookie-string" algorithm (see
+     <a href="https://tools.ietf.org/html/rfc6265#section-5.4">section 5.4</a> of
+     <a href="#refsCOOKIES">[COOKIES]</a>) with the user agent's cookie store and
+     <var>httpRequest</var>'s <a href="#concept-request-current-url" title="concept-request-current-url">current url</a>.
+
+     <li>If <var>cookies</var> is not the empty string, append
+     `<code title="http-cookies">Cookie</code>`/<var>cookies</var> to <var>httpRequest</var>'s
+     <a href="#concept-request-header-list" title="concept-request-header-list">header list</a>.
+    </ol>
+ 
    <li><p>If <var>httpRequest</var>'s <a href="#concept-request-header-list" title="concept-request-header-list">header list</a>
    contains a <a href="#concept-header" title="concept-header">header</a> whose
    <a href="#concept-header-name" title="concept-header-name">name</a> is
@@ -2909,11 +2919,14 @@ The <i title="">credentials flag</i> is one too.
  <!-- XXX xref HTTP cache -->
 
  <li>
-  <p>If <var>credentials flag</var> is set and <var>response</var>'s
-  <a href="#concept-response-header-list" title="concept-response-header-list">header list</a> contains one or more
-  <a href="#concept-header" title="concept-header">headers</a> <a href="#concept-header-name" title="concept-header-name">named</a>
-  `<code title="">Set-Cookie</code>`, update the cookies. <a href="#refsCOOKIES">[COOKIES]</a>
-  <!-- XXX proper hook? -->
+  <p>If <var>credentials flag</var> is set and the user agent is not configured to block cookies for
+  <var>request</var> (see <a href="https://tools.ietf.org/html/rfc6265#section-7">section 7</a> of
+  <a href="#refsCOOKIES">[COOKIES]</a>), then run the "set-cookie-string" parsing algorithm (see
+  <a href="https://tools.ietf.org/html/rfc6265#section-5.2">section 5.2</a> of
+  <a href="#refsCOOKIES">[COOKIES]</a>) on the <a href="#concept-header-value" title="concept-header-value">value</a>
+  of each <var>header</var> <a href="#concept-header-name" title="concept-header-name">named</a> `<code title="">Set-Cookie</code>`
+  in <var>response</var>'s <a href="#concept-response-header-list" title="concept-response-header-list">header list</a>, if any, and
+  <var>request</var>'s <span title="concept-response-current-url">current url</span>.
 
   <p class="note">This is a fingerprinting vector.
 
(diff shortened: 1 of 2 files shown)


mikewest:
WDYT, @annevk?

@mnot, FYI. I think we can clarify Fetch without updating the RFC.

Overview.src.html
```diff
@@ -2594,12 +2594,22 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
-   <!-- XXX proper hook? -->
+   <li>
+    <p>If the user agent is not configured to block cookies for <var>httpRequest</var> (see
+    <a href="https://tools.ietf.org/html/rfc6265#section-7">Section 7</a> of
+    <span data-anolis-ref>COOKIES</span>), then:
 
+    <ol>
+     <li><p>Let <var>cookies</var> be the "cookie-string" computed by passing the user agent's
+     cookie store and <var>httpRequest</var>'s <span title=concept-request-current-url>current
+     url</span> to the algorithm defined in Section 5.4 of
```
annevk: xref section 5.4?

Overview.src.html
```diff
@@ -2594,12 +2594,22 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
   <ol>
-   <li><p>Modify <var>httpRequest</var>'s
-   <span title=concept-request-header-list>header list</span>
-   per HTTP State Management Mechanism.
-   <span data-anolis-ref>COOKIES</span>
-   <!-- XXX proper hook? -->
+   <li>
+    <p>If the user agent is not configured to block cookies for <var>httpRequest</var> (see
+    <a href="https://tools.ietf.org/html/rfc6265#section-7">Section 7</a> of
```
annevk: Lowercase "Section" throughout unless it's at the start of a sentence.

Overview.src.html
```diff
@@ -2594,12 +2594,22 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+    <span data-anolis-ref>COOKIES</span>), then:
```
annevk: s/then/run these subsubsteps/

Overview.src.html
```diff
@@ -2847,11 +2857,13 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <!-- XXX xref HTTP cache -->
 
  <li>
-  <p>If <var>credentials flag</var> is set and <var>response</var>'s
-  <span title=concept-response-header-list>header list</span> contains one or more
-  <span title=concept-header>headers</span> <span title=concept-header-name>named</span>
-  `<code title>Set-Cookie</code>`, update the cookies. <span data-anolis-ref>COOKIES</span>
-  <!-- XXX proper hook? -->
+  <p>If <var>credentials flag</var> is set, and the user agent is not configured to block cookies for
```
annevk: No need for a comma after set I think.

Overview.src.html
```diff
@@ -2847,11 +2857,13 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <var>httpRequest</var> (see <a href="https://tools.ietf.org/html/rfc6265#section-7">Section 7</a> of
+  <span data-anolis-ref>COOKIES</span>), then execute the "set-cookie-string" algorithm defined in
```
annevk: s/then execute/run/

Overview.src.html
annevk: Also, perhaps we can make that a parenthesized bit again "(see section x of y)". Below too.

Overview.src.html
```diff
@@ -2847,11 +2857,13 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <a href="https://tools.ietf.org/html/rfc6265#section-5.2">Section 5.2</a> of
+  <span data-anolis-ref>COOKIES</span> on the <span title=concept-header-value>value</span>
+  of each <var>header</var> <span title=concept-header-name>named</span> `<code title>Set-Cookie</code>`
+  contained in <var>response</var>'s <span title=concept-response-header-list>header list</span>.
```
annevk: Perhaps it's better to list there being `Set-Cookie` headers as the requirement and then have "then for each such header, run x on its value" or something like that. Now it's not super clear what happens if there are no such headers.

mikewest: I guess I see that as implicit in the `for-each` logic. Looping over the elements of an empty set just means that the body of the loop never executes. *shrug* Happy to add it in if you feel strongly about it, but adding a third condition to the `if` part of this sentence seems like overkill.

Overview.src.html
mikewest: Is there some xref syntax I missed? Or is a bare link enough (as below)?

mikewest:
Addressed some of the feedback, thanks @annevk!

Overview.src.html
annevk: The alternative would be s/contained // and adding ", if any" at the end. I think that's somewhat more consistent with what we do elsewhere.

mnot:
Awesome.

Overview.html
```diff
@@ -2656,12 +2656,22 @@ <h3 id="http-network-or-cache-fetch"><span class="secno">5.5 </span>HTTP-network
...
+   <li>
+    <p>If the user agent is not configured to block cookies for <var>httpRequest</var> (see
+    <a href="https://tools.ietf.org/html/rfc6265#section-7">section 7</a> of
+    <a href="#refsCOOKIES">[COOKIES]</a>), then run these substeps:
 
+    <ol>
+     <li><p>Let <var>cookies</var> be the result of the user agent's cookie store and
+     <var>httpRequest</var>'s <a href="#concept-request-current-url" title="concept-request-current-url">current url</a> into the
+     "cookie-string" algorithm (see <a href="https://tools.ietf.org/html/rfc6265#section-5.4">section 5.4</a>
+     of <a href="#refsCOOKIES">[COOKIES]</a>).
```
annevk: This sentence reads a little weird I think. Perhaps 'be the result of running the "cookie-string" algorithm (...) with the user agent's cookie store and _httpRequest_'s current url".

Overview.html
```diff
@@ -2909,11 +2919,14 @@ <h3 id="http-network-fetch"><span class="secno">5.6 </span>HTTP-network fetch</h
...
-  `<code title="">Set-Cookie</code>`, update the cookies. <a href="#refsCOOKIES">[COOKIES]</a>
-  <!-- XXX proper hook? -->
+  <p>If <var>credentials flag</var> is set and the user agent is not configured to block cookies for
+  <var>httpRequest</var> (see <a href="https://tools.ietf.org/html/rfc6265#section-7">section 7</a> of
+  <a href="#refsCOOKIES">[COOKIES]</a>), then run the "set-cookie-string" parsing algorithm (see
+  <a href="https://tools.ietf.org/html/rfc6265#section-5.2">section 5.2</a> of
+  <a href="#refsCOOKIES">[COOKIES]</a>) on the <a href="#concept-header-value" title="concept-header-value">value</a>
+  of each <var>header</var> <a href="#concept-header-name" title="concept-header-name">named</a> `<code title="">Set-Cookie</code>`
+  contained in <var>response</var>'s <a href="#concept-response-header-list" title="concept-response-header-list">header list</a>, and
+  <var>response</var>'s <a href="#concept-response-url" title="concept-response-url">url</a>.
```
annevk: The algorithm defined in the RFC doesn't seem to take two arguments. It does talk about request-uri [sic] at some point.

This algorithm will also segfault if we ever support putting cookies in synthetic responses. I think we want to point that out in an HTML comment.

mikewest: > The algorithm defined in the RFC doesn't seem to take two arguments.

It needs `request-uri` at the bottom in order to call into the storage model. I agree that it's not written in a way that makes its requirements clear. If you squint a bit, you'll see that this algorithm is executed "When a user agent receives a Set-Cookie header field in an HTTP response", and intends to retain that context through the algorithm. Given the way things in Fetch are structured, we need to provide that context somehow. Passing in the URL seems like the clearest way of doing so. When we spin up 6265bis, I'd be happy to tweak that text to make the requirements clear.

> This algorithm will also segfault if we ever support putting cookies in synthetic responses.

That would allow a service worker to set cookies for some other origin, wouldn't it? That seems like a bad thing to support. I can add a comment to that effect?

annevk: @mikewest so yes, I can see how it needs `request-uri`. Are you sure that passing the response URL is correct then? Given that this is executed post service-workers we could just use the request URL no? (And given that it's executed post service-workers synthetic responses are not actually a probably to concern ourselves with.)

mikewest:
Addressed your last two comments, and squished. Sorry I let this slip at the end of last week...

annevk:
Ta!

PR #228, opened 2016-03-01
Handle an ArrayBuffer allocation failure

yutakahirano:


Overview.src.html
```diff
@@ -2904,6 +2904,8 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
      <li>
       <p><span title=concept-enqueue-ReadableStream>Enqueue</span> a <code>Uint8Array</code>
       object wrapping an <code>ArrayBuffer</code> containing <var>bytes</var> to <var>stream</var>.
+      If that threw an exception, <span title=concept-fetch-terminate>terminate</span> the ongoing
+      fetch with <i>fatal</i> and abort these subsubsteps.
```
annevk: Should we not error stream with the exception? Perhaps it makes sense to have an abstract operation for this since we enqueue this kind of Uint8Array three times throughout the specification and they all need the same error handling.

annevk: At least, I think they all need the same error handling.

yutakahirano: Inserted "error the stream with that exception".

> Perhaps it makes sense to have an abstract operation for this since we enqueue this kind of Uint8Array three times throughout the specification and they all need the same error handling.

They have additional error handling steps that are different each other (abort [sub]steps, terminate the fetch, etc). So I think having an abstract operation will not greatly simplify the description.

annevk:
In your implementation, are these all inlined without a shared operation? It seems a shared operation would help clarify matters, but I guess I'm okay with landing this as-is for now.

PR #236, opened 2016-03-07
Define the WebSocket client handshake in terms of Fetch

### LINKED ISSUE #235: Define how WebSocket "fetching" works
A lot of the things specified in Fetch apply to WebSocket too, but since there's no easy way to update the protocol specification a lot of the details are scrambled in various drafts.

To clear this up we should specify all the monkey patching here, in a single place, and then point to HTML Standard here. Fetch then serves as its entry point to the WebSocket protocol and everything shared between Fetch and WebSocket should become a little clearer.

Bits elsewhere that need to be put here:

* HSTS
* https://w3c.github.io/webappsec-mixed-content/#websockets-integration
* https://w3c.github.io/webappsec-upgrade-insecure-requests/#websockets-integration
* https://w3c.github.io/webappsec-csp/#directive-connect-src

Perhaps even the entire bootstrapping WebSocket fetch could just be done in terms of Fetch. That would be ideal.

annevk:
Fixes #235.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-8-march-2016">Living Standard — Last Updated 8 March 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-9-march-2016">Living Standard — Last Updated 9 March 2016</h2>
 
 <dl>
  <dt>Participate:
@@ -93,6 +93,10 @@ fetching.
    <li><a href="#structured-cloning-of-headers,-request,-and-response-objects"><span class="secno">6.5 </span>Structured cloning of <code>Headers</code>, <code>Request</code>, and <code>Response</code> objects</a></li>
    <li><a href="#fetch-method"><span class="secno">6.6 </span>Fetch method</a></li>
    <li><a href="#garbage-collection"><span class="secno">6.7 </span>Garbage collection</a></ol></li>
+ <li><a href="#websocket-protocol"><span class="secno">7 </span>WebSocket protocol alterations</a>
+  <ol>
+   <li><a href="#websocket-connections"><span class="secno">7.1 </span>Connections</a></li>
+   <li><a href="#websocket-opening-handshake"><span class="secno">7.2 </span>Opening handshake</a></ol></li>
  <li><a class="no-num" href="#background-reading">Background reading</a>
   <ol>
    <li><a class="no-num" href="#http-header-layer-division">HTTP header layer division</a></li>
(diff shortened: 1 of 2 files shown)


annevk:
@mikewest could you review this?

@domenic should Fetch abstract the entire WebSocket protocol or should HTML reference a mixture of Fetch and the WebSocket protocol? That is, should Fetch reintroduce the hooks HTML needs, perhaps even putting the security limitations on them directly so HTML no longer needs to do that. I think I have a slight preference for that, but it would be a larger change.

domenic:
This looks much less painful than anticipated.

The need for "obtain a WebSocket connection" being overridden isn't completely clear. Why is step 1 omitted? And why is _resource name_ excluded?

> should Fetch abstract the entire WebSocket protocol or should HTML reference a mixture of Fetch and the WebSocket protocol? That is, should Fetch reintroduce the hooks HTML needs, perhaps even putting the security limitations on them directly so HTML no longer needs to do that. I think I have a slight preference for that, but it would be a larger change.

I'm not really clear what the difference would be, concretely, but in the abstract wrapping everything in fetch sounds a bit nicer. On the other hand, there are a lot of concepts in [WSP] referenced by HTML, and if these two are the only ones you need to override, then just having a no-op forwarding layer between Fetch and WSP would be a bit annoying.

annevk:
> The need for "obtain a WebSocket connection" being overridden isn't completely clear.

Note that "establish a WebSocket connection" in the RFC is the process of obtaining a connection, coupled with sending a request and validating the response. It's all things intertwined.

> Why is step 1 omitted? And why is resource name excluded?

Step 1 is omitted because we already validated the input. Resource name is not actually used to obtain a connection. It's used for the handshake that is separately overridden (the larger section with all the header appending).

> On the other hand, there are a lot of concepts in [WSP] referenced by HTML, and if these two are the only ones you need to override, then just having a no-op forwarding layer between Fetch and WSP would be a bit annoying.

I think I would import some of the subsetting that HTML does around error handling. But yes, there will be some forwarding too.

domenic:
> Note that "establish a WebSocket connection" in the RFC is the process of obtaining a connection, coupled with sending a client handshake request and validating the response. It's all things intertwined. Fetch starts out with the request and does things with that, and then at some point obtains a connection. The layering is different.

Ah, I see, the structure of that plaintext document makes it hard to understand. OK then.

mikewest:
This looks good, thank you Anne.

Some thoughts:

1. How do you intend for this to play together with Service Workers? As written, we'd be exposing the initial request to the worker, which might give them an opportunity to do something unexpected.

2. It would be nice if the HSTS spec mentioned WebSockets. As is, I think the fact that Fetch upgrades the initial request from HTTP to HTTPS means that the WebSocket upgrade will automagically continue over a secure connection (because the scheme will be `https`, so `secure` will be `true` in "obtain a websocket connection").

3. We'll have to figure out what the CSP hook looks like, as `connect-src` looks for `ws:whatever` today. I think Fetch is fine here, as we're passing the whole request to CSP. 

Otherwise, LGTM.

annevk:
1. Service workers are dealt with by setting the skip-service-worker flag.
2. HSTS is indeed applied to WebSocket. (It was already by the HTML Standard, this moves it to the Fetch layer from the API layer.)
3. I think the way type/destination etc. are set this should look similar to XMLHttpRequest so I think connect-src would fall out naturally. But yeah, you might need to update it to no longer look for `ws:` and such.

PR #237, opened 2016-03-09
Integrating support for PasswordCredential objects.

mikewest:
TODO: Actually describe the change once folks are on board with the direction.

w3c/webappsec-credential-management#11

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-9-march-2016">Living Standard — Last Updated 9 March 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-10-march-2016">Living Standard — Last Updated 10 March 2016</h2>
 
 <dl>
  <dt>Participate:
@@ -889,7 +889,7 @@ special value for <cite>HTML</cite>. <a href="#refsHTML">[HTML]</a>
 <p>A <a href="#concept-request" title="concept-request">request</a> has an associated
 <dfn id="concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</dfn>, which is
 "<code title="">omit</code>", "<code title="">same-origin</code>", "<code title="">include</code>", or
-"<code title="">attached-credential</code>". Unless stated otherwise, it is "<code title="">omit</code>".
+"<code title="">credential</code>". Unless stated otherwise, it is "<code title="">omit</code>".
 
 <p class="note no-backref">When <a href="#concept-request" title="concept-request">request</a>'s
 <a href="#concept-request-mode" title="concept-request-mode">mode</a> is "<code title="">navigate</code>", its
@@ -1586,7 +1586,7 @@ these <a href="#concept-header" title="concept-header">headers</a>:
   <p>Indicates whether a resource can be shared when
   <a href="#concept-request" title="concept-request">request</a>'s
   <a href="#concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</a> is
-  "<code title="">include</code>" or "<code title="">attached-credential</code>".
+  "<code title="">include</code>" or "<code title="">credential</code>".
 
   <p class="note">For a <a href="#cors-preflight-request">CORS-preflight request</a>,
   <a href="#concept-request" title="concept-request">request</a>'s
@@ -2323,7 +2323,7 @@ indicates an attempt to authenticate.
 
     <ul>
      <li><var>request</var>'s <a href="#concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</a>
-     is "<code title="">include</code>" or "<code title="">attached-credential</code>"
+     is "<code title="">include</code>" or "<code title="">credential</code>"
      <li><var>request</var>'s  <a href="#concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</a>
      is "<code title="">same-origin</code>" and <var>request</var>'s
      <a href="#concept-request-response-tainting" title="concept-request-response-tainting">response tainting</a> is
@@ -2382,7 +2382,7 @@ indicates an attempt to authenticate.
       <ul>
        <li>The <i title="">CORS flag</i> is set.
        <li><var>request</var>'s <a href="#concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</a>
-       is <em>neither</em> "<code>include</code>" <em>nor</em> "<code>attached-credential</code>".
+       is <em>neither</em> "<code>include</code>" <em>nor</em> "<code>credential</code>".
        <li><var>request</var>'s <a href="#concept-request-window" title="concept-request-window">window</a> is
        "<code>no-window</code>".
       </ul>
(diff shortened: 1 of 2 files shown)


mikewest:
This is the general direction I'm thinking, @annevk. WDYT?

mikewest:
https://github.com/w3c/webappsec-credential-management/commit/eb1d232252771429b9bac3549aa1c69c9a34f477 shows the normative changes to the credential management spec which would be necessary. Certainly is simpler from that perspective.

Overview.html
```diff
@@ -4108,9 +4136,18 @@ <h3 id="request-class"><span class="secno">6.3 </span>Request class</h3>
...
- <a href="#concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</a> to
- <var>credentials</var>.
+ <li><p>If <var>credentials</var> is non-null:
+   
+   <ol>
+     <li>If <var>credentials</var> is a
+     <a href="https://w3c.github.io/webappsec-credential-management/#interfaces-credential-types-passwordcredential">PasswordCredential</a>,
+     set <var>request</var>'s <a href="#concept-request-credentials-mode" title="concept-request-credentials-mode">credentials mode</a> to
+     "<code title="">attached-credential</code>", and <var>request</var>'s
+     <a href="#concept-request-attached-credential" title="concept-request-attached-credential">attached credential</a> to
```
AxelNennker: Is the same-origin protection here implicit somewhere or should it be spelled out?
https://w3c.github.io/webappsec-credential-management/#monkey-patching-fetch-3

mikewest: That's in the proposed corollary patch to the CM spec at https://github.com/w3c/webappsec-credential-management/commit/eb1d232252771429b9bac3549aa1c69c9a34f477

mikewest: For clarity: the body extraction would fail if the request and the credential weren't same-registrable domain. That would result in a network error being returned from HTTP-network-or-cache fetch. If y'all would prefer to throw during creation, we could also do that check in https://fetch.spec.whatwg.org/#dom-request, somewhere after step 12.

Overview.src.html
```diff
@@ -2490,6 +2496,25 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 
  <li><p>Let <var>contentLengthValue</var> be null.
 
+ <li>
+   <p>If <var>httpRequest</var>'s <span title=concept-request-attached-credential>attached
+   credential</span> is <em>not</em> null, run these substeps:
+
+   <ol>
+     <li>If <var>httpRequest</var>'s <span title=concept-request-body>body</span> is <em>not</em>
+     null, return a <span title=concept-network-error>network error</span>.
```
annevk: How can this situation arise?

Overview.src.html
```diff
@@ -4118,10 +4155,11 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
  <li><p>If either <var>init</var>'s <code title>body</code> member is present and is non-null or
- <var>inputBody</var> is non-null, and <var>request</var>'s
+ <var>inputBody</var> is non-null, then <span data-anolis-spec=webidl>throw</span> a
+ <code title>TypeError</code> if either <var>request</var>'s
  <span title=concept-request-method>method</span> is `<code title>GET</code>` or
- `<code title>HEAD</code>`, <span data-anolis-spec=webidl>throw</span> a
- <code title>TypeError</code>.
+ `<code title>HEAD</code>`, or <var>request</var>'s
+ <span title=concept-request-attached-credential>attached credential</span> is non-null.
```
annevk: This is becoming quite hard to parse.

Overview.html
annevk: What about redirects?

annevk:
So is the intention that "attached-credential" means "include" + Credential? There's no way to get "omit" + Credential? Will we never want that?

mikewest:
> So is the intention that "attached-credential" means "include" + Credential? There's no way to get "omit" + Credential? Will we never want that?

I think we can basically consider a `PasswordCredential` object as a "credential", just as we would cookies, etc. Introducing a mode where we'd send one but not the other doesn't seem like it would be useful, as sending credentials is a fairly clear signal that you're trying to authenticate your current session, and that requires sending the state you currently have (session id, etc), and accepting any state that the server returns (auth cookie, etc).

Overview.src.html
mikewest: What would make more sense to you? Splitting it into two explicit conditions?

Overview.src.html
mikewest: I don't think it can, but it's not clear how to write an `ASSERT(x)` in Fetch-code. :)

annevk: I'm happy adopting the ECMAScript convention of "Assert: ..."

annevk:
That seems reasonable. I wonder if there's a clearer term that signifies this is "include" plus something else. I came up with these so far but none is a favorite ("attached-credential" isn't either):

* include-plus-body
* include-plus-credential
* credential
* credential-body

If I was forced to pick I'd probably go with "credential".

Overview.html
AxelNennker: I prefer to let creation of a not-same-origin request fail

mikewest:
I'm happy to spell the concept however you like. :)

Overview.src.html
annevk: Maybe 
```
<p>If all of the following are true
<ul class=brief>
 <li><p>init's body member is present and is non-nonull or inputBody is non-null
 <li><p>request's method is GET or HEAD, or request's attached credential is non-null
</ul>
<p>then throw a ... exception.</p>
```
Basically use `<ul class=brief>` as something that is part of the paragraph even though it's not. Started using that here and there and I think it makes these kind of statements easier to read.

Overview.html
annevk: We have generally deferred same-origin checks to the network layer. I don't think there's a strong reason to not do that here.

mikewest: @annevk: Are you saying that you'd prefer for Fetch to retain responsibility for the Credential comparison? Or that you're fine with the other spec being responsible for the comparison that's being triggered at the network layer rather than the object creation layer?

mikewest: > What about redirects?

Right. I forgot about those. Two options:

* The existing spec sets the redirect mode to "error". We could do that here.

* We could add custom logic for this new credential mode.

The former is probably simpler.

annevk: Yeah, that seems reasonable.

Overview.src.html
mikewest: I rearranged this a bit, hopefully it's a little more clear?

Overview.src.html
mikewest: SGTM. Done.

Overview.html
mikewest: Done.

mikewest:
> I'm happy to spell the concept however you like. :)

I went with "credential" for the "credential mode", but this is probably worth chatting about a bit. Maybe something like "deferred" or "network-bound" or something along those lines would be a clearer indication of the way that this credential mode differs from "include"?

Alternatively, we could just set the credential mode to "include", and have a `hasWhateverWeCallIt` flag on the `Request` object that allowed inspection of the request's state.

WDYT?

Overview.html
```diff
@@ -2552,6 +2558,26 @@ <h3 id="http-network-or-cache-fetch"><span class="secno">5.5 </span>HTTP-network
 
  <li><p>Let <var>contentLengthValue</var> be null.
 
+ <li>
+   <p>If <var>httpRequest</var>'s <a href="#concept-request-attached-credential" title="concept-request-attached-credential">attached
+   credential</a> is <em>not</em> null, run these substeps:
+
+   <ol>
+     <li>Assert: <var>httpRequest</var>'s <a href="#concept-request-body" title="concept-request-body">body</a> is
+     <em>not</em> null.
```
annevk: You want to assert it is null, I think.

annevk:
Once FIDO enters the picture and we overload this with `FIDOCredential` objects. Would you expect a service worker to want to distinguish between whether a request has password or FIDO "attached"? Perhaps "password" and "fido" would make more sense than "credential" long term.

annevk:
"password" still does not clearly indicate it includes the requirements of "include", but we can point that out in a note and as you say it's somewhat reasonable to not distinguish.

mikewest:
Changed "credential" to "password", fixed the assertion, and added a note about "password"'s behavior.

I'm not entirely convinced that it's worthwhile to distinguish between the types of credentials that might be attached to a request, but there doesn't seem to be any harm in specificity.

wanderview:
Does the `Request.credentials` getter return the PasswordCredentials object specified in RequestInit or does it return the `'password'` string?

mikewest:
It returns the "`password`" string. There's no JS-visible mechanism to grab the credential itself. We might need to add that for FIDO, but we don't need it for any existing credentials (as they're mostly opaque objects).

wanderview:
Ah, right.  The getter was pre-existing and you're just filling in the value at construction time.

LGTM.  Thanks!

wanderview:
Actually, I see code in the wild that does stuff like this:

```
new Request(newURL, { credentials: oldRequest.credentials });
```

That will break with this current scheme because the credentials getter does not contain enough information to set on a new RequestInit.

mikewest:
Hrm. What would you like to see us do here? Should we expose a new accessor? Or just change the type of the `credentials` accessor to the new union? I think I'd prefer the latter, but I don't have a good feel for how it might break.

mikewest:
(Basically, I won't be sad if we just can't copy this kind of request, but I want to play nicely with folks' expectations. I'm happy to defer to y'all here.)

wanderview:
I honestly don't know.  Its a problem with the fetch spec.  People want to copy requests and change a single value.  But asymmetry between getters and setters makes certain attributes uncopyable.

mikewest:
@annevk: Do you have an opinion about the scenario @wanderview describes?

annevk:
@mikewest I'm not sure. See #245. 

For this particular PR, I think I would like us to work with the FIDO folks to get an idea of what their requirements are. Seems like that should be possible to figure out in not too much time and can hopefully ensure we are doing the right thing.

There's also still some rambling on dev.platform and given that Mozilla is the only other interested party in this API I'm not super comfortable merging this as of yet. Hopefully that can be resolved soonish.

mikewest:
We're not doing this anymore. Might pull it back for something else in the future, but closing it out the meantime.

PR #242, opened 2016-03-10
Define the WebSocket subprotocol check as part of response validation

annevk:
See https://github.com/whatwg/html/pull/840#discussion_r55568720 for discussion and context. Many
thanks to Simon Pieters for helping out.


zcorpan:
LGTM

Issue #250, opened 2016-03-17
ReadableStream constructor call is not quite right


domenic:
https://fetch.spec.whatwg.org/#readablestream

> Let stream be the result of calling the initial value of ReadableStream as constructor with init as an argument\. Rethrow any exceptions\. 

The strategy should be the second parameter, not a property of the first. This probably changed a while ago in streams and we didn't realize when reviewing... @yutakahirano

yutakahirano:
Oops. Thanks, I'll fix it.

annevk:
Thanks @yutakahirano!

PR #254, opened 2016-03-18
Give strategy as the second argument of ReadableStream

yutakahirano:


yutakahirano:
Fixes #250. @domenic, can you take a look?

domenic:
LGTM!

Issue #52, opened 2015-05-15
CSP Request Header and CORS preflight fetch.


horo-t:
According to the CSP spec
https://w3c.github.io/webappsec/specs/content-security-policy/#csp-request-header
> If the user agent monitors or enforces a policy that contains a directive that contains a source list, then the user agent MUST set a CSP Request Header when requesting cross-origin resources, as described in §3.4 The CSP HTTP Request Header.

But "CSP" is not a simple header
https://fetch.spec.whatwg.org/#simple-header
> A simple header is a header whose name is either one of `Accept`, `Accept-Language`, and `Content-Language`, or whose name is `Content-Type` and value, once parsed, has a MIME type (ignoring parameters) that is one of `application/x-www-form-urlencoded`, `multipart/form-data`, and `text/plain`.

So when the user agent requests a cross-origin resource which CSP is set, it must send a CORS preflight fetch.

This means when we use CSP, we can't use CDN which doesn't support CORS preflight.

Is this my understanding correct?

mikewest:
That certainly seems terrible, and isn't what I intended with that addition. @annevk, wdyt?

annevk:
I see three problems. 1. It's not clear when this header is supposed to be set and therefore it's not clear whether e.g. service workers are supposed to be able to observe it. 2. This header can be set by `fetch()` and `XMLHttpRequest` because it is not prefixed with `Sec-` so you need to handle duplicates somehow when you set it. 3. It's not clear whether including this header cross-origin is a security risk or not and therefore whether we should preflight it. So far we have answered yes to this question.

annevk:
Ping!

mikewest:
The goal is to inform a page that any redirect they perform is potentially observable (in a binary "Happened" or "Didn't happen." sense). We plugged the whole which made the endpoint completely observable, so I've been less interested in figuring out this header, as it's less valuable than it used to be.

That said:

1. It should reflect the state of the thing that's making the request, and therefore the thing that can observe redirects. That is, if a Document sends a request, and it passes through a service worker, the header should be present. If a SW intercepts a request from that Document, and then sends it's own request in order to fulfill it, then the header would be present iff the service worker had an active policy.

2. That seems fine. The value of the header is immaterial; it's `1` at the moment, but could just as easily be `1,1,1,1,1,1` or `lkjahfklj;dvsnf`. That might change in the future if we actually use the header for anything interesting, but for now, presence is enough.

[...]

PR #258, opened 2016-03-22
Integrate Client-Hints with Fetch

igrigorik:
- Last-Event-ID is defined in the HTML spec as a utf-8 string, hence we
  don't place any restrictions on its value
- Each Client Hints header has a BNF grammar that should be validated by
  the user agent; the header is defined as simple if the name of the
  header matches one of CH header names, and its value conforms to the
  defined grammar

Closes:
- https://github.com/whatwg/fetch/issues/52
- https://github.com/httpwg/http-extensions/issues/141

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -324,16 +324,34 @@ each other by `<code title>,</code>`, in order.
 <hr>
 
 <p>A <dfn>simple header</dfn> is a <span title=concept-header>header</span> whose
-<span title=concept-header-name>name</span> is `<code title>Accept</code>`,
-`<code title>Accept-Language</code>`, or `<code title>Content-Language</code>`, or whose
-<span title=concept-header-name>name</span> is `<code title>Content-Type</code>` and
-<span title=concept-header-value>value</span>, <span title=concept-header-parse>once parsed</span>,
-has a MIME type (ignoring parameters) that is
-`<code title>application/x-www-form-urlencoded</code>`, `<code title>multipart/form-data</code>`, or
-`<code title>text/plain</code>`.
+<span title=concept-header-name>name</span> is one of
+
+<ul class=brief>
+ <li>`<code title>Accept</code>`
+ <li>`<code title>Accept-Language</code>`
+ <li>`<code title>Content-Language</code>`
+ <li>`<code title>Content-Type</code>` and whose <span title=concept-header-value>value</span>,
+ <span title=concept-header-parse>once parsed</span>, has a MIME type (ignoring parameters)
+ that is `<code title>application/x-www-form-urlencoded</code>`,
+ `<code title>multipart/form-data</code>`, or `<code title>text/plain</code>`
+</ul>
+
 <!-- XXX * needs better xref
          * ignoring parameters has been the standard for a long time now
          * interesting test: "Content-Type: text/plain;" -->
+
+<p>or whose <span title=concept-header-name>name</span> is one of
+
+<ul class=brief>
+ <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#dpr>DPR</a></code>`
+ <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#downlink>Downlink</a></code>`
+ <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#save-data>Save-Data</a></code>`
+ <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>Viewport-Width</a></code>`
+ <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#width>Width</a></code>`
+</ul>
+
+<p>and whose <span title=concept-header-value>value</span>,
+<span title=concept-header-parse>once parsed</span>, is not a failure.
 
 <p>A <dfn>forbidden header name</dfn> is a <span title=concept-header>header</span>
 <span title=concept-header-name>name</span> that is one of
@@ -803,6 +821,17 @@ the empty string.
 <p class="note no-backref">This can be used to override a referrer policy associated with
 an <span data-anolis-spec=html>environment settings object</span>.
 <span data-anolis-ref>REFERRER</span>
+
+<p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-client-hints-list>client hints list</dfn>, which is a
+<span title=concept-client-hints-list>client-hints list</span>. Unless stated otherwise, it is
+the empty list.
+
+<p class="note no-backref">This will be used to override a client hints list associated with
+an <span data-anolis-spec=html>environment settings object</span>.
+<span data-anolis-ref>CLIENT-HINTS</span>
+
+<!-- XXX this needs to move to CLIENT-HINTS -->
 
 <p>A <span title=concept-request>request</span> has an associated
 <dfn>synchronous flag</dfn>. Unless stated otherwise it is unset.
@@ -1387,6 +1416,15 @@ is listed in the first column of the following table.</p>
 <span data-anolis-ref>REFERRER</span>
 
 
+<h3>Client hints list</h3>
+
+<p>A <dfn title=concept-client-hints-list>client hints list</dfn> is a list of
+<a href=http://httpwg.org/http-extensions/client-hints.html#accept-ch>Client hint tokens</a>,
+each of which is one of `<code title>dpr</code>`,
+`<code title>save-data</code>`, `<code title>viewport-width</code>`, or
+`<code title>width</code>`.
+
+
 <h3>Streams</h3>
 
 <p class="note no-backref">This section might be integrated into other standards, such as IDL.
@@ -1888,10 +1926,59 @@ if the ongoing fetch is updating the response in the HTTP cache for the request.
   for HTTP/2, and equivalent information used to prioritize dispatch and processing of
   HTTP/1 fetches.
 
+
+ <li>
+  <p>If <var>request</var> is a <span>navigation request</span>, a user agent should, for each
+  <span title=concept-header>header</span> <span title=concept-header-name>name</span>
+  (<var>hint-name</var>) in the first column of the following table, if <var>hint-name</var>
+  is not in <var>request</var>'s <span title=concept-request-header-list>header list</span>,
+  <span title=concept-header-list-append>append</span>
+  <var>hint-name</var>/<var>hint-value</var> given in the same row on the second column, to
+  <var>request</var>'s <span title=concept-request-header-list>header list</span>.
+
+  <table>
+   <tr>
+    <th><span title=hint-name>hint-name</span>
+    <th><span title=hint-value>hint-value</span>
+   <tr>
+    <td>`<code title=http-dpr>dpr</a></code>`
+    <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
+   <tr>
+    <td>`<code title=http-save-data>save-data</code>`
+    <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
+   <tr>
+    <td>`<code title=http-viewport-width>viewport-width</code>`
+    <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+  </table>
+
  <li>
   <p>If <var>request</var> is a <span>subresource request</span>, run these substeps:
 
   <ol>
+   <li>
+    <p>If the <var>request</var>'s <span title=concept-client-hint-list>client hints list</span> is not empty, run these substeps for each <var>hint-name</var> in the list:
+
+    <ol>
+     <li>
+      <p>Set <var>value</var> to the first matching statement, if any, switching on
+      <var>hint-name</var>:
+
+      <dl class=switch>
+       <dt>"<code title>dpr</code>"
+       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
+       <dt>"<code title>save-data</code>"
+       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
+       <dt>"<code title>viewport-width</code>"
+       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+       <dt>"<code title>width</code>"
+       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#width>width value</a></code>
+      </dl>
+
+     <li><p><span title=concept-header-list-append>Append</span>
+       <var>hint-name</var>/<var>value</var> to <var>request</var>'s
+       <span title=concept-request-header-list>header list</span>.
+    </ol>
+
    <li><p>Let <var>record</var> be a new
    <span title=concept-fetch-record>fetch record</span> consisting of
    <var>request</var> and this instance of the


Overview.src.html
```diff
@@ -336,13 +336,23 @@ <h4 id=terminology-headers>Headers</h4>
...
+  <li>`<code title>Content-Type</code>` and whose <span title=concept-header-value>value</span>,
+  <span title=concept-header-parse>once parsed</span>, has a MIME type (ignoring parameters)
+  that is `<code title>application/x-www-form-urlencoded</code>`,
+  `<code title>multipart/form-data</code>`, or `<code title>text/plain</code>`
+  <li>`<code title><a href=https://html.spec.whatwg.org/multipage/comms.html#last-event-id>Last-Event-ID</a></code>`
+  <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-dpr-client-hint>DPR</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-downlink-client-hint>Downlink</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-save-data-hint>Save-Data</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-viewport-width-client-hint>Viewport-Width</a></code>`, or
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-width-client-hint>Width</a></code>`, and whose value, once parsed, is not a failure.
```
annevk: value and once parsed need to be cross-referenced.

annevk:
I have some concerns here:

1. Where did this get security review?
2. Note that in https://github.com/whatwg/html/issues/689 we might change the value space of `Last-Event-ID`.
3. According to dev.platform the Client-Hints headers are still controversial. Are there any user agents other than Chromium shipping these?

Overview.src.html
igrigorik: Ack, updated.

igrigorik:
> Where did this get security review?

For Client Hints, we went through a privacy and security review as [part of our intent to ship](https://groups.google.com/a/chromium.org/d/msg/blink-dev/ATbqmznya_k/ze5wpG8rAQAJ); same for [Save-Data](https://groups.google.com/a/chromium.org/d/msg/blink-dev/-_pEwnE_e_Y/zx5KFs3VBAAJ).

> Note that in whatwg/html#689 we might change the value space of Last-Event-ID.

Ah, that makes sense. I'm happy to drop Last-Event-ID from this pull until that's resolved.

> According to dev.platform the Client-Hints headers are still controversial. Are there any user agents other than Chromium shipping these?

Chrome and Opera supports Width, Viewport-Width, DPR, Save-Data. Yandex is [shipping Save-Data](https://developers.google.com/web/updates/2016/02/save-data). Edge has it [listed as under consideration](https://dev.windows.com/en-us/microsoft-edge/platform/status/httpclienthints).

annevk:
Okay, so other than Chromium there's no other engines yet. That's a bit of a concern. It'd be better if one of Mozilla, Apple, or Microsoft expressed support.

Should we expand the scope of this PR to include https://github.com/httpwg/http-extensions/issues/156?

And yeah, doing `Last-Event-ID` separately seems better.

igrigorik:
Dropped Last-Event-ID and took a first run at integrating the proposed logic in https://github.com/httpwg/http-extensions/issues/156#issuecomment-200837480 with Fetch. I'm sure it's rough around the edges, let me know if I'm heading in the right direction.

Overview.src.html
```diff
@@ -336,13 +336,23 @@ <h4 id=terminology-headers>Headers</h4>
...
+  <span title=concept-header-parse>once parsed</span>, has a MIME type (ignoring parameters)
+  that is `<code title>application/x-www-form-urlencoded</code>`,
+  `<code title>multipart/form-data</code>`, or `<code title>text/plain</code>`
+  <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-dpr-client-hint>DPR</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-downlink-client-hint>Downlink</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-save-data-hint>Save-Data</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-viewport-width-client-hint>Viewport-Width</a></code>`, or
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#the-width-client-hint>Width</a></code>`, and whose <span title=concept-header-value>value</span>,
+  <span title=concept-header-parse>once parsed</span>, is not a failure.
+</ul>
```
annevk: I think it might be better to have `<p>or whose name is one of </p> <ul class=brief> ... </ul> <p>and value, once parsed, is not failure.</p>` after the `Content-Type` entry. It seems a little weird to have a list of header names and then have a list of header names inside that list.

Overview.src.html
```diff
@@ -336,13 +336,23 @@ <h4 id=terminology-headers>Headers</h4>
...
 <!-- XXX * needs better xref
          * ignoring parameters has been the standard for a long time now
          * interesting test: "Content-Type: text/plain;" -->
```
annevk: This needs to be kept close to the `Content-Type` text. Below it.

Overview.src.html
```diff
@@ -1851,10 +1873,38 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+    <li>`<code title=http-save-data><a href="http://httpwg.org/http-extensions/client-hints.html#save-data">Save-Data</a></code>`/<var>value</var>, unless the <span title=concept-request-header-list>header list</span> already contains a <span title=concept-header>header</span> whose <span title=concept-header-name>name</span> is `<code title=http-save-data>Save-Data</code>`
+    <li>`<code title=http-viewport-width><a href="http://httpwg.org/http-extensions/client-hints.html#save-data">Viewport-Width</a></code>`/<var>value, unless the <span title=concept-request-header-list>header list</span> already contains a <span title=concept-header>header</span> whose <span title=concept-header-name>name</span> is `<code title=http-viewport-width>Viewport-Width</code>`</var>
+  </ol>
+
+  <p>where <var>value</var> is set according to the definition in <span data-anolis-ref>CLIENT-HINTS</span>.
```
annevk: I don't think this makes much sense since this means value would be the same for all these headers.

Overview.src.html
```diff
@@ -1349,6 +1368,9 @@ <h4 id=terminology-headers>Headers</h4>
 <p class=note>The details of referrer policies are discussed in Referrer Policy.
 <span data-anolis-ref>REFERRER</span>
 
+<h3>Client Hints list</h3>
```
annevk: Needs an extra newline before. And this should read "Client hints list" I think.

Overview.src.html
```diff
@@ -1349,6 +1368,9 @@ <h4 id=terminology-headers>Headers</h4>
...
+
+<p>A <dfn title=concept-client-hints-list>client hints list</dfn> is a list of <a href=http://httpwg.org/http-extensions/client-hints.html#rfc.section.2.2.1>Client Hint tokens</a>, whose values are one of `<code title>dpr</code>`, `<code title>width</code>`, `<code title>viewport-width</code>`.
```
annevk: Needs an extra newline after.

annevk: Also, a client hint token doesn't have a value, right? It's a value. So "which is one of" seems better. And then `viewport-width` needs "or" before it.

igrigorik:
@annevk thanks for the review! Updated the pull -- ptal.

Overview.src.html
```diff
@@ -1368,9 +1376,14 @@ <h4 id=terminology-headers>Headers</h4>
...
 <span data-anolis-ref>REFERRER</span>
 
-<h3>Client Hints list</h3>
 
-<p>A <dfn title=concept-client-hints-list>client hints list</dfn> is a list of <a href=http://httpwg.org/http-extensions/client-hints.html#rfc.section.2.2.1>Client Hint tokens</a>, whose values are one of `<code title>dpr</code>`, `<code title>width</code>`, `<code title>viewport-width</code>`.
+<h3>Client hints list</h3>
+
+<p>A <dfn title=concept-client-hints-list>client hints list</dfn> is a list of
+<a href=http://httpwg.org/http-extensions/client-hints.html#rfc.section.2.2.1>Client hint tokens</a>,
+which is one of `<code title>dpr</code>`, `<code title>save-data</code>`,
```
annevk: I guess this should say "each of which". Otherwise it reads as if the list is a byte sequence.

Overview.src.html
```diff
@@ -336,17 +336,35 @@ <h4 id=terminology-headers>Headers</h4>
...
 
 <p>A <dfn>simple header</dfn> is a <span title=concept-header>header</span> whose
-<span title=concept-header-name>name</span> is `<code title>Accept</code>`,
-`<code title>Accept-Language</code>`, or `<code title>Content-Language</code>`, or whose
-<span title=concept-header-name>name</span> is `<code title>Content-Type</code>` and
-<span title=concept-header-value>value</span>, <span title=concept-header-parse>once parsed</span>,
-has a MIME type (ignoring parameters) that is
-`<code title>application/x-www-form-urlencoded</code>`, `<code title>multipart/form-data</code>`, or
-`<code title>text/plain</code>`.
+<span title=concept-header-name>name</span> is one of:
```
annevk: I think it's fine to leave off the ":" here since it's basically one long paragraph, with nested lists. (HTML cannot express that unfortunately due to parsing issues, but this will have to do.)

Overview.src.html
```diff
@@ -336,17 +336,35 @@ <h4 id=terminology-headers>Headers</h4>
...
+  <span title=concept-header-parse>once parsed</span>, has a MIME type (ignoring parameters)
+  that is `<code title>application/x-www-form-urlencoded</code>`,
+  `<code title>multipart/form-data</code>`, or `<code title>text/plain</code>`
+</ul>
+
 <!-- XXX * needs better xref
          * ignoring parameters has been the standard for a long time now
          * interesting test: "Content-Type: text/plain;" -->
 
+<p>or whose <span title=concept-header-name>name</span> is one of:
```
annevk: Same here.

Overview.src.html
```diff
@@ -1851,10 +1886,54 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   for HTTP/2, and equivalent information used to prioritize dispatch and processing of
   HTTP/1 fetches.
 
+
+ <li><p>If <var>request</var> is a <span>navigation request</span>, a user agent should <span title=concept-header-list-append>append</span> the following pairs to <var>request</var>'s
+   <span title=concept-request-header-list>header list</span>:
```
annevk: I wonder if we should do the for each here too, that might make it read a little better.

A user agent should, for each header name /x/ in the first column of the following of the following table, if /x/ is not in request's header list, append /x/ / a value given in the same row on the second column, to request's header list.

Overview.src.html
```diff
@@ -1851,10 +1886,54 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+    is `<code title=http-save-data>Save-Data</code>`
+    <li>`<code title=http-viewport-width>
+    <a href="http://httpwg.org/http-extensions/client-hints.html#viewport-width">Viewport-Width</a></code>`/
+    <var><a href="http://httpwg.org/http-extensions/client-hints.html#viewport-width">definition</a></var>,
+    unless the <span title=concept-request-header-list>header list</span> already contains a
+    <span title=concept-header>header</span> whose <span title=concept-header-name>name</span> is
+    `<code title=http-viewport-width>Viewport-Width</code>`</var>
+  </ol>
+
+  <p>where <var>value</var> is set according to the definition in <span data-anolis-ref>CLIENT-HINTS</span>.
```
annevk: This is no longer correct.

annevk:
Also, the URLs you use don't really give information on the value. And don't use HTTPS.

igrigorik:
Updated to use a table and fixed the links (good catch!). Hopefully we're getting closer :)

/cc @mnot https://httpwg.org is timing out for me. Is that a known issue?

annevk:
1. The URL fragments seem incorrect. You hyphenate the section titles but the fragment is actually shorter.
2. Those sections should probably say they define the header field value. A header field consists of a name and a value.
3. Rather than referencing them as "definition" it would be clearer to use `a suitable <a><code>Width</code> value</a>` or some such.

annevk:
If you could also patch https://github.com/whatwg/xref to add the new reference that'd be great. I'll help out a bit with the final formatting once most is in shape. And yeah, I think it's pretty close.

igrigorik:
> The URL fragments seem incorrect. You hyphenate the section titles but the fragment is actually shorter.

Oi.. Not sure what happened there. Pushed a fix.

> Those sections, in your draft, should probably say they define the header field value. A header field consists of a name and a value.

Sounds reasonable, I'll take that as a separate AI.

> Rather than referencing them as "definition" it would be clearer to use a suitable <a><code>Width</code> value</a> or some such.

Replaced "definition" with "suitable value", which links to the definition. Does that look better? I didn't want to repeat the header name.

Re, xref: https://github.com/whatwg/xref/pull/10

---

Do you want me to squash the changes?

annevk:
The problem with not repeating the name of the header (though technically you'd be giving the name of the value production, not the header name) is that then you have six or so links that have an identical name. That is terrible UX for people with screen readers.

annevk:
> I'll take that as a separate AI

Not sure I understood, should I file an issue somewhere?

igrigorik:
> The problem with not repeating the name of the header (though technically you'd be giving the name of the value production, not the header name) is that then you have six or so links that have an identical name. That is terrible UX for people with screen readers.

Ah, that's fair. Updated.

> Not sure I understood, should I file an issue somewhere?

I'll tackle it as part of https://github.com/httpwg/http-extensions/issues/156#issuecomment-205411075.

Anything else left?

Overview.src.html
```diff
@@ -336,17 +336,35 @@ <h4 id=terminology-headers>Headers</h4>
...
          * interesting test: "Content-Type: text/plain;" -->
 
+<p>or whose <span title=concept-header-name>name</span> is one of
+
+<ul class=brief>
+  <li>`<code title><a href=http://httpwg.org/http-extensions/client-hints.html#dpr>DPR</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#downlink>Downlink</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#save-data>Save-Data</a></code>`,
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>Viewport-Width</a></code>`, or
+  `<code title><a href=http://httpwg.org/http-extensions/client-hints.html#width>Width</a></code>`
```
annevk: These should all have their own `<li>`.

Having said that, I would be okay with taking your text, modifying it a bit to match existing editorial style, and then committing it (with you still listed as author of course). The only downside of that is that this PR will be red, not purple. If that is a concern, I'll continue nitpicking a bit.

igrigorik:
@annevk split the long list into separate li's and squashed the commits. In terms of next steps: whichever is easier for you. I'm happy to continue updating, or you can take it and run with it.. the "color" on the PR doesn't bother me :)

Overview.src.html
```diff
@@ -805,6 +823,15 @@ <h4 id=terminology-headers>Headers</h4>
 <span data-anolis-ref>REFERRER</span>
 
 <p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-client-hints-list>client hints list</dfn>, which is a
+<span title=concept-client-hints-list>client-hints list</span>. Unless stated otherwise, it is
+the empty list.
+
+<p class="note no-backref">This can be used to override a client hints list associated with
+an <span data-anolis-spec=html>environment settings object</span>.
+<span data-anolis-ref>CLIENT-HINTS</span>
```
annevk: CLIENT-HINTS doesn't define a client hints list associated with an environment settings object. What exactly is the purpose of this?

Overview.src.html
```diff
@@ -324,17 +324,35 @@ <h4 id=terminology-headers>Headers</h4>
...
-`<code title>Accept-Language</code>`, or `<code title>Content-Language</code>`, or whose
-<span title=concept-header-name>name</span> is `<code title>Content-Type</code>` and
-<span title=concept-header-value>value</span>, <span title=concept-header-parse>once parsed</span>,
-has a MIME type (ignoring parameters) that is
-`<code title>application/x-www-form-urlencoded</code>`, `<code title>multipart/form-data</code>`, or
-`<code title>text/plain</code>`.
+<span title=concept-header-name>name</span> is one of
+
+<ul class=brief>
+  <li>`<code title>Accept</code>`
```
annevk: Here and elsewhere, single-space indentation should be used.

Overview.src.html
```diff
@@ -1888,10 +1924,54 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+  </table>
+
  <li>
   <p>If <var>request</var> is a <span>subresource request</span>, run these substeps:
 
   <ol>
+    <li><p>If the <var>request</var>'s <span title=concept-client-hint-list>client hints list</span> is not empty, run these substeps for each <var>hint-name</var> in the list:
+      <ol>
+        <li><p>Set <var>value</var> to the first matching statement, if any, switching on <var>hint-name</var>:
```
annevk: If an `<li>` or any other element has multiple "block"-level children, those should appear on their own line, and be indented once.

Overview.src.html
igrigorik: Right, that part hasn't been specced yet. My intent is to mirror similar language as [referrer policy](https://w3c.github.io/webappsec-referrer-policy/#terms).

- We can either do that within the CH draft.
- Or, we can integrate that language directly into the HTML spec?

Any preference on one vs other? /cc @mnot

igrigorik:
@annevk ack on formatting, see 1185bc1.

Overview.src.html
annevk: Not really, as long as it's clearly defined. I guess it would be good to state in the note that this is not yet defined. Or say "will be used" or some such. Then we can update it later once things are defined.

Overview.src.html
```diff
@@ -1387,6 +1414,15 @@ <h3 id="should-response-to-request-be-blocked-due-to-mime-type?"><dfn title="sho
 <span data-anolis-ref>REFERRER</span>
 
 
+<h3>Client hints list</h3>
+
+<p>A <dfn title=concept-client-hints-list>client hints list</dfn> is a list of
+<a href=http://httpwg.org/http-extensions/client-hints.html#accept-ch>Client hint tokens</a>,
+each of which is one of `<code title>dpr</code>`,
+`<code title>save-data</code>`, `<code title>viewport-width</code>`, or
+`<code title>width</code>`.
```
annevk: It might make sense to define this concept ("client hints list") directly in CLIENT-HINTS too, especially if it is going to be reused by multiple concepts.

Overview.src.html
```diff
@@ -1888,10 +1924,59 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <tr>
+   <td>`<code title=http-dpr>dpr</a></code>`
+   <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
+  <tr>
+   <td>`<code title=http-save-data>save-data</code>`
+   <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
+  <tr>
+   <td>`<code title=http-viewport-width>viewport-width</code>`
+   <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+  </table>
```
annevk: The `<tr>`s and `<th>`/`<td>`s need to be indented one more.

Overview.src.html
```diff
@@ -1888,10 +1924,59 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
  <li>
   <p>If <var>request</var> is a <span>subresource request</span>, run these substeps:
 
   <ol>
+   <li><p>If the <var>request</var>'s <span title=concept-client-hint-list>client hints list</span> is not empty, run these substeps for each <var>hint-name</var> in the list:
```
annevk: Here the `<p>` needs to be on a newline and indented one.

Overview.src.html
```diff
@@ -1888,10 +1924,59 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+    <ol>
+     <li>
+      <p>Set <var>value</var> to the first matching statement, if any, switching on
+      <var>hint-name</var>:
+
+       <dl class=switch>
```
annevk: It seems this `<dl>` is indented one too many.

annevk: And its children too.

Overview.src.html
```diff
@@ -1888,10 +1924,59 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+        <dt>"<code title>save-data</code>"
+        <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
+        <dt>"<code title>viewport-width</code>"
+        <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+        <dt>"<code title>width</code>"
+        <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#width>width value</a></code>
+       </dl>
+
+     <li>
+      <p><span title=concept-header-list-append>Append</span>
```
annevk: This `<p>` doesn't need to be on a newline since it's the only child of the `<li>`.

Overview.src.html
igrigorik: Seems reasonable. Do you want me to drop this from here then? Or, should we remove this and add a reference once we have the other bits specc'ed in CH/HTML spec.

igrigorik:
@annevk ack on formatting, ptal. 

Also, as an aside: it'd be great to have a section in the readme, or a standalone doc, with formatting conventions.

Overview.src.html
annevk: Either way is fine with me. If you want to go with the former approach I'd appreciate you adding `<!-- XXX this needs to move to CLIENT-HINTS -->` after this `<p>`.

annevk:
Thanks for mentioning that. I created https://gist.github.com/annevk/2f51a1082850a2814204a9a8a07f9654 as a start. What do you think? It's not entirely clear to me yet where to best host it, since these guidelines are shared across many repositories, but maybe here is fine in a FORMATTING.md file and then the other repositories can point to it. (whatwg/html will need a slightly different one that I can also write I suppose.) @domenic @foolip thoughts?

domenic:
Seems OK for DOM and partially for HTML. You should probably also mention your preferences on end tag omission and attribute quoting omission.

I don't think any of those guidelines apply to streams, e.g. there we use 120 chars and Markdown-style lists (in both Bikeshed and with the Ecmarkup integration). So probably introduce this on a repository-by-repository basis.

foolip:
I suppose we'll need one style for HTML and one for all other WHATWG specs? Media Session uses 80 columns, FWIW.

Also, the 'Do not use newlines inside "inline" elements' rule seems weird, that's not what we do in HTML.

annevk:
Yeah, I figure we need multiple for now. Was just wondering about reuse but I suppose specifications that reuse can point at each other their policy just as we do now.

igrigorik:
@annevk added the XXX/TODO and squashed -- think it should be good to go!

Re, gist: yes, that's exactly what I was looking for. Now, if only we could coerce tidy into automating this for us... that would be super.

annevk:
I landed this as https://github.com/whatwg/fetch/commit/404dc3acfddb5e9870fdee333302f80ecff661dd. Purple PR was not possible anyway due to `Overview.html` not being generated.

I changed the XXX comment to a note and moved it to a more appropriate place.

igrigorik:
:+1: looks great, thanks Anne!

PR #261, opened 2016-03-23
Block 'image/*' resources loaded as script.

mikewest:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27852. Chrome implemented this change in
https://bugs.chromium.org/p/chromium/issues/detail?id=433049, and has been shipping it
for over a year.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -1748,7 +1748,6 @@ response <span title=concept-header>header</span> can be used to require checkin
  <li><p>Return <b title>allowed</b>.
 </ol>
 
-
 <h2>Fetching</h2>
 
 <div class="note no-backref">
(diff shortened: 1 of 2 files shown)


domenic:
Not necessarily part of this commit (@annevk?), but at some point we should probably move HTML's blocking of module scripts with non-JavaScript MIME types here as well.

We may also want to add a NOTE in HTML calling out that this will happen.

annevk:
@domenic can we even move that here? We'd have to introduce new fetching parameters for that I think, if we wanted to do that.

Overview.src.html
```diff
@@ -3358,6 +3365,27 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ <span title=concept-header-extract-mime-type>extracting a MIME type</span> from
+ <var>response</var>'s <span title=concept-response-header-list>header list</span>.
+
+ <li><p>Let <var>type</var> be <var>request</var>'s
+ <span title=concept-request-type>type</span>.
+
+ <li><p>If <var>type</var> is "<code title>script</code>", and <var>MIMEType</var>
+ begins with `<code title>image/</code>`, return <b title>blocked</b>.
+
+ <li><p class=XXX>"<code title>audio</code>", "<code title>video</code>" ...
```
annevk: Is this actually something we want to do? Maybe just leave this out until we actually get there?

Overview.src.html
```diff
@@ -3358,6 +3365,27 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+<ol>
+ <li><p>Let <var>MIMEType</var> be the result of
+ <span title=concept-header-extract-mime-type>extracting a MIME type</span> from
+ <var>response</var>'s <span title=concept-response-header-list>header list</span>.
+
+ <li><p>Let <var>type</var> be <var>request</var>'s
+ <span title=concept-request-type>type</span>.
+
+ <li><p>If <var>type</var> is "<code title>script</code>", and <var>MIMEType</var>
+ begins with `<code title>image/</code>`, return <b title>blocked</b>.
```
annevk: Let's do "starts with" after JavaScript. Also, ", then return ..."

Overview.src.html
```diff
@@ -2040,15 +2039,23 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
- <a href=https://w3c.github.io/webappsec-csp/#should-block-response>should <var>internalResponse</var> to <var>request</var> be blocked by Content Security Policy</a>,
- or
- <span title="should response to request be blocked due to nosniff">should <var>internalResponse</var> to <var>request</var> be blocked due to nosniff</span>,
- returns <b title>blocked</b>, set <var>response</var> and <var>internalResponse</var> to a
- <span title=concept-network-error>network error</span>.
- <span data-anolis-ref>MIX</span>
- <span data-anolis-ref>CSP</span>
+ <li>
+  <p>If <var>response</var> is not a <span title=concept-network-error>network error</span> and
+  any of the following algorithms return <b title>blocked</b>, then set <var>response</var> and
```
annevk: returns*

mikewest:
Addressed feedback and squished, thanks!

annevk:
Thank you, landed as https://github.com/whatwg/fetch/commit/b561a73f23b7139d5f4bb0de322dc37e74af0091. Ended up moving the section elsewhere since thus far only CORS is part of "Fetching" proper. Not that it matters much.

mikewest:
Wunderbar, thanks.

domenic:
> can we even move that here? We'd have to introduce new fetching parameters for that I think, if we wanted to do that.

Yeah, we would. Which seems OK? Just another destination, right?

@mikewest looking back at https://www.w3.org/Bugs/Public/show_bug.cgi?id=27852 and at this PR, it looks like what you've specced blocks `<script>`, `Worker`, `SharedWorker`, `ServiceWorker`, and `importScripts`. Good!

annevk:
@domenic yeah, I guess we could give modules their own destination. I wonder if with Web Assembly that means we'll end up with even more destinations, taking into account that these multiply due to all the different workers.

PR #265, opened 2016-03-28
Change how CORS filtered response filters the headers.

mkruisselbrink:
Rather than just directling taking the exposed headers from the
`Access-Control-Expose-Headers` header, this adds a separate list of
exposed headers to a response, initialized from that header. This makes
it possible for foreign fetch to filter headers differently.

This will be used by foreign fetch in mkruisselbrink/ServiceWorker@bf4585bc6afb9e0b1cefdeff97f69595545a1af4

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -381,9 +381,9 @@ from APIs using <span title=concept-fetch>fetch</span> that allow control over
  <li>`<code title>Set-Cookie2</code>`
 </ul>
 
-<p>A <dfn id="cors-safelisted-response-header-name">CORS-safelisted response-header name</dfn>, given a
-<span title=concept-response-cors-exposed-headers-list>CORS-exposed headers list</span> <var>list</var>, is a
-<a href="#concept-header" title="concept-header">header</a> <a href="#concept-header-name" title="concept-header-name">name</a> that is one of:
+<p>A <dfn>CORS-safelisted response-header name</dfn>, given a
+<span title=concept-response-cors-exposed-header-name-list>CORS-exposed header-name list</span> <var>list</var>, is a
+<span title=concept-header>header</span> <span title=concept-header-name>name</span> that is one of:
 
 <ul>
  <li>`<code title>Cache-Control</code>`
@@ -392,8 +392,8 @@ from APIs using <span title=concept-fetch>fetch</span> that allow control over
  <li>`<code title>Expires</code>`
  <li>`<code title>Last-Modified</code>`
  <li>`<code title>Pragma</code>`.
- <li>Any <a href="#concept-header-value" title="concept-header-value">value</a> in
- <var>list</var> that is not a <a href="#forbidden-response-header-name">forbidden response-header name</a>.
+ <li>Any <span title=concept-header-value>value</span> in
+ <var>list</var> that is not a <span>forbidden response-header name</span>.
 </ul>
 
 <hr>
@@ -1041,10 +1041,17 @@ for the <span title=concept-response>response</span>. The list is empty unless o
 specified. <span data-anolis-ref>CSP</span>
 
 <p>A <span title=concept-response>response</span> has an associated
-<dfn title=concept-response-cors-exposed-headers-list>CORS-exposed headers list</dfn>
+<dfn title=concept-response-cors-exposed-header-name-list>CORS-exposed header-name list</dfn>
 (a list of zero or more
 <span title=concept-header>header</span> <span title=concept-header-name>names</span>).
 The list is empty unless otherwise specified.
+
+<p class="note no-backref">A <span title=concept-response>response</span> will typically get its
+<span title=concept-response-cors-exposed-header-name-list>CORS-exposed header-name list</span>
+set by <span title=concept-header-parse>parsing</span> the
+`<code title=http-access-control-expose-headers>Access-Control-Expose-Headers</code>` header. This
+list is used by a <span title=concept-filtered-response-cors>CORS filtered response</span> to
+determine which headers to expose.
 
 <hr>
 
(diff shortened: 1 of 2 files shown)


Overview.src.html
```diff
@@ -381,9 +381,9 @@ <h4 id=terminology-headers>Headers</h4>
  <li>`<code title>Set-Cookie2</code>`
 </ul>
 
-<p>A <dfn>CORS-safelisted response-header name</dfn>, given a
-<span title=concept-response-header-list>header list</span> <var>list</var>, is a
-<span title=concept-header>header</span> <span title=concept-header-name>name</span> that is one of:
+<p>A <dfn id="cors-safelisted-response-header-name">CORS-safelisted response-header name</dfn>, given a
```
annevk: Why did you add an id here?

Overview.src.html
```diff
@@ -1042,6 +1040,12 @@ <h4 id=terminology-headers>Headers</h4>
 for the <span title=concept-response>response</span>. The list is empty unless otherwise
 specified. <span data-anolis-ref>CSP</span>
 
+<p>A <span title=concept-response>response</span> has an associated
+<dfn title=concept-response-cors-exposed-headers-list>CORS-exposed headers list</dfn>
```
annevk: s/headers/header-name/.

Overview.src.html
```diff
@@ -1042,6 +1040,12 @@ <h4 id=terminology-headers>Headers</h4>
...
+(a list of zero or more
+<span title=concept-header>header</span> <span title=concept-header-name>names</span>).
+The list is empty unless otherwise specified.
+
```
annevk: Can you add a note here that says this is only used for CORS filtered responses?

Overview.src.html
```diff
@@ -2036,6 +2040,18 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <span title=concept-header-parse>parsing</span>
+  `<code title=http-access-control-expose-headers>Access-Control-Expose-Headers</code>` in
+  <var>response</var>'s
+  <span title=concept-response-header-list>header list</span>.
+
+ <li>
+  <p>If <var>response</var> is not a
+  <span title=concept-network-error>network error</span> and
+  <var>response</var> is not a
+  <span title=concept-filtered-response>filtered response</span>, set
```
annevk: I think we should try to find a solution that creates a CORS filtered response in a single step. Or at least a set of steps that belong together. E.g., the check for network error and filtered response is only needed once, you could then run substeps. But maybe there's a way to do the `<dl>` differently as well that cleans this up a bit.

Overview.src.html
```diff
@@ -381,9 +381,9 @@ <h4 id=terminology-headers>Headers</h4>
  <li>`<code title>Set-Cookie2</code>`
 </ul>
 
-<p>A <dfn>CORS-safelisted response-header name</dfn>, given a
-<span title=concept-response-header-list>header list</span> <var>list</var>, is a
-<span title=concept-header>header</span> <span title=concept-header-name>name</span> that is one of:
+<p>A <dfn id="cors-safelisted-response-header-name">CORS-safelisted response-header name</dfn>, given a
+<span title=concept-response-cors-exposed-headers-list>CORS-exposed headers list</span> <var>list</var>, is a
+<a href="#concept-header" title="concept-header">header</a> <a href="#concept-header-name" title="concept-header-name">name</a> that is one of:
```
annevk: This also looks wrong. It seems like you copied output from `Overview.html` and put it in `Overview.src.html`. Did Anolis fail for some reason? If you cannot use Anolis, please only submit a patch to `Overview.src.html` and leave `Overview.html` to me. See the README for instructions.

Overview.src.html
```diff
@@ -392,10 +392,8 @@ <h4 id=terminology-headers>Headers</h4>
  <li>`<code title>Expires</code>`
  <li>`<code title>Last-Modified</code>`
  <li>`<code title>Pragma</code>`.
- <li>Any <span title=concept-header-value>value</span> resulting from
- <span title=concept-header-parse>parsing</span>
- `<code title=http-access-control-expose-headers>Access-Control-Expose-Headers</code>` in
- <var>list</var> that is not a <span>forbidden response-header name</span>.
+ <li>Any <a href="#concept-header-value" title="concept-header-value">value</a> in
+ <var>list</var> that is not a <a href="#forbidden-response-header-name">forbidden response-header name</a>.
```
annevk: Same here.

Overview.src.html
mkruisselbrink: Added a note explaining what this is and what it's used for.

Overview.src.html
mkruisselbrink: Yeah, my mistake. Also happy to leave out the Overview.html changes if that's what you prefer, although that wouldn't have prevented my from stupidly managing to copy stuff from the generated html into the source file in an attempt to undo earlier changes I made...

Overview.src.html
mkruisselbrink: For now I've just separated out the check for network error/filtered response, and then have substeps to set the CORS-exposed header-name list and create the filtered response. I agree that somehow making creating a CORS filtered response more of a single step would be nice, but I'm not sure how best to do so...

Overview.src.html
```diff
@@ -2035,22 +2046,34 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+  <span title=concept-filtered-response>filtered response</span>, run these substeps:
+
+  <ol>
+   <li>
+    <p>Set <var>response</var>'s
+    <span title=concept-response-cors-exposed-header-name-list>CORS-exposed header-name list</span>
+    to the result of <span title=concept-header-parse>parsing</span>
+    `<code title=http-access-control-expose-headers>Access-Control-Expose-Headers</code>` in
+    <var>response</var>'s
+    <span title=concept-response-header-list>header list</span>.
```
annevk: Although it does not matter in a black-box observable way, I still think we should add a conditional here for response tainting being "cors".

annevk:
This looks pretty good. It could use some rewrapping here and there to account for 100 column lines (while keeping inline elements on one line at all cost, even if it requires exceeding the limit after putting them on their own line), but I'm happy to do that for you.

Overview.src.html
mkruisselbrink: Done

mkruisselbrink:
> This looks pretty good. It could use some rewrapping here and there to account for 100 column lines (while keeping inline elements on one line at all cost, even if it requires exceeding the limit after putting them on their own line), but I'm happy to do that for you.

Thanks. I tried to do wrapping that seemed to match most of the rest of the document, but yeah, it's certainly possible that I missed something/wrapped something differently than you'd like.

annevk:
Thank you, merged as https://github.com/whatwg/fetch/commit/32411c753f46a10fd20d2ad3610244e7fdf44d4a.

PR #280, opened 2016-04-12
Update request's referrer policy when fetching

### LINKED ISSUE #266: Referrer policy: Should request's referrer policy be updated in the main fetch?
The spec says:

If request's referrer is not "no-referrer", set request's referrer to the result of invoking determine request's referrer.

The "determine request's referrer" algorithm combines request's referrer policy and environment's referrer policy, and then use it to calculate the referrer string. Is it reasonable to update the request's referrer policy to the combined referrer policy? That change is observable when the request is intercepted by the service worker.

@mikewest @jeisinger

yutakahirano:
With this change request's referrer policy is updated before calculating referrer so that the referrer policy actually used for the referrer calculation is transferred to the service worker and the default referrer policy is not transferred to the service worker.

This change fixes #266.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -2035,7 +2035,9 @@ with a <i title>CORS flag</i> and <i title>recursive flag</i>, run these steps:
  is the empty string and <var>request</var>'s
  <span title=concept-request-client>client</span> is non-null, then set <var>request</var>'s
  <span title=concept-request-referrer-policy>referrer policy</span> to <var>request</var>'s
- <span title=concept-request-client>client</span>'s associated referrer policy.
+ <span title=concept-request-client>client</span>'s
+ <a href="https://w3c.github.io/webappsec-referrer-policy/#referrer-policy">associated referrer policy</a>.
+ <span data-anolis-ref>REFERRER</span>
 
  <li>
   <p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -2033,14 +2033,18 @@ with a <i title>CORS flag</i> and <i title>recursive flag</i>, run these steps:
 
  <li><p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
  is the empty string and <var>request</var>'s
- <span title=concept-request-client>client</span> is non-null, set <var>request</var>'s
+ <span title=concept-request-client>client</span> is non-null, then set <var>request</var>'s
  <span title=concept-request-referrer-policy>referrer policy</span> to <var>request</var>'s
  <span title=concept-request-client>client</span>'s associated referrer policy.
 
- <li><p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
- is the empty string, set <var>request</var>'s
- <span title=concept-request-referrer-policy>referrer policy</span> to
- "<code>no-referrer-when-downgrade</code>".
+ <li>
+  <p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
+  is the empty string, then set <var>request</var>'s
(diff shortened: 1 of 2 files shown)


yutakahirano:
@annevk @estark37 @jeisinger

yutakahirano:
With this change item 3.2 in https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer can be rewritten as "Let policy be requests's referrer policy", I think (see also: https://github.com/w3c/webappsec-referrer-policy/issues/26).

Overview.src.html
```diff
@@ -2031,6 +2031,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <span data-anolis-ref>MIX</span>
  <span data-anolis-ref>CSP</span>
 
+ <li><p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
+ is the empty string and <var>request</var>'s
+ <span title=concept-request-client>client</span> is non-null, set <var>request</var>'s
+ <span title=concept-request-referrer-policy>referrer policy</span> to <var>request</var>'s
+ <span title=concept-request-client>client</span>'s associated referrer policy.
```
annevk: Where is "associated referrer policy" defined? Would be nice to directly reference it.

Overview.src.html
```diff
@@ -2031,6 +2031,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <span data-anolis-ref>MIX</span>
  <span data-anolis-ref>CSP</span>
 
+ <li><p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
+ is the empty string and <var>request</var>'s
+ <span title=concept-request-client>client</span> is non-null, set <var>request</var>'s
```
annevk: "then set"

Overview.src.html
```diff
@@ -2031,6 +2031,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ <span title=concept-request-referrer-policy>referrer policy</span> to <var>request</var>'s
+ <span title=concept-request-client>client</span>'s associated referrer policy.
+
+ <li><p>If <var>request</var>'s <span title=concept-request-referrer-policy>referrer policy</span>
+ is the empty string, set <var>request</var>'s
```
annevk: "then set"

Overview.src.html
```diff
@@ -2031,6 +2031,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+ <span title=concept-request-referrer-policy>referrer policy</span> to
+ "<code>no-referrer-when-downgrade</code>".
```
annevk: We should maybe add a note here that this is the historical default.

Overview.src.html

Overview.src.html

Overview.src.html

Overview.src.html
yutakahirano: I expected it would be added to https://www.w3.org/TR/html5/webappapis.html#settings-object by webappsec-referrer-policy authors in the future. Should I link to https://w3c.github.io/webappsec-referrer-policy/#referrer-policy?

annevk: Yeah, I think that would be good enough for now. Bit weird that they link to W3C HTML, which is hopelessly wrong.

annevk:
Would love a final sign-off by @estark37 or @jeisinger too just to make sure we're not missing anything. (When you look at the diff, search for "Overview.src.html" for the easier to review set of changes.)

Overview.src.html
yutakahirano: Added a link.

estark37:
This looks good to me.

annevk:
Thank you both!

PR #283, opened 2016-04-14
Response construction with ReadableStream

yutakahirano:
This change enables constructing a Response from ReadableStream, as discussed in
https://github.com/yutakahirano/fetch-with-streams/issues/63.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -3842,7 +3842,7 @@ running these steps:
 <h3>Body mixin</h3>
 
 <pre class=idl>typedef any <dfn>JSON</dfn>;
-typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString) <dfn>RequestBodyInit</dfn>;
+typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString) <dfn>BodyInit</dfn>;
 typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString or <code title=concept-ReadableStream>ReadableStream</code>) <dfn>ResponseBodyInit</dfn>;</pre>
 
 <p>To <dfn title=concept-BodyInit-extract>extract</dfn> a <span title=concept-body>body</span> and a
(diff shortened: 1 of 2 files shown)


Overview.src.html
```diff
@@ -3842,7 +3842,8 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 <h3>Body mixin</h3>
 
 <pre class=idl>typedef any <dfn>JSON</dfn>;
-typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString) <dfn>BodyInit</dfn>;</pre>
+typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString) <dfn>RequestBodyInit</dfn>;
+typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString or <code title=concept-ReadableStream>ReadableStream</code>) <dfn>ResponseBodyInit</dfn>;</pre>
```
annevk: Why not `typedef (<span>BodyInit</span> or <span title=concept-ReadableStream>ReadableStream</span>) <dfn>ResponseBodyInit</dfn>` or just inline that for the `Response` constructor since it's only used once? Changing `BodyInit` is problematic since other specifications use it too.

yutakahirano: Fixed. I'd like to define ResponseBodyInit here because with ResponseBodyInit it's easier to understand why we have ReadableStream clause in "extract a body and a Content-Type value" algorithm.

annevk: Fair.

Overview.src.html
```diff
@@ -3842,7 +3842,8 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 <h3>Body mixin</h3>
 
 <pre class=idl>typedef any <dfn>JSON</dfn>;
-typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString) <dfn>BodyInit</dfn>;</pre>
+typedef (<span data-anolis-spec=fileapi>Blob</span> or BufferSource or <span data-anolis-spec=xhr>FormData</span> or <span data-anolis-spec=url>URLSearchParams</span> or USVString) <dfn>BodyInit</dfn>;
+typedef (<span>BodyInit</span> or <code title=concept-ReadableStream>ReadableStream</code>) <dfn>ResponseBodyInit</dfn>;</pre>
```
annevk: I was about to merge, but you need to use `<span>` inside `<pre>`, not `<code>`. Otherwise the coloring gets weird. A `<pre class=idl>` block is code already in a way.

yutakahirano: I see, thanks. Fixed.

annevk:
\o/

yutakahirano:
Thank you!

PR #291, opened 2016-05-03
Editorial: clarify cache match conditions

### LINKED ISSUE #281: "There is a cache match for request if origin is..."
https://fetch.spec.whatwg.org/#cors-preflight-cache

> There is a cache match for request if origin is request's origin, url is request's current url, and either credentials is false and request's credentials mode is not "include" or credentials is true\.

The logical grouping here is ambiguous.

annevk:
Fixes #281.


annevk:
@mnot thanks for reporting, is this better?

mnot:
LGTM, if that matches reality. Thx!

Issue #264, opened 2016-03-24
Provide more information in the spec on how withCredentials and Access-Control-Allow-Credentials interact


roryhewitt:
@annevk following on from our discussion at #251 I think it would make sense to include some more information in the spec on how `withCredentials` and `Access-Control-Allow-Credentials` interact. Based also, perhaps on [this thread](https://lists.w3.org/Archives/Public/www-tag/2016Feb/0007.html) from @mnot and @sicking.

Basically, I think there should be two separate sections - one for client-side developers discussing why they should (or should not) specify `withCredentials=  true` and _exactly_ what it will affect (cookies, Authorization header, in-URL credentials etc.) and a separate section for back-end developers, indicating the circumstances under which they should return the `Access-Control-Allow-Credentials` header.

As an example of why I think we need separate sections, one of the first websites found via Google when I search for CORS is http://www.html5rocks.com/en/tutorials/cors/. This site (which is generally excellent as a CORS tutorial) includes the following erroneous information:

> The `Access-Control-Allow-Credentials` header works in conjunction with the `withCredentials` property on the XMLHttpRequest 2 object. Both these properties must be set to true in order for the CORS request to succeed. If `.withCredentials` is true, but there is no `Access-Control-Allow-Credentials` header, the request will fail (and vice versa).

In this case, the vice-versa case mentioned is incorrect - if `Access-Control-Allow-Credentials: true` is returned, the request will succeed, whether or not `.withCredentials` is specified (ignoring any ACAO issues, obviously).

For my part, this was one of the first sites I looked at back in 2012 to gain more information about CORS, and this statement caused me to include all sorts of server-side code to check for the existence of cookies, Authorization header etc. in order to determine whether to return `Access-Control-Allow-Credentials: true`, since my fear was that if I returned it in every case, the request would fail.

Sure, you could say that I should have looked at the the actual [W3C spec](https://www.w3.org/TR/cors/#resource-sharing-check), which correctly doesn't say that, but nor does it say exactly what counts as _credentials_ (beyond a somewhat vague "cookies, HTTP authentication, and client-side SSL certificates"). And a newbie to CORS (from either server-side or client-side) is likely to refer to 'user-friendly' websites as they are to the actual spec. More likely, actually. Finalkly, of course, that site refers (as do most 'CORS tutorials'!) to the W3C spec...

I'm thinking that we could provide some scenarios discussing the following:

[...]

PR #293, opened 2016-05-03
Explain CORS protocol and credentials interaction

annevk:
WIP for #264.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-4-may-2016">Living Standard — Last Updated 4 May 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-5-may-2016">Living Standard — Last Updated 5 May 2016</h2>
 
 <dl>
  <dt>Participate:
@@ -119,7 +119,8 @@ fetching.
 
 <h2 class="no-num short" id="goals">Goals</h2>
 
-<p>To unify fetching across the web platform this specification supplants a number of algorithms and specifications:
+<p>To unify fetching across the web platform this specification supplants a number of algorithms and
+specifications:
 
 <ul class="brief no-backref">
  <li>HTML Standard's fetch and potentially CORS-enabled fetch algorithms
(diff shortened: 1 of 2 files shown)


annevk:
Did not have enough time to add all the interesting examples to the table today, but this should give a general outline.

@roryhewitt thoughts?

Overview.src.html
```diff
@@ -1773,6 +1782,76 @@ <h3 id=http-cors-protocol>CORS protocol</h3>
...
+client.withCredentials = true
+/* &hellip; */</pre>
+
+ <p>Nowadays, <code title>fetch("./", { credentials:"include" }).then(/* &hellip; */)</code>
+ suffices.
+</div>
+
+<p><span title=concept-request>Request</span>'s
+<span title=concept-request-credentials-mode>credentials mode</span> is not necessarily observable
+on the server, only when <span>credentials</span> exist for
```
domenic: s/,/;/

Overview.src.html
```diff
@@ -1773,6 +1782,76 @@ <h3 id=http-cors-protocol>CORS protocol</h3>
...
+<span title=concept-request>request</span> can it be observed. In such a scenario the
+<span>CORS request</span> would include <span>credentials</span>, but the
+<span>CORS-preflight request</span> would not. It never does.
+
+<p>The server developer therefore needs to decide whether or not resources "tainted" with
+<span>credentials</span> can be shared. Generally speaking, this is rather unsafe and extreme care
```
domenic: comma after unsafe

roryhewitt:
@annevk, this looks better. Thanks! I have some examples if you want 'em.

Overview.src.html
mnot: I found this really confusing on the first two reads, because of your use of "resource" -- which server-side means something very different than what you intend. How about:

    ...whether or not responses "tainted" with credentials...

and similar below.

annevk:
@roryhewitt examples are most welcome, especially if they're funny.

Overview.src.html
annevk: The whole section uses "resource" in this way. Would need some more substantive rewriting to avoid it.

annevk: I can maybe try doing that here.

annevk:
This is now ready I think. Another quick review appreciated (I ended up changing quite a bit due to @mnot's feedback). @roryhewitt, if the examples could be part of this, happy to take them. Otherwise we can also include them through a separate PR.

roryhewitt:
@annevk, what you have done is great, but I'd love to see some real user examples (from both a client and server perspective. The data in this document is great, but its understandably from a browser developer perspective, so it uses terminology that a client-side (JavaScript) developer or a server-side developer may not understand.

Attached is a document (*CORS request processing - Simple*) I wrote which explains (to a client-side developer) what they need to do, if anything, related to CORS, depending on their requirements. As such, it doesn't refer to e.g. "credentials mode", but instead refers to the "withCredentials" flag that an end-user would see in their JavaScript.

I thought I had a separate document referencing 'complex' CORS (including the preflight OPTIONS request), but I can't find it right now, but I can always write it. I don't know if examples like this could be included _somewhere_ in the WHATWG ecosphere - perhaps not directly in the fetch standard, but maybe referenced in there?

There's also a separate PowerPoint presentation about CORS which explains it (as best I can) to other technical people - that's just for 'fun'...


[CORS Overview.pptx](https://github.com/whatwg/fetch/files/249276/CORS.Overview.pptx)


[CORS request processing - Simple.docx](https://github.com/whatwg/fetch/files/249265/CORS.request.processing.-.Simple.docx)

Overview.html
```diff
@@ -1923,6 +1923,32 @@ <h4 id="cors-protocol-and-credentials"><span class="secno">4.2.5 </span>CORS pro
   <td>Omitted
   <td>❌
```
roryhewitt: This doesn't appear display correctly in Chrome - it just shows up as a question mark...

image

annevk:
You might have some font issues if even em dash shows up as a question mark. Or maybe the encoding is not utf-8 when you check it locally?

annevk:
I filed a separate issue for adding examples, I might work on that later.

Issue #241, opened 2016-03-10
Merge HTML5 and Fetch definitions of 'HTTPS state'


jwatt:
We have both:

https://fetch.spec.whatwg.org/#concept-response-https-state

and

https://html.spec.whatwg.org/multipage/webappapis.html#https-state

It would be good if fetch defined HTTPS state in a way that HTML5 could reuse.

domenic:
This doesn't really make sense. The one in HTML is a value stored on the settings object (~ the window) that is passed to certain Fetch requests made with that settings object. You can't get rid of HTML's since otherwise you wouldn't know what to pass to fetch.

In JavaScript programming terms, basically Fetch defines something like `specFetch(..., httpsState, ...)` and HTML defines `const currentSettingsObject = { ..., httpsState: <something>, ... }` and then later calls `specFetch(..., currentSettingsObject.httpsState, ... })`.

annevk:
I think what @jwatt meant is to define the set of values in one place.

jwatt:
Yeah, sorry, that could have been clearer.

PR #294, opened 2016-05-04
Editorial: make an HTTPS state value its own thing

annevk:
Fixes the Fetch part of #241.


annevk:
@jwatt, does this look okay?

jwatt:
It does, thanks!

PR #295, opened 2016-05-04
Add "only-if-cached" cache mode back

### LINKED ISSUE #159: Only-if-cached
As discussed in #39, it might be possible to bring back the `only-if-cached` mode if it's restricted to same-origin.

annevk:
Fixes #159. The difference with the previous attempt is that it now has
a “same-origin” mode restriction (enforced through the Fetch API).

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-13-may-2016">Living Standard — Last Updated 13 May 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-18-may-2016">Living Standard — Last Updated 18 May 2016</h2>
 
 <dl>
  <dt>Participate:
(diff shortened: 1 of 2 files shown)


annevk:
@ehsan @tyoshino @mnot FYI (and if you'd like to review, please).

ehsan:
LGTM!

Overview.html
```diff
@@ -990,8 +990,14 @@ <h4 id="requests"><span class="secno">3.1.5 </span>Requests</h4>
 
   <dt>"<code title="">force-cache</code>"
   <dd>Fetch uses any response in the HTTP cache matching the request, not paying attention to
-  staleness. If there was no response, it creates a normal request updates the HTTP cache with the
-  response.
+  staleness. If there was no response, it creates a normal request and updates the HTTP cache with
+  the response.
+
+  <dt>"<code title="">only-if-cached</code>"
+  <dd>Fetches uses any response in the HTTP cache matching the request, not paying attention to
```
mnot: Fetch uses

Overview.src.html
```diff
@@ -3029,9 +3036,18 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
- <li><p>If <var>response</var> is null, set <var>response</var> to the result of making an
- <span title=concept-http-network-fetch>HTTP-network fetch</span> using <var>httpRequest</var> with
- the <i title>credentials flag</i> set if set.
+ <li>
+  <p>If <var>response</var> is null, run these substeps:
+
+  <ol>
+   <li><p>If <var>httpRequest</var>'s <span title=concept-request-cache-mode>cache mode</span> is
+   "<code title>only-if-cached</code>", then return a
+   <span title=concept-network-error>network error</span>.
```
mnot: Http does this with a 504 - http://httpwg.org/specs/rfc7234.html#rfc.section.5.2.1.7

Considered aligning with that?

mnot:
what about the privacy aspect (cache probing)?

Overview.src.html
annevk: The difference would be that it would not surface as a rejection in the API meaning you'd have to inspect the response status code to know what's going on (which could theoretically be the same).

annevk:
Cache probing would only be possible same-origin since this feature is restricted to same-origin. It might make it slightly easier if there was an XSS exploit to figure out capability URLs perhaps, but that still seems rather hard and with an XSS exploit you likely have access to sensitive information anyway.

mnot:
LGTM

annevk:
Thanks!

PR #298, opened 2016-05-05
More wildcards in CORS when used without credentials

### LINKED ISSUE #251: Allow * for Access-Control-Allow-Headers and Access-Control-Allow-Methods
This has almost certainly been discussed before, but would it be possible to allow * (allow-all) as a separate value for the **Access-Control-Allow-Headers** CORS response header?

This would allow _all_ non-simple headers passed in the request to be added to the browser's preflight cache. This si currently possible by simply mirroring back the value of the **Access-Control-Request-Headers** request header, but this would be much simpler. The browser would need to track the request headers passed and add them all to their preflight cache (rather than simply parse them out from the **Access-Control-Allow-Headers** response header, assuming that's what they currently do), but that's not too hard to do.

So the spec would become the following:

Access-Control-Allow-Headers = "Access-Control-Allow-Headers" ":" #field-name | "*"

### LINKED ISSUE #252: Allow * for Access-Control-Expose-Headers
This has almost certainly been discussed before, but would it be possible to allow * (allow-all) as a separate value for the **Access-Control-Expose-Headers** CORS response header?

This would allow all non-simple headers passed in the request to be 'exposable' to client-side code.

When the original CORS spec was written, there was an assumption (I assume!) that the preflight OPTIONS response would be served by the same application that would serve the subsequent CORS GET/POST/whatever request. Therefore, that application would have full knowledge of what headers the API might eventually respond with, and which of them should be exposed or suppressed.

However, with the advent of CDN's and load balancers, it may be that the code that responds to the preflight OPTIONS request is **not** the actual application that will process the subsequent GET/POST/whatever request. Indeed, website owners might want the OPTIONS response to be served from non-application code, because that code may be quicker to respond to the client.

(Full disclosure: I work for Akamai, a large CDN. Whilst my views don't represent any official Akamai position, we would obviously like this, since it would make our (and out customer's) lives easier. I have also implemented CORS solutions using an F5 LTM load balancer, and run into this problem).

So the spec would become the following:

**Access-Control-Expose-Headers = "Access-Control-Expose-Headers" ":" #field-name | "*"**

If Access-Control-Expose-Headers: * is returned, browser would allow all headers to be returned using e.g. XMLHttpRequest getResponseHeader().

annevk:
Fixes #251 and fixes #252.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -1374,7 +1374,7 @@ functionality.
   <p>Let <var>connection</var> be the result of establishing an HTTP connection to
   <var>origin</var>. <a href="#refsHTTP">[HTTP]</a> <a href="#refsTLS">[TLS]</a>
 
-  <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.</p>
+  <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.
 
   <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), return failure.
 
@@ -1420,7 +1420,7 @@ clearly stipulates that <a href="#concept-connection" title="concept-connection"
 </ol>
 
 <p>A <a class="external" data-anolis-spec="url" href="https://url.spec.whatwg.org/#concept-url-port" title="concept-url-port">port</a> is a <dfn id="bad-port">bad port</dfn> if it
-is listed in the first column of the following table.</p>
+is listed in the first column of the following table.
 
 <table>
  <tr><th>Port<th>Typical service
@@ -2005,7 +2005,7 @@ fetch(url).then(success, failure)</pre>
  `<a href="#http-access-control-allow-origin"><code title="http-access-control-allow-origin">Access-Control-Allow-Origin</code></a>` response header.
  If its value is either `<code>https://foo.invalid</code>` or `<code>*</code>`, the user agent will
  invoke the <code>success</code> callback. If it has any other value, or is simply missing, the user
- agent will invoke the <code>failure</code> callback.</p>
+ agent will invoke the <code>failure</code> callback.
 </div>
 
 <div class="example">
@@ -2482,7 +2482,7 @@ with a <i title="">CORS flag</i> and <i title="">recursive flag</i>, run these s
    <li>
     <p>If <var>request</var>'s
     <a href="#concept-request-response-tainting" title="concept-request-response-tainting">response tainting</a> is
-    "<code title="">cors</code>", then run these substeps:</p>
+    "<code title="">cors</code>", then run these substeps:
 
     <ol>
      <li><p>Let <var>headerNames</var> be the result of
@@ -2496,12 +2496,12 @@ with a <i title="">CORS flag</i> and <i title="">recursive flag</i>, run these s
      <a href="#concept-response-cors-exposed-header-name-list" title="concept-response-cors-exposed-header-name-list">CORS-exposed header-name list</a>
      to all unique <a href="#concept-header" title="concept-header">header</a>
      <a href="#concept-header-name" title="concept-header-name">names</a> in <var>response</var>'s
-     <a href="#concept-response-header-list" title="concept-response-header-list">header list</a>.</li>
+     <a href="#concept-response-header-list" title="concept-response-header-list">header list</a>.
 
      <li><p>Otherwise, if <var>headerNames</var> is <em>not</em> `<code title="">*</code>`, then set
      <var>response</var>'s
      <a href="#concept-response-cors-exposed-header-name-list" title="concept-response-cors-exposed-header-name-list">CORS-exposed header-name list</a>
-     to <var>headerNames</var>.</p>
+     to <var>headerNames</var>.
     </ol>
 
    <li>
@@ -2580,7 +2580,7 @@ with a <i title="">CORS flag</i> and <i title="">recursive flag</i>, run these s
  <li>
   <p>If <var>request</var>'s <a href="#concept-request-current-url" title="concept-request-current-url">current url</a>'s
   <a class="external" data-anolis-spec="url" href="https://url.spec.whatwg.org/#concept-url-scheme" title="concept-url-scheme">scheme</a> is "<code title="">http</code>" or
-  "<code title="">https</code>", run these substeps:</p>
+  "<code title="">https</code>", run these substeps:
 
   <ol>
    <li><p>If <var>request</var>'s <a href="#concept-request-body" title="concept-request-body">body</a> is
@@ -3971,7 +3971,7 @@ fetch("/music/pk/altes-kamuffel.flac")
  <pre>var meta = { "Content-Type": "text/xml", "Breaking-Bad": "&lt;3" }
 new Headers(meta)</pre>
 
- <p>See <a href="https://github.com/whatwg/fetch/issues/164">issue #164</a> for discussion.</p>
+ <p>See <a href="https://github.com/whatwg/fetch/issues/164">issue #164</a> for discussion.
 </div>
 
 <pre class="idl">[<a href="#dom-headers" title="dom-Headers">Constructor</a>(optional <a href="#headersinit">HeadersInit</a> <var>init</var>),
@@ -5288,7 +5288,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <h3 id="websocket-opening-handshake"><span class="secno">7.2 </span>Opening handshake</h3>
 
 <p>To <dfn id="concept-websocket-establish" title="concept-websocket-establish">establish a WebSocket connection</dfn>, given a
-<var>url</var>, <var>protocols</var>, and <var>client</var>, run these steps:</p>
+<var>url</var>, <var>protocols</var>, and <var>client</var>, run these steps:
 
 <ol>
  <li>
@@ -5325,7 +5325,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
  <li>
   <p>Let <var>keyValue</var> be a nonce consisting of a randomly selected 16-byte value that has
   been base64-encoded (see <a href="https://tools.ietf.org/html/rfc4648#section-4">section 4</a> of
-  <a href="#refsRFC4648">[RFC4648]</a>).</p>
+  <a href="#refsRFC4648">[RFC4648]</a>).
 
   <p class="example">If the randomly selected value was the byte sequence 0x01 0x02 0x03 0x04 0x05
   0x06 0x07 0x08 0x09 0x0a 0x0b 0x0c 0x0d 0x0e 0x0f 0x10, <var>keyValue</var> would be
@@ -5356,7 +5356,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
  to <var>request</var>'s <a href="#concept-request-header-list" title="concept-request-header-list">header list</a>.
 
  <li><p>Let <var>response</var> be the result of <a href="#concept-fetch" title="concept-fetch">fetching</a>
- <var>request</var>.</li>
+ <var>request</var>.
 
  <li><p>If <var>response</var> is a <a href="#concept-network-error" title="concept-network-error">network error</a> or its
  <a href="#concept-response-status" title="concept-response-status">status</a> is not <code title="">101</code>,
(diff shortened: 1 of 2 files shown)


annevk:
@tyoshino review appreciated. @roryhewitt, @craigfrancis, @mozfreddyb, you might be interested in reviewing too. @sicking, I included your suggestion around `Authorization`.

Overview.html
```diff
@@ -1856,12 +1858,14 @@ <h4 id="http-new-header-syntax"><span class="secno">4.2.4 </span>HTTP new-header
 <pre>Access-Control-Request-Method    = <a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.1.1">method</a>
 Access-Control-Request-Headers   = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a>
 
-Access-Control-Allow-Origin      = origin-or-null / "*"
+wilrdcard                        = "*"
```
MichaelKohler: I highly suspect this to be a typo and it should read "wildcard"?

Overview.src.html
```diff
@@ -1785,12 +1787,14 @@ <h4 id=http-new-header-syntax>HTTP new-header syntax</h4>
 <pre>Access-Control-Request-Method    = <span data-anolis-spec=http>method</span>
 Access-Control-Request-Headers   = #<span data-anolis-spec=http>field-name</span>
 
-Access-Control-Allow-Origin      = origin-or-null / "*"
+wilrdcard                        = "*"
```
MichaelKohler: typo in "wilrdcard"

annevk: Thank you, fixed.

annevk:
I thought a bit about this and I think one thing we should probably change is to make the CORS-preflight cache support wildcards directly. That could have a pretty dramatic impact on reducing chattiness.

roryhewitt:
@annevk that would be nice. +1 from me.

Overview.html
```diff
@@ -1857,12 +1859,14 @@ <h4 id="http-new-header-syntax"><span class="secno">4.2.4 </span>HTTP new-header
...
+Access-Control-Allow-Origin      = origin-or-null / wildcard
 Access-Control-Allow-Credentials = %x74.72.75.65 ; "true", case-sensitive
-Access-Control-Expose-Headers    = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a>
+Access-Control-Expose-Headers    = field-name-or-wildcard
```
roryhewitt: Shouldn't this be **#field-name-or-wildcard** (with a leading #?)

Overview.html
```diff
@@ -1857,12 +1859,14 @@ <h4 id="http-new-header-syntax"><span class="secno">4.2.4 </span>HTTP new-header
...
+field-name-or-wildcard           = <a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a> / wildcard
 Access-Control-Max-Age           = <a class="external" data-anolis-spec="http-caching" href="https://tools.ietf.org/html/rfc7234#section-1.2.1">delta-seconds</a>
-Access-Control-Allow-Methods     = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.1.1">method</a>
-Access-Control-Allow-Headers     = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a></pre>
+Access-Control-Allow-Methods     = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.1.1">method</a> / wildcard
+Access-Control-Allow-Headers     = #field-name-or-wildcard</pre>
```
roryhewitt: Should probably add back link to IETF?

Overview.html
```diff
@@ -3644,6 +3683,18 @@ <h3 id="cors-preflight-fetch"><span class="secno">5.7 </span>CORS-preflight fetc
    <a href="#concept-network-error" title="concept-network-error">network error</a>.
 
    <li><p>If one of <var>request</var>'s
+   <a href="#concept-request-header-list" title="concept-request-header-list">header list</a>'s
+   <a href="#concept-header-name" title="concept-header-name">names</a> is a
+   <a href="#cors-non-wildcard-request-header-name">CORS non-wildcard request-header name</a> and is not in <var>headerNames</var>, then
+   return a <a href="#concept-network-error" title="concept-network-error">network error</a>.
+
+   <li><p>If <var>headerNames</var> contains `<code>*</code>`, then remove all occurances of
```
vigneshshanmugam: occurrences?

Overview.html
```diff
@@ -1857,12 +1859,14 @@ <h4 id="http-new-header-syntax"><span class="secno">4.2.4 </span>HTTP new-header
 <pre>Access-Control-Request-Method    = <a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.1.1">method</a>
 Access-Control-Request-Headers   = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a>
 
-Access-Control-Allow-Origin      = origin-or-null / "*"
+wildcard                         = "*"
+Access-Control-Allow-Origin      = origin-or-null / wildcard
 Access-Control-Allow-Credentials = %x74.72.75.65 ; "true", case-sensitive
-Access-Control-Expose-Headers    = #<a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a>
+Access-Control-Expose-Headers    = field-name-or-wildcard
+field-name-or-wildcard           = <a class="external" data-anolis-spec="http" href="https://tools.ietf.org/html/rfc7230#section-3.2">field-name</a> / wildcard
```
tyoshino: Shouldn't this be:
#field-name / wildcard

Overview.html
tyoshino: and shouldn't this be just field-name-or-wildcard (no hash)?

Overview.html
```diff
@@ -2452,13 +2463,29 @@ <h3 id="main-fetch"><span class="secno">5.1 </span>Main fetch</h3>
...
   <ol>
-   <li><p>If <var>request</var>'s
-   <a href="#concept-request-response-tainting" title="concept-request-response-tainting">response tainting</a> is
-   "<code title="">cors</code>", then set <var>response</var>'s
-   <a href="#concept-response-cors-exposed-header-name-list" title="concept-response-cors-exposed-header-name-list">CORS-exposed header-name list</a>
-   to the result of <a href="#concept-header-parse" title="concept-header-parse">parsing</a>
-   `<a href="#http-access-control-expose-headers"><code title="http-access-control-expose-headers">Access-Control-Expose-Headers</code></a>` in
-   <var>response</var>'s <a href="#concept-response-header-list" title="concept-response-header-list">header list</a>.
+   <li>
+   	<p>If <var>request</var>'s
```
tyoshino: broken indentation

Overview.html
```diff
@@ -2452,13 +2463,29 @@ <h3 id="main-fetch"><span class="secno">5.1 </span>Main fetch</h3>
...
+    <a href="#concept-request-response-tainting" title="concept-request-response-tainting">response tainting</a> is
+    "<code title="">cors</code>", then run these substeps:</p>
+
+    <ol>
+     <li><p>Let <var>headerNames</var> be the result <a href="#concept-header-parse" title="concept-header-parse">parsing</a>
```
tyoshino: the result of parsing

Overview.html
tyoshino: ah, is this intended to allow putting * and Authorization together?

annevk: Yeah.

Overview.html
annevk: Yeah, it probably doesn't make sense to allow either here.

Overview.html
annevk: It should have the leading #, but since this is not an IETF production this should not link.

Overview.html
annevk: Thank you!

annevk:
I now changed it so that the CORS-preflight cache can store wildcards which should reduce chattiness a whole lot for non-credentialed requests (at the point user agents implement this, of course).

annevk:
@roryhewitt @tyoshino would you be so kind to look over the latest changes? I think this is ready, but I just want to be sure.

roryhewitt:
@annevk I think the second example in section 4.2.6 should have the last line changed to something like this:

> This is because bar.invalid needs to explicitly share each header by listing their names in the `Access-Control-Expose-Headers` response header. Alternatively, bar.invalid could have returned `Access-Control-Expose-Headers: *` which would share all response headers.

Basically showing developers that they can specify `Access-Control-Expose-Headers: *`.

annevk:
Ended up writing an extra paragraph.

roryhewitt:
@annevk Perfect!

Overview.src.html
```diff
@@ -2380,13 +2401,30 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <li>
+    <p>If <var>request</var>'s
+    <span title=concept-request-response-tainting>response tainting</span> is
+    "<code title>cors</code>", then run these substeps:</p>
+
+    <ol>
+     <li><p>Let <var>headerNames</var> be the result of
+     <span title=concept-header-parse>parsing</span>
+     `<code title=http-access-control-expose-headers>Access-Control-Expose-Headers</code>` in
+     <var>response</var>'s <span title=concept-response-header-list>header list</span>.
```
tyoshino: paragraph not enclosed

tyoshino:
Sorry for delay.

lgtm but I want to make sure that it's fine that
- a user-agent may end up sending no credential even when credentials mode is set to include
- in such case, the server cannot know that the client is using `credentials mode == include` only by inspecting the received request, and therefore it's not safe to use the wildcard.

This means that developers are basically required to coordinate use of the wildcard between client code and server. Right?

annevk:
@tyoshino yeah, I think that coordination has effectively always been required (it also applies to the wildcard case for origin). It might help if we start communicating that more clearly, but that would be a new request header and a separate discussion.

tyoshino:
ok. lgtm

annevk:
Thank you (and the others)!

roryhewitt:
@annevk forgive my ignorance, but I'm not sure how this all goes together...

If these new wildcard values are now in the master branch (and are what I see when I go to https://fetch.spec.whatwg.org/#http-new-header-syntax), does that mean that they are **officially supported** by all (major) browsers? Or is this just a new version of the Fetch spec, and as browsers add support for them, they will publicize this fact?

annevk:
@roryhewitt the latter. I wish it were that easy!

tonyjin:
Also not sure how this all goes together, but I noticed the following two don't match:
https://www.w3.org/TR/cors/#access-control-allow-headers-response-header
https://fetch.spec.whatwg.org/#http-new-header-syntax

Specifically:
Access-Control-Allow-Headers: "Access-Control-Allow-Headers" ":" #field-name (in W3)
Access-Control-Allow-Headers     = #field-name-or-wildcard (in Fetch Spec)

As I was writing this I realized the W3 recommendation was published in Jan 2014, so I guess we'll have to wait for a newer version to be published?

Does the latest fetch spec's definition of Access-Control-Allow-Headers supersede the W3 recommendation?

annevk:
@tonyjin yes it does. W3C's copy of CORS is obsolete and should not be used. Unfortunately they have so far refused to clearly mark it as such.

roryhewitt:
Hey Tony,

Yes, the fetch spec on WHATWG (including the CORS headers) supersedes the
W3C spec.

Unfortunately, it appears to be a low priority to get this 'publicized' by,
for instance, updating the W3C spec to make it clear.

Note also that, although the latest fetch spec includes wildcards for the
ACAM, ACAH and ACEH headers, I'm not sure whether these special values are
actually supported by any browsers yet...

Perhaps @annevk can comment?

Rory

tonyjin:
Thanks for clarifying and working on this @annevk & @roryhewitt - I ran into this PR after getting frustrated at adding Access-Control-Allow-Headers: [a bajillion headers] in nginx 🙄

PR #310, opened 2016-05-31
Grammar change: "that is check" to that checks"

styfle:
"A CORS-preflight request is a CORS request that checks to see if the
CORS protocol is understood."

Alternatively, you could say a "A CORS-preflight request is a CORS request that is checked to see if the
CORS protocol is understood."

But I think the former is more accurate.


annevk:
I think it was supposed to be "a check", but this works. What name shall I list in the acknowledgments section?

styfle:
Added, thanks :+1:

annevk:
Ta, merged as https://github.com/whatwg/fetch/commit/bf9c1c1fbb81404cfdb40789cf6ee83215229b00.

PR #316, opened 2016-06-05
Link to definition of BufferSource

isonmad:
This typedef's definition was missing.


annevk:
I would prefer not merging this. The rationale is that we don't link other types defined in IDL either.

annevk:
Closing this. Sorry that your first contribution didn't make it.

PR #322, opened 2016-06-09
Make sure all responses have a URL list

### LINKED ISSUE #312: Non-http fetch doesn't set response URL
XHR needs to know the response URL (https://fetch.spec.whatwg.org/#concept-response-url). However, datauri/about/etc fetches don't set it.

We should probably set the response URL in basic fetch itself.

annevk:
In #146 I tried to let service worker responses that were not synthetic
to keep their url list. However, the way I fixed that regressed url
lists for a number of responses: HTTP cache responses, about URL
responses, etc.

This fix instead tries to keep the old setup and only overwrites the
url list of a response if it’s empty.

Fixes #312.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-7-june-2016">Living Standard — Last Updated 7 June 2016</h2>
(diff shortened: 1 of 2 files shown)


annevk:
Would appreciate review from @Manishearth and @wanderview.

Manishearth:
LGTM

Manishearth:
Implementing this PR doesn't cause additional failures in Servo. It doesn't cause any additional test passes, either, which is curious; we probably want more tests for this.

wanderview:
LGTM.

@Manishearth, to my knowledge neither chrome nor firefox implement this "if url list is empty" check yet.  For us this is:

https://bugzilla.mozilla.org/show_bug.cgi?id=1222008

PR #329, opened 2016-06-28
Add steps to intercept requests using foreign fetch.

mkruisselbrink:
If response is still null after a regular service worker has tried handling it
this gives a foreign fetch handler a chance to handle the request.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -2767,8 +2767,11 @@ indicates an attempt to authenticate.
   </ol>
 
  <li>
-  <p>If <var>response</var> is null and <var>request</var>'s
-  <span>skip-service-worker flag</span> is unset, run these substeps:
+  <p>If <var>response</var> is null, <var>request</var>'s <span>skip-service-worker flag</span> is
+  unset, <var>request</var> is a <span>subresource request</span> and <var>request</var>'s
+  <span title=concept-request-origin>origin</span> is not the same as <var>request</var>'s
+  <span title=concept-request-url>url</span>'s <span data-anolis-spec=html>origin</span>,
+  run these substeps:
 
   <ol>
    <li>
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -2720,7 +2720,7 @@ indicates an attempt to authenticate.
  <li><p>Let <var>actualResponse</var> be null.
 
  <li>
-  <p>If <var>request</var>'s <span>skip-service-worker flag</span> is unset, run these substeps:
+  <p>If <var>request</var>'s <span>skip-service-worker flag</span> is unset, then run these substeps:
 
   <ol>
    <li>
@@ -2728,22 +2728,22 @@ indicates an attempt to authenticate.
     null or <var>request</var>'s <span title=concept-request-client>client</span>'s
     <span data-anolis-spec=html title=concept-settings-object-global>global object</span> is not a
     <code data-anolis-spec=sw>ServiceWorkerGlobalScope</code> object,
-    set <var>response</var> to the result of invoking
+    then set <var>response</var> to the result of invoking
     <span data-anolis-spec=sw>handle fetch</span> for <var>request</var>.
     <span data-anolis-ref>HTML</span>
     <span data-anolis-ref>SW</span>
 
    <li>
     <p>If <var>response</var> is null,
-    <var>request</var> is a <span>subresource request</span> and <var>request</var>'s
-    <span title=concept-request-origin>origin</span> is not the same as <var>request</var>'s
+    <var>request</var> is a <span>subresource request</span>, and <var>request</var>'s
+    <span title=concept-request-origin>origin</span> is not <span data-anolis-spec=html>same origin</span> with <var>request</var>'s
     <span title=concept-request-url>url</span>'s <span data-anolis-spec=html>origin</span>,
-    set <var>response</var> to the result of invoking
+    then set <var>response</var> to the result of invoking
     <span data-anolis-spec=sw>handle foreign fetch</span> for <var>request</var>.
     <span data-anolis-ref>SW</span>
 
    <li>
-    <p>If <var>response</var> is not null, run these substeps:
+    <p>If <var>response</var> is not null, then run these substeps:
 
     <ol>
      <li><p><span title="Read a request">Read <var>request</var></span>.
@@ -2753,7 +2753,7 @@ indicates an attempt to authenticate.
      <span title=concept-internal-response>internal response</span> otherwise.
 
      <li>
-      <p>If one of the following conditions is true, return a
+      <p>If one of the following conditions is true, then return a
       <span title=concept-network-error>network error</span>:
 
       <ul class=brief>
@@ -2776,7 +2776,6 @@ indicates an attempt to authenticate.
      <li><p>Execute
      <a href=https://w3c.github.io/webappsec-csp/#set-response-csp-list>set <var>response</var>'s CSP list</a>
      on <var>actualResponse</var>. <span data-anolis-ref>CSP</span>
-
     </ol>
   </ol>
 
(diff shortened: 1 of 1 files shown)


annevk:
1. This lacks a same-origin check that I think we want. The fetch needs to actually be foreign. I'm not sure if you're handling that on the other side, but I think it would make more sense here. Happy to discuss of course.
2. A lot of this text is copy-and-pasted from the previous major step. If the requirements are actually identical, we should do the work to make that clear without duplication. I haven't quite figured out yet whether the "network error" requirements need to be the same, but CSP definitely makes sense.

mkruisselbrink:
> This lacks a same-origin check that I think we want. The fetch needs to actually be foreign. I'm not sure if you're handling that on the other side, but I think it would make more sense here. Happy to discuss of course.

Yeah, that check does exist on the service worker spec side as well, but I agree that it makes sense to have the check here as well. I also added the "is this a subresource" check here, to make it clearer which requests actually can be intercepted by foreign fetch.

> A lot of this text is copy-and-pasted from the previous major step. If the requirements are actually identical, we should do the work to make that clear without duplication. I haven't quite figured out yet whether the "network error" requirements need to be the same, but CSP definitely makes sense.

Duplication is indeed unfortunate. Maybe one way to deal with this would be to move all the shared substeps to a separate step, so it roughly would be something like:

1. If [....] invoke handle fetch.
2. If response is null and [...], invoke handle foreign fetch.
3. If response is not null:
  1. Set actualResponse
  2. Do the network error stuff
  3. Do the CSP stuff

annevk:
> Yeah, that check does exist on the service worker spec side as well, but I agree that it makes sense to have the check here as well.

I think in the end we should only have it on one side.

With regards to duplication, what you suggest seems reasonable. An alternative might be to group what you suggest under a common skip-service-worker flag check. Not entirely sure what works better.

mkruisselbrink:
Okay, factored out the common post-processing and skip-service-worker flag logic (which has the side effect of also fixing that previously things like "set response's CSP list" were called even when response was null).

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
-    <p><span title="Read a request">Read <var>request</var></span>.
-
-   <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
-   <span title=concept-filtered-response>filtered response</span>, and to <var>response</var>'s
-   <span title=concept-internal-response>internal response</span> otherwise.
+   <li>
+    <p>If <var>response</var> is null,
+    <var>request</var> is a <span>subresource request</span> and <var>request</var>'s
+    <span title=concept-request-origin>origin</span> is not the same as <var>request</var>'s
```
annevk: "same" needs to reference "same origin" defined in HTML.

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
  <li>
-  <p>If <var>request</var>'s <span>skip-service-worker flag</span> is unset and
-  either <var>request</var>'s <span title=concept-request-client>client</span> is
-  null or <var>request</var>'s <span title=concept-request-client>client</span>'s
-  <span data-anolis-spec=html title=concept-settings-object-global>global object</span> is not a
-  <code data-anolis-spec=sw>ServiceWorkerGlobalScope</code> object, run these substeps:
-  <span data-anolis-ref>HTML</span>
-  <span data-anolis-ref>SW</span>
+  <p>If <var>request</var>'s <span>skip-service-worker flag</span> is unset, run these substeps:
```
annevk: then run*

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
 
   <ol>
    <li>
-    <p>Set <var>response</var> to the result of invoking
+    <p>If <var>request</var>'s <span title=concept-request-client>client</span> is
+    null or <var>request</var>'s <span title=concept-request-client>client</span>'s
+    <span data-anolis-spec=html title=concept-settings-object-global>global object</span> is not a
+    <code data-anolis-spec=sw>ServiceWorkerGlobalScope</code> object,
```
annevk: then set*

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
     <span data-anolis-ref>SW</span>
 
-    <p><span title="Read a request">Read <var>request</var></span>.
-
-   <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
-   <span title=concept-filtered-response>filtered response</span>, and to <var>response</var>'s
-   <span title=concept-internal-response>internal response</span> otherwise.
+   <li>
+    <p>If <var>response</var> is null,
+    <var>request</var> is a <span>subresource request</span> and <var>request</var>'s
```
annevk: comma before and

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+    <span title=concept-request-origin>origin</span> is not the same as <var>request</var>'s
+    <span title=concept-request-url>url</span>'s <span data-anolis-spec=html>origin</span>,
+    set <var>response</var> to the result of invoking
```
annevk: then set*

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+    <span data-anolis-spec=sw>handle foreign fetch</span> for <var>request</var>.
+    <span data-anolis-ref>SW</span>
 
    <li>
-    <p>If one of the following conditions is true, return a
-    <span title=concept-network-error>network error</span>:
+    <p>If <var>response</var> is not null, run these substeps:
```
annevk: then run*

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+     <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
+     <span title=concept-filtered-response>filtered response</span>, and to <var>response</var>'s
+     <span title=concept-internal-response>internal response</span> otherwise.
 
-     <li><var>request</var>'s <span title=concept-request-redirect-mode>redirect mode</span> is not
-     "<code title>manual</code>" and <var>response</var>'s
-     <span title=concept-response-type>type</span> is "<code title>opaqueredirect</code>".
+     <li>
+      <p>If one of the following conditions is true, return a
+      <span title=concept-network-error>network error</span>:
```
annevk: then return*

Overview.src.html
```diff
@@ -2720,50 +2720,64 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+
+       <li><var>request</var>'s <span title=concept-request-redirect-mode>redirect mode</span> is not
+       "<code title>follow</code>" and <var>response</var>'s
+       <span title=concept-response-url-list>url list</span> has more than one item.
+      </ul>
+
+     <li><p>Execute
+     <a href=https://w3c.github.io/webappsec-csp/#set-response-csp-list>set <var>response</var>'s CSP list</a>
+     on <var>actualResponse</var>. <span data-anolis-ref>CSP</span>
+
```
annevk: remove this newline

annevk:
I like it!

mkruisselbrink:
Addressed your comments.

annevk:
Landed as https://github.com/whatwg/fetch/commit/8510b2f52d576af6a6c683620878eb3e39e14954. Thanks for working on this!

PR #335, opened 2016-07-15
Call out to Referrer Policy to set policy on redirect

estark37:
Before following a redirect, Fetch should call out to Referrer Policy to
update the referrer policy on the request, if necessary. (In particular,
this applies when the redirect response contains a Referrer-Policy
header.)

This integration is described in
https://w3c.github.io/webappsec-referrer-policy/#integration-with-fetch,
though it needs to be updated per
https://github.com/w3c/webappsec-referrer-policy/issues/59.


annevk:
Thank you, landed as https://github.com/whatwg/fetch/commit/a8f1cd5ae2d080ee4e1e8e02b68e8ec6ae2c2833.

PR #340, opened 2016-07-22
Make it clearer that cache mode "default" follows normal logic

### LINKED ISSUE #336: Make it more clear that cache mode "default" does respect Cache-Control headers and such
Basically, when you go to the HTTP cache with a request whose cache mode is "default", and the request also has Cache-Control or Pragma headers, those need to be respected by the HTTP cache, just as they always were.

Alternatively we could make those headers and their values change the cache mode:
- If the cache mode is “default” and a `Cache-Control` or `Pragma` Header is present, whose value is “no-cache”, then set the cache mode from “Default” to no-cache.
- If the cache mode is “default” and a `Cache-Control` or `Pragma` Header is present, whose value is “no-store”, then change the cache mode from “Default” to no-store.

That's not observable so either approach is probably fine. Thoughts anyone?

annevk:
If the cache mode is “default” and you set a Cache-Control header, that
should be taken into account when talking to the HTTP cache subsystem.
It’s how the cache works by “default”, after all.

Fixes #336.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -3329,12 +3329,29 @@ The <i title="">credentials flag</i> is one too.
  <li><p>Let <var>response</var> be null.
 
  <li>
-  <p>If <var>httpRequest</var>'s
-  <a href="#concept-request-cache-mode" title="concept-request-cache-mode">cache mode</a> is neither
-  "<code title="">no-store</code>" nor "<code title="">reload</code>", and there is a <em>complete</em>
-  <a href="#concept-response" title="concept-response">response</a> in the HTTP cache for
-  <var>httpRequest</var> run these substeps:
-  <!-- XXX xref "HTTP cache" -->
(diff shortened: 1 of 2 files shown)


annevk:
@ehsan does this look okay?

annevk:
@mnot any thoughts? I was wondering, would it help to start a  thread on the HTTP mailing list about better integration between the documents?

wanderview:
FWIW, I thought we already had "default" doing whatever the cache-control headers said to do.  That's what we implement.

annevk:
I know, but it's not clear from the specification as new implementations got it wrong. That's why I'm trying to get these changes reviewed.

annevk:
Acknowledge when landing this: Ali Alabbas.

I'm starting to think there's probably room for a more expansive section on the HTTP cache to also deal with the credentials issue. I'll wait a bit for @mnot to poke around before doing anything about it though.

annevk:
This is superseded by the rewrite in #496.

PR #346, opened 2016-07-29
Referrer policies: add the missing policies (fixes #342)

### LINKED ISSUE #342: Support same-origin, strict-origin, strict-origin-when-cross-origin in request' referrer policy
There're 3 new states "same-origin, strict-origin, strict-origin-when-cross-origin" have added to Referrer Policy specs [1]
These states should be added to request's referrer policy [2]
[1] https://w3c.github.io/webappsec-referrer-policy/
[2] https://fetch.spec.whatwg.org/#referrer-policies

Thanks

fmarier:
The following policy states were recently added to the Referrer Policy
specification:

* same-origin
* strict-origin
* strict-origin-when-cross-origin

They were originally added in
https://github.com/w3c/webappsec-referrer-policy/pull/19 and
https://github.com/w3c/webappsec-referrer-policy/pull/56,
and discussed on the W3C WebAppSec list at
https://lists.w3.org/Archives/Public/public-webappsec/2016Mar/0085.html.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -1475,15 +1475,19 @@ is listed in the first column of the following table.
 <h3>Referrer policies</h3>
 
 <p>A <dfn title=concept-referrer-policy>referrer policy</dfn> is the empty string,
-"<code>no-referrer</code>", "<code>no-referrer-when-downgrade</code>", "<code>origin</code>",
-"<code>origin-when-cross-origin</code>", or "<code>unsafe-url</code>".
+"<code>no-referrer</code>", "<code>no-referrer-when-downgrade</code>", "<code>same-origin</code>",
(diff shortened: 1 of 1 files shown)


fmarier:
I looked through all mentions of the word "referrer" in the spec and that seemed like the only place that needed to be updated, but it was so simple I'm afraid I may have missed something.

annevk:
You missed the list just before the enumeration. Each value is mentioned at least twice at the moment.

fmarier:
> You missed the list just before the enumeration.

Hm, not sure how I managed to miss that, but I've just pushed a fix.

annevk:
Landed as https://github.com/whatwg/fetch/commit/eb08febe5f56f7beefa21e91366c4fa5a545e7d8, thanks!

PR #350, opened 2016-08-03
Add ReadableStreamOperations for ServiceWorker

yutakahirano:
This change adds some ReadableStream operations that are used in https://github.com/slightlyoff/ServiceWorker/pull/934.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -1595,6 +1595,11 @@ run these steps:
  <li><p>Return <var>reader</var>.
 </ol>
 
+<p>To <dfn title=concept-read-chunk-from-ReadableStream>read a chunk</dfn> from a
+<code title=concept-ReadableStream>ReadableStream</code> object with <var>reader</var>, return the
+result of calling
+<code data-anolis-spec=streams>ReadFromReadableStreamReader</code>(<var>reader</var>).
+
 <p>To <dfn title=concept-read-all-bytes-from-ReadableStream>read all bytes</dfn> from a
 <code title=concept-ReadableStream>ReadableStream</code> object with <var>reader</var>, run these
 steps:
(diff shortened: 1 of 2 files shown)


yutakahirano:
@annevk if you have time.

@jungkees fyi

I noticed my wording at slightlyoff/ServiceWorker#934 was a bit inconsistent with existing operations. I will create a PR replacing "cancel a reader" with "cancel a ReadableStream" etc. after this change is landed. I'm sorry for the mess...

Overview.src.html
```diff
@@ -1589,6 +1589,17 @@ <h3 id="should-response-to-request-be-blocked-due-to-mime-type?"><dfn title="sho
...
+<p>To <dfn title=concept-read-chunk-from-ReadableStream>read a chunk</dfn> from a
+<code title=concept-ReadableStream>ReadableStream</code> object with <var>reader</var>, run these
+steps:
+
+<ol>
+ <li><p>Let <var>promise</var> be the result of calling
+ <code data-anolis-spec=streams>ReadFromReadableStreamReader</code>(<var>reader</var>).
+
+ <li><p>Return <var>promise</var>.
+</ol>
```
annevk: I think we should inline this and the other operation you introduce by replacing "run these steps" with "return the result of calling ..."

annevk:
This looks good, should I rebase it for you?

Overview.src.html
```diff
@@ -1589,6 +1589,11 @@ <h3 id="should-response-to-request-be-blocked-due-to-mime-type?"><dfn title="sho
  <li><p>Return <var>reader</var>.
 </ol>
 
+<p>To <dfn title=concept-read-chunk-from-ReadableStream>read a chunk</dfn> from a
+<code title=concept-ReadableStream>ReadableStream</code> object with <var>reader</var>, return the
+result of calling
+<code data-anolis-spec=streams>ReadFromReadableStreamReader</code>(<var>reader</var>).
```
annevk: Actually, maybe we should not use `<code>` here? I don't think we do that in HTML.

annevk: `<span>` seems better for abstract operations.

yutakahirano: Hm, that should be fixed separately with other abstract operations, I think.

yutakahirano:
Filed https://github.com/whatwg/fetch/issues/361.

annevk:
Sounds good, thanks!

PR #359, opened 2016-08-09
A request needs to be created with a URL

### LINKED ISSUE #351: Request's current url is undefined for a new request
Step 8 of the Request constructor says `Set request to a new request whose url is request's current url`. A request's current url is defined as `A request has an associated current url. It is a pointer to the last URL in request's url list.`. If the Request constructor receives a string as the input, we are told to create `a new request`. The url list is defined as `A request has an associated url list (a list of one or more URLs). Unless stated otherwise, it is a list containing a copy of request's url.` And finally, the request's url is defined as `A request has an associated url (a URL).` with no default value, and a note suggesting that it should be an alias of the first URL in the url list, which yields a circular definition.

annevk:
Fixes #351.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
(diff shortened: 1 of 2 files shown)


annevk:
@jdm I think this is what you were looking for, right?

It seems like a fairly safe change to me as well and fixes a bug with request's current url not being anything as you pointed out.

annevk:
(Search for "Overview.src.html" in the diff to get the more readable version.)

jdm:
Much better, thank you.

PR #369, opened 2016-08-16
Allowed header in CORS with Access-Control-Expose-Headers

udnisap:
allowed headers in CORS also depends on the response headers
http://stackoverflow.com/questions/38953864/how-to-get-headers-of-the-response-from-fetch/38954244#38954244

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -418,6 +418,9 @@ each other by `<code title>,</code>`, in order.
 <p>and whose <span title=concept-header-value>value</span>,
 <span title=concept-header-parse>once parsed</span>, is not a failure.
 
+<p>or whose <span title=concept-header-name>header name</span> is in the commar seperated list of Access-Control-Expose-Headers 
+header of the response.
+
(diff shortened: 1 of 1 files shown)


annevk:
Safelisted request headers don't depend on the response. I think you're confusing this with https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name which already takes that header into account.

udnisap:
May be I have the wrong wording. What I wanted to express was that other than the safe listed headers if you want to have your custom header available for the web client, you can send that header name in `Access-Control-Expose-Headers ` header.

For example, Lets say the server wants the clients to be able to access header `A` of the response of a CORS request. Server can send `Access-Control-Expose-Headers : A` on the response and the fetch implementation makes the header available through `response.headers.get('A')`.

annevk:
Yeah, that is already covered by the standard (see my pointer and the places that make use of it) and unrelated to request headers.

udnisap:
Sorry my bad. I got it confused. I checked this but couldnt figure it out before.

annevk:
No worries, appreciate you looking things over.

PR #377, opened 2016-08-26
Make Request constructor more forgiving

### LINKED ISSUE #245: referrer same-origin constraint is a footgun for people trying to "copy" a Request
Recently I saw a website in the wild attempting to do this:

```
var request(newURL, {
  referrer: oldRequest.referrer,
  // copy other attributes as well
});
```

This will work just fine during development on localhost, because .referrer will most likely always be same-origin.  When the site is posted on twitter, for example, it will be visited through a t.co redirector.  This results in a t.co referrer which is cross-origin.

So the site that worked fine in local development will blow up when its published to twitter.  This seems like a bit of a footgun.

We could make new Request() silently ignore the value if its invalid instead of throwing.  This is somewhat similar to using bad header values.  They just get ignored.

annevk:
Folks are using the Request constructor in unexpected ways and
therefore it is throwing in unexpected ways (in particular when mode is
“navigate” or when setting referrer to a cross-origin URL). This will
make it throw less, while not really being less useful.

Fixes #245.

### DIFF
--- a/Overview.html
+++ b/Overview.html
@@ -7,7 +7,7 @@
 
 <p><a class="logo" href="https://whatwg.org/"><img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-fetch.svg" width="100"></a>
 <h1 id="cors">Fetch</h1>
-<h2 class="no-num no-toc" id="living-standard-—-last-updated-15-september-2016">Living Standard — Last Updated 15 September 2016</h2>
+<h2 class="no-num no-toc" id="living-standard-—-last-updated-27-september-2016">Living Standard — Last Updated 27 September 2016</h2>
 
 <dl>
  <dt>Participate:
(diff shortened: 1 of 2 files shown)


annevk:
@wanderview you blocked this in #245 but then did not follow up to feedback. Thoughts appreciated for what we should do here.

jdm:
It's disappointing that the web-platform-tests were not updated to reflect this change.

jdalton:
@jdm Are you up for a pull request?

jdm:
That's my plan.

annevk:
@jdm @jdalton sorry about that, it's a policy we only started relatively recently. Future changes should land together with WPT changes.

annevk:
I filed these browser bugs:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1341223
* https://bugs.chromium.org/p/chromium/issues/detail?id=694430
* https://bugs.webkit.org/show_bug.cgi?id=168649

I also wanted to further fix the test, but got stuck, see https://github.com/w3c/web-platform-tests/issues/2618 for that.

PR #383, opened 2016-09-13
Add request's reserved client and target client id

jungkees:
Add request's reserved client and target client id

This patch defines the request's reserved client and the target client
id. The reserved client of a request is set to an environment or an
environment settings object during a navigation or a worker creation
request, respectively. The target client id of a request is set to the
target browsing context's active document's environment settings
object's id during a navigation. The given values are primarily used
to provide the reserved client's id and target client's id to service
workers' fetch event handlers.

This removes the request's target browsing context as the reserved
client provides the same information.

Related issue: https://github.com/w3c/ServiceWorker/issues/870
Related change: https://github.com/whatwg/html/pull/1776

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -662,8 +662,9 @@ outlawing <span title="forbidden method">forbidden methods</span> and
 <span data-anolis-spec=html>environment settings object</span>).
 
 <p>A <span title=concept-request>request</span> has an associated
-<dfn title=concept-request-reserved-client-id data-export data-dfn-for=request>reserved client
-id</dfn> (null or a string). Unless stated otherwise it is null.
+<dfn title=concept-request-reserved-client data-export data-dfn-for=request>reserved client</dfn>
+(null or an <span data-anolis-spec=html>environment</span> or an
+<span data-anolis-spec=html>environment settings object</span>). Unless stated otherwise it is null.
 
 <p>A <span title=concept-request>request</span> has an associated
 <dfn title=concept-request-target-client-id data-export data-dfn-for=request>target client id</dfn>
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -663,12 +663,12 @@ outlawing <span title="forbidden method">forbidden methods</span> and
 
 <p>A <span title=concept-request>request</span> has an associated
 <dfn title=concept-request-reserved-client data-export data-dfn-for=request>reserved client</dfn>
-(null or an <span data-anolis-spec=html>environment</span> or an
+(null, an <span data-anolis-spec=html>environment</span>, or an
 <span data-anolis-spec=html>environment settings object</span>). Unless stated otherwise it is null.
 
 <p>A <span title=concept-request>request</span> has an associated
 <dfn title=concept-request-target-client-id data-export data-dfn-for=request>target client id</dfn>
-(null or a string). Unless stated otherwise it is null.
+(a string). Unless stated otherwise it is the empty string.
 
 <p>A <span title=concept-request>request</span> has an associated
 <dfn title=concept-request-window data-export data-dfn-for=request>window</dfn>
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -684,10 +684,6 @@ outlawing <span title="forbidden method">forbidden methods</span> and
 explicitly set <span title=concept-request>request</span>'s
 <span title=concept-request-window>window</span>.
 
-<p>A <span title=concept-request>request</span> has an associated
-<dfn title=concept-request-target-browsing-context data-export data-dfn-for=request>target browsing context</dfn>
-(null or a <span data-anolis-spec=html>browsing context</span>). Unless stated otherwise it is null.
-
 <p id=keep-alive-flag>A <span title=concept-request>request</span> has an associated
 <dfn>keepalive flag</dfn>. Unless stated otherwise it is unset.
 
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -662,6 +662,14 @@ outlawing <span title="forbidden method">forbidden methods</span> and
 <span data-anolis-spec=html>environment settings object</span>).
 
 <p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-reserved-client-id data-export data-dfn-for=request>reserved client
+id</dfn> (null or a string). Unless stated otherwise it is null.
+
+<p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-target-client-id data-export data-dfn-for=request>target client id</dfn>
+(null or a string). Unless stated otherwise it is null.
+
+<p>A <span title=concept-request>request</span> has an associated
 <dfn title=concept-request-window data-export data-dfn-for=request>window</dfn>
 ("<code>no-window</code>", "<code>client</code>", or an
 <span data-anolis-spec=html>environment settings object</span> whose
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -666,20 +666,20 @@ outlawing <span title="forbidden method">forbidden methods</span> and
 (null, an <span data-anolis-spec=html>environment</span>, or an
 <span data-anolis-spec=html>environment settings object</span>). Unless stated otherwise it is null.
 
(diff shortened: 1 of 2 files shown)


jungkees:
@annevk, please take a look in line with the related changes to HTML and SW.

annevk:
So how can this be an environment settings object? Isn't this information what's used for the eventual creation of one?

annevk:
Also, if a reserved client knows the target browsing context, should we still provide that as a slot?

jungkees:
> So how can this be an environment settings object? Isn't this information what's used for the eventual creation of one?

I defined an algorithm setting up and returning an environment settings object that has different algorithms than those of browsing context/worker's environment settings object: https://github.com/whatwg/html/pull/1776/commits/df5262bf83f7f10b5620f14b625b29b3c5987cd0#diff-36cd38f49b9afa08222c0dc9ebfe35ebR77922
The reserved client's id will be used to set the actual environment settings object's id when it's created later. Also, I wanted it to be a type of an environment settings object as we should be able to enumerate these client objects in the service worker steps: `get(id)`, `matchAll()`, etc.

> Also, if a reserved client knows the target browsing context, should we still provide that as a slot?

The reserved client's target browsing context just returns request's target browsing context which was set in https://html.spec.whatwg.org/#process-a-navigate-fetch step 2. The reason why reserved client should hold this information is that when a service worker gets this client later from `clients.get(id)` or `clients.matchAll()`, it won't have any reference to the original request. (We plan to use the state of the active document of the target browsing context for WinodwClient's properties associated with the reserved client.)

annevk:
This setup sounds a little weird. This object is passed along with a request but it echoes data from a request? Why wouldn't we just store that data on the object and create it together with the request?

jungkees:
Do you mean we may use the target browsing context passed to "process a navigate fetch" directly to set up the object and delete the target browsing context internal slot from request? If so, I think that will work. FYI, this is the added step to set up the reserved object: https://github.com/whatwg/html/pull/1776/commits/df5262bf83f7f10b5620f14b625b29b3c5987cd0#diff-36cd38f49b9afa08222c0dc9ebfe35ebR82116 If I understood correctly, we may pass the target browsing context to this algorithm.

annevk:
I think what I'm saying is that if this object lives on a request, a request should not be used to create it.

jungkees:
I understood. So, rather than passing a request to create an object, I'll change the object creation algorithm to take a request's url and target browsing context. Does that make sense?

annevk:
I guess. It seems a little weird to duplicate the URL field. I wonder if we're not better off if we keep some of these concepts distinct and not try to group them all together.

An alternative would be that we just add some properties to request (perhaps grouped together) and at a different stage use those to create an environment settings object.

jungkees:
> It seems a little weird to duplicate the URL field.

request' url and request's reserved client's creation URL?

> I wonder if we're not better off if we keep some of these concepts distinct and not try to group them all together.

Could you be more specific with the idea? I'm curious if you don't want to use an environment settings object to represent a reserved client in the first place?

> An alternative would be that we just add some properties to request (perhaps grouped together) and at a different stage use those to create an environment settings object.

Which environment settings object do you mean? One for reserved client? Or one with the resulting fetched resource?

annevk:
> request' url and request's reserved client's creation URL?

Right. Also, what happens if we hit a redirect?

> I'm curious if you don't want to use an environment settings object to represent a reserved client in the first place?

Yeah, it doesn't seem like it's entirely the right fit. Maybe it's better if we had a client concept that's either an environment settings object or a "reserved client", or some such.

jungkees:
@annevk, I updated it back to use a reserved client (an environment) instead of just a reserved client id. Please check out https://github.com/whatwg/html/pull/1776#issuecomment-253411812

Also, the last patch suggests we remove the target browsing context internal slot of the request. I presumed there's no usage of accessing this state yet and reserved client's target browsing context may be used for that purposed if necessary onwards.

Overview.src.html
```diff
@@ -662,6 +662,15 @@ <h4 id=terminology-headers>Headers</h4>
 <span data-anolis-spec=html>environment settings object</span>).
 
 <p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-reserved-client data-export data-dfn-for=request>reserved client</dfn>
+(null or an <span data-anolis-spec=html>environment</span> or an
+<span data-anolis-spec=html>environment settings object</span>). Unless stated otherwise it is null.
+
+<p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-target-client-id data-export data-dfn-for=request>target client id</dfn>
+(null or a string). Unless stated otherwise it is null.
```
annevk: Can't we just use the empty string here? It's usually better to not change data types if not necessary.

Overview.src.html
```diff
@@ -662,6 +662,15 @@ <h4 id=terminology-headers>Headers</h4>
 <span data-anolis-spec=html>environment settings object</span>).
 
 <p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-reserved-client data-export data-dfn-for=request>reserved client</dfn>
+(null or an <span data-anolis-spec=html>environment</span> or an
+<span data-anolis-spec=html>environment settings object</span>). Unless stated otherwise it is null.
```
jungkees: Done.

Overview.src.html
jungkees: Right. Missed that point. Addressed.

annevk:
We should wait with this until HTML lands right? Otherwise the cross-references will lead nowhere. But it looks good. The only thing worth doing might be adding a note to each field mentioning that they are only useful for navigation (and worker?) requests and can be left alone otherwise.

jungkees:
> We should wait with this until HTML lands right? Otherwise the cross-references will lead nowhere.

Right. HTML also cross-references to fetch (request's reserved client and target clientid). So, ideally they should land at the same time when both of them are good-to-go.

> The only thing worth doing might be adding a note to each field

Will do!

Overview.src.html
```diff
@@ -662,6 +662,26 @@ <h4 id=terminology-headers>Headers</h4>
...
+
+<p>A <span title=concept-request>request</span> has an associated
+<dfn title=concept-request-target-client-id data-export data-dfn-for=request>target client id</dfn>
+(a string). Unless stated otherwise it is the empty string.
+
+<p class="note no-backref">This is only used by <span>navigation requests</span>. It has the
+<span data-anolis-spec=html>id</span> of the <span data-anolis-spec=html>target browsing
+context</span>'s <span data-anolis-spec=html>active document</span>'s
+<span data-anolis-spec=html>environment settings object</span> captured before entering the
+<span title=concept-fetch>fetch</span> algorithm.
```
annevk: I was about to merge, but I don't understand this. The target client ID isn't that a reserved ID of sorts?

jungkees: We discussed this during the f2f last month. This is used to expose the target client's id to fetch event's handler without the handle fetch having to cross the process border to get the target browsing context's document's environment's id from its own steps. The reserved client's target browsing context is used within SW's clients' method (like `clients.get(reservedClientId)`) to get the target's active document's related window properties later.

jungkees: So to speak, it's used to initialize `fetchEvent.targetClientid` without crossing the process border during running the handle fetch steps.

annevk: But this ID goes away, right? This is the ID of the global that disappears somewhere during the navigation.

jungkees: Right. But having discussed with @jakearchibald, this is also a value we'd want to expose to devs to help maintaining their Caches (i.e. for the per-client cache control).

annevk: Okay, thanks for confirming. Once the HTML PR has a commit message and is landed I'll merge this with a couple of minor tweaks.

jungkees: Sounds great. Thanks!

jungkees:
@annevk, I updated the commit message of this PR in the OP, too.

annevk:
Thanks @jungkees, all good now.

jungkees:
Looks good. Thanks!

PR #387, opened 2016-09-14
Treat data URLs as same-origin, except for workers

### LINKED ISSUE #381: Same-origin data-URL flag only for fetch and XHR?
AFAIK, this flag is only used by XHR and fetch.
It does not seem to be used by the HTML specification at all.
Is it on purpose or just a lack of synchronization?

Doing a quick test, crossorigin-attribute-enabled scripts seem to load fine with data URLs in Chrome and Firefox.

annevk:
HTML gives data URLs a unique origin when navigating to them to prevent
a class of XSS attacks.

Since browsers already largely allow data URLs in all other contexts
this commit aligns with that, opting them into being same-origin
elsewhere.

Workers however are still prevented. It would create problems for
shared workers and potentially also for dedicated workers.

Fixes #381.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -2410,10 +2410,7 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
    <span title=concept-request-origin>origin</span> and <i>CORS flag</i> is unset
    <dt><var>request</var>'s
    <span title=concept-request-current-url>current url</span>'s
-   <span data-anolis-spec=url title=concept-url-scheme>scheme</span> is "<code>data</code>" and
-   <var>request</var>'s <span title=concept-request-destination>destination</span> is not
-   "<code>sharedworker</code>" or "<code>worker</code>"
-   <!-- "serviceworker" already prevents anything non-HTTP(S) -->
+   <span data-anolis-spec=url title=concept-url-scheme>scheme</span> is "<code>data</code>"
    <dt><var>request</var>'s <span title=concept-request-mode>mode</span> is
    "<code title>navigate</code>" or "<code title>websocket</code>"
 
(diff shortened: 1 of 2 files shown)


annevk:
Once this lands I should update HTML to remove mention of the flag.

domenic:
> HTML gives data URLs a unique origin when navigating to them to prevent
a class of XSS attacks.

more accurately "gives Documents created from data URLs", right?

annevk:
Yeah, that'd be a little better.

mikewest:
This patch does what you want it to do, and I think I'm happy with moving Chrome in this direction. LGTM.

Overview.src.html
```diff
@@ -2426,6 +2423,16 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
      using <var>request</var>.
     </ol>
 
+    <p class="note no-backref">HTML will assign any documents and dedicated workers created from a
+    <span data-anolis-spec=url title=concept-url>URL</span> whose
+    <span data-anolis-spec=url title=concept-url-scheme>scheme</span> is "<code>data</code>" an
+    <span data-anolis-spec=html>opaque origin</span> and it prevents creation of shared workers from
```
mikewest: Tiny nit: perhaps `s/and it prevents/, and prevents/`

annevk:
Proposed commit message:
```
Treat data URLs as same-origin

By-and-large browsers treat data URLs as same-origin, though there 
are some inconsistencies. This change will treat all data URLs, 
regardless of origin, as same-origin from the perspective of Fetch.

HTML will assign a unique opague origin to documents and dedicated
workers created from a data URL.

HTML will also forbid shared workers to be created from data URLs.

See https://github.com/whatwg/html/pull/1782 for the changes made to
HTML.

Service workers already prevent anything but HTTP(S) URLs from 
creating them.

Fixes #381.
```

PR #388, opened 2016-09-16
allow Request to outlive environment settings object

igrigorik:
Related discussion in https://github.com/w3c/beacon/pull/27. 

@annevk ptal, is this on the right track? I have a few ???'s and TODO's that I could use help with :)

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -679,13 +679,13 @@ explicitly set <span title=concept-request>request</span>'s
 <dfn title=concept-request-target-browsing-context data-export data-dfn-for=request>target browsing context</dfn>
 (null or a <span data-anolis-spec=html>browsing context</span>). Unless stated otherwise it is null.
 
-<p>A <span title=concept-request>request</span> has an associated
-<dfn>keepalive flag</dfn>. Unless stated otherwise it is unset.
+<p>A <span title=concept-request>request</span> has an associated <dfn>keepalive flag</dfn>. Unless
+stated otherwise it is unset.
 
 <p class="note no-backref">This can be used to allow the request to outlive the
-<span data-anolis-spec=html>environment settings object</span> - e.g.
-<code>navigator.sendBeacon</code> and the HTML <code>img</code> element set this flag. Requests
-with this flag set are subject to additional processing requirements.
+<span data-anolis-spec=html>environment settings object</span>, e.g.,
+<code>navigator.sendBeacon</code> and the HTML <code>img</code> element set this flag. Requests with
+this flag set are subject to additional processing requirements.
 
 <p>A <span title=concept-request>request</span> has an associated
 <dfn data-export data-dfn-for=request>skip-service-worker flag</dfn>. Unless stated otherwise it is
@@ -1336,8 +1336,7 @@ functionality.
 <dfn title=concept-fetch-group-terminate data-export data-dfn-for="fetch group">terminated</dfn>,
 for each associated <span title=concept-fetch-record>fetch record</span> whose
 <span title=concept-fetch-record-request>request</span>'s <span>done flag</span> or
-<span>keepalive flag</span> is unset,
-<span title=concept-fetch-terminate>terminate</span> the
+<span>keepalive flag</span> is unset, <span title=concept-fetch-terminate>terminate</span> the
 <span title=concept-fetch-record>fetch record</span>'s
 <span title=concept-fetch-record-fetch>fetch</span> with reason <i title>fatal</i>.
 
@@ -3100,14 +3099,13 @@ steps:
  <var>httpRequest</var>'s
  <span title=concept-request-header-list>header list</span>.
 
- <li><p>If <var>contentLengthValue</var> is non-null,
- the <var>httpRequest</var>'s <span>keepalive flag</span> is set,
- and <var>contentLengthValue</var> is greater than user-agent-defined maximum body size,
- then return a <span title=concept-network-error>network error</span>.
+ <li><p>If <var>contentLengthValue</var> is non-null, <var>httpRequest</var>'s
+ <span>keepalive flag</span> is set, and <var>contentLengthValue</var> is greater than a
+ user-agent-defined maximum, then return a <span title=concept-network-error>network error</span>.
 
  <p class="note no-backref">The above user-agent-defined value ensures that requests that are
- allowed to outlive the <span data-anolis-spec=html>environment settings object</span> and
- contain a body have bounded size and are not allowed to stay alive indefinitely.
+ allowed to outlive the <span data-anolis-spec=html>environment settings object</span> and contain a
+ body, have a bounded size and are not allowed to stay alive indefinitely.
 
  <li><p>If <var>httpRequest</var>'s <span title=concept-request-referrer>referrer</span> is a
  <span data-anolis-spec=url title=concept-url>URL</span>, then
@@ -4233,7 +4231,7 @@ typedef (<span>BodyInit</span> or <span title=concept-ReadableStream>ReadableStr
 
 <p>To <dfn title=concept-BodyInit-extract>extract</dfn> a <span title=concept-body>body</span> and a
 `<code title>Content-Type</code>` <span title=concept-header-value>value</span> from
-<var>object</var> with optional <var>keepalive</var> flag, run these steps:
+<var>object</var>, with an optional <var>keepalive flag</var>, run these steps:
 
 <ol>
  <li><p>Let <var>stream</var> be the result of
@@ -4241,6 +4239,7 @@ typedef (<span>BodyInit</span> or <span title=concept-ReadableStream>ReadableStr
  <code title=concept-ReadableStream>ReadableStream</code> object.
 
  <li><p>Let <var>Content-Type</var> be null.
+
  <li><p>Let <var>action</var> be null.
 
  <li>
@@ -4299,14 +4298,14 @@ typedef (<span>BodyInit</span> or <span title=concept-ReadableStream>ReadableStr
    <dd><p>Set <var>stream</var> to <var>object</var>.
   </dl>
 
- <li>
-  <p>If <var>keepalive</var> flag is set and <var>object</var>'s type is a
-  <code title=concept-ReadableStream>ReadableStream</code> object, then throw a
-  <code title>TypeError</code>.
+ <li><p>If <var>keepalive flag</var> is set and <var>object</var>'s type is a
+ <code title=concept-ReadableStream>ReadableStream</code> object, then throw a
+ <code title>TypeError</code>.
 
  <li>
-  <p>If <var>action</var> is non-null, run <var>action</var> <span data-anolis-spec=html>in
-  parallel</span>:
+  <p>If <var>action</var> is non-null, run <var>action</var>
+  <span data-anolis-spec=html>in parallel</span>:
+
   <ol>
    <li><p>Whenever one or more bytes are available, let <var>bytes</var> be the bytes and
    <span title=concept-enqueue-ReadableStream>enqueue</span> a <code>Uint8Array</code> object
@@ -4756,7 +4755,8 @@ constructor must run these steps:
  <span title=concept-request-integrity-metadata>integrity metadata</span> to it.
 
  <li><p>If <var>init</var>'s <code title>keepalive</code> member is present, then set
- <var>request</var>'s <span>keepalive flag</span> to it.
+ <var>request</var>'s <span>keepalive flag</span> if <var>init</var>'s <code title>keepalive</code>
+ member is true, and unset it otherwise.
 
  <li>
   <p>If <var>init</var>'s <code title>method</code> member is present, let
@@ -4825,11 +4825,11 @@ constructor must run these steps:
 
   <ol>
    <li><p>Let <var>Content-Type</var> be null.
-   <li>
-    <p>If <var>init</var>'s <code title>keepalive</code> member is present and is set to
-    <code>true</code>, then set <var>inputBody</var> and <var>Content-Type</var> to the
-    result of <span title=concept-BodyInit-extract>extracting</span> <var>init</var>'s
-    <code title>body</code> member with <var>keepalive</var>. Rethrow any exceptions.
+
+   <li><p>If <var>init</var>'s <code title>keepalive</code> member is present and is true, then set
+   <var>inputBody</var> and <var>Content-Type</var> to the result of
+   <span title=concept-BodyInit-extract>extracting</span> <var>init</var>'s <code title>body</code>
+   member, with <var>keepalive flag</var> set. Rethrow any exceptions.
 
    <li><p>Otherwise, set <var>inputBody</var> and <var>Content-Type</var> to the result of
    <span title=concept-BodyInit-extract>extracting</span> <var>init</var>'s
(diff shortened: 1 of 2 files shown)


annevk:
I'm not entirely sure actually. @yutakahirano can hopefully help.

yutakahirano:
concept-request relies on ReadableStream which relies on the realm. So we cannot read data from the request body when the environment settings object is invalid, I guess. cc: @domenic

domenic:
Yeah, that's true, if someone sends a ReadableStream body but the window has been collected, you're in trouble.

igrigorik:
Ah, well.. In that case I guess we should throw an error if body object's type is `ReadableStream` and `keepAlive` flag is set? In effect, if you want `keepAlive` semantics than you have to provide the body upfront. Also, that also eliminates the timer question.

Before I update the pull.. Does that make sense? Any other gotchas?

igrigorik:
TPAC notes:
- Streams are still work-in-progress, for v1 we can throw an error if body is a ReadableStream and keepAlive flag is enabled (same behavior for upload streams)
- Throw `TypeError`  if/when size is exceeded.

yutakahirano:
The problem is that the concept layer also depends on ReadableStream. Even when you create a Request object from a DOMString body, the body is held as a ReadableStream (see https://fetch.spec.whatwg.org/#concept-body and https://fetch.spec.whatwg.org/#concept-bodyinit-extract).

If we need to have a fetch operation work beyond the realm destruction, we need to define a stream that work in such a circumstance.

wanderview:
> The problem is that the concept layer also depends on ReadableStream. Even when you create a Request object from a DOMString body, the body is held as a ReadableStream (see https://fetch.spec.whatwg.org/#concept-body and https://fetch.spec.whatwg.org/#concept-bodyinit-extract).

I thought one of the goals was to let the browser optimize such that if it never has to execute js to read the stream, then it could perform off thread copying, etc.  A DOMString body should fall into this category, no?

igrigorik:
Updated / simplified the proposed change in 3b71edd -- getting warmer? 

This doesn't address @yutakahirano's comment about Streams support for outliving real destruction.. but that smells like a Streams API feature? @domenic any thoughts or recommendations on this one?

domenic:
Yeah, I'm not sure what to do about that issue :-/. I think it might be worth noting in the spec somehow, and I'll try to think of a nicer solution next week...

igrigorik:
@domenic sgtm, thanks!
@annevk how do you feel about current changeset? Flag name, processing logic, etc.

Overview.src.html
```diff
@@ -4224,7 +4224,7 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
 
 <p>To <dfn title=concept-BodyInit-extract>extract</dfn> a <span title=concept-body>body</span> and a
 `<code title>Content-Type</code>` <span title=concept-header-value>value</span> from
-<var>object</var>, run these steps:
+<var>object</var> with optional <var>keepalive-body-size</var>, run these steps:
```
annevk: `keepaliveBodySize` (`Content-Type` is an exception that should maybe be modified at some point).

Overview.src.html
```diff
@@ -4487,6 +4493,7 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   readonly attribute <span>RequestCache</span> <span title=dom-Request-cache>cache</span>;
   readonly attribute <span>RequestRedirect</span> <span title=dom-Request-redirect>redirect</span>;
   readonly attribute DOMString <span title=dom-Request-integrity>integrity</span>;
+  readonly attribute boolean <span title=dom-Request-keepAlive>keepAlive</span>;
```
annevk: keepAlive or keepalive? We should stay consistent. https://en.wikipedia.org/wiki/Keepalive suggests the time has come where it is one word, but HTTP sets precedent in the other direction unfortunately.

Overview.src.html
```diff
@@ -4630,10 +4638,14 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <span title=concept-request-cache-mode>cache mode</span>,
  <span title=concept-request-redirect-mode>redirect mode</span> is
  <var>request</var>'s
- <span title=concept-request-redirect-mode>redirect mode</span>, and
+ <span title=concept-request-redirect-mode>redirect mode</span>,
  <span title=concept-request-integrity-metadata>integrity metadata</span> is
  <var>request</var>'s
- <span title=concept-request-integrity-metadata>integrity metadata</span>.
+ <span title=concept-request-integrity-metadata>integrity metadata</span>, and
+ <span title=concept-request-keep-alive-flag>keep-alive flag</span> is
```
annevk: This flag hasn't been defined it seems?

Overview.src.html
```diff
@@ -4734,6 +4746,10 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <var>request</var>'s
  <span title=concept-request-integrity-metadata>integrity metadata</span> to it.
 
+ <li><p>If <var>init</var>'s <code title>keepAlive</code> member is present, set
```
annevk: then set*

Overview.src.html
```diff
@@ -4291,6 +4291,12 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   </dl>
 
  <li>
+  <p>If <var>keepalive-body-size</var> is provided and <var>object</var>'s type is
+  <code title=concept-ReadableStream>ReadableStream</code>, or <var>object</var>'s size is greater
+  than <var>keepalive-body-size</var>, throw a <code title>TypeError</code> and abort the
```
annevk: then throw* and no need to say the remaining steps are aborted. That follows from throwing (new convention we're trying to use).

Overview.src.html
```diff
@@ -4802,7 +4818,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   <ol>
    <li><p>Let <var>Content-Type</var> be null.
 
-   <li><p>Set <var>inputBody</var> and <var>Content-Type</var> to the result of
+   <li><p>If <var>init</var>'s <code title>keepAlive</code> member is present and is set to
+   <code>true</code>, then set <var>inputBody</var> and <var>Content-Type</var> to the
+   result of <span title=concept-BodyInit-extract>extracting</span> <var>init</var>'s
+   <code title>body</code> member with <var>keepalive-body-size</var> set by user agent policy.
```
annevk: I suggest we introduce keepalive-body-size in an earlier step so it's more clear it's user-agent defined.

Overview.src.html
```diff
@@ -4802,7 +4818,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   <ol>
    <li><p>Let <var>Content-Type</var> be null.
 
-   <li><p>Set <var>inputBody</var> and <var>Content-Type</var> to the result of
+   <li><p>If <var>init</var>'s <code title>keepAlive</code> member is present and is set to
```
annevk: You need a newline after the `<li>` since it contains two `<p>`s.

Overview.src.html
```diff
@@ -4802,7 +4818,17 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
+   <code>true</code>, then set <var>inputBody</var> and <var>Content-Type</var> to the
+   result of <span title=concept-BodyInit-extract>extracting</span> <var>init</var>'s
+   <code title>body</code> member with <var>keepalive-body-size</var> set by user agent policy.
+   Rethrow any exceptions.
+   <p class="note no-backref">This step ensures that requests that are allowed to outlive the
```
annevk: Newline before the `<p>` and probably also should be indented one space more.

Overview.src.html
```diff
@@ -4291,6 +4291,12 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
   </dl>
 
  <li>
+  <p>If <var>keepalive-body-size</var> is provided and <var>object</var>'s type is
+  <code title=concept-ReadableStream>ReadableStream</code>, or <var>object</var>'s size is greater
```
annevk: a `ReadableStream` object*

annevk: Also, how do we determine _object_'s size? As I understand it that can be quite hard for `FormData`.

Overview.src.html
igrigorik: I'll rename to lowercase "keepalive" throughout.

Overview.src.html
igrigorik: > <p>A <span title=concept-request>request</span> has an associated
<dfn>keep-alive flag</dfn>. Unless stated otherwise it is unset.

I think that should do the trick?

Overview.src.html
igrigorik: I can change that, but (fyi) it's inconsistent with surrounding language.. e.g.

>  <li><p>If <var>init</var>'s <code title>integrity</code> member is present, set
 <var>request</var>'s <span title=concept-request-integrity-metadata>integrity metadata</span> to it.

I'll keep as is for now, let me know if you want me to change it.

igrigorik:
@annevk thanks for the review, I think I addressed most of it in 759a8fa, modulo...

> Also, how do we determine object's size? As I understand it that can be quite hard for FormData.

That's a great question.. I was hoping you could point me towards a handy hook somewhere. :) 

How does 5.5 step 5 derive it?

> If httpRequest's body is non-null, set contentLengthValue to httpRequest's body's total bytes, utf-8 encoded.

Presumably this step has the same issue as you outlined above?

Overview.src.html
annevk: Yeah, though "keepalive" per earlier comment and maybe a note following it that describes it would be good. Ideally all request fields have a note describing them, but I haven't gotten around to do that just yet.

Overview.src.html
annevk: Please change it. Existing language should also be updated at some point. But it's better if new language does it from the start for git blame and such.

annevk:
Yeah, I think it does. The main difference is that currently we don't need to know the size synchronously as it's not exposed anywhere. This would expose it synchronously and I don't think user agents would be happy with that.

igrigorik:
That's true. Well, I guess the other approach we have here is:

- Keep the existing logic to throw TypeError if body is a ReadableStream.
- Add **new** step in `HTTP-network-or-cache fetch` after we obtain `contentLengthValue`, which checks that the size is less than the ~UA-set size for requests with keepalive flag set. If the condition doesn't hold, return a network error.

~section 5.5, step 7:

> If `contentLengthValue` is non-null, the `httpRequest`'s keepalive flag is set, and `contentLengthValue` is greater than the UA-set maximum body size for such requests, return a network error. 

WDYT?

annevk:
"user-agent-defined maximum body size", but yes, that could work. I'd love @esprehn or someone else from Blink to sign off on that though.

Overview.src.html
```diff
@@ -4630,10 +4638,14 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
...
  <var>request</var>'s
- <span title=concept-request-redirect-mode>redirect mode</span>, and
+ <span title=concept-request-redirect-mode>redirect mode</span>,
  <span title=concept-request-integrity-metadata>integrity metadata</span> is
  <var>request</var>'s
- <span title=concept-request-integrity-metadata>integrity metadata</span>.
+ <span title=concept-request-integrity-metadata>integrity metadata</span>, and
+ <span title=concept-request-keepalive-flag>keepalive flag</span> is
+ <var>request</var>'s
+ <span title=concept-request-keepalive-flag>keepalive flag</span>
```
annevk: title can be removed here too.

Overview.src.html
```diff
@@ -4734,6 +4746,10 @@ <h4 id="should-response-to-request-be-blocked-due-to-nosniff?"><dfn title="shoul
  <var>request</var>'s
  <span title=concept-request-integrity-metadata>integrity metadata</span> to it.
 
+ <li><p>If <var>init</var>'s <code title>keepalive</code> member is present, then set
+ <var>request</var>'s
+ <span title=concept-request-keepAlive>keepalive flag</span> to it.
```
annevk: And here.

igrigorik:
@annevk updated, PTAL. In the meantime, I'll ping some Blink folks for a review -- stay tuned.

esprehn:
Yeah I think that works. Btw what makes it clear that the fetch result promise will never resolve on a keepalive request if the page is destroyed?

annevk:
I don't think anything at the moment. We don't have a clear "page is destroyed" hook yet. What does that mean, by the way? Browsing context going away (`iframe` element removal), navigation, something else? There's a lot around the lifetime of the various global objects that's extremely poorly defined at the moment.

tyoshino:
the spec change lgtm. we still have issues with our implementation for the sendBeacon and this, but I agree this is what we should pursue. I guess Chromium could start implementing this in Q117.

annevk:
@youennf any feedback from WebKit?

Apart from some minor nits I can take care of when merging this looks okay to me now.

youennf:
Makes sense to me.
I am not sure how we might handle the FormData case in WebKit though.

I guess the keep-alive attribute might be useful for service workers although for some of the keep-alive use-cases, there might be no need to resolve/create any response.

igrigorik:
@tyoshino @youennf thanks for the review! @annevk sounds like you got it from here.. that said, let me know if you want my help with anything.

annevk:
@igrigorik the only thing that remains doing is writing tests. I assume that will happen as part of implementation work, but if you all could make sure they end up in https://github.com/w3c/web-platform-tests that'd be great.

annevk:
Also, thank you for your work on this!

igrigorik:
@annevk yep, thanks! I'll a tests TODO on my list..

wanderview:
I'm sorry I didn't provide more feedback sooner, but can someone explain why we need to expose a `keepalive` flag to script?  Why can't the browser just consume the `Request` immediately when `sendBeacon()` is called, store an internal request, and do the fetch later?

It feels really weird to me to ask devs to specify a flag for something the browser can probably do automatically in implementation.

annevk:
How would this work with sendBeacon()? E.g., what this enables is a keepalive HEAD. Not possible today.

wanderview:
> How would this work with sendBeacon()?

The same way we effectively keep Request objects in Cache API beyond the end of an environment settings object.  The browser has to maintain an internal representation in some way.  It doesn't need to keep the JS object alive.

But maybe I don't understand what you mean by "keepalive HEAD".

annevk:
E.g., how would you write `fetch(url, {keepalive:true, method:"HEAD"})` using the `navigator.sendBeacon()` API?

wanderview:
Anne explained further in IRC that this is about not canceling the fetch operation if the environment settings object is destroyed.  I got confused because the PR and spec text talk more about the lifecycle of (R|r)equest objects.

Anyway, carry on.

igrigorik:
From https://github.com/w3c/beacon/pull/27#issuecomment-253551671: 

> I realize I'm late to the party here, but IMO it would have been nicer to make sendBeacon() drain the Request to an internal storage space and then send it out later from there. Then the Request and its JS context don't need to stick around unless the upload body is a JS sourced stream.

@tyoshino @annevk @youennf any thoughts or opinions on the above?

annevk:
I think @wanderview misunderstood what was going on due to some unclear wording in this PR's title. See comments above.

igrigorik:
OK, never mind.. Onwards! :-)

Issue #164, opened 2015-11-29
OpenEndedDictionary clarification


mseddon:
In the spec we are told:

OpenEndedDictionary<T> is a future IDL construct. Expect it to be used as such:

``` javascript
var meta = { "Content-Type": "text/xml", "Breaking-Bad": "<3" }
new Headers(meta)
```

Sadly the currently linked [IDL draft](https://heycam.github.io/webidl/) does not specify what this means.  Can someone clarify how it differs from a regular dictionary?

annevk:
Sorry, I was hoping we'd have this resolved by now.

@heycam, when will https://www.w3.org/Bugs/Public/show_bug.cgi?id=27008 be resolved in some way? Is @bzbarsky's solution of "optional object" plus prose really the best we can do?

heycam:
I think that's the most straightforward thing to do. I'll reply in the bug.

bzbarsky:
@annevk My solution is the best we can do for the semantics Domenic described in that bug.  It's not clear to me whether those are the semantics you actually want (and in particular he seems to be trying to avoid coercions in ways that I'm not sure the spec should actually avoid in this case).

bzbarsky:
@annevk Oh, and the other slightly open question is what the behavior should actually be.  As in, Domenic's semantics involve calling the actual "append" method (which may be overridden by the web page, subclassed, etc) whereas the current spec language is calling the internal "append" method instead.  This difference is super-important: if you're going to call the internal method, then of course you want to do the type coercion before you do that, but if you're going to call `this.append` then of course you _don't_ want to do the type coercion before doing so.

annevk:
@domenic, what do you recommend?

annevk:
(@mseddon, how would you like to be acknowledged?)

mseddon:
@annevk I am unsure what you mean by that?

annevk:
I attempt to acknowledge each contribution, however minor, here: https://fetch.spec.whatwg.org/#acknowledgments.

annevk:
@heycam I left a suggestion in that bug for how this could be something that IDL does for you automatically.

mseddon:
Oh, I see.  Thanks :)

My name is Matt Seddon, although I would also very much like to stress that I am really reporting this issue on behalf of Henry Story, @bblfish.

bblfish:
[...]

PR #398, opened 2016-10-23
Convert to bikeshed

dontcallmedom:
A first stab at converting fullscreen to bikeshed.

Verifications I have made after the conversion:
* all the links to external specs from the current fetch appear in the same order in the new one (which has more of those due to autolinking done in bikeshed from WebIDL)
* a visual inspection of the HTMLDiff (ignoring the transformation of "'s" to "’s")

The conversion doesn't preserve all the ids yet; I'm not sure how critical that is.

Bikeshed complains about the invalid IDL in the definition of `HeadersInit` and thus doesn't apply some of its magic to the fragment; not sure how that should be handled. 

It also tries to interpret the `'self'` in content policy as a CSS property description (and naturally fails to do so); I haven't figured out how to get rid of that spurious error yet.

It also warns about missing ids on examples and headings, as well as non-reused `<var>` - I don't think any of those are particular important to fix in the short term at least.

The bikeshed markup can probably be simplified some more; I've focused on fixing the results of the automatic conversion, not optimizing it for author convenience which can hopefully be done at a later stage.

I cherry-picked and hand-coded the 2 commits that were merged after I got started, so this might need more git magic before merging, but I thought I would bring this is now to avoid further divergence.


annevk:
Thank you for working on this!

So I think I need to merge #383 before this is done at least since @jungkees has been at it for a while and we just finished the HTML side (it's rather small thankfully). Seems reasonable for other PRs to be adjusted post-merge since they haven't been super active.

> The conversion doesn't preserve all the ids yet; I'm not sure how critical that is.

Depends on the IDs. IDs for the reference section and items in it are fine, but IDs for definitions and the main sections should remain stable. Fetch has many incoming links.

> Bikeshed complains about the invalid IDL in the definition of HeadersInit and thus doesn't apply some of its magic to the fragment; not sure how that should be handled.

Don't mark it up as IDL for now. We'll fix this as part of #164.

> It also tries to interpret the 'self' in content policy as a CSS property description (and naturally fails to do so);

@mikewest probably knows how to fix this.

> It also warns about missing ids on examples and headings

Would it help if I made an editorial change to the standard to fix these beforehand?

> not optimizing it for author convenience which can hopefully be done at a later stage.

That sounds good. That makes the diff more readable.

dontcallmedom:
I'll work on the ids then.

re 'self', I've looked at https://github.com/w3c/webappsec-csp/blob/master/index.src.html and the trick seems to be escaping with `` - will fix.

I'll also un-markup the WebIDL for now.

Re ids in headings and examples, I would prefer if we fixed that after merging - they're only warnings, and fixing this in base doc won't help much.

annevk:
I landed the PR I mentioned. The other thing that needs to be done is that the renaming of the files should be a separate commit from the conversion changes. That confuses git blame the least as I understand it.

If @foolip has the bandwidth I'd appreciate if he could review it as well once @dontcallmedom considers it ready for review given that the conversion he did of Fullscreen went great.

mikewest:
@dontcallmedom: Right, `Markup Shorthands: css off` is the way to be.

Yay! This should vastly improve the linkability of the Fetch spec via Shepard, and means I can cross "Figure out how Anolis uses titles." off my todo list! :)

dontcallmedom:
I have now integrated the commits from the last merged-in PR, and split to convert commit into rename then convert.
@foolip can you take a look?

foolip:
There are still conflicts. What I did in https://github.com/whatwg/fullscreen/pull/49 was to write a script for the conversion to do as much as possible and then I verified that IDs and links did not change. If any of these steps were manual, I'd suggest also diffing the text. Will take a look.

foolip:
What is "Add request's reserved client and target client id" doing on the branch?

dontcallmedom:
@foolip I used a [derivative of your script to do a big part of the conversion](https://github.com/whatwg/fetch/pull/398/commits/7b275ebd40b886e735a57356ef6894c506c1ec6b), but there was still lots of manual changes as well. I also htmldiff'd the resulting file and verified that all ids of the original file showed up in the resulting file.
For links, I checked that all external links didn't change.

dontcallmedom:
@foolip @annevk asked to integrate that just-landed PR to the conversion

foolip:
OK, so I'll try to compare to commit fb87b70df0af3a5e6bc7bb9a91b4dc7eb4924acb. If this is intended to land as more than one commit (separate rename and rewrite?) can you force push a new branch with those commits?

dontcallmedom:
@foolip the current branch has separate rename  (eb9e5d9) and rewrite (7b275eb) commits, so I'm not sure I understand your request

foolip:
```
$ git log --oneline origin/master..
c949f0d Add request's reserved client and target client id
61fd8a3 Preserve ids
e2df02a Avoid error on not-yet-valid WebIDL
38e9ee5 Avoid interpretation of 'self' as a CSS propdesc
8968b9c Editorial: note for `invalid` argument delete()
280cf75 Allow fetch() to outlive document/workers
d35bb19 Use backwards-compatible ids where possible
1521473 Re-add interal links to interface objects and Origin header
53eb7d7 Fix quirks of automated conversion
6db2670 Fix references
3cfec23 Escape internal slots syntax to avoid bikeshed confusing
0b30857 Add line return after <pre> for bikeshed's sake
ca2e292 Set alias for references
21976de Add metadata
7b275eb Convert to bikeshed using convert-to-bikeshed.html
eb9e5d9 Rename for bikeshed conversion
```

Should all commits after eb9e5d9 be a single commit then?

foolip:
This sanity check passes:
```JavaScript
  for (var e of document.querySelectorAll('*')) {
    if (e.id)
      console.log('id: ' + e.id);
    if (e.hasAttribute('href') && e.className != 'self-link')
      console.log('href: ' + e.getAttribute('href'));
  }
```

annevk:
@dontcallmedom you will need to rebase your branch so that the eventual two commits (rename, convert) can be applied on top of master.

foolip:
Bah, I was comparing the wrong thing, I assumed that Overview.html was updated on the branch but it isn't touched. Trying again with the generated fetch.html.

dontcallmedom:
#399 has the simplified mergeable commit tree

foolip:
OK, so the logs from https://github.com/whatwg/fetch/pull/398#issuecomment-255728651 reveal a number of problems, links that have changed by accident, ids that have changed, and so on.

It would be good to see this diff be as small as possible, with any intentional changes done before or after this PR.

dontcallmedom:
@foolip can you share your logs (either by email or here)? I didn't detect breaking changes with my own comparison (although there are changes due to the fact that bikeshed produces more links/ids), but could easily have missed some.

This PR is supposed to do only backwards-compatibility changes, with just a couple of exceptions (due to a buggy `data-analis-spec` IIRC).

foolip:
Sure, [here it is](https://github.com/whatwg/fetch/files/547836/fetch.txt).

foolip:
Comparing the copied-from-browser-to-plaintext with `git diff --word-diff -w` I see "BodyInit;typedef" and the addition of lots of references that weren't there, in particular "[HTTP-SEMANTICS] [HTTP-COND] [HTTP-CACHING] [HTTP-AUTH]" appended to "[HTTP]".

dontcallmedom:
re addition of references, that was unavoidable since bikeshed doesn't support having one reference name matching several specs (afaict).

The BodyInit;typedef thing is because bikeshed seems to skip a line return in the pre, will investigate.

Am looking at the ids and href

foolip:
This kind of this is possible if one doesn't like the names that specref uses:
```
<pre class=biblio>
{
    "CSS": {
        "aliasOf": "CSS2"
    }
}
</pre>
```

dontcallmedom:
@foolip but that only renames existing references (which I've done in my pull request), it doesn't let you associate one name to several specs (or does it?)

foolip:
Oh, so the problem is that [HTTP] and [HTTPVERBSEC] are multiple links. I'll defer to @annevk about what to do about it. Looks like you've already reduced the number of references, but "CORS protocol and HTTP caches" still seems to refer to more than it needs to?

dontcallmedom:
right, I trimmed the list where it was obvious, but didn't where it wasn't to me, in the spirit of staying as close to the original spec as possible.

dontcallmedom:
@foolip I've pushed in #399 an updated fetch - I did spot and fixed quite a few errors thanks to your diff. I also reverted some of the fixes that I had done while converting the spec to facilitate comparison - I'll bring them as a separate PR later.

The diff is still not empty, due among other things to:
* bikeshed using #biblio rather than #ref 
* bikeshed adding ids to examples
* bikeshed auto-linking the plural form of "navigation requests" (I guess I could `<span>` it to avoid it and fix it in my PR, but seems superfluous)
* bikeshed NOT autolinking to the prose description of the constructor (or at least, me not knowing how to make it happen)
* different ids (and thus autolinks) generated on IDL fragments

If you regenerate a diff, I can probably annotate it with my explanations of that helps

dontcallmedom:
strike the bit about constructors, I've figured it out

annevk:
> bikeshed adding ids to examples

I figured things like this is why we might want to do an initial PR to reduce these differences. @foolip did something similar for Fullscreen: https://github.com/whatwg/fullscreen/commit/6f43f8e972c6b456a2a8fe8fda0501f9b6390a4d.

dontcallmedom:
@annevk that makes more sense now that I understand how you want to merge this. I'm not sure if this is still worth it at this point though, esp given that I don't think we can get the diff to be entirely empty. 

I guess you or @foolip should let me know if you feel that'd still be preferable though.

annevk:
Slight preference towards adding IDs beforehand, but not a dealbreaker for me.

PR #399, opened 2016-10-24
Convert to bikeshed (simplified)

dontcallmedom:


annevk:
What about the output? Overview.html becomes fetch.html and should be here too.

dontcallmedom:
I wasn't sure how you wanted to handle that piece; I'll update with that change in.

annevk:
I missed that you updated this PR (best to add a comment as I don't think notifications get triggered). I'll take a look and maybe @foolip can too.

annevk:
I ran @foolip's comparison script mentioned in https://github.com/whatwg/fullscreen/commit/f9df3ea73f1f32490699e0a925d7e0822c2084a1 and looked at the diff between the two documents. These seem wrong:
```diff
-href: https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-target-browsing-context
+href: https://html.spec.whatwg.org/multipage/infrastructure.html#concept-request-target-browsing-context
```
I wonder if for these we should make use of bikeshed's header "`<dfn>`" feature to avoid the new ID:
```diff
 id: should-response-to-request-be-blocked-due-to-mime-type?
+id: should-response-to-request-be-blocked-due-to-mime-type
 id: should-response-to-request-be-blocked-due-to-nosniff?
+id: should-response-to-request-be-blocked-due-to-nosniff
```
In #origin-header the `Origin` header mentioned in the note has lost its hyperlink.
```diff
-href: #http-origin
```
In a note in the Response class section a link points to the wrong thing now:
```diff
-href: #dom-response-trailer
+href: #concept-response-trailer
```
I noticed a couple of good fixes too:
```diff
+href: https://url.spec.whatwg.org/#concept-url-origin
+href: #concept-header-name
+href: #concept-request-current-url
+href: #concept-request-cache-mode
```

dontcallmedom:
I've updated the pull request:
* fixed the link for target browsing context
* removed the duplicate ids on "should…" using `<h3 dfn>`
* fixed the link on `<code>trailer</code>`

I've resisted fixing the link on the `Origin` (assuming this is the note about supplanting the IETF Origin spec) - it feels wrong to point to the header defined in fetch in this context. But if you'd rather see that fixed separately, I'll happily oblige.

annevk:
For the Translation I prefer what I have in DOM:
```
!Translation (non-normative): <span title=Japanese><a href=https://triple-underscore.github.io/DOM4-ja.html lang=ja hreflang=ja rel=alternate>日本語</a></span>
```
That way it appears under Commits and is slightly less offensive ("likely out-of-date" is not very friendly).

Anyway, I can also fix that up when I adjust the commits slightly.

@foolip any comments?

foolip:
When my console log output looks good, maybe compare the plaintexts with `git diff --word-diff -w` as a final sanity check? (If you use `git diff` outside of a repo it works with any two files.)

It might be noisy and wrong, but it might also be worth logging headings, `<var>`, `<code>` and other things that have some semantic meaning and affect how things look, to see if things changed accidentally.

annevk:
After doing some cleanup:

* Wonder whether I should fix the remaining issues with HTML linking first
* And maybe fix other linking issues
* Wonder if there is more spurious usage of `for`
* Found a bug with respect to the non-relative flag as URL changed that term

This is generally in good shape though, so maybe that can all be follow up...

Anyone else wants to review?

annevk:
@foolip I found one example using word-diff that had incorrect formatting due to a missing newline after a `<pre>` bikeshed insists upon. I didn't include all my further cleanup in the comparison, but that should be mostly good.

I think I'm happy to merge this by end-of-day.

The only thing I plan on changing is the commit messages (need to start with "Editorial: ") and I plan on squashing all my commits into the second commit.

foolip:
Sounds like all is well, I won't review further since I'd just use the same scripts again.

annevk:
I forgot to update the README and remove the Makefile. Now done. I also prepared the commits.

annevk:
Thanks @dontcallmedom!

PR #404, opened 2016-10-28
Reference RFC 7578 instead of RFC 2388

eehakkin:
RFC 2388 has been obsoleted by RFC 7578. This change reflects a similar
change applied to the HTML spec.

See https://github.com/whatwg/html/issues/398 for discussion.


annevk:
Thank you!

Issue #409, opened 2016-11-04
Editorial: remove direct w3c.github.io links


annevk:
It seems these are all for WebAppSec drafts and could use the bikeshed autolinking system instead. @mikewest interested in looking at this since I think you put most of these in?

mikewest:
Would you like to link to the `/TR/` document instead? I haven't linked to those just because we haven't set up autopublication via Travis yet. I generally publish manually, and not as often as I should... That's something we should address, it just hasn't been on anyone's priority list...

/cc @hillbrad @dveditz @wseltzer who might have spare cycles. :)

annevk:
No I don't. What I want is to use Shepherd's autolinking feature.

annevk:
I really rather avoid linking to TR/ since that seems worse than the editor's draft.

annevk:
To be extra clear, this is meant as editorial cleanup of the source, I expect the output to be identical.

mikewest:
Ah. I see, and agree. I'll send you a PR.

annevk:
https://github.com/w3c/webappsec-upgrade-insecure-requests/commit/6fabd5bd2f132ea04be1199bc0cedeb3b797080a also helps with this.

PR #410, opened 2016-11-04
Use auto-linking for CSP3.

mikewest:
Rather than hard-coding links to CSP, we should use the magic of Bikeshed
automatically link to the right spot with the right text.

Addresses part of whatwg/fetch#409.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1223,7 +1223,7 @@ Unless stated otherwise it is `<code>OK</code>`.
 
 <p>A <a for=/>response</a> has an associated
 <dfn export for=response id=concept-response-csp-list>CSP list</dfn>, which is a
-list of <a href=https://w3c.github.io/webappsec-csp/#policy>Content Security Policy objects</a>
+list of <a lt="policy" for="/">Content Security Policy objects</a>
 for the <a for=/>response</a>. The list is empty unless otherwise
 specified. [[!CSP]]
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2393,7 +2393,7 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
  <a lt="block bad port">should fetching <var>request</var> be blocked due to a bad port</a>,
  <a href=https://w3c.github.io/webappsec-mixed-content/#should-block-fetch>should fetching <var>request</var> be blocked as mixed content</a>,
  or
- <a href=https://w3c.github.io/webappsec-csp/#should-block-request>should fetching <var>request</var> be blocked by Content Security Policy</a>
+ [[csp-3#should-block-request]]
```
annevk: The output of this is "Content Security Policy §4.1.3 Should request be blocked by Content Security Policy?" which seems suboptimal. Are you not exporting this as a dfn? Hardwiring a dependency on csp-3 seems troublesome as well.

mikewest: Hrm. CSP marks up headers as algorithms:

```
<h3 id="foo" algorithm>
  Foo the bar.
</h3>
```

That syntax doesn't seem to export a definition; it does export the header, which the syntax in this patch links to. I agree that the linking text is a little verbose... But CSP would be (even more!) verbose if I added a distinct `<dfn>` with the same text as the heading.

@tabatkins: is there a better syntax I should be using?

annevk: You can use `<h3 id=foo algorithm dfn export>` maybe?

tabatkins: Yes, if you tag a heading with one of the dfn-type attributes (like `dfn`, here) it becomes a definition too.  (And as Anne's example shows, if you're using `dfn`, don't forget to `export` it too unless you want it private.)

PR #413, opened 2016-11-08
Export 'keepalive flag', and tie it to 'request'.

mikewest:


fetch.bs
```diff
@@ -744,7 +744,8 @@ explicitly set <a for=/>request</a>'s
 <a for=request>window</a>.
 
 <p id=keep-alive-flag>A <a for=/>request</a> has an associated
-<dfn>keepalive flag</dfn>. Unless stated otherwise it is unset.
+<dfn for=request export id=concept-keepalive-flag>keepalive flag</dfn>. Unless stated otherwise it
```
annevk: I don't think this ID was ever in use. Let's not add it.

mikewest: Hrm. It's inconsistent with the rest of the link structure. *shrug* But whatever. Happy to remove it if you're shifting to a different pattern going forward.

annevk:
Thanks!

PR #418, opened 2016-11-23
Simplify HeadersInit

annevk:
Since Headers has an iterator now through its iterable<> declaration
there’s no real need to special case.

It might be slightly faster to have it there, but this is not
performance-critical code and I’d rather have it be simple.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4015,7 +4015,7 @@ fetch("/music/pk/altes-kamuffel.flac")
 
 <h3 id=headers-class>Headers class</h3>
 
-<pre class=idl>typedef (Headers or sequence&lt;sequence&lt;ByteString>> or record&lt;ByteString, ByteString>) HeadersInit;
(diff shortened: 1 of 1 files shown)


annevk:
@bzbarsky @wanderview @domenic thoughts?

wanderview:
Does this have any change on behavior visible to script?  If not, then I I have no objections.

annevk:
It does. If you overwrite the iterator of a Headers object that would be invoked instead by the binding layer.

bzbarsky:
Right, the behavior also changes if people have messed with array iterators (so the inner sequence produces unexpected values).

It's probably OK to not worry about those edge cases.

domenic:
Looks good; tests would probably help.

annevk:
I created https://github.com/w3c/web-platform-tests/pull/4240. It seems we already had many tests for IDL records since everyone implemented that somehow before IDL got them.

annevk:
Tests landed. Waiting with landing this until a Shepherd issue is resolved.

PR #419, opened 2016-11-23
Enforce limit on inflight keepalive bytes

igrigorik:
Requests with keepalive flag set are allowed to outlive the environment
settings object. We want to make sure that such requests do not
negatively impact the user experience when a page is unloaded, etc.

This limits the amount of (body) bytes that can be inflight at any point
when the request has the keepalive flag set; this flag is set by
sendBeacon.

Background: https://github.com/w3c/beacon/pull/39

***
[Preview](https://s3.amazonaws.com/pr-preview/igrigorik/fetch/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/68a9867..igrigorik:master:1bd7e76.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3180,10 +3180,17 @@ steps:
 
  <li>
   <p>If <var>contentLengthValue</var> is non-null, <var>httpRequest</var>'s
-  <a>keepalive flag</a> is set, and <var>contentLengthValue</var> is greater than a
-  user-agent-defined maximum, then return a <a>network error</a>.
+  <a>keepalive flag</a> is set:
 
-  <p class="note no-backref">The above user-agent-defined maximum ensures that requests that are
+  <ul>
+    <li>Let <var>group</var> be <var>httpRequest</var>'s <a>client</a>'s <a>fetch group</a>.
+    <li>Let <var>inflight keepalive bytes</var> be the sum of <a>request</a> <a>body</a>'s
+    <a>total bytes</a> for each <a>record</a> in <var>group</var> whose <a>done flag</a> is unset.
+    <li>If the sum of <var>contentLengthValue</var> and <var>inflight keepalive bytes</var> is
+    greater than 64KB, then return a <a>network error</a>.
+  </ul>
+
+  <p class="note no-backref">The above limit ensures that requests that are
   allowed to outlive the <a>environment settings object</a> and contain
   a body, have a bounded size and are not allowed to stay alive indefinitely.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -3178,31 +3178,35 @@ steps:
  <var>httpRequest</var>'s
  <a for=request>header list</a>.
 
-<li>
+ <li>
   <p>If <var>contentLengthValue</var> is non-null and <var>httpRequest</var>'s
-  <a for=request>keepalive flag</a> is set:
+  <a for=request>keepalive flag</a> is set, then:
 
-  <ul>
-    <li>Let <var>inflight keepalive bytes</var> be zero.
-    <li>Let <var>group</var> be <var>httpRequest</var>'s <a>client</a>'s <a>fetch group</a>.
-    <li>Let <var>inflight records</var> be the set of <a>records</a> in <var>group</var> whose
-    <a for=request>keepalive flag</a> is set, and whose <a>done flag</a> is unset.
-    <li>
-      <p>For each <var>rec</var> in <var>inflight records</var>:
+  <ol>
+   <li><p>Let <var>inflight keepalive bytes</var> be zero.
 
-      <ul class=brief>
-        <li>Let <var>inflight request</var> be <var>rec</var>'s <a>request</a>.
-        <li>Increment <var>inflight keepalive bytes</var> by <var>inflight request</var>'s
-        <a for=request>body</a>'s <a for=body>total bytes</a>.
-      </ul>
+   <li><p>Let <var>group</var> be <var>httpRequest</var>'s <a>client</a>'s <a>fetch group</a>.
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -3122,10 +3122,17 @@ steps:
 
 <li>
   <p>If <var>contentLengthValue</var> is non-null, <var>httpRequest</var>'s
-  <a>keepalive flag</a> is set, and <var>contentLengthValue</var> is greater than a
-  user-agent-defined maximum, then return a <a>network error</a>.
+  <a>keepalive flag</a> is set:
```
annevk: s/,/ and/

fetch.bs
```diff
@@ -3122,10 +3122,17 @@ steps:
...
 
-  <p class="note no-backref">The above user-agent-defined maximum ensures that requests that are
+  <ul>
+    <li>Let <var>group</var> be <var>httpRequest</var>'s <a>client</a>'s <a>fetch group</a>.
+    <li>Let <var>inflight keepalive bytes</var> be the sum of <a>request</a> <a>body</a>'s
+    <a>total bytes</a> for each <a>record</a> in <var>group</var> whose <a>done flag</a> is unset.
```
annevk: I get what you're trying to say, but this sentence does not read well. Maybe "sum of each X's Y in Z"?

I also think this is lacking some `for=""` attributes since the request you're talking about here is that of the fetch group, as is the record.

fetch.bs
```diff
@@ -3122,10 +3122,17 @@ steps:
...
+    <li>If the sum of <var>contentLengthValue</var> and <var>inflight keepalive bytes</var> is
+    greater than 64KB, then return a <a>network error</a>.
```
annevk: KiB, presumably? I wonder if we should use `<abbr>` for the occasion.

igrigorik:
@annevk ack, updated. I've split up the sum step.. hopefully that makes it more clear.

annevk:
You need to use `<ol>` instead of `<ul>` and generally things are indented with only a single space and `<li>` are separated by a newline (though not from the opening `<ol>` and closing `</ol>`).

I also think some steps can be grouped together again, but I'm happy to make that change later on.

Is there someone else that can review this? And do we have tests for this somewhere?

igrigorik:
@tyoshino could you please take a look? 

@annevk please feel free to wordsmith as needed.

tyoshino:
looks good

annevk:
To be clear, still need an update on the testing scenario.

igrigorik:
@annevk sorry, forgot about that. Are you looking for web-platform tests? No one has implemented this yet, so no, we don't have those.. yet. We (Chrome) are looking to implement this (and associated sendBeacon changes) in short order, so that should be resolved fairly soon.

igrigorik:
*bump* 

@annevk anything else I can help unblock on this one?

annevk:
I'm sorry. I totally forgot about this. Yeah I am looking for web-platform-tests. If they are being written as part of the implementation I suppose that might be sufficient, though not ideal. Perhaps a simple test can be added?

We have a new policy where each standard PR also has a WPT PR and then the former lands after the latter.

igrigorik:
@annevk PTAL: https://github.com/w3c/web-platform-tests/pull/4788

annevk:
Thanks, I'll rebase this PR meanwhile. Tests can probably land pretty quickly if nobody else has feedback and my minor nits are fixed.

annevk:
Hmm, I'm not sure how to rebase this since you added the commits on master.

annevk:
The tests moved to https://github.com/w3c/web-platform-tests/pull/4878. @igrigorik I'm still blocked on you sorting out how to rebase this.

igrigorik:
@annevk apologies about the delay. Updated.. ptal.

annevk:
I made some formatting changes. Please review. Also note that we're currently blocked on https://github.com/tabatkins/bikeshed/issues/936.

igrigorik:
@annevk lgtm, good to merge once Bikeshed issue is resolved.

annevk:
@igrigorik also, it's fine filing bugs now we have an agreed upon standard change and tests. Can you do that? Just file a bug against everyone that hasn't implemented this with a pointer to this PR and the tests.

igrigorik:
AFAIK, Edge already has it implemented.

- Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=695939
- Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1342484

annevk:
We also need one for WebKit, right?

igrigorik:
@annevk indeed. Opened & updated the list above.

annevk:
I'd also appreciate one more final review since I had to make a couple more changes for Bikeshed.

igrigorik:
Took another scrub.. LGTM.

annevk:
Great, thanks for reporting the bugs!

mnot:
I don't suppose a different name for this could be used? Calling it "keepalive" is going to be confusing for some HTTP people...

annevk:
I don't think any code has shipped thus far, though Edge has implemented. So if we have a better name we could rename, but we'd have to be quick I think and @igrigorik should confirm.

igrigorik:
@mnot it's request level keepalive... which—I think—is a good semantic fit with what we're trying to communicate here. I don't think we have have to monopolize keepalive for connection level semantics; most developers will never touch connection keepalive.

mnot:
_most developers will never touch connection keepalive_ 

That's a surprising assertion, but OK...

mnot:
One more possibly stupid question - why is this in `HTTP-network-or-cache fetch` and not `HTTP-network fetch`?

annevk:
No stupid questions. I don't think there's an observable difference and I suspect @igrigorik did it this way because the former already had a handle on the body size.

Issue #403, opened 2016-10-27
Export all <dfn>s that Service Worker uses


domenic:
https://github.com/w3c/ServiceWorker/blob/c8b78bfd84136ce577564f58f0b3fbef1db183eb/docs/index.bs#L68-L135

annevk:
A lot of those are exported already. I guess I can go through manually, but that's a lot of make work... Sigh.

It would be nice actually if Service Worker started making use of bikeshed. E.g., linking to "Handle Foreign Fetch" or "Handle Fetch" doesn't really work.

annevk:
Going through them I'm having a hard time finding one without export...

I think the problem might be that Service Worker doesn't link to them correctly, as that specification is generally a big mess when it comes to bikeshed...

domenic:
I tried deleting all of them from the link-defaults block and Bikeshed couldn't find any of them. Maybe Fetch just hadn't been indexed yet, or maybe they were all phrased slightly wrong. I can try to be more precise in a follow-up to this thread as I continue to try to update Service Worker for HTML.

annevk:
They probably don't use the correct for attributes and such. I think Fetch has been indexed and exporting these for quite a while now.

PR #420, opened 2016-11-24
Export various things for service worker

jakearchibald:
For linking to from the service worker spec


annevk:
I'll keep this open for now in case there's other instances. This will end up fixing #403 then.

jakearchibald:
I'm done!

PR #421, opened 2016-11-24
Emptying a readable stream should be for readablestream

jakearchibald:


annevk:
Would you mind adding this to #420? All these meta changes are fine to group imo.

jakearchibald:
Shall do

PR #424, opened 2016-11-25
Clarify Body package data algorithm with multipart/form-data MIME type

### LINKED ISSUE #392: Clarification of Body package data algorithm with bytes, FormData and multipart/form-data MIME type
The Body package data algorithm with bytes, FormData and multipart/form-data MIME type step 1 says

> Parse bytes, using the value of the `boundary` parameter from MIME type and utf-8 as encoding, per the rules set forth in Returning Values from Forms: multipart/form-data. [RFC2388]

That is not very detailed parsing specification.
While implementing the algorithm for Chromium, I ran into interpretation issues.
My main question is how to parse different kind of multipart parts.

Parts whose Content-Disposition header field contains a filename parameter must surely be parsed into **Blob**s. That is what Mozilla does.

What about multipart parts which have a Content-Type header field but whose Content-Disposition header field does not contain a filename parameter? To my understanding, it is not possible to construct such parts with **FormData** because **USVString**s will never generate a Content-Type header field and **Blob**s have a default file name ("blob"). Should these parts be parsed into **Blob**s in spite of the missing filename parameter?

Should a multipart part with a name "_charset_" have no effect on decoding encoding?
I think that probably yes, because RFC2388 does not even describe that name (but RFC7578 does) and because  the body package data algorithm with bytes, FormData and application/x-www-form-urlencoded MIME type runs the application/x-www-form-urlencoded parser steps with a use_charset_flag implicitly unset thus it does not change decoding encoding based on the "_charset_" component.

If string parts can never have a Content-Type header field, they can never have a charset parameter either. The multipart part with a name "_charset_" does not seem to have an effect on decoding, either. Does it thus follow that string parts must always be decoded using utf-8 encoding?

eehakkin:
Fixes #392.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4364,7 +4364,7 @@ runs the associated steps:
   <ol>
    <li>
     <p>Parse <var>bytes</var>, using the value of the `<code>boundary</code>` parameter from
-    <var>MIME type</var> and <a>UTF-8</a> as encoding, per the rules set forth in
+    <var>MIME type</var>, per the rules set forth in
     <cite>Returning Values from Forms: multipart/form-data</cite>. [[!RFC7578]]</p>
 
     <p>Each part whose `<code>Content-Disposition</code>` header contains a `<code>filename</code>`
@@ -4377,9 +4377,9 @@ runs the associated steps:
 
     <p>Each part whose `<code>Content-Disposition</code>` header does not contain a
     `<code>filename</code>` parameter must be parsed into an <a>entry</a> whose value is the
-    <a>UTF-8 decoded</a> content of the part. This is done regardless of the presence or the value
-    of a `<code>Content-Type</code>` header and regardless of the presence or the value of a
-    `<code>charset</code>` parameter.</p>
+    <a>UTF-8 decoded</a> content of the part. <span class=note>This is done regardless of the
+    presence or the value of a `<code>Content-Type</code>` header and regardless of the presence or
+    the value of a `<code>charset</code>` parameter.</span></p>
 
     <p class=note>A part whose `<code>Content-Disposition</code>` header contains a
     `<code>name</code>` parameter whose value is `<code>_charset_</code>` is parsed like any other
--- a/fetch.bs
+++ b/fetch.bs
@@ -23,7 +23,9 @@ Translate IDs: typedefdef-bodyinit bodyinit,typedefdef-responsebodyinit response
 <script src=https://resources.whatwg.org/dfn.js defer id=head></script>
 
 <pre class=anchors>
-url:https://xhr.spec.whatwg.org/#concept-formdata-entry;spec:XHR;type:dfn;text:entries
+url:https://xhr.spec.whatwg.org/#concept-formdata-entry;spec:xhr;type:dfn;
+    text:entry
+    text:entries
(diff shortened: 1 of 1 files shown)


annevk:
@youennf what do you think? This seems like an improvement. I have a couple nits I'll push in a commit once everyone is on board with this.

yutakahirano:
lg

annevk:
@youenff ping.

annevk:
@youennf ping.

youennf:
Sorry for the delay, looks like an improvement to me as well!

I would split some sentences in two so that they are more readable.
entry might need a link reference.
And I guess entry has a "value" and not "values"

eehakkin:
I split some sentences and corrected plurality errors.
Bikeshed correctly links <a>entry</a> to https://xhr.spec.whatwg.org/#concept-formdata-entry.
Don't know what happens with Travis.

annevk:
I think you can satisfy Travis by rebasing at this point. Sorry about that.

fetch.bs
```diff
@@ -4363,10 +4363,30 @@ runs the associated steps:
...
+    the `<code>Content-Type</code>` header of the part if the part has such header or
+    `<code>text/plain</code>` (the default defined by [[!RFC7578]] section 4.4) otherwise.</p>
+
+    <p>Each part whose `<code>Content-Disposition</code>` header does not contain
+    a `<code>filename</code>` parameter must be parsed into an <a>entry</a>
+    whose value is a {{USVString}} whose content is the <a>utf-8</a> decoded content of the part.
+    This is done regardless of the presence or the value of a `<code>Content-Type</code>` header
+    and regardless of the presence or the value of a `<code>charset</code>` parameter.</p>
+
+    <p>A part whose `<code>Content-Disposition</code>` header contains a `<code>name</code>`
```
annevk: I think this would be clearer as a note (class=note) since it's just a statement of fact.

fetch.bs
```diff
@@ -4363,10 +4363,30 @@ runs the associated steps:
...
+    a `<code>filename</code>` parameter must be parsed into an <a>entry</a>
+    whose value is a {{File}} whose content is the content of the part. The {{File/name}} attribute
+    of the {{File}} must have the value of the `<code>filename</code>` parameter of the part.
+    The {{Blob/type}} attribute of the {{File}} must have the value of
+    the `<code>Content-Type</code>` header of the part if the part has such header or
+    `<code>text/plain</code>` (the default defined by [[!RFC7578]] section 4.4) otherwise.</p>
+
+    <p>Each part whose `<code>Content-Disposition</code>` header does not contain
+    a `<code>filename</code>` parameter must be parsed into an <a>entry</a>
+    whose value is a {{USVString}} whose content is the <a>utf-8</a> decoded content of the part.
```
annevk: Use "whose value is the <a>UTF-8 decoded</a> content of the part" instead.

eehakkin: Done.

fetch.bs
eehakkin: Done.

annevk:
@eehakkin are you okay with the changes I made? Looks pretty good now.

eehakkin:
@annevk: Yes, sure. I am okay with the changes you made.

annevk:
Great, thanks!

Issue #88, opened 2015-07-22
Developer-controlled streams for request/response


annevk:
First https://github.com/yutakahirano/fetch-with-streams needs to be integrated. So developers can access the stream of request/response. Next is giving them control over the stream.
- What if you do `fetch(url, { body })` where body is a stream which gets random stuff (not bytes) enqueued in it?
- Result: when fetch reads from body, it notices a non-`BufferSource`, and ... does what?
  - Returned promise rejects
  - Cancel the stream
  - On the wire, it just terminates somehow? How does HTTP work?
- Require "cors-with-forced-preflight" (you need a preflight in most cases anyway, and it's also required for progress events, which streams enable) (in particular you need it for everything that isn't a POST with one of the `<form>` MIME types)
- Always do content-encoding: chunked, and do uploads as chunked encoding, which we need to spec in some way

mnot:
> On the wire, it just terminates somehow? How does HTTP work?

For H1, you'd just close the connection. For H2, you'd cancel the stream.

> Always do content-encoding: chunked, and do uploads as chunked encoding, which we need to spec in some way

As noted elsewhere, that's not very interoperable with existing, deployed servers; while *some* will accept chunked encoding, *most* will 411 (Length Required) if there's no Content-Length, because they don't want to infinitely buffer the request. 

One way around this would be to specify that if Content-Length is set, you'll only write that many bytes to the wire, and error if it tries to write more.

annevk:
Basic tests: https://github.com/w3c/web-platform-tests/pull/4362. Fetch Standard PR: #425.

amn:
@annevk: Is `chunked` a valid value for `Content-Encoding` HTTP header? I think not? Perhaps you meant `Transfer-Encoding: chunked`?

annevk:
@amn yeah, that's what we ended up with in step 4 and 5 of https://fetch.spec.whatwg.org/#concept-http-network-fetch.

ioquatix:
I have played around with and implemented bi-directional streaming for both HTTP/1 and HTTP/2 in `async-http`.

While it's not possible to demonstrate in a web browser, here is a simple example of streaming from the server to the client: https://utopia-falcon-heroku.herokuapp.com/welcome/index

In addition, using the `async-http` API directly, it's possible to write applications that stream in both directions, and in theory there is no reason why a web-browser couldn't do it too.

> For H1, you'd just close the connection. For H2, you'd cancel the stream.

[...]

PR #425, opened 2016-11-28
Uploading a Request made from a ReadableStream body

yutakahirano:
This change enables developers to create a request from a ReadableStream body. fetch() reads data from the body and sends it to the server.

Fixing https://github.com/yutakahirano/fetch-with-streams/issues/66.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -15,7 +15,7 @@ Markup Shorthands: css off
 !Commits: [SNAPSHOT-LINK]
 !Commits: <a href=https://twitter.com/fetchstandard>@fetchstandard</a>
 !Translation (non-normative): <span title=Japanese><a href=https://triple-underscore.github.io/Fetch-ja.html lang=ja hreflang=ja rel=alternate>日本語</a></span>
-Translate IDs: typedefdef-bodyinit bodyinit,typedefdef-responsebodyinit responsebodyinit,dictdef-requestinit requestinit,typedefdef-requestinfo requestinfo,enumdef-requesttype requesttype,enumdef-requestdestination requestdestination,enumdef-requestmode requestmode,enumdef-requestcredentials requestcredentials,enumdef-requestcache requestcache,enumdef-requestredirect requestredirect,dictdef-responseinit responseinit,enumdef-responsetype responsetype
+Translate IDs: typedefdef-bodyinit bodyinit,dictdef-requestinit requestinit,typedefdef-requestinfo requestinfo,enumdef-requesttype requesttype,enumdef-requestdestination requestdestination,enumdef-requestmode requestmode,enumdef-requestcredentials requestcredentials,enumdef-requestcache requestcache,enumdef-requestredirect requestredirect,dictdef-responseinit responseinit,enumdef-responsetype responsetype
 </pre>
 
 <script src=https://resources.whatwg.org/file-issue.js async></script>
@@ -571,13 +571,16 @@ user-agent-defined <a for=header>value</a> for the
 <p>A <dfn export id=concept-body>body</dfn> consists of:
 
 <ul>
- <li><p>A <dfn export for=body id=concept-body-stream>stream</dfn> (a {{ReadableStream}} object).
+ <li><p>A <dfn export for=body id=concept-body-stream>stream</dfn> (null or a {{ReadableStream}}
+ object).
 
  <li><p>A <dfn export for=body id=concept-body-transmitted>transmitted bytes</dfn>
  (an integer), initially 0.
 
  <li><p>A <dfn export for=body id=concept-body-total-bytes>total bytes</dfn> (an
  integer), initially 0.
+
+ <li><p>A <dfn export for=body id=concept-body-source>source</dfn>, initially null.
 </ul>
 
 <p>A <a for=/>body</a> <var>body</var> is said to be
@@ -2984,6 +2987,10 @@ in addition to <a>HTTP fetch</a> above.
  <a lt="include credential">includes credentials</a>, then return a
  <a>network error</a>.
 
+ <li><p>If <var>actualResponse</var>'s <a for=response>status</a> is not <code>303</code>,
+ <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+ <a for=request>body</a>'s <a for=body>source</a> is null, then return a <a>network error</a>.
+
  <li>
   <p>If <i>CORS flag</i> is set and <var>actualResponse</var>'s
   <a for=response>location URL</a>
@@ -3007,6 +3014,15 @@ in addition to <a>HTTP fetch</a> above.
  <code>303</code>, set <var>request</var>'s <a for=request>method</a>
  to `<code>GET</code>` and <var>request</var>'s <a for=request>body</a>
  to null.
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>body</a> is non-null, then set <var>request</var>'s
+  <a for=request>body</a> to the first part of <a lt=extract for=BodyInit>extracting</a>
+  <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
+
+  <p class="note no-backref"><var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>'s
+  nullity has already been checked. The <a lt=extract for=BodyInit>extracting</a> operation cannot
+  throw as it was called for the same <a for=body>source</a> before.
 
  <li><p>Append <var>actualResponse</var>'s
  <a for=response>location URL</a> to <var>request</var>'s
@@ -3035,17 +3051,31 @@ steps:
 <i>authentication-fetch flag</i>.
 
 <ol>
- <li>
-  <p>Let <var>httpRequest</var> be <var>request</var> if
-  <var>request</var>'s <a for=request>window</a> is "<code>no-window</code>"
-  and <var>request</var>'s <a for=request>redirect mode</a> is
-  "<code>error</code>", and the result of <a lt=clone for=request>cloning</a>
-  <var>request</var> otherwise.
+ <li><p>Let <var>httpRequest</var> be null.
 
-  <p class="note no-backref">A <var>request</var> is typically cloned as it needs to be possible to
-  add <a>headers</a> and read its
-  <a for=request>body</a> without affecting <var>request</var>. As
-  <var>request</var> is reused with redirects, authentication, and proxy authentication.
+ <li><p>If <var>request</var>'s <a for=request>window</a> is "<code>no-window</code>" and
+ <var>request</var>'s <a for=request>redirect mode</a> is "<code>error</code>", then set
+ <var>httpRequest</var> to <var>request</var>.
+
+ <li>
+  <p>Otherwise, run these substeps:
+
+  <ol>
+   <li><p>Set <var>httpRequest</var> to a copy of <var>request</var> except for its
+   <a for=request>body</a>.
+
+   <li><p>Let <var>body</var> be <var>request</var>'s <a for=request>body</a>.
+
+   <li><p>Set <var>httpRequest</var>'s <a for=request>body</a> to <var>body</var>.
+
+   <li><p>If <var>body</var> is non-null, then set <var>request</var>'s <a for=request>body</a> to a
+   new <a for=/>body</a> whose <a for=body>stream</a> is null and whose <a for=body>source</a> is
+   <var>body</var>'s <a for=body>source</a>.
+  </ol>
+
+  <p class="note no-backref">Here we do not <a for=request>clone</a> <var>request</var> in order
+  to reduce memory consumption. <var>request</var> can be reused with redirects, authentication,
+  and proxy authentication.
 
  <li>
   <p>Let <i>credentials flag</i> be set if one of
@@ -3325,6 +3355,22 @@ steps:
   <ol>
    <li class=XXX><p>Needs testing: multiple `<code>WWW-Authenticate</code>` headers, missing,
    parsing issues.
+
+   <li>
+    <p>If <var>request</var>'s <a for=request>body</a> is non-null, then run these subsubsteps:
+
+    <ol>
+     <li><p>If <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is null,
+     then return a <a>network error</a>.
+
+     <li>
+      <p>Set <var>request</var>'s <a for=request>body</a> to the first part of
+      <a lt=extract for=BodyInit>extracting</a> <var>request</var>'s <a for=request>body</a>'s
+      <a for=body>source</a>.
+
+      <p class="note no-backref">The <a lt=extract for=BodyInit>extracting</a> operation cannot
+      throw as it was called for the same <a for=body>source</a> before.
+    </ol>
 
    <li>
     <p>If <var>request</var>'s
@@ -4184,9 +4230,7 @@ method, when invoked, must run these steps:
 <h3 id=body-mixin>Body mixin</h3>
 
 <pre class=idl>
-typedef (Blob or BufferSource or FormData or URLSearchParams or USVString) BodyInit;
-
-typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
+typedef (Blob or BufferSource or FormData or URLSearchParams or ReadableStream or USVString) BodyInit;</pre>
 
 <p>To <dfn id=concept-bodyinit-extract for=BodyInit export>extract</dfn> a <a for=/>body</a> and a
 `<code>Content-Type</code>` <a for=header>value</a> from
@@ -4199,7 +4243,9 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
 
  <li><p>Let <var>Content-Type</var> be null.
 
-<li><p>Let <var>action</var> be null.
+ <li><p>Let <var>action</var> be null.
+
+ <li><p>Let <var>source</var> be null.
 
  <li>
   <p>Switch on <var>object</var>'s type:
@@ -4212,6 +4258,8 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
     <p>If <var>object</var>'s {{Blob/type}}
     attribute is not the empty byte sequence, set <var>Content-Type</var> to its value.
 
+    <p>Set <var>source</var> to <var>object</var>.
+
    <dt><code>BufferSource</code>
    <dd>
     <p><a for=ReadableStream>Enqueue</a> a <code>Uint8Array</code> object
@@ -4219,6 +4267,8 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
     to <var>stream</var> and <a abstract-op>close</a>
     <var>stream</var>. If that threw an exception,
     <a abstract-op>error</a> <var>stream</var> with that exception.
+
+    <p>Set <var>source</var> to <var>object</var>.
 
    <dt>{{FormData}}
    <dd>
@@ -4232,6 +4282,8 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
     <a><code>multipart/form-data</code> boundary string</a> generated by the
     <a><code>multipart/form-data</code> encoding algorithm</a>.
 
+    <p>Set <var>source</var> to <var>object</var>.
+
    <dt>{{URLSearchParams}}
    <dd>
     <p>Set <var>action</var> to an action that runs the
@@ -4243,11 +4295,15 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
     <p>Set <var>Content-Type</var> to
     `<code>application/x-www-form-urlencoded;charset=UTF-8</code>`.
 
+    <p>Set <var>source</var> to <var>object</var>.
+
    <dt><code>USVString</code>
    <dd>
     <p>Set <var>action</var> to an action that runs <a>UTF-8 encode</a> on <var>object</var>.
 
     <p>Set <var>Content-Type</var> to `<code>text/plain;charset=UTF-8</code>`.
+
+    <p>Set <var>source</var> to <var>object</var>.
 
    <dt>{{ReadableStream}}
    <dd><p>Set <var>stream</var> to <var>object</var>.
@@ -4272,8 +4328,8 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
    <a abstract-op>close</a> <var>stream</var>.
   </ol>
 
- <li><p>Let <var>body</var> be a <a for=/>body</a> whose
- <a for=body>stream</a> is <var>stream</var>.
+ <li><p>Let <var>body</var> be a <a for=/>body</a> whose <a for=body>stream</a> is
+ <var>stream</var> and whose <a for=body>source</a> is <var>source</var>.
 
  <li><p>Return <var>body</var> and <var>Content-Type</var>.
 </ol>
@@ -4914,7 +4970,7 @@ run these steps:
 
 <h3 id=response-class>Response class</h3>
 
-<pre class=idl>[Constructor(optional ResponseBodyInit? body = null, optional ResponseInit init),
+<pre class=idl>[Constructor(optional BodyInit? body = null, optional ResponseInit init),
  Exposed=(Window,Worker)]
 interface Response {
   [NewObject] static Response error();
--- a/fetch.bs
+++ b/fetch.bs
@@ -4857,9 +4857,17 @@ constructor must run these steps:
    {{Headers}} object. Rethrow any exception.
   </ol>
 
- <li><p>If <var>inputBody</var> is non-null, <var>inputBody</var>'s <a for=body>source</a> is
- null, and <var>r</var>'s <a for=Request>request</a>'s <a for=request>mode</a> is neither
- "<code>same-origin</code>" nor "<code>cors</code>", then throw a <code>TypeError</code>.
+ <li>
+  <p>If <var>inputBody</var> is non-null and <var>inputBody</var>'s <a for=body>source</a> is
+  null, then run these substeps:
+
+  <ol>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -589,6 +589,9 @@ user-agent-defined <a for=header>value</a> for the
 
  <li><p>A <dfn export for=body id=concept-body-total-bytes>total bytes</dfn> (an
  integer), initially 0.
+
+ <li><p>A <dfn export for=body id=concept-body-replayable>replayable flag</dfn> (a boolean),
+ initially false.
```
annevk: I know this is annoying, and at some point we're going to burn all the flags, but for now we should keep consistency and only set and unset flags and don't say they're booleans.

fetch.bs
```diff
@@ -3025,6 +3028,11 @@ in addition to <a>HTTP fetch</a> above.
  <a lt="include credential">includes credentials</a>, then return a
  <a>network error</a>.
 
+ <li><p>If <var>request</var>'s <a for=request>method</a> is not <code>303</code>,
```
annevk: I think you mean to check response's status here, no?

fetch.bs
annevk: Also, I think we want to call this cloneable maybe? That is what the operation we want to fail is called.

fetch.bs
```diff
@@ -4240,9 +4252,7 @@ method, when invoked, must run these steps:
 <h3 id=body-mixin>Body mixin</h3>
 
 <pre class=idl>
-typedef (Blob or BufferSource or FormData or URLSearchParams or USVString) BodyInit;
-
-typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
+typedef (Blob or BufferSource or FormData or URLSearchParams or USVString or ReadableStream) BodyInit;</pre>
```
annevk: I think we should continue to put USVString last.

fetch.bs
yutakahirano: Thanks, fixed.

fetch.bs

fetch.bs
yutakahirano: Replaced boolean/true/false with (none)/set/unset.

We actually [clone a body](https://fetch.spec.whatwg.org/#concept-body-clone) regardless of the flag. So calling it cloneable  is confusing, I think.

annevk:
Okay, so you are saying that `request.clone()` will continue to work. However, I don't understand why https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch should continue to clone. The whole reason we have additional checks is that we can be more conservative there.

fetch.bs
```diff
@@ -3025,6 +3027,11 @@ in addition to <a>HTTP fetch</a> above.
  <a lt="include credential">includes credentials</a>, then return a
  <a>network error</a>.
 
+ <li><p>If <var>actualResponse</var>'s <a for=response>status</a> is not <code>303</code>,
+ <var>request</var>'s <a for=request>body</a> is not null, and <var>request</var>'s
```
annevk: non-null

fetch.bs
```diff
@@ -3366,6 +3373,10 @@ steps:
    <li class=XXX><p>Needs testing: multiple `<code>WWW-Authenticate</code>` headers, missing,
    parsing issues.
 
+   <li><p>If <var>request</var>'s <a for=request>body</a> is not null and <var>request</var>'s
```
annevk: non-null

yutakahirano:
> However, I don't understand why https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch should continue to clone. The whole reason we have additional checks is that we can be more conservative there.

As concept body has the body data as a ReadableStream, we need to clone it to replay uploading regardless of the source data type (String, ArrayBuffer, ReadableStream, ...). There are some options I could think of to handle this issue:

1. Stop holding the body data as a ReadableStream.
2. Define a new concept, say "replayable stream", to handle replaying
3. Keep teeing, but adding a flag to forbid user-visible teeing to allow implementers to omit teeing.

I chose the last option as it looked easiest.

fetch.bs

fetch.bs

annevk:
I think we should way more carefully explain that setup then, but I don't like it as much as just being explicit about it as we have done thus far. Maybe 1 is the better solution if that is what implementations actually do.

yutakahirano:
Hmm, https://github.com/whatwg/fetch/pull/425/commits/2ab2fcefbf89b26ffeea13ae2148bb7ab7a19292 introduces the original object (_source_) instead of _replayable flag_. Does that look better to you? With the change we don't need teeing stream for body in [http-network-or-cache-fetch](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch), but apparently cloning a request is needed for headers according to the comment. Should I create _clone-request-without-teeing-body_?

annevk:
> Should I create clone-request-without-teeing-body?

Yeah, I think that might be needed. There used to be some handwavy language along the lines of "clone except for body", but maybe that's gone now due to other changes.

fetch.bs
```diff
@@ -3367,6 +3383,23 @@ steps:
...
 
    <li>
+    <p>If <var>request</var>'s <a for=request>body</a> is non-null, then run these subsubsteps:
+
+    <ol>
+     <li><p>If <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is null,
+     then return a <a>network error</a>.
+
+     <li>
+      <p>
```

fetch.bs
```diff
@@ -4268,6 +4301,8 @@ typedef (BodyInit or ReadableStream) ResponseBodyInit;</pre>
     <p>If <var>object</var>'s {{Blob/type}}
     attribute is not the empty byte sequence, set <var>Content-Type</var> to its value.
 
+    <p>Set <var>source</var> to object.
```

yutakahirano:
Removed teeing from [http-network-or-cache-fetch](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch). PTAL again.

fetch.bs
```diff
@@ -3093,16 +3093,36 @@ steps:
 <i>authentication-fetch flag</i>.
 
 <ol>
+ <li><p>Let <var>httpRequest</var> be a new <a for=/>request</a>.
```
annevk: It seems you should just let it be null, no? Since every substep overrides what it is.

fetch.bs
```diff
@@ -3093,16 +3093,36 @@ steps:
...
+ <var>httpRequest</var> to <var>request</var>.
+
  <li>
-  <p>Let <var>httpRequest</var> be <var>request</var> if
-  <var>request</var>'s <a for=request>window</a> is "<code>no-window</code>"
-  and <var>request</var>'s <a for=request>redirect mode</a> is
-  "<code>error</code>", and the result of <a lt=clone for=request>cloning</a>
-  <var>request</var> otherwise.
+  <p>Otherwise, if <var>request</var>'s <a for=request>body</a> is null, then set
+  <var>httpRequest</var> to a copy of <var>request</var> except for its <a for=request>body</a>.
```
annevk: It seems you can combine this step with the next step. And only if request's body is non-null do the dance of moving the pointer across.

annevk:
Apologies for the delay, I've been traveling/vacationing.

Another thing I forgot to ask for is tests. We're trying to get to a place where all specification changes are accompanied by tests in w3c/web-platform-tests.

fetch.bs
yutakahirano: I thought the current structure was easier to read because it saved one nesting level. Don't you think so?

annevk: I'm not sure. The fact that the note needs to be duplicated makes me think it's not.

fetch.bs
```diff
@@ -3049,6 +3055,17 @@ in addition to <a>HTTP fetch</a> above.
  to `<code>GET</code>` and <var>request</var>'s <a for=request>body</a>
  to null.
 
+ <li>
+  <p>If <var>request</var>'s <a for=request>body</a> is non-null, then set <var>request</var>'s
+  <a for=request>body</a> to the first part of <a lt=extract for=BodyInit>extracting</a>
+  <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
```
annevk: I'm still a little confused by the new proposed setup. Now body can have a source, but body itself can also be a byte sequence? I'm not sure that makes sense.

fetch.bs
```diff
@@ -3076,17 +3093,36 @@ steps:
...
+  <ol>
+   <li><p>Set <var>httpRequest</var> to a copy of <var>request</var> except for its
+   <a for=request>body</a>.
+
+   <li><p>Set <var>httpRequest</var>'s <a for=request>body</a> to <var>request</var>'s
+   <a for=request>body</a>.
+
+   <li><p>Set <var>request</var>'s <a for=request>body</a> to a new <a for=/>body</a> whose
+   <a for=body>stream</a> is null and whose <a for=body>source</a> is <var>httpRequest</var>'s
+   <a for=request>body</a>'s <a for=body>source</a>.
```
annevk: I think it's worth pointing out here why the request isn't simply cloned.

fetch.bs
```diff
@@ -3049,6 +3055,17 @@ in addition to <a>HTTP fetch</a> above.
...
+ <li>
+  <p>If <var>request</var>'s <a for=request>body</a> is non-null, then set <var>request</var>'s
+  <a for=request>body</a> to the first part of <a lt=extract for=BodyInit>extracting</a>
+  <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
+
+  <p class="note no-backref"><var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>'s
+  nullity has already been checked.
+
+  <p class="note no-backref">The <a lt=extract for=BodyInit>extracting</a> operation cannot throw
+  as it was called for the same <a for=body>source</a> before.
```

fetch.bs

yutakahirano:
I'll write tests soon...

guest271314:
Does `body` at `body:/* ReadableStream instance */` call `.getReader().read()` on the value set as to `body`? If yes, before passing the accumulated value to `fetch()` call?

annevk:
@guest271314 I appreciate you're interested in learning how this all works, but a better place to ask questions is on IRC (#whatwg, Freenode). Pull requests are meant for reviewing changes to a standard.

fetch.bs
```diff
@@ -577,7 +577,7 @@ user-agent-defined <a for=header>value</a> for the
 <ul>
  <li>
   <p>A <dfn export for=body id=concept-body-stream>stream</dfn> (a
-  {{ReadableStream}} object).
+  {{ReadableStream}} object or null).
```

yutakahirano:
Can you take a look again?

annevk:
I didn't realize you were waiting for me, apologies. Let me clean up the PR a bit and evaluate where we stand.

annevk:
Aside from that note, the other thing I'd like to understand is what the plan is surrounding the contents of these streams. When should we fail upon non-Uint8Array object usage? Is that going to be fixed as part of #441 for both requests and responses?

annevk:
Looking at the test again it would be better if we fixed #441 straight away as well. This change to the specification doesn't make it clear that only Uint8Array objects will work for Request/Response ReadableStream objects.

annevk:
Also, didn't we decide that this would require "cors"? Otherwise it seems you can maybe do more careful timing attacks?

fetch.bs
```diff
@@ -3035,17 +3051,31 @@ steps:
...
+   <li><p>Set <var>httpRequest</var>'s <a for=request>body</a> to <var>body</var>.
+
+   <li><p>If <var>body</var> is non-null, then set <var>request</var>'s <a for=request>body</a> to a
+   new <a for=/>body</a> whose <a for=body>stream</a> is null and whose <a for=body>source</a> is
+   <var>body</var>'s <a for=body>source</a>.
+  </ol>
+
+  <p class="note no-backref">Here we do not <a for=request>clone</a> <var>request</var> in order
+  to reduce memory consumption. <var>request</var> can be reused with redirects, authentication,
+  and proxy authentication.
```
yutakahirano: I'd say "request's body's source is null" or "request's body is made from ReadableStream" insetad of "request's body's source is a ReadableStream object".

yutakahirano:
> Also, didn't we decide that this would require "cors"? Otherwise it seems you can maybe do more careful timing attacks?

I don't remember that. Can you point a discussion? no-cors is already timing-attack fragile, though I'm fine with that restriction.

annevk:
It seems we should also say something about https://github.com/yutakahirano/fetch-with-streams/issues/57 in this commit.

annevk:
We also need to set the use-CORS-preflight flag, right? Also, we should handle/mention the scenario where the response is not HTTP/1.1 and you're trying to use chunked.

fetch.bs

yutakahirano:
The last commit deals with https://github.com/yutakahirano/fetch-with-streams/issues/65.

yutakahirano:
> We also need to set the use-CORS-preflight flag, right?

Done.

> Also, we should handle/mention the scenario where the response is not HTTP/1.1 and you're trying to use chunked.

Do we need to introduce a new member in response? Or is something like "If response is transferred via HTTP/1.0 or older, then..." enough?

annevk:
I think a sentence like that in step 4 of https://fetch.spec.whatwg.org/#http-network-fetch is fine for now.

yutakahirano:
Done.

fetch.bs
```diff
@@ -3089,6 +3122,11 @@ steps:
   a body, have a bounded size and are not allowed to stay alive indefinitely.
 
 
+ <li><p>If <var>httpRequest</var>'s <a for=request>body</a> is non-null and <var>httpRequest</var>'s
+ <a for=request>body</a>'s <a for=request>source</a> is null, then <a for="header list">append</a>
+ `<code>Transfer-Encoding</code>`/`<code>chunked</code>` to <var>httpRequest</var>'s
+ <a for=request>header list</a>.
```
annevk: We should probably do this after the connection is established since we shouldn't do this over an HTTP/2 connection per https://tools.ietf.org/html/rfc7540#section-8.1.

yutakahirano: Fixed though I'm not sure "_connection_ is not an HTTP/2 connection" is the right wording.

annevk:
This looks good to me. I think you did a great job. I'm a little worried because we lack tests for all the various complicated scenarios, but I hope we'll address those during the implementation stage.

Could @tyoshino maybe do a quick review as well just to make sure we didn't miss anything?

annevk:
@tyoshino ping. Or otherwise a review from @domenic would be okay too.

oliverjanik:
What's the status, here?

yutakahirano:
Chrome hasn't shipped the feature yet.

danielkorczak:
Anything in the rumor mill for this one? 

Otherwise, at the very least, I'm going to have to ask for this on my Christmas list!

davidfowl:
Is this feature available in any browsers as yet?

annevk:
We should probably remove this from the standard as it's not implemented, nobody appears to have plans, and what is in the standard today is incomplete.

ioquatix:
But it’s super useful!

davidfowl:
@annevk Agreed but it makes me very sad as we're extremely close to being able to potentially do full duplex streaming with the fetch APIs which would basically make it a real replacement for things like websockets (which in referred to as a legacy API in another issue).

danielkorczak:
@davidfowl

So at this point, we’re waiting on HTTP/3 standards being finalized and then server implementations of HTTP/3?

yutakahirano:
We'll re-assess whether we (Chromium) would implement this in the near future.

davidfowl:
I’ve been following up with the edge folks about helping to push this as well (since they now work on chromium).

annevk:
@davidfowl reportedly you'll hit middle-ware snags trying to do full-duplex HTTP and therefore still need WebSocket. So as long as that's the case and as long as "Web Transport" isn't really a thing, WebSocket is here to stay.

And then while it would indeed be nice to stream request bodies, since you likely have to support resumption of some kind splitting the payload yourself and sending it as a series of requests over H/2 or H/3 might not be that much worse.

ioquatix:
WebSockets are a mess to implement on the server side, HTTP streaming is semantically much simpler. So prefer this proposed approach.

What problems did you run into?

davidfowl:
> @davidfowl reportedly you'll hit middle-ware snags trying to do full-duplex HTTP and therefore still need WebSocket. So as long as that's the case and as long as "Web Transport" isn't really a thing, WebSocket is here to stay.

@annevk Not necessarily, using https limits what middleware can do and things like GRPC already take advantage of HTTP/2 to support streaming. That said, I don't really have an issue with websockets staying, but until the websocket over h2/h3 specs are implemented I'd like to experiment with cheaper ways to do duplex streaming (over HTTP/2, QUIC/HTTP/3 streams instead of a TCP connection). This is a path to doing so. 

> And then while it would indeed be nice to stream request bodies, since you likely have to support resumption of some kind splitting the payload yourself and sending it as a series of requests over H/2 or H/3 might not be that much worse.

The use case I have is for [SignalR](https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr?view=aspnetcore-2.2&tabs=visual-studio) a real time framework that is transport agnostic (much like socket.io). It already runs over websockets, Server-Sent-Events (EventSource APIs in the browser) and long polling. I'd like to explore an H/2/H/3 bidirectional streaming transport which would pretty much work just like the other transports I've called out but be cheaper in terms of client and server resources.

oliverjanik:
Websocket or H2/H3 ideas aside, upload progress is the last major thing missing from fetch that xhr has. 

There's no way to track upload progress using fetch at the moment, which is a major blocker for adoption from what i'm seeing.

ioquatix:
WebSocket over H2 already exists. There is an RFC and client/server is implemented in https://github.com/socketry/async-websocket

annevk:
@davidfowl I thought HTTPS would do so as well, but reportedly corporate deployment of middleware is significant enough to pose a problem. This might well change over time though and I don't necessarily think it should block us from offering the ability to attempt full-duplex HTTP.

@oliverjanik see the `FetchObserver` idea. All it needs is time and flushing out.

davidfowl:
@ioquatix do browsers support it?

ioquatix:
Apparently it does but I could not get it to work in my testing.

bradisbell:
In case abandoning this feature is being considered, I'd like to pile on a use case:

The existence of this API has implications for streaming media sourced from browsers and browser-like clients.  Without a way to make a simple HTTP PUT request with a stream from a browser, all of us working in this space have been hacking around the problem by proxying the stream data via Web Socket servers or repeatedly making requests with chunks every second or so.

For 20 years, we have been streaming media via HTTP PUT on the source end, and HTTP GET on the receiving end.  This works most everywhere, except for the source end on browsers, which is important.  Implementing ReadableStream as a request body would remove the need for a lot of hackarounds.

I suspect there are many others with different use cases who are generating data on-the-fly where a normal HTTP request would be more appropriate for sending data to the server.  In at least some of these use cases, we don't always have control over the servers, which already support HTTP and may not be able to support other protocols.  Please, when considering this stream-as-request-body feature, keep in mind that it goes far beyond file uploads.  And, thank you for your efforts!

yutakahirano:
I started a thread on a chromium mailing list: https://groups.google.com/a/chromium.org/forum/#!topic/blink-network-dev/bsVgOxNCzFc

You feedback will be appreciated there too.

@annevk, can we talk about this topic with other vendors at TPAC?

yutakahirano:
> @annevk, can we talk about this topic with other vendors at TPAC?

@annevk: ^^^

annevk:
Yeah, that seems fine to me.

yutakahirano:
I've proposed the topic for SW WG (https://github.com/w3c/ServiceWorker/issues/1460).

yutakahirano:
Mozilla: Interested
Apple: Interested
Google: Interested

TODO(@yutakahirano ): Explicitly state that bi-directional streaming is not supported.

ioquatix:
> [yutaka] we aren't trying to support bi-directional streams; we won't read response headers until we finish sending the body. This also means we won't support redirect

... why not? Bidirectional streaming is really important for so many use cases.

davidfowl:
I’m glad to hear there’s an interest here, 2 streaming unidirectional requests are a big stepping stone 😁. The ability to do bidirectional steaming from a single request would avoid the need to use sticky sessions on your load balancer to make the up and down request hit the same server.

Something to think about in the future.

yutakahirano:
I understand that bi-directional streaming is important for some users, but at least for Chromium it's very difficult to implement.

bradisbell:
@yutakahirano 

> we won't read response headers until we finish sending the body

How will the client handle cases of authentication problems?  That is, if the server sends a `401 Unauthorized` or some other error instead of a `100 Continue`, I would expect that this response will be accessible in JavaScript in the same way that any other response is.  Is that still the case?

> This also means we won't support redirect

Could you elaborate on this?  Suppose I'm `PUT`-ing a stream and server the server immediately sends a redirect upon receiving request headers.  (That is, no `100 Continue`, but a `307 Temporary Redirect`.)  I assume the stream in the request body would be closed, and the response from the Fetch would be available to JavaScript.  Is this the case?

annevk:
No, it would network error. Exposing redirects would have to be opt-in and is tracked by https://github.com/whatwg/fetch/issues/601.

yutakahirano:
We're considering requiring http2 for this feature. If you have any opinions please leave them at https://github.com/whatwg/fetch/issues/966.

benbucksch:
http://w3c-test.org/fetch/api/basic/request-upload.any.html
Firefox 76 FAIL:
```Fetch with POST with ReadableStream ... expected "Test" but got "[object ReadableStream]"```

annevk:
Yeah, this isn't implemented yet and there are several outstanding issues.

drzraf:
I concur. It's a great opportunity to make the web a bit more bidirectional.
Another use-case: A form is posted (to an openstack POST-handler), attachments are encrypted on-the-fly. In the current situation, the memory is ~ twice the file which is unsuitable for multi-GB uploads.

fetch.bs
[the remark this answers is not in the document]
BingyuLiu: Fetch post-call Always return 415 ？？？？？？？？？？WHY

BingyuLiu:
Fetch post-call Always return 415 ？？？？？？？？？？WHY

BingyuLiu:
var    requestOptions = {
          mode: 'no-cors',
          method: 'POST',
          headers: { 'Content-Type': 'application/json' },
          body: JSON.stringify(jsonObj)
        };

fetch(requestUrl, requestOptions).then(....

BingyuLiu:
Anyone can please help here, thanks！

yutakahirano:
Sorry for the long silence. 

We, Chrome, did try to prove that this feature is useful, with gRPC/web. But gRPC/web lost its interest to this feature, given we now have WebTransport. As a result, I'm not confident that I can persuade Chromium API owners to ship this feature, given the spec/impl complexity needed for this feature.

FWIW gRPC/web was the only web developer who wanted to use this feature with HTTP/1.1, so I believe we all are now happy to drop the support of this feature on HTTP/1.1.

We have two paths forward:

 1. Get more support from other browser vendors and web developers, and ship this feature.
 2. Give up this feature and remove the logic from the spec.

Thoughts?

bradisbell:
> FWIW gRPC/web was the only web developer who wanted to use this feature with HTTP/1.1, so I believe we all are now happy to drop the support of this feature on HTTP/1.1.

I had no idea this was available on HTTP/1.1... we've been needing this for streaming media for a very long time, and yet previous discussions stated that it was locked to HTTP/2.  Existing servers, such as Icecast and FFmpeg, do not support HTTP/2 and have no need to.  They also do not necessarily need to support HTTPS.

I would strongly urge reconsidering.  Please keep HTTP/1.1 streaming request bodies available, and *tell us how to use them in Chromium**.  It does not seem to be available with current builds.  If there is some sort of flag we need, what is it?  And then, we should go back to all of the threads where this was discussed and inform folks that it is available:

 - https://groups.google.com/a/chromium.org/g/blink-network-dev/c/bsVgOxNCzFc
 - https://stackoverflow.com/a/41222366/362536
 - https://github.com/whatwg/fetch/issues/966
 - https://stackoverflow.com/q/35899536/362536
 - https://github.com/whatwg/fetch/issues/439
 - https://github.com/whatwg/fetch/issues/88

drzraf:
> I had no idea this was available on HTTP/1.1...  [...] I would strongly urge reconsidering. Please keep HTTP/1.1 streaming request bodies available, and _tell us how to use them in Chromium_. It does not seem to be available with current builds.

The specs is here, but implementations are missing. Chrome devs is just saying _we will implement it, but since it's too complex for our code base, could you please alter the specs to make it HTTP2+ only and then only we'll ship it_. This feature has never been part of a Chrome release.

bradisbell:
@drzraf Ok, thank you for that clarification.

ioquatix:
If it's helpful, https://github.com/socketry/async-http supports full bi-directional HTTP/1.1 and HTTP/2 client and server. So you can test with real world code.

davidfowl:
I'm ready to add support for SignalR in .NET, I'd love to chrome and edge to support this. It would be enough to justify doing adding this transport.

yutakahirano:
>> I had no idea this was available on HTTP/1.1... [...] I would strongly urge reconsidering. Please keep HTTP/1.1 streaming request bodies available, and tell us how to use them in Chromium. It does not seem to be available with current builds.
>
> This feature has never been part of a Chrome release. The specs is here, but implementations are missing. Chrome devs told they would implement it, but since it's too complex for Chrome's code base, could the specs please be altered to make it HTTP2+ and then only it may be shipped inside Chrome.

FTR Chrome [ran an origin trial for this feature](https://groups.google.com/a/chromium.org/g/blink-dev/c/mNVaRrvBZG0/m/djXjGWXtAAAJ). Web developers were able to use the feature at that time.

annevk:
@yutakahirano can you please open a new issue to track this? I suppose part of it is covered by #966 but deciding on the fate of this feature seems worthy of its own issue.

yutakahirano:
Sure!

yutakahirano:
Here it is: #1438

PR #426, opened 2016-11-28
Meta: run HTML checker on generated HTML

annevk:
Also fix the HTML and HTTP header definitions and HTTP header links in
the process.


annevk:
Many thanks to @sideshowbarker for leading the way in https://github.com/whatwg/html/pull/2093 by the way.

PR #432, opened 2016-12-08
Link directly to the concept-settings-object-referrer-policy anchor.

Ms2ger:


fetch.bs
```diff
@@ -2370,10 +2371,8 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
  <a for=request>client</a> is non-null, then set <var>request</var>'s
  <a for=request>referrer policy</a> to <var>request</var>'s
  <a for=request>client</a>'s
- <a lt="environment settings object">referrer policy</a>.
```
annevk: I think the only change required here is to change `lt` to `for` (and remove the comment).

Ms2ger:
Thanks!

Issue #362, opened 2016-08-10
Dispatch to foreign fetch for redirects


annevk:
Currently we set the skip-service-worker flag so a new location of a redirect does not get intercepted by a same-origin service worker. However, a cross-origin service worker should likely intercept that, otherwise we make service migration harder.

See https://github.com/slightlyoff/ServiceWorker/issues/793 for context.

@mkruisselbrink will fix this (and will soon be able to assign himself).

annevk:
This should be quite a bit easier now thanks to #363.

sirdarckcat:
Just to clarify. Would this mean that a 30X redirect on any subresource HTTP request would trigger the service worker of the target URL and dispatch a foreign fetch event?

annevk:
No, only requests that are cross-origin, but same-origin with a foreign-fetch service worker. Same-origin we cannot expose without some kind of server-side opt-in.

PR #435, opened 2016-12-13
Changing skip-service-worker flag to use-service-workers enum

### LINKED ISSUE #303: Only specific APIs should skip the fetch event when called within a service worker
Currently it is not clear whether resource fetches for notifications (e.g. icons) shown from a service worker should trigger the service worker onfetch handler. As a use case it makes sense to enable offlining these resources.

My understanding is that fetches initiated directly from a service worker are not passed through the service worker onfetch handler, which seems reasonable. However, what about the fetches triggered indirectly by actions initiated in the service worker, like showing a notification?

CC @annevk @jakearchibald @beverloo @slightlyoff

jakearchibald:
This fixes #303. I don't think there's anything else to do with this PR, but I need to make changes to SW to set this flag on particular APIs.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -705,7 +705,7 @@ explicitly set <a for=/>request</a>'s
 this flag set are subject to additional processing requirements.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export>use-service-workers value</dfn>, that is "<code>all</code>",
+<dfn for=request export>service-workers mode</dfn>, that is "<code>all</code>",
 "<code>foreign</code>", or "<code>none</code>". Unless stated otherwise it is
 "<code>all</code>".
 
@@ -2801,13 +2801,13 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
  <li><p>Let <var>actualResponse</var> be null.
 
  <li>
-  <p>If <var>request</var>'s <a>use-service-workers value</a> is not "<code>none</code>", then run these
+  <p>If <var>request</var>'s <a>service-workers mode</a> is not "<code>none</code>", then run these
   substeps:
 
   <ol>
    <li>
     <p>If <var>request</var>'s <a for=request>client</a> is null or
-    <var>request</var>'s <a>use-service-workers value</a> is "<code>all</code>", then set
+    <var>request</var>'s <a>service-workers mode</a> is "<code>all</code>", then set
     <var>response</var> to the result of invoking <a for=/>handle fetch</a> for <var>request</var>.
     [[!HTML]] [[!SW]]
 
@@ -2899,7 +2899,7 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
 
    <li>
     <p>If <var>request</var>'s <a for=request>redirect mode</a> is
-    "<code>follow</code>", then set <var>request</var>'s <a>use-service-workers value</a> to
+    "<code>follow</code>", then set <var>request</var>'s <a>service-workers mode</a> to
     "<code>none</code>".
 
     <p class="note no-backref">Redirects coming from the network (as opposed to from a service
@@ -5194,7 +5194,7 @@ method, must run these steps:
 
  <li>If <var>request</var>'s <a for=request>client</a>'s
  <a for="environment settings object">global object</a> is a {{ServiceWorkerGlobalScope}} object,
- set <var>request</var>'s <a>use-service-workers value</a> to "<code>foreign</code>".
+ set <var>request</var>'s <a>service-workers mode</a> to "<code>foreign</code>".
 
  <li><p>Let <var>responseObject</var> be a new {{Response}} object and a new associated
  {{Headers}} object whose <a for=Headers>guard</a> is
@@ -5347,7 +5347,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
  <li><p>Let <var>request</var> be a new <a for=/>request</a>, whose
  <a for=request>url</a> is <var>requestURL</var>,
  <a for=request>client</a> is <var>client</var>,
- <a>use-service-workers value</a> is "<code>none</code>",
+ <a>service-workers mode</a> is "<code>none</code>",
  <a>synchronous flag</a> is set,
  <a for=request>mode</a> is "<code>websocket</code>",
  <a for=request>credentials mode</a> is
--- a/fetch.bs
+++ b/fetch.bs
@@ -2917,7 +2917,7 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
    <li>
     <p>If <var>request</var>'s <a for=request>redirect mode</a> is
     "<code>follow</code>", then set <var>request</var>'s <a>service-workers mode</a> to
-    "<code>none</code>".
+    "<code>foreign</code>".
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
Technically, this and skip-service-worker could become an enum of "none", "local" and "all", but it feels like that'd break specs already using it. Is that the right thing to do?

jakearchibald:
Related SW PR https://github.com/w3c/ServiceWorker/pull/1025

annevk:
I think I prefer the enum approach. There's not too many consumers as far as I know.

jakearchibald:
Cool, I'll update the PR

annevk:
Thanks. There's also #362 which relates to this flag somewhat.

fetch.bs
```diff
@@ -728,11 +728,9 @@ explicitly set <a for=/>request</a>'s
 this flag set are subject to additional processing requirements.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export>skip-service-worker flag</dfn>. Unless stated otherwise it is
-unset.
-
-<p>A <a for=/>request</a> has an associated <dfn export>skip-local-service-worker flag</dfn>. Unless
-stated otherwise it is unset.
+<dfn export>use-service-workers value</dfn>, that is "<code>all</code>",
```
annevk: This needs `for=request` too. Seems we forgot that until now.

annevk:
So where do we check for the `foreign` value? If we only ever set it, it seems like we could also reuse another value, no?

annevk:
Do we have WPT tests around this?

jakearchibald:
> So where do we check for the foreign value? If we only ever set it, it seems like we could also reuse another value, no?

Yeah, when it comes to performing foreign fetch we're only really concerned that the value *isn't* "none". Although I'm not sure what else could be reused.

> Do we have WPT tests around this?

No, I guess this is something I should learn and do 😄

annevk:
> Yeah, when it comes to performing foreign fetch we're only really concerned that the value isn't "none".

And it cannot be "all" either I guess? I guess "foreign" is fine then. Can @mkruisselbrink have a look or is he still away?

annevk:
As for the tests, the new rule is that they land together with the specification change. Let me know if you need help.

annevk:
And only https://w3c.github.io/ServiceWorker/ seems to use the skip-service-worker flag currently. We should maybe prepare a PR there too so they can all land at the same time.

jakearchibald:
https://github.com/w3c/ServiceWorker/pull/1025

fetch.bs
```diff
@@ -728,8 +728,9 @@ explicitly set <a for=/>request</a>'s
 this flag set are subject to additional processing requirements.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export>skip-service-worker flag</dfn>. Unless stated otherwise it is
-unset.
+<dfn export>use-service-workers value</dfn>, that is "<code>all</code>",
```
annevk: What do you think about `<dfn export for=request>service-workers mode</dfn>`? I think that would be more consistent with what we have thus far.

jakearchibald: SGTM, I wasn't happy with the naming

fetch.bs
```diff
@@ -2793,16 +2794,14 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
  <li><p>Let <var>actualResponse</var> be null.
 
  <li>
-  <p>If <var>request</var>'s <a>skip-service-worker flag</a> is unset, then run these
+  <p>If <var>request</var>'s <a>service-workers mode</a> is not "<code>none</code>", then run these
   substeps:
 
   <ol>
    <li>
     <p>If <var>request</var>'s <a for=request>client</a> is null or
```
mkruisselbrink: Do we still need the client is null check, or is just relying on the service-workers mode enough now? And in particular it seems wrong that service-workers mode==foreign and client==null will now result in handle fetch being called.

fetch.bs
```diff
@@ -728,8 +728,9 @@ explicitly set <a for=/>request</a>'s
 this flag set are subject to additional processing requirements.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export>skip-service-worker flag</dfn>. Unless stated otherwise it is
-unset.
+<dfn for=request export>service-workers mode</dfn>, that is "<code>all</code>",
```
mkruisselbrink: Would it make sense to add some non-normative text somewhere to explain what this mode actually means? Of course you can figure that out by reading the algorithms, but I find the non-normative notes that many of the requests attributes have to be quite helpful in quickly figuring out what a particular thing is for.

mkruisselbrink:
> > Do we have WPT tests around this?
>
> No, I guess this is something I should learn and do

Hmm, I should probably also work on actually upstreaming the existing foreign fetch tests we have in blink.  Although not sure what WPT tests would be desired for this specific change, as this change by itself doesn't seem to have any behavioral changes?

fetch.bs
annevk: Yeah, sounds good.

fetch.bs
annevk: I think that relying on service-workers mode is enough. At least, I thought the idea was that having a client no longer mattered.

annevk:
>  Although not sure what WPT tests would be desired for this specific change, as this change by itself doesn't seem to have any behavioral changes?

The fact that certain fetches from the service worker can now invoke the service worker is a normative change that will result in behavioral changes.

jakearchibald:
I agree this is a change that needs tests. Will pick this up after the Christmas break.

jakearchibald:
Tests: https://github.com/jakearchibald/web-platform-tests/commit/0fa7353b8412182feac56c350542b3318ebd53ab

@annevk would be grateful for a review. Be kind!

annevk:
@jakearchibald probably best to create a PR on web-platform-tests for review. Overall I think it's fine, but since you need permissions you'll have to use the -manual suffix I think.

jakearchibald:
https://github.com/w3c/web-platform-tests/pull/4518

fetch.bs
jakearchibald: Done

jakearchibald:
@annevk tests merged. Happy to merge this?

fetch.bs
```diff
@@ -705,8 +705,27 @@ explicitly set <a for=/>request</a>'s
...
+"<code>foreign</code>", or "<code>none</code>". Unless stated otherwise it is
+"<code>all</code>".
+
+<div class=note>
+ This determines which service workers will receive a {{fetch!!event}} event for this fetch.
```
annevk: This should start with a `<p>`.

annevk:
One more question, in a commit you stated #362 is fixed by this, but that seems unlikely as service workers are skipped for redirects with this change. So are we leaving that for that issue?

Could you write a commit message for this?

jakearchibald:
@annevk uhhhh good catch. Sorry about that. As far as I can tell it just needed that "none" changed to "foreign".

jakearchibald:
Commit message:

```
Add "service-workers mode" flag to allow particular fetches within a service worker to trigger "fetch" events.

Fixes: #303, #362.
Related service worker PR: https://github.com/w3c/ServiceWorker/pull/1025.
Tests: https://github.com/w3c/web-platform-tests/pull/4518.
```

annevk:
Thanks! (Adjusted the message a bit to fit style guidelines.)

PR #440, opened 2016-12-20
Remove an obsolete note for ReadableStream type

yutakahirano:
Once we planned to have both ReadableStream and ReadableByteStream and wrote a note in the spec for a possible naming change in the future. But the plan was abandoned and we only have ReadableStream. Let's remove the obsolete note.


yutakahirano:
This is a TODO in https://github.com/yutakahirano/fetch-with-streams/issues/46 and I've forgotten to do that (Sorry!)

PR #442, opened 2016-12-22
Add "fetch" as a destination for "connect-src" bound features

yoavweiss:
Following the discussion on https://github.com/w3c/preload/issues/80 I believe there's agreement to add an alias to the empty destination in Fetch. 

This is a stab in that direction (I'm not sure that's the right way to do that though, so guidance would be appreciated).

/cc @annevk @igrigorik

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2226,6 +2226,9 @@ the request.
  `<code>Accept-Language</code>`/an appropriate <a for=header>value</a>
  to <var>request</var>'s <a for=request>header list</a>.
 
+ <li><p>If <var>request</var>'s <a for=request>destination</a> is <code>fetch</code>, then set it
+ to the empty string.
+
  <li>
   <p>If <var>request</var>'s <a for=request>priority</a> is
   null, use <var>request</var>'s
@@ -2237,9 +2240,6 @@ the request.
   <p class=note>The user-agent-defined object could encompass stream weight and dependency
   for HTTP/2, and equivalent information used to prioritize dispatch and processing of
   HTTP/1 fetches.
-
- <li><p>If <var>request</var>'s <a for=request>destination</a> is <code>fetch</code>, then set it
- to the empty string.
 
  <li>
   <p>If <var>request</var> is a <a>navigation request</a>, a user agent should, for each
--- a/fetch.bs
+++ b/fetch.bs
@@ -750,6 +750,7 @@ the empty string,
 the empty string,
 "<code>document</code>",
 "<code>embed</code>",
+"<code>fetch</code>",
 "<code>font</code>",
 "<code>image</code>",
 "<code>manifest</code>",
@@ -1084,10 +1085,10 @@ Unless stated otherwise, it is "<code>basic</code>".
 <hr>
 
 <p>A <dfn export>subresource request</dfn> is a <a for=/>request</a>
-whose <a for=request>destination</a> is "<code>font</code>",
-"<code>image</code>", "<code>manifest</code>", "<code>media</code>",
-"<code>script</code>", "<code>style</code>", "<code>xslt</code>", or the empty
-string.
+whose <a for=request>destination</a> is "<code>fetch</code>",
+"<code>font</code>", "<code>image</code>", "<code>manifest</code>",
+"<code>media</code>", "<code>script</code>", "<code>style</code>",
+"<code>xslt</code>", or the empty string.
 
 <p>A <dfn export>potential-navigation-or-subresource request</dfn> is a
 <a for=/>request</a> whose
@@ -4490,7 +4491,7 @@ dictionary RequestInit {
 };
 
 enum RequestType { "", "audio", "font", "image", "script", "style", "track", "video" };
-enum RequestDestination { "", "document", "embed", "font", "image", "manifest", "media", "object", "report", "script", "serviceworker", "sharedworker", "style",  "worker", "xslt" };
+enum RequestDestination { "", "document", "embed", "fetch", "font", "image", "manifest", "media", "object", "report", "script", "serviceworker", "sharedworker", "style",  "worker", "xslt" };
 enum RequestMode { "navigate", "same-origin", "no-cors", "cors" };
 enum RequestCredentials { "omit", "same-origin", "include" };
 enum RequestCache { "default", "no-store", "reload", "no-cache", "force-cache", "only-if-cached" };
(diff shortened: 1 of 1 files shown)


igrigorik:
That'll do the trick for preload, I'm not sure if/what implications this has for other parts of Fetch however.

annevk:
You've only added it to a non-normative table here? The description of the field, enum, and processing do not appear to be altered. Those would have to change.

yoavweiss:
You're right. Added "fetch" to the various destination definitions. Please let me know if I missed something.

Also, going over CSP, I noticed that it refers to a "subresource" destination where I'd expect an empty string destination, which makes me think that used to be the value at some point. Is it worth while to go back to that value rather than "fetch"?

annevk:
If I remember correctly "subresource" was much broader so probably not.

annevk:
I still don't see any normalization of "fetch" to "" happening in your commits.

fetch.bs
```diff
@@ -791,7 +792,7 @@ the empty string,
    <td><code>child-src</code>
    <td>HTML's <code>&lt;iframe></code> and <code>&lt;frame></code>
   <tr>
-   <td>""
+   <td>"" or "<code>fetch</code>"
```
annevk: I think `("<code>fetch</code>")` would be clearer since you can only use it as input, right?

yoavweiss: I guess so. I'm not familiar with `()` signifying input-only

yoavweiss:
> I still don't see any normalization of "fetch" to "" happening in your commits.

Where would such normalization take place? As part of the Fetch algorithms in section 4? Or is there a central place where this can be defined?

Also, I now noticed that an empty string destination can mean both connect-src bound features as well as download. I think it might be better to define download as an explicit destination (but it's probably best to do that in a separate PR)

fetch.bs
annevk: This whole table is descriptive. It doesn't have any impact on implementers. But saying "or 'fetch'" means you can see that in the wild for such requests, while they would always be the empty string.

annevk:
> Where would such normalization take place? As part of the Fetch algorithms in section 4?

I suppose, this is a new thing so you get to design it to some extent, but I don't see where else we could have it.

> I think it might be better to define download as an explicit destination (but it's probably best to do that in a separate PR)

Yes, please do that separately.

We should also have a test for a change landing in WPT. Something that shows that "fetch" as input becomes "" later on. I guess for now that can only be done through `<link rel=preload>` coupled with a service worker?

yoavweiss:
> We should also have a test for a change landing in WPT. Something that shows that "fetch" as input becomes "" later on. I guess for now that can only be done through <link rel=preload> coupled with a service worker?

Yeah, I can add such a test to preload's WPT tests

yoavweiss:
Added such a conversion step to "to perform a fetch". Let me know if it makes sense.

fetch.bs
```diff
@@ -2237,6 +2238,9 @@ the request.
   for HTTP/2, and equivalent information used to prioritize dispatch and processing of
   HTTP/1 fetches.
 
+ <li>
+  <p>If <var>request</var>'s <a for=request>destination</a> is
+  <code>fetch</code>, set it to be the empty string.
```
annevk: then set it to the empty string*

(plus then, minus be)

Also, this can be on the same line as the `<li>` (apart from 100 columns wrapping) since it's the only child.

annevk: Also, maybe move this step to be before priority, since priority is a later-defined request field.

Looks good with those nits I think. Let me know when there's a WPT PR.

yoavweiss: addressed

annevk:
So actually, we probably should not modify the enum as it's only used for reflection and destination can never be "fetch" at that point.

fetch.bs
```diff
@@ -1084,10 +1085,10 @@ Unless stated otherwise, it is "<code>basic</code>".
...
 
 <p>A <dfn export>subresource request</dfn> is a <a for=/>request</a>
-whose <a for=request>destination</a> is "<code>font</code>",
-"<code>image</code>", "<code>manifest</code>", "<code>media</code>",
-"<code>script</code>", "<code>style</code>", "<code>xslt</code>", or the empty
-string.
+whose <a for=request>destination</a> is "<code>fetch</code>",
+"<code>font</code>", "<code>image</code>", "<code>manifest</code>",
+"<code>media</code>", "<code>script</code>", "<code>style</code>",
+"<code>xslt</code>", or the empty string.
```
annevk: This is wrong. It should be normalized at this point, no?

I guess we should fix it differently although I'm not entirely sure how. Ideally whenever a request is created the destination is already correct.

Perhaps we should define a normalize operation for destinations. It takes a potentialDestination. If it is "fetch", we return the empty string. If it is any other value, it returns input. And then standards that want to accept "fetch" as value use that operation.

yoavweiss: If I properly understand your suggestion here, you'd prefer to keep the destination as is (i.e. not add a "fetch" destination), create the concept of "potential destination", have preload use that, and just translate is to destination (by having preload call "normalize potential destination" or somesuch) before preload triggers a fetch. Is that correct?

annevk: Yeah, basically. At some point we might want to refactor that by defining an actual "constructor" for the concept of requests, but this seems like the simplest solution for now.

igrigorik: If we're special-casing preload behavior here, what's the functional difference from second point we proposed in https://github.com/w3c/preload/issues/80#issuecomment-268573775?  We can just add a clause in Preload that maps "fetch" -> "".

annevk: I think the main point is to have it here in case someone else needs to use it too. Just like Fetch defines header value normalization and not XMLHttpRequest.

igrigorik: Hmm, ok. The plumbing here seems a little weird, but I'll defer to you and Yoav. 

Yoav: any blockers from your side for this?

yoavweiss: No blockers, just need to get back to this.

annevk: @igrigorik how would you address it?

igrigorik: @annevk no brilliant insights here. I'll defer to your and Yoav's judgement on this one.. :)

yoavweiss:
Seems like the way forward for this is https://github.com/whatwg/fetch/pull/547. Closing

Issue #441, opened 2016-12-20
Define "to read a request" more specifically


yutakahirano:
https://fetch.spec.whatwg.org/#read-a-request

See https://github.com/w3c/web-platform-tests/pull/4362#issuecomment-268006360 for the background.

yutakahirano:
@annevk I would like to make a separate change from #425. Right now I'm not sure the capability of the network layer in the fetch spec. What kind of object does it consume / generate?

annevk:
It's rather opaque. Bytes go in and bytes come out. But with `Request` and `Response` objects JavaScript developers now have some control over what happens there in a way. So we need to filter when reading from them (or when creating them, but it seems like that was not the way streams ended up being designed).

annevk:
The high-level goal is that we only allow Uint8Array in streams associated with either `Request` or `Response` objects, right? That seems important to test for and implement, even if we haven't quite figured out how to require that in the specification.

yutakahirano:
My understanding is you can create a `Request` object from arbitrary ReadableStream, but you need to pass a ReadableStream each of whose chunks is a `Uint8Array` object if you want to make a `fetch()` call.

yutakahirano:
For receiving a response body, it is specified in https://fetch.spec.whatwg.org/#http-network-fetch.

> Whenever one or more bytes are transmitted from response’s message body, ...

For transferring a response from a service worker, it's specified in https://w3c.github.io/ServiceWorker/#dom-fetchevent-respondwith.

> If response’s body is non-null, run these substeps:

annevk:
> My understanding is you can create a Request object from arbitrary ReadableStream

So the only case for this is you also write the reader to consume the request yourself, right? Because any browser-supplied reader for a request, be it `fetch()` or `req.json()` or some such, will error.

Issue #330, opened 2016-07-07
How response bodies get from a service worker to the main page is not very clear


domenic:
Right now the spec says that the internal response has a ReadableStream object. Since the internal response is shared between the service worker and the page, the ReadableStream is implicitly shared.

This is not very good. For example it implies sharing arbitrary JavaScript objects, since you can enqueue arbitrary JavaScript objects in the readable stream. Right now Chrome ignores all objects except Uint8Arrays. It should instead error in some way, but exactly how is not clear, since there is no spec.

At first I thought this would be related to structured cloning of readable streams (https://github.com/whatwg/streams/issues/276 and https://github.com/whatwg/streams/issues/244) but that might not be true. Our plan for structured cloning was to lock the stream. But that will not work in this case, as we want the service worker code to be able to enqueue at the same time the main thread code is reading.

I guess what we need to do is specify how each `Response` object has a `ReadableStream` object. Then respondWith() roughly does the following:

- Create a new page-side `ReadableStream`
- Read chunks from the service worker's `Response`'s `ReadableStream`. For each chunk:
  - If it is not a Uint8Array, error the page-side `ReadableStream`
  - Otherwise, transfer it across realms/event loops, and enqueue it in the page-side `ReadableStream`.

Apologies if we already have some of this and I missed it.

yutakahirano:
- "transfer it" means copying the Uint8Array, right?
 - Does a BYOB reader offer a better way to transfer the buffer?

annevk:
Transfer means copying and then detaching on the side you copied from.

yutakahirano:
That means a script author will lose the access to the Uint8Array buffer passed to `controller.enqueue`. Is it desirable?

annevk:
I'm not sure. Given how requests and responses work with streams today it kind of makes sense to me that if you want to retain a value you clone it first. But we do need to make it deterministic when the transfer happens.

yutakahirano:
Another point:

We (Chrome) serialize the buffers to a C++ byte array and create Uint8Array objects from the byte array in the page. That means buffer's boundary is not preserved (i.e., buffers may be merged / split). As an implementer I would like the spec to allow us doing it.

yutakahirano:
@wanderview fyi

wanderview:
[...]

PR #449, opened 2017-01-05
Specify "upload body of a request"

yutakahirano:
...and replace "read a request" with it.

Fixing #441.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1109,7 +1109,7 @@ or "<code>worker</code>".
  <li><p>Return <var>newRequest</var>.
 </ol>
 
-<p>To <dfn export for=request id=concept-request-upload-body>upload body</dfn> of a
+<p>To <dfn export for=request id=concept-request-transmit-body>transmit body</dfn> for a
 <a for=/>request</a> <var>request</var>, run these steps:
 
 <ol>
@@ -1131,7 +1131,7 @@ or "<code>worker</code>".
     <ol>
      <li><p>Let <var>bytes</var> be the bytes represented by the <code>Uint8Array</code> object.
 
-     <li><p>Upload <var>bytes</var>.
+     <li><p>Transmit <var>bytes</var>.
 
      <li><p>Increase <var>body</var>'s <a for=body>transmitted bytes</a> with <var>bytes</var>'
      length.
@@ -2819,7 +2819,7 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
     <p>If <var>response</var> is not null, then run these substeps:
 
     <ol>
-     <li><p><a for=request>Upload body</a> of <var>request</var>.
+     <li><p><a for=request>Transmit body</a> for <var>request</var>.
 
      <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
      <a>filtered response</a>, and to <var>response</var>'s
@@ -3485,7 +3485,7 @@ steps:
   "<code>deprecated</code>" state value ought to be a temporary and last resort kind
   of option.
 
-  <p><a for=request>Upload body</a> of <var>request</var>.
+  <p><a for=request>Transmit body</a> for <var>request</var>.
 
  <li>
   <p>Let <var>strategy</var> be an object. The user agent may choose any object.
--- a/fetch.bs
+++ b/fetch.bs
@@ -1094,6 +1094,8 @@ or "<code>worker</code>".
 <p class=note>See <a for=/>handle fetch</a> for usage of these terms.
 [[!SW]]
 
+<hr>
+
 <p>To <dfn export for=request id=concept-request-clone>clone</dfn> a
 <a for=/>request</a> <var>request</var>, run these steps:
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1121,6 +1111,46 @@ or "<code>worker</code>".
...
+     <li><p>Run the above step again.
+    </ol>
+
+   <li><p>When <var>read</var> is fulfilled with an object whose <code>done</code>
+   property is true, <a>queue a fetch task</a> on <var>request</var> to
+   <a>process request end-of-body</a> for <var>request</var>.
+
+   <li><p>When <var>read</var> is fulfilled with a value that matches with neither of the
+   above patterns, or <var>read</var> is rejected, <a lt=terminated for=fetch>terminate</a> the
+   ongoing fetch with reason <i>fatal</i>.
```
annevk: This seems a little weird since we don't have a handle on the ongoing fetch and it's not passed to this algorithm. Perhaps it's okay for now though, but we should maybe file a follow up issue to see if we can fix that somehow.

fetch.bs
```diff
@@ -1121,6 +1111,46 @@ or "<code>worker</code>".
...
+
+ <li>
+  <p>Let <var>read</var> be the result of <a lt="read a chunk" for=ReadableStream>reading a
+  chunk</a> from <var>body</var>'s <a for=body>stream</a>.
+
+  <ul>
+   <li>
+    <p>When <var>read</var> is fulfilled with an object whose <code>done</code>
+    property is false and whose <code>value</code> property is a
+    <code>Uint8Array</code> object, run these substeps:
```
annevk: Is this the right way of phrasing this, @domenic?

fetch.bs
```diff
@@ -215,16 +215,6 @@ for <var>request</var>.
 
 <hr>
```
annevk: This can also be removed it seems if we move this algorithm.

yutakahirano: And can we remove two references to this item in Main Fetch section?

annevk: References to this `<hr>`? I'm not sure I understand.

yutakahirano: Ah, sorry, I misunderstood.

fetch.bs
domenic: It seems solid for a web spec. ES-style specs could use more formal internal slot inspection but that doesn't seem necessary here.

fetch.bs
```diff
@@ -1121,6 +1109,46 @@ or "<code>worker</code>".
  <li><p>Return <var>newRequest</var>.
 </ol>
 
+<p>To <dfn export for=request id=concept-request-upload-body>upload body</dfn> of a
```
domenic: Nit: I would phrase this as "upload a body for a request".

annevk: I think we should avoid "upload" entirely. The protocol specification doesn't use it either and it doesn't make much sense for non-HTTP URLs (although I guess those will fail to function anyway).

yutakahirano: Then which term should we use? In other words, what are the "primitive" operations on the network layout which I can use?

annevk: What was wrong with read?

annevk: If read cannot be used, "transmit" would be okay.

fetch.bs
annevk: I guess we can make it formal once IDL formalizes promises.

fetch.bs

annevk:
@yutakahirano I'd appreciate if you could mention what was wrong with "read". And also what the plan is for responses.

yutakahirano:
> @yutakahirano I'd appreciate if you could mention what was wrong with "read". And also what the plan is for responses.

For me, send / upload / transmit are easier to understand the intention. When you read something, you can do anything (including nothing) to it and so I feel it vague.

For responses, isn't the 15th step of HTTP-network fetch enough?

annevk:
@yutakahirano that only covers a response coming from the network. What about a response coming from a service worker?

yutakahirano:
> @yutakahirano that only covers a response coming from the network. What about a response coming from a service worker?

It's written in https://w3c.github.io/ServiceWorker/#dom-fetchevent-respondwith.

annevk:
Interesting. I think a lot of that would be more logically placed in Fetch. Otherwise we have to repeat all that text for foreign fetch service workers. Would you agree with that?

yutakahirano:
> Interesting. I think a lot of that would be more logically placed in Fetch. Otherwise we have to repeat all that text for foreign fetch service workers. Would you agree with that?

Yes, but as in [a comment](https://github.com/whatwg/fetch/issues/330#issuecomment-234462140), I hope we can simplify the description with pipeTo and transform streams someday (IIUC [transform stream](https://streams.spec.whatwg.org/#ts) is not ready). Is it OK to move the algorithm at that time?

annevk:
Maybe, it depends on whether that or foreign fetch happens earlier. Can you open a new tracking issue against Fetch for it? (I'm fine not doing it right away.)

annevk:
I'll also let @domenic have another look before landing this.

yutakahirano:
> Can you open a new tracking issue against Fetch for it? (I'm fine not doing it right away.

Done: #450

fetch.bs
domenic: Oh, I see, I was focusing on the Uint8Array issue, but the "object whose done property is false" part is the imprecise part. In particular, what if `obj.done` or `obj.value` is a throwing getter? Let me push a commit to fix.

domenic: Nevermind, these objects are created by the streams, and can never have getters or non-booleans or similar. The only "bad" thing that can happen is value not being a Uint8Array, and we have that covered.

PR #456, opened 2017-01-14
export disturbed and locked Body definitions

isonmad:


annevk:
I'm curious, what do you need these for? Also, did you intentionally not add your name to the Acknowledgments section?

isonmad:
The ServiceWorker spec talks about a Response being disturbed, but since https://github.com/w3c/ServiceWorker/commit/bec528d602bcbcfcf63cb16066cbcfa033733d46 it mistakenly links to [the readablestream definition of 'disturbed'](https://fetch.spec.whatwg.org/#concept-readablestream-disturbed) and 'locked' instead of these.

And I just didn't think of that, oops.

annevk:
@isonmad ah great, thanks for getting that fixed!

PR #458, opened 2017-01-17
Fix body handling on the Request constructor

yutakahirano:
As Request.body is exposed, we need to specify how to handle the body property of the input Request more correctly. After

```
let r = new Request(request, ...);
```

successfully returns,

 - `request.body` should remain unchanged.
 - `request.bodyUsed` should be true.

Fixing https://github.com/yutakahirano/fetch-with-streams/issues/60.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4925,34 +4925,25 @@ constructor must run these steps:
   <a for=request>body</a> is non-null, run these substeps:
 
   <ol>
-   <li><p>Let <var>dummyStream</var> be an <a>empty</a>
-   {{ReadableStream}} object.
-
-   <li><p>Set <var>input</var>'s <a for=Request>request</a>'s
-   <a for=request>body</a> to a new <a for=/>body</a> whose
-   <a for=body>stream</a> is <var>dummyStream</var>.
+   <li>
+    <p>Let «<var>ws</var>, <var>rs</var>» be a
+    <a href=https://streams.spec.whatwg.org/#ts-model>transform stream</a> such that
+    from <var>rs</var> one can read exactly the same data written to <var>ws</var>.
+    <p class="note no-backref"><var>ws</var> is a {{WritableStream}} object and <var>rs</var>
+    is a {{ReadableStream}} object.
 
    <li>
-    <p>Let <var>reader</var> be the result of <a lt="get a reader" for=ReadableStream>getting a reader</a>
-    from <var>dummyStream</var>.
+    <p>Call <a href=https://streams.spec.whatwg.org/#rs-pipe-to>pipeTo</a> on
+    <var>input</var>'s <a for=Request>request</a>'s <a for=request>body</a>'s
+    <a for=body>stream</a> with <var>ws</var>.
     <p class="note no-backref">This operation will not throw an exception.
 
-   <li>
-    <p><a lt="read all bytes" for=ReadableStream>Read all bytes</a> from
-    <var>dummyStream</var> with <var>reader</var>.
-    <p class="note no-backref">This operation makes <var>dummyStream</var> disturbed.
-  </ol>
+   <li><p>Set <var>r</var>'s <a for=Request>request</a>'s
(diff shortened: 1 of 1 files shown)


yutakahirano:
@domenic, can you take a look? I want to use a transform stream and pipeTo to avoid manual stream creation. Does it make sense to you?

I filed https://github.com/whatwg/streams/issues/658 for this change.

Thanks!

domenic:
This seems a little confusing to me because it mixes the conceptual level (linking to the model section for [transform stream](https://streams.spec.whatwg.org/#transform-stream)) with the concrete level (linking to the specific algorithm for [pipeTo](https://streams.spec.whatwg.org/#rs-pipe-to)). Maybe it is OK because we are just saying "everyone knows what an identity transform stream is". But it's not 100% clear to me, e.g. what the error propagation behavior is.

So in other words, the intent seems good, but it might not be precise enough. I'm curious what @annevk thinks.

Maybe one direction would be to make things even more vague, e.g. not mention the pipeTo algorithm at all, but just kind of vaguely describe how to get `rs` from `inputBody`?

annevk:
So my main problem is, what happens if the developer overrides `pipeTo()`? I guess you want to use some kind of safe `pipeTo()`? Maybe Streams should define a way to invoke that. If `pipeTo()` is safe, then invoking it with a conceptual stream is probably okay. But if `pipeTo()` needs to be abstracted anyway, we might as well abstract the identity transform stream too.

domenic:
The idea would be to create an abstract op for pipeTo (https://github.com/whatwg/streams/issues/658) so that calling it always calls the spec algorithm. Not sure where that falls on your safe vs. abstracted spectrum :)

annevk:
Sorry I missed that. I think that's okay. Could Streams not also define an identity transform stream that can be passed to that algorithm? Perhaps with a note that it's only for use by specifications until someone figures out a use case to expose it to JavaScript?

domenic:
Yeah, maybe that is the way to go. Then we can define error handling behavior more concretely.

We definitely have a use case for it, we just haven't yet specified all the underlying infrastructure (e.g. there are no algorithms for transform streams yet).

fetch.bs
```diff
@@ -4910,49 +4912,33 @@ constructor must run these steps:
...
-    <p><a lt="read all bytes" for=ReadableStream>Read all bytes</a> from
-    <var>dummyStream</var> with <var>reader</var>.
-    <p class="note no-backref">This operation makes <var>dummyStream</var> disturbed.
+    <p>Let <var>rs</var> bs a {{ReadableStream}} object from which one can read the exactly
+    same data as one could read from <var>inputBody</var>'s <a for=body>stream</a>.
+    <p class="note no-backref">This will be specified more precisely once
+    <a href=https://streams.spec.whatwg.org/#ts-model>transform stream</a> and
+    <a href=https://streams.spec.whatwg.org/#pipe-chains>piping</a> are precisely defined.
+    This makes <var>inputBody</var>'s <a for=body>stream</a> <a for=ReadableStream>locked</a> and
+    <a for=ReadableStream>disturbed</a> immediately.
```
annevk: I suggest we file an issue against Fetch to track this. And make this a class=XXX rather than a note.

annevk: And point to the issue here.

domenic:
So I am now a bit confused. Do we want streams to define these things in vague terms? Or is the vague phrasing here, linking to the vague "model" section of streams, good enough for now (until transform streams are defined)? It sounds like we don't need a pipeTo abstract operation anymore?

annevk:
I think it's okay to be vague for now, but I don't want us to be vague forever, so I want some kind of tracking URL for making it less vague.

We also need tests for the overall change.

fetch.bs
yutakahirano: Fixed.

annevk: Thanks, this PR looks good to me now, but we will need tests per the new process.

yutakahirano:
https://github.com/w3c/web-platform-tests/pull/4612

PR #464, opened 2017-01-24
Block 'ftp:' requests from non-'ftp:' clients.

mikewest:
Usage of the 'ftp:' protocol when requesting subresources from non-'ftp:'
clients has slowly declined over the last few years to the point where
it represents a [negligable amount of traffic][1]. The protocol does not
support modern requirements, like encryption, and there's interest from
at least one browser vendor in [removing support for FTP from their
network stack entirely][2].

To that end, this patch alters Fetch to block FTP subresources from
webby clients. That is, a page delivered from `http://example.com/`
will receive a network error response to requests like those generated
from ``.

[1]: https://www.chromestatus.com/metrics/feature/timeline/popularity/531
[2]: https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/bIJdwwoQ98U/-F1aL2FgBAAJ

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2408,11 +2408,11 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
 
  <li>
   <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
-  "<code>ftp</code>" and <var>request</var>'s <a for=request>client</a>'s
+  "<code>ftp</code>", <var>request</var>'s <a for=request>client</a>'s
   <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>", and
-  <var>request</var>'s <a for=request>reserved client</a> is either <code>null</code> or an
-  <a>environment</a> whose <a for=environment>target browsing context</a> is a
-  <a>nested browsing context</a>, then set <var>response</var> to a <a>network error</a>.
+  <var>request</var>'s <a for=request>reserved client</a> is either null or an <a>environment</a>
+  whose <a for=environment>target browsing context</a> is a <a>nested browsing context</a>, then set
+  <var>response</var> to a <a>network error</a>.
 
  <li>
   <p>Set <var>request</var>'s <a for=request>current url</a>'s
@@ -3147,7 +3147,7 @@ steps:
  <!-- https://chromium.googlesource.com/chromium/src/+/master/net/http/http_network_transaction.cc#982 -->
 
  <li><p>If <var>httpRequest</var>'s <a for=request>body</a> is non-null and <var>httpRequest</var>'s
- <a for=request>body</a>'s <a for=request>source</a> is non-null, then set
+ <a for=request>body</a>'s <a for=body>source</a> is non-null, then set
  <var>contentLengthValue</var> to <var>httpRequest</var>'s <a for=request>body</a>'s
  <a for=body>total bytes</a>, <a>UTF-8 encoded</a>.
 
@@ -3502,7 +3502,7 @@ steps:
 
  <li><p>If <var>connection</var> is not an HTTP/2 connection, <var>request</var>'s
  <a for=request>body</a> is non-null, and <var>request</var>'s <a for=request>body</a>'s
- <a for=request>source</a> is null, then <a for="header list">append</a>
+ <a for=body>source</a> is null, then <a for="header list">append</a>
  `<code>Transfer-Encoding</code>`/`<code>chunked</code>` to <var>request</var>'s
  <a for=request>header list</a>.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -2407,6 +2407,12 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
   have it expose less sensitive information.
 
  <li>
+  <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
+  "<code>ftp</code>" and <var>request</var>'s <a for=request>client</a>'s
+  <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>",
+  set <var>response</var> to a <a>network error</a>.
+
+ <li>
   <p>Set <var>request</var>'s <a for=request>current url</a>'s
   <a for=url>scheme</a> to "<code>https</code>" if
   all of the following conditions are true:
--- a/fetch.bs
+++ b/fetch.bs
@@ -2411,8 +2411,8 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
   "<code>ftp</code>" and <var>request</var>'s <a for=request>client</a>'s
   <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>", and
   <var>request</var>'s <a for=request>reserved client</a> is either <code>null</code> or an
-  <a>environment</a> whose <a for=environment>target browsing context</a> is a <a>nested browsing
-  context</a>, then set <var>response</var> to a <a>network error</a>.
+  <a>environment</a> whose <a for=environment>target browsing context</a> is a
+  <a>nested browsing context</a>, then set <var>response</var> to a <a>network error</a>.
 
  <li>
   <p>Set <var>request</var>'s <a for=request>current url</a>'s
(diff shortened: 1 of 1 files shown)


mikewest:
Hey @annevk. This seemed like a reasonable spot to put the FTP check we briefly discussed on blink-dev@. I don't think the ordering particularly matters, except insofar as it should come after CSP so that violation reports are generated correctly. I considered putting the check into the `ftp` case of Basic Fetch's switch, but that seemed a little more hidden than reasonable. WDYT?

As noted on IRC, tests will be a little annoying, as WPT doesn't actually support FTP, and @jgraham (reasonably) doesn't consider such support to be a priority. I'll put something together, but I think it'll be ugly.

annevk:
Would this not also block HTTP->FTP navigation and FTP `<iframe>`s?

mikewest:
Blocking `<iframe>` was intentional, but I didn't mean to block top-level navigation. That said, it's not clear to me that we have a mechanism for determining whether or not a given navigation request is top-level anymore. Is that something we could tack on from HTML's navigation algorithm, or is there a mechanism I'm missing?

mikewest:
Regardless of whether it's alive, `client` would only tell me if a top-level document initiated the request, though, right? It wouldn't tell me if it was _targeting_ a top-level context or a nested context.

annevk:
Good point. We added "reserved client" and "target client id" for service workers, but neither can answer this question I think.

domenic:
Is there a reason we want to allow top-level navigation to ftp: resources?

mikewest:
@domenic: Folks still use the browser to download resources from FTP, and navigate FTP servers. *shrug* Chrome's network team is interested in moving FTP support out of the browser and into something else (bundled extension/ChromeOS app, perhaps?), so I think the end goal is indeed to disallow navigation to `ftp:`. Killing off subresource requests is a necessary step to give them the ability to make that change, but I don't think it's on their short-term roadmap.

mikewest:
Updated the patch to check the `reserved client`'s `target browsing context` based on the discussion with @jungkees.

fetch.bs
```diff
@@ -2407,6 +2407,14 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
   have it expose less sensitive information.
 
  <li>
+  <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
+  "<code>ftp</code>" and <var>request</var>'s <a for=request>client</a>'s
+  <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>",
+  and <var>request</var>'s <a for=request>reserved client</a> is either <code>null</code>
+  or an <a>environment</a> whose <a for=environment>target browsing context</a> is a
+  <a>nested browsing context</a>, set <var>response</var> to a <a>network error</a>.
```

Also, this doesn't seem to be formatted per the 100 column rule?

annevk:
@youennf any thoughts from WebKit on this?

@travisleithead any thoughts from Edge?

@mcmanus @valenting any thoughts from Gecko/Necko?

fetch.bs
mikewest: Done and done.

fetch.bs
```diff
@@ -2409,10 +2409,10 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
...
   <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
   "<code>ftp</code>" and <var>request</var>'s <a for=request>client</a>'s
-  <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>",
-  and <var>request</var>'s <a for=request>reserved client</a> is either <code>null</code>
-  or an <a>environment</a> whose <a for=environment>target browsing context</a> is a
-  <a>nested browsing context</a>, set <var>response</var> to a <a>network error</a>.
+  <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>", and
+  <var>request</var>'s <a for=request>reserved client</a> is either <code>null</code> or an
+  <a>environment</a> whose <a for=environment>target browsing context</a> is a <a>nested browsing
+  context</a>, then set <var>response</var> to a <a>network error</a>.
```
annevk: No newlines in inline elements. Sorry for the difference with HTML.

mikewest: Hrm. I thought y'all just didn't do newlines inside the tag, but that newlines inside the text were allowed. *shrug* I'll fix it up.

valenting:
Tracking Firefox changes in [bug 1340198](https://bugzilla.mozilla.org/show_bug.cgi?id=1340198)

annevk:
@mikewest given that Mozilla is tentatively on board I'm okay with landing this. https://github.com/w3c/web-platform-tests/issues/4751 is the web-platform-tests issue we should reference from the commit? While you file bugs against Safari and Edge?

mikewest:
That's the correct issue, yes. I'll file bugs against the other browsers.

I can also upstream the (not very useful) test that I'll land to verify Blink's console error message; might be useful for others, even if it doesn't verify any actual behavior (because no FTP server is present).

fetch.bs
```diff
@@ -2407,6 +2407,14 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
   have it expose less sensitive information.
 
  <li>
+  <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
+  "<code>ftp</code>" and <var>request</var>'s <a for=request>client</a>'s
```
annevk: Can this "and" be replaced by a comma? (I can do it, just want someone to double check.)

mikewest: Yes, I think it can be replaced by a comma.

hlsantos:
I don't get this decision.  What makes it your right to decide what a private or public client/server system uses?  That's not "open" programming.  FTP is still a variable protocol used at all scales.  Sure, perhaps for legacy operations, but what isn't legacy today?   Even a browser is a legacy UI and HTTP method download is legacy!   Normally, in lieu of some security reason,  you would make this optional for end-users to decide.   Just because you measured "low traffic," (you didn't measure mine), in my design opinion, you don't take features out and break web access and backend operations to switch their load/storage resources.  geez

annevk:
Moving away from non-secure FTP and HTTP is a security decision. Everything should move to HTTPS.

hlsantos:
Why not also pull HTTP? 

That should not be your decision.  Just consider that not everything needs to be secured. Not everything needs to be encrypted with high overhead processing.  It should be left to the implementations and their operations.  There are public and private environments.  Don't assume everything is public and cloud-based.   There are many reasons why people would use a FTP resource.  There are private systems that have operated and still operate just fine for a long time, until now.   

The fact is, there is an extreme amount of overhead with just allowing for HTTPS especially today with enhanced ciphers encrypted packets going through DPI (Deep Packet Inspection), couple this with connectivity movement towards broadband, the overhead and network latency is far worst with higher delays, timeouts, resets, etc.    In my technical opinion, this will soon be a major problem.

Chrome's future needs to be better managed.  These poor decisions puts pressure for Chrome derivatives that do not follow highly subjective, to me,  market inexperienced,  main branch design change decisions.  To use a "low traffic" decision to pull FTP, imo, this was not a legit and reasonable reason.

annevk:
This is not a decision made solely by Chrome. You're leaving comments on the forum of a standards body. I also don't see us reverting this decision at this point.

hlsantos:
This is an example of how this "standards body" really didn't represent the market place,  its users, its implementations.  It made a unreasonable decision to pull a still viable FTP protocol standard resource still used across the globe and at all scales.   

So now I have to see what this means, what browsers are using the same base code, which are following suit, whether it is deprecated or disabled or completely pulled, begin to inform customers about this change, go thru dozens of company support resource sites to change download links, moving all overhead and files to a HTTPS server, and also begin doing what I have never done before with Chrome -- begin recommending non-chrome browsers to customers.

I just wanted to put it on the "record."  Thanks for your feedback.

Hector Santos, CTO
Santronics Software, Inc.

hlsantos:
I see, the user can still use the "ftp:" resource for a url from the address bar and the browser, I suppose has a built-in ftp client to display the ftp output as a web page with a standard folder/file tree display.    

But a link with a "ftp:" reference within a HTTP: generated page will not work?

If that the issue?  

Is this a feature disabled in the code or pulled from the code?  If disabled, is there a config switch for it?

Why wasn't address bar method  pulled?

PR #470, opened 2017-01-27
Clarified that origins are compared as being same-origin

### LINKED ISSUE #468: Clarification: how are origins compared for equality?
In https://fetch.spec.whatwg.org/#concept-main-fetch, main fetch step 11 includes a test:

> request’s current url’s origin is request’s origin and CORS flag is unset 

which includes an equality test between origins, but it is not clear which equality is meant. The lhs is "request’s current url’s origin", so its domain is always `null` (https://url.spec.whatwg.org/#origin). The rhs is the "request’s origin", which may be set by https://fetch.spec.whatwg.org/#dom-request, step 7:

> Let origin be current settings object’s origin.

or by https://fetch.spec.whatwg.org/#concept-fetch, step 2:

> If request’s origin is "client", set request’s origin to request’s client’s origin.

which may have a domain.

I think what's intended is that the origins are to be compared as being same-origin (https://html.spec.whatwg.org/multipage/browsers.html#same-origin) that is the domain is ignored. However, the way the spec is currently written, it looks like syntactic identity is being used, so the test will fail if the client has set `document.domain`.

Should the test here be "request’s current url’s origin is **the same as** request’s origin and CORS flag is unset"?

asajeffrey:
Origins should be compared as being same-origin in the main fetch algorithm.

Fixes #468.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2436,7 +2436,7 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
 
   <dl class=switch>
    <dt><var>request</var>'s <a for=request>current url</a>'s
-   <a for=url>origin</a> is <var>request</var>'s
+   <a for=url>origin</a> is <a>same origin</a> with <var>request</var>'s
    <a for=request>origin</a> and <i>CORS flag</i> is unset
    <dt><var>request</var>'s
    <a for=request>current url</a>'s
@@ -5586,6 +5586,7 @@ however, it is perfectly fine to do so.
 <p>Thanks to
 Adam Barth,
 Adam Lavin,
+Alan Jeffrey,
 Alexey Proskuryakov,
 Andrew Sutherland,
 Ángel González,


asajeffrey:
cc @annevk

fetch.bs
```diff
@@ -2436,7 +2436,9 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
 
   <dl class=switch>
    <dt><var>request</var>'s <a for=request>current url</a>'s
-   <a for=url>origin</a> is <var>request</var>'s
+   <a for=url>origin</a> is the
+   <a href="https://html.spec.whatwg.org/multipage/#same-origin">same</a>
```
annevk: You can change this to `<a>same origin</a>`with phrasing "current url's origin is [same origin] with request's origin". That will end up linking to HTML's definition once generated.

asajeffrey: Done.

annevk:
Thanks!

PR #476, opened 2017-01-31
Stop lowercasing header names

### LINKED ISSUE #203: Headers.get/getAll/has do not byte-lowercase the name
The algorithms for insertion/deletion/appending all explicitly byte-lowercase the provided name. The algorithms for retrieving header values only say "whose name is _name_".

### LINKED ISSUE #304: Optional disable setting headers keys in lowercase
Hi, I'm trying to use `fetch` in Chrome but I think it's not working for me for one reason. I'll explain my use case and you can tell me if I can do something to use `fetch`.

I'm developing a Chrome extension (a content script) that reads from `localStorage` an auth `token`. This way I can make request agains a server. I'm not managing that server and I'm not the owner. I can't change anything on the server. The thing is that I'm trying this:

```
fetch(url, {
  method: "GET",
  headers: {
    "Token": "auth-token-content-goes-hear"
  },
})
```

But when I see the request on Chrome Dev Tools I see `token` instead of `Token` and the server is responding with `401`. My first implementation is doing the same with jQuery and the server is returning results.

I understand the standard is put headers in lowercase. But maybe could be posible set a flag to skip it. In situations like mine (user can not change server) is not posible to use `fetch`. 

BTW great work! I appreciate all the work you are doing here 💪 

Thanks!

annevk:
Instead compare header names using a byte-case-insensitive match and
lowercase them when exposed through the API.

Fixes #203 and fixes #304.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -162,13 +162,6 @@ comparing them exactly, byte for byte, except that the bytes in the range 0x41 t
 inclusive, are considered to also match their corresponding byte in the range 0x61 to
 0x7A, inclusive.
 
-<p>A <dfn>case-insensitive byte sequence</dfn> is a byte sequence that when compared to
-another byte sequence does so in a <a>byte-case-insensitive</a> manner.
-
-<p id=example-case-insensitive-byte-sequence class=example>The
-<a>case-insensitive byte sequences</a>
-`<code>Content-Type</code>` and `<code>content-TYPE</code>` are equal.
-
 <hr>
 
 <p id=fetch-url>A <dfn>response URL</dfn> is a <a for=/>URL</a> for which implementations need not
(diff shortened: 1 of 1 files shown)


annevk:
See https://github.com/w3c/wptserve/issues/112 for what is blocking tests. See https://github.com/w3c/web-platform-tests/commit/e94ccfed3a3ec91cb357fd6684caa2d065b34ae8 for some test sketches.

The change shouldn't affect any JavaScript API behavior. They will still expose header names as lowercase even though the case given by the developer is used on the network. (Though see https://github.com/whatwg/xhr/issues/108 and https://github.com/whatwg/xhr/issues/109 for discussions on that front with regards to XMLHttpRequest, which ideally is somewhat aligned.)

If we want a discussion on whether to allow case normalization for known headers to happen (Firefox and Safari do this) we should do that through a separate issue. For now I noted it in the source that this is a thing.

I would appreciate review from @younnf and @jdm.

You can view the specification with this PR applied at https://fetch.spec.whatwg.org/branch-snapshots/annevk/header-names/. Hopefully that simplifies reviewing. (Click in the title of the red blob to hide it.)

annevk:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1282426
* https://bugs.chromium.org/p/chromium/issues/detail?id=687155
* Need more tests for the other browsers.

mcmanus:
http header names are case insensitive, and h2 forces them lower case. seems weird to have case sensitive comparisons in that case

annevk:
@mcmanus the only thing this patch does is to preserve the case, as enough users of h1 complained that lowercasing header names didn't work for them (Firefox never implemented that bit of Fetch). Comparisons are done using byte case-insensitive comparison, as is called out throughout this patch. Is there a particular instance where this is done wrong?

mcmanus:
Its all good - I misread part of the patch

fetch.bs
```diff
@@ -278,66 +271,59 @@ for consistency.
...
+
+<ol>
+ <li>
+  <p>If <var>list</var> <a for="header list">contains</a> <var>name</var>, then set <var>name</var>
+  to the first such <a>header</a>'s <a for=header>name</a>.
+
+  <p class="note no-backref">This reuses the casing of the <a for=header>name</a> of the
+  <a>header</a> already in the <a for=/>header list</a>, if any. If there are multiple matched
+  <a>headers</a> their <a for=header>names</a> will all be identical.
+  <!-- XXX Firefox and Safari adjust known header names too. -->
```
youennf: I like the simplicity that provided case-insensitive header names to the spec.
Maybe a serialization context could have stored the casing information to leave the header names case-insensitive.
But anyway, this looks good to me as the PR is taking great care of making the specific casing unobservable from JS scripts.

I would tend to write the above statements as SHOULD/MUST/MAY, something like:
All matching headers must use the same casing.
The casing of the first header already in the list should be used for all matching headers.
The user agent may use its own casing for well-known headers.

annevk: Note that all browsers currently make it observable through `getAllResponseHeaders()` so we might still get stuck there one way or another.

If we are to allow the user agent to pick casing for well-known headers I think we should probably standardize that and get everyone on board. Seems like it would be best done as a new issue.

fetch.bs
```diff
@@ -362,6 +348,8 @@ a <a for=/>header list</a> (<var>list</var>), run these steps:
    <a for=header>combined value</a> given <var>name</var> and
    <var>list</var>.
 
+   <li><p>Set <var>name</var> to <var>name</var>, <a>byte-lowercased</a>.
```
yutakahirano: Shouldn't we lowercase _name_ before sorting? Imagine _headers_ contains "a", "B", "c" and "D" as names. _names_ will be ["B", "D", "a", "c"] and after lowercased it will be ["b", "d", "a", "c"].

fetch.bs
```diff
@@ -2257,10 +2244,8 @@ the request.
...
 
- <li><p>If <var>request</var>'s <a for=request>header list</a> does not
- contain a <a>header</a> whose
- <a for=header>name</a> is
- `<code>Accept-Language</code>`, user agents should
+ <li><p>If <var>request</var>'s <a for=request>header list</a>
+ <a for="header list">does not contain</a> `<code>Accept-Language</code>`, user agents should
  <a for="header list">append</a>
  `<code>Accept-Language</code>`/an appropriate <a for=header>value</a>
  to <var>request</var>'s <a for=request>header list</a>.
```
yutakahirano: Should "_hint_name_ is not in _request_'s header list" below be fixed?

annevk:
Thanks @yutakahirano, I believe I addressed your feedback.

fetch.bs
youennf: Isn't getAllResponseHeaders outputting lowercase header names with the proposed PR?

annevk: It currently is, yes, but https://github.com/whatwg/xhr/issues/109 is still unresolved and nobody implements lowercasing there.

annevk:
I'm assuming the text looks okay to everyone now. I'm mainly waiting on tests and the infrastructure needed for that.

annevk:
Ended up filing https://bugs.webkit.org/show_bug.cgi?id=168768 against WebKit.

Issue #481, opened 2017-02-03
Constructing a Headers from another Headers is lossy


bzbarsky:
We've changed the constructor signature to take:

    sequence<sequence<ByteString>> or record<ByteString, ByteString>

so when passing a Headers we will iterate over it, instead of just copying its list of stuff.

But iteration over headers is specced to run through https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine which means it will reorder the headers, combine them from multiple header lines to one line, and so forth.  This process is lossy and the result may well behave differently from the original when sent to a server.  Yes, per HTTP spec it probably mostly should not.  No, that's not guaranteed even per spec (e.g. behavior for non-list headers that are duplicate is ... well, the RFCs don't really define it very well).  No, not all servers implement HTTP correctly.

annevk:
Hmm. Thoughts @youennf?

Note that this only affects headers that developers control themselves. Browsers don't really manipulate this list.

bzbarsky:
Note that this affects cases when `Request` or `Response` are constructed via their respective constructors.  Those constructors involve calling https://fetch.spec.whatwg.org/#concept-headers-fill with the thing that was passed in, which will have the same issue, right?

youennf:
Yes, the process is lossy. We might end up loosing some header casing information as well for well known header names. WebKit internal structure is a map which is lossy.
It seems good that the header API is sanitising things and makes things consistent amongst browsers.

As of Request/Response constructors, we could special case the transfer of headers in a "lossless" way. We could also put the burden on web developers to be extra careful at how they use the API in those edge scenarios.

annevk:
Given the change needed to the `Request` constructor in #484, perhaps we should just make `HeadersInit` accept a `Headers` object again and copy things in order whenever that takes place... I don't really care strongly one way or another though.

PR #484, opened 2017-02-08
Regression: make Request constructor account for changes to fill

### LINKED ISSUE #483: Type confusion in Request constructor leading to underdefined behavior
https://fetch.spec.whatwg.org/#dom-request step 28 says:

> Let _headers_ be a copy of r’s Headers object and its associated header list.

OK, so _headers_ is a Headers object.

Step 29 says:

> If _init_’s `headers` member is present, set _headers_ to init’s `headers` member.

Let's say it's not present.

Step 32 says: 

> Fill _r_’s Headers object with _headers_. Rethrow any exceptions.

with a link to <https://fetch.spec.whatwg.org/#concept-headers-fill>.

But that algorithm expects the input to be a sequence or a record, and it's neither: it's a Headers.  Yes, you can create a sequence from a Headers by round-tripping through to-JS and from-JS conversion, but if that's the expectation that needs to be clearly spelled out here, because that's obviously observably different from just copying over the things from the Headers instance that was passed in.

annevk:
In particular, now the fill algorithm no longer takes Headers objects,
the Request constructor algorithm will need to deal with that situation
itself.

Regressed in 00fce0d96e799f5bc8bc53b30fcec56c4e0b948c.

(Also fix some minor linking issues Bikeshed found.)

Fixes #483.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/regression-fix/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/686a1ad...d91f5df.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4963,8 +4963,13 @@ constructor must run these steps:
    "<code>request-no-cors</code>".
   </ol>
 
- <li><p><a lt=fill for=Headers>Fill</a> <var>r</var>'s <a for=Request>headers</a> with
- <var>headers</var>. Rethrow any exceptions.
+ <li><p>If <var>headers</var> is a <code>Headers</code> object, then for each <var>header</var> in
+ its <a for=Headers>header list</a>, retaining order, <a for=Headers>append</a> <var>header</var>'s
+ <a for=header>name</a>/<var>header</var>'s <a for=header>value</a> to <var>r</var>'s {{Headers}}
+ object.
+
+ <li><p>Otherwise, <a for=Headers>fill</a> <var>r</var>'s {{Headers}} object with
+ <var>headers</var>.
 
  <li><p>Let <var>inputBody</var> be <var>input</var>'s
  <a for=Request>request</a>'s <a for=request>body</a>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4963,10 +4963,10 @@ constructor must run these steps:
    "<code>request-no-cors</code>".
   </ol>
 
- <li><p>If <var>headers</var> is a <code>Headers</code> object, then for each <var>header</var> in
- its <a for=Headers>header list</a>, retaining order, <a for=Headers>append</a> <var>header</var>'s
- <a for=header>name</a>/<var>header</var>'s <a for=header>value</a> to <var>r</var>'s {{Headers}}
- object.
+ <li><p>If <var>headers</var> is a <code>Headers</code> object, then <a for=list>for each</a>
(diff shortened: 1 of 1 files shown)


bzbarsky:
That looks pretty reasonable, thank you!

annevk:
Thanks for the review. I'll leave this open for a bit in case #481 gets any more arguments one way or another as that would definitely affect the outcome here.

rakuco:
How is this change supposed to be tested? Are we basically interested in making sure that creating a `Request` with a `Headers` object has the same behavior as if we use a sequence or record instead?

In other words,

```js
const headers = [
  ["X-fOo", "BAZ"],
  ["X-FOO", "bar"],
  ["SomeHeader", "quux"]
];
fetch("....", {"headers": headers})
```

will send a request with `X-fOo: BAZ, bar` and `SomeHeader: quux` among its headers, and so will

```js
let r = new Request("....");
r.headers.append("X-fOo", "BAZ");
r.headers.append("X-FOO", "bar");
r.headers.append("SomeHeader", "quux");
fetch(r)
```

which would, before this PR, actually be expected to send `someheader: quux` and `x-foo: baz, bar`, and `someheader` would always show up before `x-foo`.

If that's the case, I can get the expected behavior in Blink and WebKit (WebKitGTK+ 2.18), but Gecko (Firefox Nightly) always sends `x-foo: bar` (dropping "BAZ" altogether). WebKit and Blink always seem to do a case-insensitive sort in the headers set above before sending them though, while Gecko doesn't, but I guess that's not an issue.

annevk:
I don't think this can really be tested, this was just a bug in the algorithm. In particular, https://fetch.spec.whatwg.org/#concept-headers-fill doesn't accept a `Headers` object, so we needed to account for that somehow.

Firefox dropping `BAZ` seems like a bug and that's indeed something worth testing (and filing a bug on).

rakuco:
> I don't think this can really be tested, this was just a bug in the algorithm.

OK, thanks for the clarification.

> Firefox dropping `BAZ` seems like a bug and that's indeed something worth testing (and filing a bug on).

This is already tested in `XMLHttpRequest/getallresponseheaders-cl.htm`. I could file a new Gecko bug if you prefer, but I thought this would already be tracked by Firefox's WPT imports.

annevk:
We don't automatically file new bugs against ourselves for new failures unfortunately. For all browsers it's still best to explicitly file bugs. I can file this one.

annevk:
Turns out this is filed as https://bugzilla.mozilla.org/show_bug.cgi?id=1366945.

Issue #473, opened 2017-01-31
Cannot use [SameObject] for trailers


annevk:
Per https://github.com/heycam/webidl/pull/217 [SameObject] is no longer suitable. Fixing this should also give us an opportunity to write some tests.

annevk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1339096
https://bugs.webkit.org/show_bug.cgi?id=168232
https://bugs.chromium.org/p/chromium/issues/detail?id=691599
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10939642/

PR #488, opened 2017-02-13
Response's trailer cannot use [SameObject]

annevk:
Since an attribute that returns a promise can be made to return a
rejecting promise as well. (We really need [Cached] or some such.)

Tests: https://github.com/w3c/web-platform-tests/issues/4811.


annevk:
This ends up fixing #473.

PR #494, opened 2017-02-22
Breaking: redo value parsing as value extraction

### LINKED ISSUE #474: "parse a header value" should be more careful with types
It's passed either a header or header list. Seems we could make the algorithm deal with that a little better than just ignore the differences.

annevk:
We now have two separate algorithms that can be used rather than one
with type confusion:

* “extracting header values” (takes a header)
* “extracting header list values” (takes a name and a header list)

Fixes #474.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/value-extraction/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/ff8a3e9..annevk/value-extraction:a9d8f59.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -399,7 +399,7 @@ each other by `<code>,</code>`, in order.
  <li>`<code>Accept-Language</code>`
  <li>`<code>Content-Language</code>`
  <li>`<code>Content-Type</code>` and whose <a for=header>value</a>,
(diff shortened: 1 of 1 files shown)


annevk:
@mikewest would you take a look at this? I suspect if anyone you might have a dependency on this algorithm already.

mikewest:
I'm pretty sure I depend on this in a few places. But it looks like a
reasonable change, so I'll do the trivial updates. :)

LGTM.

Issue #373, opened 2016-08-24
Go deeper into HTTP caching


mnot:
My first inclination is to break out some parts of HTTP-network-or-cache Fetch into new sections and then grow them.

Specifically - 
- bullets 17 and 18 ("If httpRequest's cache mode is neither...", "Otherwise, if httpRequest's cache mode is either...") should be shifted to a new "HTTP Cache Lookup" section, that specifies http://httpwg.org/specs/rfc7234.html#constructing.responses.from.caches algorithmically. 
- after bullet 19 ("If response is null, run these substeps..."), insert a new item (or reference to a new section, "HTTP Cache Storage" or similar, along the lines specified in http://httpwg.org/specs/rfc7234.html#response.cacheability. 

If that makes sense, I'll start a branch, and probably some tests too.

Question - Is the Streams work largely done? I suspect some of the fine points here are going to need them.

annevk:
Yeah, that aligns with my tentative plan for the cache. Either something like "HTTP-cache fetch" or a new Infrastructure section.

Streams is still being improved and uploads isn't done yet. We can probably find some temporary wording though if it's required anywhere and we lack something.

mnot:
Am starting in https://github.com/mnot/fetch/tree/http-cache -- not done yet, but would love any feedback you have so far.

annevk:
One thing that would be good to take into account is #307. Effectively a browser needs at least two global instances of an HTTP cache, as defense-in-depth for CORS. That's why I was considering having the cache lookup be in its own section.

I have editorial quibbles, but maybe those are better to do once you're ready?

Issue #307, opened 2016-05-20
Vary HTTP cache on credentials mode


annevk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1274555 by @ziyunfei for details. Best to make this explicit.

annevk:
@mnot any ideas on how to best do this? Basically say that aside from the request we also use some credentials boolean as a cache key?

mnot:
Yeah. I'll try to rough something in (kick me if it takes too long).

annevk:
Note that #340 is also outstanding. I'll leave both for a bit to find out your approach. If you lack the time I guess I'll study the HTTP cache specification in a bit more detail and have another go at the whole thing.

mnot:
Thinking out loud:

I'd really like to fix this for all caches, not just browser caches -- although it might be that the first step is spec'ing it in Fetch, it'd be great if the approach taken can be backported to 7234.

I don't think using _credentials flag_ or even _credentials mode_ is adequate. For example, with either, if I browse using two different client certs or session cookies, there is going to be intermingling between them in the cache.

Instead, the actual credential has to be part of the cache key.

That means:
 - HTTP authentication
 - Cookies
 - TLS client certs

The sticky part here is cookies; if we start making them part of the cache key for all responses, cache efficiency is going to bomb, because lots of cookies aren't credentials.

Flagging content with credentials (e.g., with a new `Set-Cookie` parameter, or using the `Key` header) probably won't be sufficient to meet the security requirements here.

Doing the opposite -- flagging responses whose cookies *aren't* credentials -- might work. `Cache-Control: public` sort of works this way already for the `Authentication` header -- WDYT?

annevk:
The main invariant we are trying to protect here is that if you load something with credentials that has `Access-Control-Allow-Origin: *` due to navigation or some such, that response does not end up reused by a `fetch()` for the same URL without credentials (because then the CORS check would suddenly pass while it should not).

If the user has multiple identities they are better off using something like https://blog.mozilla.org/tanvi/2016/06/16/contextual-identities-on-the-web/ I think, although offering servers more tools to more cleanly separate identities seems like a useful endeavor as well, but I think that's somewhat separate from this distinction here, which mainly follows from CORS.

mnot:
![preview](https://cloud.githubusercontent.com/assets/74384/23386504/ea387cb6-fdaa-11e6-8192-4b648f76fd70.jpeg)

mnot:
[...]

Issue #144, opened 2015-10-26
Handling Partial Content / 206


mnot:
Discussed in #97 - breaking out into a separate issue.

mnot:
Some discussion at TPAC. There are many use cases for range requests, initial approach might be to identify bare minimum and only support those. E.g.,

0. Pass through range requests to the network
1. Serve range requests from a complete cache entry
2. Cache partial content when requested
3. Synthesise range requests to complete partial content (for full or partial requests)
4. ...

(0) and (1) are a straw-man target for now. Thoughts?

annevk:
So from those:

1. Already happens.
2. Is something for the Cache API to decide.
3. Already happens in step 8 of https://fetch.spec.whatwg.org/#concept-http-network-fetch as far as I can tell.
4. This does not currently happen.

Here are some things I'm wondering about:

1. Should we change https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch step 15 to talk about one or more partial responses?
2. Should step 17.1 explicitly talk about a complete response?
3. Is there any freshening of headers when you get a 206 response? https://tools.ietf.org/html/rfc7234#section-4.3.4 seems to suggest there isn't. HTTP Range Requests also does not talk about this as far as I can tell.
4. Is therefore the only thing we need to introduce a 206 clause (similar to step 17 for 304) when cache mode is either "default" or "no-cache" that does recombining?

mnot:
1. You can take the approach that every 206 is combined into one cache entry as soon as it occurs, or you can alternatively store multiple 206s and combine them at some later time. HTTP doesn't care which, as long as the effect is the same; see http://httpwg.github.io/specs/rfc7234.html#combining.responses 
2. Step 14 already requires a complete response.
3. see http://httpwg.github.io/specs/rfc7233.html#combining.byte.ranges
4. maybe, yeah...

mnot:
> Is something for the Cache API to decide.

I mean the HTTP cache, not the SW cache. Or do you mean it's effectively an implementation decision?

annevk:
> I mean the HTTP cache, not the SW cache.

Ah I see. Maybe we should talk about that too.

code-tree:
> 2. Serve range requests from a complete cache entry

Any hope that this will be implemented? It certainly would help with [a problem](http://stackoverflow.com/questions/37612177/cannot-play-cached-audio-from-service-worker) I currently have with the Audio element and range headers

jakearchibald:
[...]

PR #496, opened 2017-03-01
Http cache

mnot:
For #373.

Also should help #336 (since HTTP clearly says this, and there's a direct reference now).

I think it addresses #307 (although Anne may want a different approach).

Touches #144.

***
[Preview](https://s3.amazonaws.com/pr-preview/mnot/fetch/http-cache.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/3b4cb54...mnot:edef79f.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3119,6 +3119,9 @@ steps:
 <p class=note><i>CORS flag</i> is still a bookkeeping detail. As is
 <i>authentication-fetch flag</i>.
 
+<p class=note>Some implementations might support caching of partial content, as per <cite>HTTP
+Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by browser caches.
+
 <ol>
  <li><p>Let <var>httpRequest</var> be null.
 
@@ -3345,7 +3348,7 @@ steps:
 
  <li><p>Let <var>response</var> and <var>storedResponse</var> be null.
 
- <li><p>Let the <var>revalidatingFlag</var> and <var>partialFlag</var> be unset.
+ <li><p>Let the <var>revalidatingFlag</var> be unset.
 
  <li>
   <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is neither "<code>no-store</code>"
@@ -3369,11 +3372,6 @@ steps:
    <li><p>If <var>storedResponse</var> requires validation (i.e., it is not fresh), then set the
    <var>revalidatingFlag</var>.
 
-   <li><p>If <var>storedResponse</var> is not complete, then set the <var>partialFlag</var>.
-
-   <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>only-if-cached</code>"
-   and the <var>partialFlag</var> is set, then abort these substeps.
-
    <li>
     <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>force-cache</code>" or
     "<code>only-if-cached</code>", then set <var>response</var> to <var>storedResponse</var> and
@@ -3382,14 +3380,8 @@ steps:
     <p class=note>As mandated by HTTP, this still takes the `<code>Vary</code>` <a>header</a>
     into account.
 
-   <li><p>If both the <var>revalidatingFlag</var> and <var>partialFlag</var> are set,
-   <a for="header list">append</a> `<code>If-Range</code>` with an appropriate value to
-   <var>httpRequest</var>'s <a for=request>header list</a>, as per the
-   "<a href=https://tools.ietf.org/html/rfc7233#section-3.2>If-Range</a>" chapter of
-   <cite>HTTP Range Requests</cite>. [[!HTTP-RANGE]]
-
    <li>
-    <p>If the <var>revalidatingFlag</var> is set and the <var>partialFlag</var> is unset, then:
+    <p>If the <var>revalidatingFlag</var> is set, then:
 
     <ol>
      <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
@@ -3406,15 +3398,6 @@ steps:
     <p class=note>See also the
     "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
     chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
-
-   <li><p>If the <var>revalidatingFlag</var> is unset and the <var>partialFlag</var> is set, then
-   <a for="header list">append</a> `<code>Range</code>` with an appropriate value to
-   <var>httpRequest</var>'s <a for=request>header list</a>, as per the
-   "<a href=https://tools.ietf.org/html/rfc7233#section-3.1>Range</a>" chapter of
-   <cite>HTTP Range Requests</cite>. [[!HTTP-RANGE]]
-
-   <li><p>If both the <var>revalidatingFlag</var> and <var>partialFlag</var> are unset, then set
-   <var>response</var> to <var>storedResponse</var>.
   </ol>
 
  <li>
@@ -3451,24 +3434,6 @@ steps:
      <li><p>Set <var>response</var> to <var>storedResponse</var>.
 
     </ol>
-
-   <li>
-    <p>If the <var>partialFlag</var> is set and <var>forwardResponse</var>'s
-    <a for=response>status</a> is <code>206</code>, run these substeps:
-
-    <ol>
-     <li>
-      <p>Update <var>storedResponse</var>'s <a for=response>header list</a> using
-      <var>forwardResponse</var>'s <a for=response>header list</a>, as per the
-      "<a href=https://tools.ietf.org/html/rfc7234#section-3.3>Combining Partial Content</a>" chapter
-      of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
-
-      <p class="note">This updates the stored response in cache as well.
-
-     <li><p>Set <var>response</var> to <var>storedResponse</var>.
-
-    </ol>
-    <!-- TODO: account for partial request failure -->
 
    <li>
     <p>If <var>response</var> is null, then run these substeps:
--- a/fetch.bs
+++ b/fetch.bs
@@ -3345,7 +3345,7 @@ steps:
 
  <li><p>Let <var>response</var> and <var>storedResponse</var> be null.
 
- <li><p>Let the <var>revalidating flag</var> and <var>partial flag</var> be unset.
+ <li><p>Let the <var>revalidatingFlag</var> and <var>partialFlag</var> be unset.
 
  <li>
   <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is neither "<code>no-store</code>"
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
 
- <li>
-  <p>If <var>response</var>'s <a for=response>status</a> is <code>304</code> and
-  <var>httpRequest</var>'s <a for=request>cache mode</a> is either
-  "<code>default</code>" or "<code>no-cache</code>", run these substeps:
+   <li><p>If <var>httpRequest</var>'s <var>method</var> is unsafe and <var>forwardResponse</var>'s
+   <a for=response>status</a> begins with <code>2</code> or <code>3</code>, invalidate appropriate
+   stored responses in the HTTP cache, as per the 
+   "<a href=https://tools.ietf.org/html/rfc7234#section-4.4>Invalidation</a>" chapter of 
+   <cite>HTTP Caching</cite> [[!HTTP-CACHING]], and set <var>storedResponse</var> to null.
```
annevk: This step looks a little scary. 1 "unsafe" can change over time. Are we expecting implementations to change their result over time? If so we should probably reference "unsafe". I guess we also want tests for this.

Also, I think I'm mostly treating status as an integer, so a range check would make more sense, e.g., in the range 200 to 399, inclusive.

fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
+  <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is neither "<code>no-store</code>"
+  nor "<code>reload</code>", run these substeps:
 
   <ol>
    <li>
-    <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is
-    "<code>force-cache</code>" or "<code>only-if-cached</code>", then set
-    <var>response</var> to the <a for=/>response</a> in the HTTP cache for
-    <var>httpRequest</var>.
+    <p>Let <var>storedResponse</var> be the result of selecting a response from the HTTP cache,
```
annevk: I think it would be good to declare storedResponse upfront just like partial/revalidating since you end up using it outside these substeps.

annevk: And then here you just say "Set storedResponse" to the result of selecting a ..., if any." (meaning it keeps its initial value otherwise).

fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
+     "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Freshening Stored Responses upon Validation</a>"
+     chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
 
-   <li><p>Update <var>cachedResponse</var>'s <a for=response>header list</a>
-   using <var>response</var>'s <a for=response>header list</a>, as per the
-   "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Freshening Stored Responses upon Validation</a>"
-   chapter of <cite>HTTP Caching</cite>. [[!HTTP-CACHING]]
+     <li><p>Set <var>response</var> to <var>storedResponse</var>.
+
+     <p class="note">This updates the stored response in cache as well.
```
annevk: It seems this note is more fitting for the paragraph above? Also, the formatting is a little off. If you have multiple `<p>` descendants, they need to be on their own line and indented by one.

fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
-    <a for=response>status</a> which is most likely <code>200</code> now.
+    <ol>
+     <li><p>Update <var>storedResponse</var>'s <a for=response>header list</a> using
+     <var>forwardResponse</var>'s <a for=response>header list</a>, as per the 
+     "<a href=https://tools.ietf.org/html/rfc7234#section-3.3>Combining Partial Content</a>" chapter
+     of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+
+     <li><p>Set <var>response</var> to <var>storedResponse</var>.
+
+     <p class="note">This updates the stored response in cache as well.
```
annevk: Same here. It seems this note should be for one step earlier.

fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
+    </ol>
+    <!-- TODO: account for partial request failure -->
+
+    <li><p>If <var>response</var> is null, run these substeps:
```
annevk: `<p>` needs to be on a new line due to the sibling `<ol>`.

fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
+
+      <li><p>Set <var>response</var> to <var>forwardResponse</var>.
+
+      <li><p>Store <var>httpRequest</var> and <var>forwardResponse</var> in the HTTP cache, as per 
+      the "<a href=https://tools.ietf.org/html/rfc7234#section-3>Storing Responses in Caches</a>"
+      chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+
+      <p class=note>The cache key used to store the response must include the value of the
+      <var>credentials</var> flag, to ensure that a stored response to a request with credentials is
+      not used to satisfy a request without credentials.
```
annevk: Same comment as earlier, no requirements in notes. Also a formatting glitch here due to multiple `<p>` inside the `<li>`.

fetch.bs
mnot: Where is "if any"? Phrase ends with "or null if no stored response is selected."

fetch.bs
```diff
@@ -3342,86 +3345,149 @@ steps:
...
+    <var>httpRequest</var>'s <a for=request>header list</a>, as per the 
+    "<a href=https://tools.ietf.org/html/rfc7233#section-3.1>Range</a>" chapter of 
+    <cite>HTTP Range Requests</cite> [[!HTTP-RANGE]].
+
+    <li><p>If both the <var>revalidation</var> and <var>partial</var> flags are unset, set
+    <var>response</var> to <var>storedResponse</var>.
+  </ol>
 
  <li>
+  <!-- If response is still null, we require a forward request. -->
```
mnot: Yes. Should I factor that out, explain it, or?

fetch.bs
annevk: I meant that if you declare storedResponse upfront outside these substeps as "Let storedResponse be null" (to make it clearer what its scope is) you can rephrase this too.

fetch.bs
annevk: I meant to change from "forward" to "forwarded".

fetch.bs
mnot: Hm. I can reference the part of HTTP that defines the registry:
  http://httpwg.org/specs/rfc7231.html#method.registry
... or I can reference the registry directly:
  http://www.iana.org/assignments/http-methods/http-methods.xhtml

Preference?

annevk: I think http://httpwg.org/specs/rfc7231.html#safe.methods would be the most logical. I mostly meant that the term was not referenced. That it's variable is also somewhat problematic, but maybe okay.

mnot:
I think all feedback is addressed. Tests are next; planning to start with the direct requirements added here, then work on those incorporated by reference from the RFCs.

annevk:
That is great, while you work on tests, would you mind clicking the checkbox that enables me to push to your branch (should be in the sidebar)? I was mainly thinking of collapsing your commits into one and maybe adding one or more follow up cleanup commits for you to review.

mnot:
Box ticked.

annevk:
For some reason the box got unticked after I force pushed your collapsed changes with a revised commit message. Can you tick it again as I have a bunch of changes I wanted to push in a new commit? I'm guessing you created this branch well before GitHub introduced this feature, leading to this somewhat weird situation.

mnot:
Weird. Done.

annevk:
Thanks, fixup pushed.

annevk:
Curiously enough it seems again unchecked. I'll contact GitHub support about it now.

annevk:
(Sigh, if you want to troubleshoot the push access revocation issue, you need to contact GitHub yourself. I gave them a rather detailed explanation of what went down over several emails, but that didn't cut it...)

mnot:
Any preference on where to put the tests -- e.g., http/cache, fetch/cache, or? Assuming I'll be testing with fetch, not XHR.

mnot:
I think I've got a rough work list for tests:

* [ ] Do credentials form part of the cache key?
* [ ] Are If-None-Match and If-Modified-Since sent correctly?
* [ ] Are cached responses invalidated?
* [ ] Are response headers updated on a 304?
* [ ] Are stale responses used when there's a network error?
* [ ] Is partial content handled correctly, and If-Range, etc. generated?
* [ ] Are non-200 status codes cached?
* [ ] Is Vary handled correctly?
* [ ] Are Cache-Control request directives respected?
* [ ] Is the precedence of Expires and max-age handled correctly?
* [ ] Is heuristic caching done?
  * [ ] Is the 'public' response directive honoured?
* [ ] Is CC: no-store handled correctly?
* [ ] is no-cache handled correctly in requests and responses?

The rest seems to be pretty well covered y fetc/api/request-cache.js.

I've already written and stashed a few tests. Some notes:

### Invalidation

Chrome partially implements; it invalidates the cache upon a successful PUT or unrecognised method (e.g., FOO). Firefox and Safari TP do not.

### Credentials in the Cache Key

Firefox implements; Chrome and Safari TP appear not to. This could be because it's spec'd and tested based upon the credentials flag; they may be looking for the actual credentials. Will play with it a bit more.

### Using Stale Responses on a Network Error

This isn't implemented, AFAICT, although the responses generated by the framework are extremely stale, and that may be playing into it. It might be best to take this out of the spec for now; it's allowed but not mandated by HTTP.

mnot:
Fixed my tests for invalidation; support is much better than I thought.

Firefox and Safari TP invalidate everything correctly, but they over-invalidate (i.e., they do so even when the response is unsuccessful - 500 in this case).

Safari also shows a bit of variance when hitting 'reload', but I see that in other tests too.

Chrome only invalidates on POST and DELETE, regardless of whether the response is successful.

See:
  https://github.com/mnot/web-platform-tests/tree/http-cache

annevk:
fetch/cache or fetch/http-cache seems okay to me. I'd prefer fetch/ over http/ since some of the requirements are outlined here and we're testing the integration.

mnot:
I think my tests are done. Before I do a pull, one question: I have an additional page of tests for partial content (i.e., range requests), but as far as I can tell, they're not supported by any browser, except Safari TP -- and its support is extremely basic.

My inclination is to pull the specification of partial content support in the Fetch HTTP cache; it's an optional feature of HTTP caching, and I don't think we're going to convince other browsers to implement (unless someone suddenly gets VERY interested).

Thoughts? See also #144 (which I'll update in a moment).

annevk:
It seems reasonable to omit processing requirements, but we should at least have a note indicating we are aware of the feature and that it's simply not sufficiently supported or some such.

Having the tests somewhere seems good though in case browsers get interested at some point. Especially given the surprise you got from mcmanus.

mnot:
I think tests are ready (sorry for the last minute additions). I'm thinking we should remove the incorporation of credentials into the cache key from this pull, at least until #307 progresses. Make sense?

annevk:
I guess that's fair.

mnot:
Done.

mnot:
Firefox bugs:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1348701
- https://bugzilla.mozilla.org/show_bug.cgi?id=1348702
- https://bugzilla.mozilla.org/show_bug.cgi?id=1348703
- https://bugzilla.mozilla.org/show_bug.cgi?id=1348706
- https://bugzilla.mozilla.org/show_bug.cgi?id=1348707

mnot:
Chrome bugs:
- https://bugs.chromium.org/p/chromium/issues/detail?id=74981
- https://bugs.chromium.org/p/chromium/issues/detail?id=703008
- https://bugs.chromium.org/p/chromium/issues/detail?id=477567

mnot:
Edge bugs:
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10739540/
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11325681/
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11325686/
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11325891/

annevk:
Can you click the checkbox again so I can push to your branch?

I think this is all good now. Are the tests ready to land as well?

fetch.bs
```diff
@@ -3340,88 +3346,104 @@ steps:
...
- <a for=request>header list</a> with resume
- <a>headers</a>.
- <!-- XXX xref partial, modify, resume headers -->
+    <p class=note>As mandated by HTTP, this still takes the `<code>Vary</code>` <a>header</a>
+    into account.
 
- <li>
-  <p>If <var>response</var> is null, run these substeps:
+   <li>
+    <p>If the <var>revalidatingFlag</var> is set, then:
```
yoavweiss: If I'm reading this correctly, nothing assigns `storedResponse` to `response` if revalidation is not needed, and the cache-mode is not `force-cache` or `only-if-cached`.
Am I reading this correctly? If so, is this intentional? I'd expect such assignment to happen if no revalidation is needed.

mnot: Ah, thanks - that got acccidentally removed in 2ae9df076. Will fix.

mnot:
Checkbox clicked. Yes, I think this is ready, and so are tests; doubtless there will be more of each, but it's good for now.

annevk:
🍸

youennf:
That is a lot of test :) 
Is WebKit all fine?

mnot:
AIUI Webkit doesn't do the HTTP caching layer -- or has this changed? I tested Safari TP, see here:
  https://www.mnot.net/blog/2017/03/16/browser-caching
In a nutshell, pretty well, a few places to look. Can file bugs if you point me to the right place to do it.

youennf:
WebKit has a bunch of HTTP caches, most notably https://trac.webkit.org/raw-attachment/wiki/March%202015%20Meeting/Disk%20cache.pdf and memory cache.

This is interesting even if it is related to the underlying HTTP library.
https://bugs.webkit.org is the usual place for feedback.

annevk:
@mnot https://bugs.webkit.org/enter_bug.cgi?product=WebKit in particular (random component, prefix subject with "HTTP cache: ").

mnot:
Safari TP bugs:
* https://bugs.webkit.org/show_bug.cgi?id=170714
* https://bugs.webkit.org/show_bug.cgi?id=170715
* https://bugs.webkit.org/show_bug.cgi?id=170716

PR #499, opened 2017-03-07
Do not always set the recursive flag in HTTP-redirect fetch

annevk:
Also keep “includes credentials” steps next to each other by moving
another step and update origin terminology.

Fixes https://github.com/whatwg/html/issues/2396.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/http-redirect-fetch/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/3b4cb54...c958914.html)


annevk:
@jdm ping!

PR #501, opened 2017-03-07
Make combine work for XMLHttpRequest and WebSocket

annevk:
Unfortunately as established in https://github.com/whatwg/xhr/issues/108 setRequestHeader() uses `, ` whereas fetch() uses `,` as value separator. This introduces a legacySpaceFlag for combine that XMLHttpRequest and WebSocket can use. New code and CORS (in Access-Control-Request-Headers) can continue not passing this flag.

Tests were fixed in https://github.com/w3c/web-platform-tests/pull/5008.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/legacy-space/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/ba175cf...4dc529a.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -315,20 +315,24 @@ all <a>headers</a> whose <a for=header>name</a> is a <a>byte-case-insensitive</a
  <a for=header>value</a> is <var>value</var> to <var>list</var>.
 </ol>
 
-<p>To <dfn export id=concept-header-list-combine>combine</dfn> a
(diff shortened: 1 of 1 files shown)


annevk:
Closing this in favor of #504. We're going to try to get unified behavior between Fetch and XMLHttpRequest, even if that's currently not the case, favoring XMLHttpRequest over Fetch due to its long legacy.

PR #502, opened 2017-03-07
Make combined value and sort and combine work for XMLHttpRequest

annevk:
See https://github.com/whatwg/xhr/issues/109.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/combined-value/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/ba175cf...5a20d7d.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -331,9 +331,9 @@ all <a>headers</a> whose <a for=header>name</a> is a <a>byte-case-insensitive</a
 <p class="note no-backref"><a>Combine</a> is used by {{XMLHttpRequest}} and the
 <a lt="establish a WebSocket connection" for=websocket>WebSocket protocol handshake</a>.
 
-<p>To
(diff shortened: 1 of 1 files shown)


youennf:
It is probably better to keep fetch and XHR consistent here as WebKit (and probably other browsers) might share the same code path for header handling

annevk:
@wanderview what do you think about changing the Fetch API implementation in Gecko to use `, ` rather than `,` as a value separator in headers? Thereby making it consistent with XMLHttpRequest?

@tyoshino @yutakahirano can you answer the same question for Chrome?

Thanks!

annevk:
@wanderview seemed okay with making this change.

@youennf could you perhaps help with identifying web-platform-tests that will need to change for this? I'll try to update PRs tomorrow.

annevk:
#504 is the new PR for this.

PR #504, opened 2017-03-10
Change combine and combined value to use 0x2C 0x20 

annevk:
Instead of just 0x2C, use 0x2C 0x20 somewhat consistently (except where we can't and point it out) as that is what XMLHttpRequest implementations have always done and nobody likes too much change.

Fixes the Fetch part of https://github.com/whatwg/xhr/issues/108 and https://github.com/whatwg/xhr/issues/109.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/combine/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/89f7361...e3978ef.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -315,20 +315,20 @@ all <a>headers</a> whose <a for=header>name</a> is a <a>byte-case-insensitive</a
  <a for=header>value</a> is <var>value</var> to <var>list</var>.
 </ol>
 
-<p>To <dfn export id=concept-header-list-combine>combine</dfn> a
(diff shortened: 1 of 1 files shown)


annevk:
Tests: https://github.com/w3c/web-platform-tests/issues/5115.

(There have also been tests modified on the XMLHttpRequest side. Those already landed.)

annevk:
Bugs filed are referenced in https://github.com/whatwg/xhr/issues/108.

PR #510, opened 2017-03-20
Editorial: clarify that an internal response cannot be filtered

### LINKED ISSUE #386: Can an internal response be another filtered response?
Relevant spec: https://fetch.spec.whatwg.org/#concept-filtered-response

Is it possible that the internal response of a filtered response is yet another filtered response? If so, how many layers of filtered responses are there?

annevk:
Fixes #386.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/internal-response/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/b59c72b...36bfd7b.html)


KiChjang:
Changes LGTM.

PR #511, opened 2017-03-20
Clarify requirements on a CORS server

### LINKED ISSUE #172: Status code to use in a response indicating rejection of a cross-origin access
**tl;dr:** see the title.

**Long and trivial version:**

The CORS protocol can work independently from the response representation. I.e. attaching/omitting the `Access-Control-Allow-Origin` header can instruct UAs whether or not to allow the script to see the response without making any change on the other headers and status-line.

So, to define the CORS algorith, technically, we don't have to define any dedicated HTTP status code for representing rejection due to a bad Origin value. Actually, the current spec doesn't say anything (neither requirement nor recommendation) about a status code to include in a response representing cross-origin access rejection.

Does this design imply that the server should:
- serve the contents anyway
- but decide whether or not to include `Allow-Control-Allow-Origin` to instruct the UA how to handle the response in CORS enabled case
- and then delegate the final decision to the UA. It's possible that the UA doesn't use the `Allow-Control-Allow-Origin` in case the request was issued in the `no-CORS` mode.
  - this (the server sees Origin header but the UA is in `no-CORS` mode) happens since XHR and Fetch works with omit-Origin-header-flag unset.

Regardless the answer is yes or no, in practice, developers want to reject unexpected cross origin requests at minimum cost. We basically want to keep the connection alive, so disconnect is not the best idea. Then, we generate a response w/o `Allow-Control-Allow-Origin`, body. I.e. the response no longer represent any valid response from the application layer. What status code should we put in the response? Maybe 403 is appropriate. But I wonder if there any discussion about this topic has happened in past.

annevk:
Fixes #172.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/cors-server/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/b59c72b...346e45d.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1925,6 +1925,13 @@ if the <a>CORS protocol</a> is understood. It uses `<code>OPTIONS</code>` as
  of the response by listing their <a lt=name for=header>names</a>.
 </dl>
 
+<hr>
+
+<p>In case a server does not wish to participate in the <a>CORS protocol</a> its HTTP response to
+the <a lt="CORS request">CORS</a> or <a>CORS-preflight request</a> must not include any of the above
+<a>headers</a>. The server is encouraged to use the <code>403</code> <a for=/>status</a> in such
+HTTP responses.
+
 
 <h4 id=http-new-header-syntax>HTTP new-header syntax</h4>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -1927,7 +1927,7 @@ if the <a>CORS protocol</a> is understood. It uses `<code>OPTIONS</code>` as
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1925,6 +1925,13 @@ if the <a>CORS protocol</a> is understood. It uses `<code>OPTIONS</code>` as
  of the response by listing their <a lt=name for=header>names</a>.
 </dl>
 
+<hr>
+
+<p>In case a server does not wish to participate in the <a>CORS protocol</a> its HTTP response to
```
tyoshino: comma before "its"?

PR #523, opened 2017-04-09
Aborting fetch

jakearchibald:
One day this will be complete and I will retire.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/cancelation/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/a4290d1...8e1d2d5.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1189,37 +1189,44 @@ or "<code>worker</code>".
   <p><a>In parallel</a>, while true:
 
   <ol>
-   <li><p>Wait for <var>read</var> to be fulfilled or rejected.
-
    <li>
-    <p>If <var>read</var> is fulfilled with an object whose <code>done</code> property is false
-    and whose <code>value</code> property is a <code>Uint8Array</code> object, then run these steps:
+    <p>Run these steps, but abort if the ongoing fetch is <a for=fetch>terminated</a>:</p>
 
     <ol>
-     <li><p>Let <var>bs</var> be the <a>byte sequence</a> represented by the <code>Uint8Array</code>
-     object.
+     <li><p>Wait for <var>read</var> to be fulfilled or rejected.
 
      <li>
-      <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
-      <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
-      <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
-      for <var>request</var>.
+      <p>If <var>read</var> is fulfilled with an object whose <code>done</code> property is false
+      and whose <code>value</code> property is a <code>Uint8Array</code> object, then run these
+      steps:
 
-      <p class="note no-backref">This step blocks until <var>bs</var> is fully transmitted.
+      <ol>
+       <li><p>Let <var>bs</var> be the <a>byte sequence</a> represented by the
+       <code>Uint8Array</code> object.
 
-     <li><p>Set <var>read</var> to the result of <a lt="read a chunk" for=ReadableStream>reading a
-     chunk</a> from <var>body</var>'s <a for=body>stream</a> with <var>reader</var>.
+       <li>
+        <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
+        <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
+        <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
+        for <var>request</var>.
+
+        <p class="note no-backref">This step blocks until <var>bs</var> is fully transmitted.
+
+       <li><p>Set <var>read</var> to the result of <a lt="read a chunk" for=ReadableStream>reading a
+       chunk</a> from <var>body</var>'s <a for=body>stream</a> with <var>reader</var>.
+      </ol>
+
+     <li><p>Otherwise, if <var>read</var> is fulfilled with an object whose <code>done</code>
+     property is true, then <a>queue a fetch task</a> on <var>request</var> to
+     <a>process request end-of-body</a> for <var>request</var> and abort these in-parallel steps.
+
+     <li><p>Otherwise, <a lt=terminated for=fetch>terminate</a> the ongoing fetch with reason
+     <i>fatal</i>.
     </ol>
 
-   <li><p>Otherwise, if <var>read</var> is fulfilled with an object whose <code>done</code> property
-   is true, then <a>queue a fetch task</a> on <var>request</var> to
-   <a>process request end-of-body</a> for <var>request</var> and abort these steps.
-
-   <li><p>Otherwise, <a lt=terminated for=fetch>terminate</a> the ongoing fetch with reason
-   <i>fatal</i> and abort these steps.
+   <li><p>If the ongoing fetch is <a for=fetch>terminated</a>, then abort these in-parallel steps.
   </ol>
 </ol>
-
 
 <h4 id=responses>Responses</h4>
 
@@ -1488,21 +1495,38 @@ for each associated <a for="fetch group">fetch record</a> whose
 <var>origin</var> and <var>credentials</var>, run these steps:
 
 <ol>
- <li><p>If <a for=connection>connection pool</a> contains a
- <a>connection</a> whose <b>origin</b> is <var>origin</var> and
- <b>credentials</b> is <var>credentials</var>, return that
+ <li><p>If <a for=connection>connection pool</a> contains a <a>connection</a> whose <b>origin</b> is
+ <var>origin</var> and <b>credentials</b> is <var>credentials</var>, then return that
  <a>connection</a>.
 
+ <li><p>Let <var>connection</var> be null.
+
  <li>
-  <p>Let <var>connection</var> be the result of establishing an HTTP connection to
-  <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
+  <p>Run these steps, but abort if the ongoing fetch is <a for=fetch>terminated</a>:</p>
 
-  <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.
+  <ol>
+   <li>
+    <p>Set <var>connection</var> to the result of establishing an HTTP connection to
+    <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+    [[!TLS]]
 
-  <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), return failure.
+    <p>If <var>credentials</var> is false, then do <em>not</em> send a TLS client certificate.
 
- <li><p>Add <var>connection</var> to the <a for=connection>connection pool</a>
- with <b>origin</b> being <var>origin</var> and <b>credentials</b> being <var>credentials</var>.
+    <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), then return
+    failure.
+  </ol>
+
+ <li>
+  <p>If the ongoing fetch is <a for=fetch>terminated</a>, then:
+
+  <ol>
+   <li><p>If <var>connection</var> is not null, close <var>connection</var>.
+
+   <li><p>Return failure.
+  </ol>
+
+ <li><p>Add <var>connection</var> to the <a for=connection>connection pool</a> with <b>origin</b>
+ being <var>origin</var> and <b>credentials</b> being <var>credentials</var>.
 
  <li><p>Return <var>connection</var>.
 </ol>
@@ -1776,11 +1800,8 @@ from a {{ReadableStream}} object with <var>reader</var>, run these steps:
 </ol>
 
 <p>To <dfn export for=ReadableStream id=concept-cancel-readablestream>cancel</dfn> a
-{{ReadableStream}} object with <var>reader</var> and <var>reason</var>, return the result of calling
-<a abstract-op>ReadableStreamCancel</a>(<var>reader</var>, <var>reason</var>).
-
-<p class="note no-backref">Because the reader grants exclusive access, the actual mechanism of how
-to read cannot be observed. Implementations could use more direct mechanism if convenient.
+{{ReadableStream}} object <var>stream</var> with <var>reason</var>, return the result of calling
+<a abstract-op>ReadableStreamCancel</a>(<var>stream</var>, <var>reason</var>).
 
 <p>To <dfn export for=ReadableStream id=concept-tee-readablestream>tee</dfn> a {{ReadableStream}}
 object <var>stream</var>, run these steps:
@@ -2295,126 +2316,135 @@ the request.
 [[!HTTP-CACHING]]
 
 <ol>
- <li><p>If <var>request</var>'s <a for=request>window</a> is
- "<code>client</code>", set <var>request</var>'s
- <a for=request>window</a> to <var>request</var>'s
- <a for=request>client</a>, if <var>request</var>'s
- <a for=request>client</a>'s
- <a for="environment settings object">global object</a> is a
- {{Window}} object, and to "<code>no-window</code>"
- otherwise.
-
- <li><p>If <var>request</var>'s <a for=request>origin</a> is
- "<code>client</code>", set <var>request</var>'s
- <a for=request>origin</a> to  <var>request</var>'s
- <a for=request>client</a>'s <a for="environment settings object">origin</a>.
-
  <li>
-  <p>If <var>request</var>'s <a for=request>header list</a>
-  <a for="header list">does not contain</a> `<code>Accept</code>`, then:
+  <p>Run these steps, but abort if the ongoing fetch is <a for=fetch>terminated</a>:</p>
 
   <ol>
-   <li><p>Let <var>value</var> be `<code>*/*</code>`.
+   <li><p>If <var>request</var>'s <a for=request>window</a> is
+   "<code>client</code>", set <var>request</var>'s
+   <a for=request>window</a> to <var>request</var>'s
+   <a for=request>client</a>, if <var>request</var>'s
+   <a for=request>client</a>'s
+   <a for="environment settings object">global object</a> is a
+   {{Window}} object, and to "<code>no-window</code>"
+   otherwise.
 
-   <li><p>If <var>request</var> is a <a>navigation request</a>, a user agent should set
-   <var>value</var> to
-   `<code>text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8</code>`.
+   <li><p>If <var>request</var>'s <a for=request>origin</a> is
+   "<code>client</code>", set <var>request</var>'s
+   <a for=request>origin</a> to  <var>request</var>'s
+   <a for=request>client</a>'s <a for="environment settings object">origin</a>.
 
    <li>
-    <p>Otherwise, a user agent should set <var>value</var> to the first matching statement, if any,
-    switching on <var>request</var>'s <a for=request>destination</a>:
-    <!-- https://github.com/whatwg/fetch/issues/43#issuecomment-97909717 -->
+    <p>If <var>request</var>'s <a for=request>header list</a>
+    <a for="header list">does not contain</a> `<code>Accept</code>`, then:
 
-    <dl class=switch>
-     <dt>"<code>image</code>"
-     <dd>`<code>image/png,image/svg+xml,image/*;q=0.8,*/*;q=0.5</code>`
+    <ol>
+     <li><p>Let <var>value</var> be `<code>*/*</code>`.
 
-     <dt>"<code>style</code>"
-     <dd>`<code>text/css,*/*;q=0.1</code>`
-    </dl>
+     <li><p>If <var>request</var> is a <a>navigation request</a>, a user agent should set
+     <var>value</var> to
+     `<code>text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8</code>`.
 
-   <li><p><a for="header list">Append</a>
-   `<code>Accept</code>`/<var>value</var> to <var>request</var>'s
-   <a for=request>header list</a>.
-  </ol>
+     <li>
+      <p>Otherwise, a user agent should set <var>value</var> to the first matching statement, if
+      any, switching on <var>request</var>'s <a for=request>destination</a>:
+      <!-- https://github.com/whatwg/fetch/issues/43#issuecomment-97909717 -->
 
- <li><p>If <var>request</var>'s <a for=request>header list</a>
- <a for="header list">does not contain</a> `<code>Accept-Language</code>`, user agents should
- <a for="header list">append</a>
- `<code>Accept-Language</code>`/an appropriate <a for=header>value</a>
- to <var>request</var>'s <a for=request>header list</a>.
+      <dl class=switch>
+       <dt>"<code>image</code>"
+       <dd>`<code>image/png,image/svg+xml,image/*;q=0.8,*/*;q=0.5</code>`
 
- <li>
-  <p>If <var>request</var>'s <a for=request>priority</a> is null, then use <var>request</var>'s
-  <a for=request>initiator</a> and <a for=request>destination</a> appropriately in setting
-  <var>request</var>'s <a for=request>priority</a> to a user-agent-defined object.
+       <dt>"<code>style</code>"
+       <dd>`<code>text/css,*/*;q=0.1</code>`
+      </dl>
 
-  <p class=note>The user-agent-defined object could encompass stream weight and dependency
-  for HTTP/2, and equivalent information used to prioritize dispatch and processing of
-  HTTP/1 fetches.
+     <li><p><a for="header list">Append</a>
+     `<code>Accept</code>`/<var>value</var> to <var>request</var>'s
+     <a for=request>header list</a>.
+    </ol>
 
- <li>
-  <p>If <var>request</var> is a <a>navigation request</a>, a user agent should, for each
-  <a for=/>header</a> <a for=header>name</a> (<var>hintName</var>) in the first column of the
-  following table, if <var>request</var>'s <a for=request>header list</a>
-  <a for="header list">does not contain</a> <var>hintName</var>, then
-  <a for="header list">append</a>
-  <var>hintName</var>/the value given in the same row on the second column, to <var>request</var>'s
-  <a for=request>header list</a>.
+   <li><p>If <var>request</var>'s <a for=request>header list</a>
+   <a for="header list">does not contain</a> `<code>Accept-Language</code>`, user agents should
+   <a for="header list">append</a>
+   `<code>Accept-Language</code>`/an appropriate <a for=header>value</a>
+   to <var>request</var>'s <a for=request>header list</a>.
 
-  <table>
-   <tbody><tr>
-    <th><a for=header>Name</a>
-    <th><a for=header>Value</a>
-   <tr>
-    <td>`<code>dpr</code>`
-    <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
-   <tr>
-    <td>`<code>save-data</code>`
-    <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
-   <tr>
-    <td>`<code>viewport-width</code>`
-    <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
-  </table>
-
- <li>
-  <p>If <var>request</var> is a <a>subresource request</a>, run these substeps:
-
-  <ol>
    <li>
-    <p>If <var>request</var>'s <a for=client>client hints list</a> is not empty, then run these
-    substeps for each <var>hintName</var> in the list:
+    <p>If <var>request</var>'s <a for=request>priority</a> is null, then use <var>request</var>'s
+    <a for=request>initiator</a> and <a for=request>destination</a> appropriately in setting
+    <var>request</var>'s <a for=request>priority</a> to a user-agent-defined object.
+
+    <p class=note>The user-agent-defined object could encompass stream weight and dependency
+    for HTTP/2, and equivalent information used to prioritize dispatch and processing of
+    HTTP/1 fetches.
+
+   <li>
+    <p>If <var>request</var> is a <a>navigation request</a>, a user agent should, for each
+    <a for=/>header</a> <a for=header>name</a> (<var>hintName</var>) in the first column of the
+    following table, if <var>request</var>'s <a for=request>header list</a>
+    <a for="header list">does not contain</a> <var>hintName</var>, then
+    <a for="header list">append</a>
+    <var>hintName</var>/the value given in the same row on the second column, to
+    <var>request</var>'s <a for=request>header list</a>.
+
+    <table>
+     <tbody><tr>
+      <th><a for=header>Name</a>
+      <th><a for=header>Value</a>
+     <tr>
+      <td>`<code>dpr</code>`
+      <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
+     <tr>
+      <td>`<code>save-data</code>`
+      <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
+     <tr>
+      <td>`<code>viewport-width</code>`
+      <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+    </table>
+
+   <li>
+    <p>If <var>request</var> is a <a>subresource request</a>, run these substeps:
 
     <ol>
      <li>
-      <p>Set <var>value</var> to the first matching statement, if any, switching on
-      <var>hintName</var>:
+      <p>If <var>request</var>'s <a for=client>client hints list</a> is not empty, then run these
+      substeps for each <var>hintName</var> in the list:
 
-      <dl class=switch>
-       <dt>"<code>dpr</code>"
-       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
-       <dt>"<code>save-data</code>"
-       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
-       <dt>"<code>viewport-width</code>"
-       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
-       <dt>"<code>width</code>"
-       <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#width>width value</a>
-      </dl>
+      <ol>
+       <li>
+        <p>Set <var>value</var> to the first matching statement, if any, switching on
+        <var>hintName</var>:
 
-     <li><p><a for="header list">Append</a> <var>hintName</var>/<var>value</var> to
-     <var>request</var>'s <a for=request>header list</a>.
+        <dl class=switch>
+         <dt>"<code>dpr</code>"
+         <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#dpr>dpr value</a>
+         <dt>"<code>save-data</code>"
+         <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#save-data>save-data value</a>
+         <dt>"<code>viewport-width</code>"
+         <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
+         <dt>"<code>width</code>"
+         <dd>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#width>width value</a>
+        </dl>
+
+       <li><p><a for="header list">Append</a> <var>hintName</var>/<var>value</var> to
+       <var>request</var>'s <a for=request>header list</a>.
+      </ol>
+
+     <li><p>Let <var>record</var> be a new
+     <a for="fetch group">fetch record</a> consisting of
+     <var>request</var> and this instance of the
+     <a for=/>fetch</a> algorithm.
+
+     <li><p>Append <var>record</var> to <var>request</var>'s
+     <a for=request>client</a>'s
+     <a for=fetch>fetch group</a> list of
+     <a for="fetch group">fetch records</a>.
     </ol>
-
-   <li><p>Let <var>record</var> be a new
-   <a for="fetch group">fetch record</a> consisting of
-   <var>request</var> and this instance of the
-   <a for=/>fetch</a> algorithm.
-
-   <li><p>Append <var>record</var> to <var>request</var>'s
-   <a for=request>client</a>'s
-   <a for=fetch>fetch group</a> list of
-   <a for="fetch group">fetch records</a>.
   </ol>
+
+ <li><p>If the ongoing fetch is <a for=fetch>terminated</a>, then let <var>reason</var> be the
+ provided termination reason, and return a <a>network error</a> with
+ <a for=response>termination reason</a> set to <var>reason</var>.
 
  <li><p>Return the result of performing a <a for=main>main fetch</a>
  using <var>request</var>.
@@ -2432,81 +2462,91 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
 <ol>
  <li><p>Let <var>response</var> be null.
 
- <li><p>If <var>request</var>'s <a>local-URLs-only flag</a> is set and
- <var>request</var>'s <a for=request>current url</a> is
- not <a lt="is local">local</a>, set
- <var>response</var> to a <a>network error</a>.
-
- <li><p>Execute <a href=https://w3c.github.io/webappsec-csp/#report-for-request>Report Content Security Policy violations for <var>request</var></a>.
- [[!CSP]]
-
- <li><p><a href=https://w3c.github.io/webappsec-upgrade-insecure-requests/#upgrade-request>Upgrade <var>request</var> to a potentially secure URL, if appropriate</a>.
- [[!UPGRADE]]
-
- <li><p>If
- <a lt="block bad port">should fetching <var>request</var> be blocked due to a bad port</a>,
- <a href=https://w3c.github.io/webappsec-mixed-content/#should-block-fetch>should fetching <var>request</var> be blocked as mixed content</a>,
- or
- <a href=https://w3c.github.io/webappsec-csp/#should-block-request>should fetching <var>request</var> be blocked by Content Security Policy</a>
- returns <b>blocked</b>, set <var>response</var> to a
- <a>network error</a>.
- [[!MIX]]
- [[!CSP]]
-
- <li><p>If <var>request</var>'s <a for=request>referrer policy</a> is the empty string and
- <var>request</var>'s <a for=request>client</a> is non-null, then set <var>request</var>'s
- <a for=request>referrer policy</a> to <var>request</var>'s <a for=request>client</a>'s
- <a for="environment settings object">referrer policy</a>.
- [[!REFERRER]]
-
  <li>
-  <p>If <var>request</var>'s <a for=request>referrer policy</a>
-  is the empty string, then set <var>request</var>'s
-  <a for=request>referrer policy</a> to
-  "<code>no-referrer-when-downgrade</code>".
+  <p>Run these steps, but abort if the ongoing fetch is <a for=fetch>terminated</a>:</p>
 
-  <p class="note no-backref">We use "<code>no-referrer-when-downgrade</code>" because it is the
-  historical default.
+  <ol>
 
- <li>
-  <p>If <var>request</var>'s <a for=request>referrer</a>
-  is not "<code>no-referrer</code>", set <var>request</var>'s
-  <a for=request>referrer</a> to the result of invoking
-  <a>determine <var>request</var>'s referrer</a>.
-  [[!REFERRER]]
+   <li><p>If <var>request</var>'s <a>local-URLs-only flag</a> is set and
+   <var>request</var>'s <a for=request>current url</a> is
+   not <a lt="is local">local</a>, set
+   <var>response</var> to a <a>network error</a>.
 
-  <p class="note no-backref">As stated in <cite>Referrer Policy</cite>, user agents can
-  provide the end user with options to override <var>request</var>'s
-  <a for=request>referrer</a> to "<code>no-referrer</code>" or
-  have it expose less sensitive information.
+   <li><p>Execute <a href=https://w3c.github.io/webappsec-csp/#report-for-request>Report Content Security Policy violations for <var>request</var></a>.
+    [[!CSP]]
 
- <li>
-  <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
-  "<code>ftp</code>", <var>request</var>'s <a for=request>client</a>'s
-  <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>", and
-  <var>request</var>'s <a for=request>reserved client</a> is either null or an <a>environment</a>
-  whose <a for=environment>target browsing context</a> is a <a>nested browsing context</a>, then set
-  <var>response</var> to a <a>network error</a>.
+   <li><p><a href=https://w3c.github.io/webappsec-upgrade-insecure-requests/#upgrade-request>Upgrade <var>request</var> to a potentially secure URL, if appropriate</a>.
+   [[!UPGRADE]]
 
- <li>
-  <p>Set <var>request</var>'s <a for=request>current url</a>'s
-  <a for=url>scheme</a> to "<code>https</code>" if
-  all of the following conditions are true:
+   <li><p>If
+   <a lt="block bad port">should fetching <var>request</var> be blocked due to a bad port</a>,
+   <a href=https://w3c.github.io/webappsec-mixed-content/#should-block-fetch>should fetching <var>request</var> be blocked as mixed content</a>,
+   or
+   <a href=https://w3c.github.io/webappsec-csp/#should-block-request>should fetching <var>request</var> be blocked by Content Security Policy</a>
+   returns <b>blocked</b>, set <var>response</var> to a
+   <a>network error</a>.
+   [[!MIX]]
+   [[!CSP]]
 
-  <ul class=brief>
-   <li><var>request</var>'s <a for=request>current url</a>'s
-   <a for=url>scheme</a> is "<code>http</code>"
-   <li><var>request</var>'s <a for=request>current url</a>'s
-   <a for=url>host</a> is a
-   <a for=/>domain</a>
-   <li>Matching <var>request</var>'s <a for=request>current url</a>'s
-   <a for=url>host</a> per
-   <a href=https://tools.ietf.org/html/rfc6797#section-8.2>Known HSTS Host Domain Name Matching</a>
-   results in either a superdomain match with an asserted <code>includeSubDomains</code>
-   directive or a congruent match (with or without an asserted <code>includeSubDomains</code>
-   directive) [[!HSTS]]
-  </ul>
- <!-- Per Mike West HSTS happens "probably after" Referrer -->
+   <li><p>If <var>request</var>'s <a for=request>referrer policy</a> is the empty string and
+   <var>request</var>'s <a for=request>client</a> is non-null, then set <var>request</var>'s
+   <a for=request>referrer policy</a> to <var>request</var>'s <a for=request>client</a>'s
+   <a for="environment settings object">referrer policy</a>.
+   [[!REFERRER]]
+
+   <li>
+    <p>If <var>request</var>'s <a for=request>referrer policy</a>
+    is the empty string, then set <var>request</var>'s
+    <a for=request>referrer policy</a> to
+    "<code>no-referrer-when-downgrade</code>".
+
+    <p class="note no-backref">We use "<code>no-referrer-when-downgrade</code>" because it is the
+    historical default.
+
+   <li>
+    <p>If <var>request</var>'s <a for=request>referrer</a>
+    is not "<code>no-referrer</code>", set <var>request</var>'s
+    <a for=request>referrer</a> to the result of invoking
+    <a>determine <var>request</var>'s referrer</a>.
+    [[!REFERRER]]
+
+    <p class="note no-backref">As stated in <cite>Referrer Policy</cite>, user agents can
+    provide the end user with options to override <var>request</var>'s
+    <a for=request>referrer</a> to "<code>no-referrer</code>" or
+    have it expose less sensitive information.
+
+   <li>
+    <p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
+    "<code>ftp</code>", <var>request</var>'s <a for=request>client</a>'s
+    <a for=environment>creation URL</a>'s <a for=url>scheme</a> is not "<code>ftp</code>", and
+    <var>request</var>'s <a for=request>reserved client</a> is either null or an <a>environment</a>
+    whose <a for=environment>target browsing context</a> is a <a>nested browsing context</a>, then
+    set <var>response</var> to a <a>network error</a>.
+
+   <li>
+    <p>Set <var>request</var>'s <a for=request>current url</a>'s
+    <a for=url>scheme</a> to "<code>https</code>" if
+    all of the following conditions are true:
+
+    <ul class=brief>
+     <li><var>request</var>'s <a for=request>current url</a>'s
+     <a for=url>scheme</a> is "<code>http</code>"
+     <li><var>request</var>'s <a for=request>current url</a>'s
+     <a for=url>host</a> is a
+     <a for=/>domain</a>
+     <li>Matching <var>request</var>'s <a for=request>current url</a>'s
+     <a for=url>host</a> per
+     <a href=https://tools.ietf.org/html/rfc6797#section-8.2>Known HSTS Host Domain Name
+     Matching</a> results in either a superdomain match with an asserted
+     <code>includeSubDomains</code> directive or a congruent match (with or without an asserted
+     <code>includeSubDomains</code> directive) [[!HSTS]]
+    </ul>
+   <!-- Per Mike West HSTS happens "probably after" Referrer -->
+  </ol>
+
+  <li><p>If the ongoing fetch is <a for=fetch>terminated</a>, then let <var>reason</var> be the
+  provided termination reason, and return a <a>network error</a> with
+  <a for=response>termination reason</a> set to <var>reason</var>.
 
  <li><p>If <var>request</var>'s <a>synchronous flag</a> is unset and
  <i>recursive flag</i> is unset, run the remaining steps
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4683,6 +4705,7 @@ dictionary RequestInit {
   DOMString integrity;
   boolean keepalive;
   any window; // can only be set to null
+  CancelationSignal signal;
```
jakearchibald: This is `CancelationSignal` rather than `FetchSignal`, as it'd be nice to allow the generic signal here. I'm guessing I'll be able to do something like "If signal is a `FetchSignal`…" later to handle things like priority changes.

fetch.bs
```diff
@@ -5343,6 +5377,13 @@ method, must run these steps:
  {{Headers}} object whose <a for=Headers>guard</a> is
  "<code>immutable</code>".
 
+ <li><p>Add the following steps to <var>requestObject</var>'s <a for=Request>signal</a>'s
+ <a>cancelation callbacks</a>:
```
jakearchibald: No idea if this is an acceptable way to add a sub-algorithm to a list of algorithms.

fetch.bs
```diff
@@ -5343,6 +5377,13 @@ method, must run these steps:
...
+
+ <ol>
+  <li><a lt=terminated for=fetch>Terminate the request</a>
```
jakearchibald: This seems really hand-wavey (what is "the request"?) but it's how XHR does it.

fetch.bs
annevk: Yeah, that would work.

fetch.bs
annevk: Something like that is how I'd expect it to work. "cancelation callbacks" should probably have for=signal or some such, but I suspect we'll end up renaming these terms a bit before it's all done.

fetch.bs
annevk: Yeah, I was hoping we could fix that a bit as part of doing this, but I don't have great suggestions as to how.

fetch.bs
```diff
@@ -4363,6 +4363,28 @@ method, when invoked, must run these steps:
...
+
+<h3 id=fetch-controller>Fetch controller</h3>
+
+<pre class=idl>
+[Constructor(), Exposed=(Window,Worker)]
+interface FetchController : CancelationController {
+  [SameObject] readonly attribute FetchSignal signal;
+  // Nothing additional exposed here yet
+};
+</pre>
```
annevk: I suggest we remove these two interfaces if we're not actually going to use them. We can add them when we add functionality that requires them.

fetch.bs
```diff
@@ -4705,6 +4728,9 @@ omitted from <a enum><code>RequestMode</code></a> as it cannot be used nor obser
 is itself associated with <a for=Request>request</a>'s
 <a for=request>header list</a>.
 
+<p>A {{Request}} object has an associated <dfn for=Request>signal</dfn> (a {{CancelationSignal}}
+object), which is initially null.
```
annevk: Needs to say "null or a X object" if you make it null.

fetch.bs
```diff
@@ -4924,10 +4954,14 @@ constructor must run these steps:
    to <var>method</var>.
   </ol>
 
+ <li><p>If <var>init</var>'s <code>signal</code> member is present, let <var>signal</var> be it.
```
annevk: Cannot use let twice for the same variable.

fetch.bs
```diff
@@ -4924,10 +4954,14 @@ constructor must run these steps:
...
+
  <li><p>Let <var>r</var> be a new {{Request}} object associated with <var>request</var> and
  a new associated {{Headers}} object whose <a for=Headers>guard</a>
  is "<code>request</code>".
 
+ <li><p>Set <var>r</var>'s <a for=Request>signal</a> to <var>signal</var>.
```
annevk: Reading your additions it seems you could maybe set _r_'s signal directly each time? Or is _r_ not set up yet?

fetch.bs
```diff
@@ -5329,11 +5363,11 @@ method, must run these steps:
...
- <li><p>Let <var>request</var> be the associated
- <a for=/>request</a> of the result of invoking the initial value of
- {{Request}} as constructor with <var>input</var> and
- <var>init</var> as arguments. If this throws an exception, reject
- <var>p</var> with it and return <var>p</var>.
+ <li><p>Let <var>requestObject</var> be the result of invoking the initial value of {{Request}} as
+ constructor with <var>input</var> and <var>init</var> as arguments. If this throws an exception,
+ reject <var>p</var> with it and return <var>p</var>
+
+ <li><p>Let <var>request</var> be <var>requestObject</var>'s <a for=Request>request</a>.
```
annevk: This change is just editorial, right?

fetch.bs
jakearchibald: The signal that appears on request objects will at some point be a `FetchSignal`. Is it fine for it to be a superclass of that initially?

fetch.bs
jakearchibald: It introduces *requestObject* as a variable, so I can get its signal in the proceeding steps.

fetch.bs
annevk: Well, if you pass in the superclass or a differently subclass, presumably we'd reflect that instance directly on the constructed `Request` object and not some copy. And therefore that would have to specify the superclass.

fetch.bs
```diff
@@ -5343,10 +5402,22 @@ method, must run these steps:
...
+
+ <ol>
+  <li><p><a lt=terminated for=fetch>Terminate</a> <var>ongoingFetch</var> with reason <i>fatal</i>.
+ </ol>
+
  <li>
   <p>Run the following <a>in parallel</a>:
 
-  <p><a for=/>Fetch</a> <var>request</var>.
+  <p><a for=/>Fetch</a> <var>request</var> and let <var>ongoingFetch</var> be the ongoing fetch.
```
jakearchibald: I'm still not happy with this. The termination of *ongoingFetch* happens on the main thread, but I'm not sure how to get from this "in parallel" to the "abort algorithm".

jakearchibald: (the steps which call the abort algorithms are main thread)

wanderview: Can you set a flag that is shared between main thread and the in parallel steps?  Then you can have various "cancellation points" where you check that flag.  Somewhat similar to how the implementation is done in places.  Of course, cancelling the network stack doesn't work that way, but maybe you could poll the flag there.

jakearchibald: That sort-of seems to be how it's spec'd already, but using language "if fetch is terminated". I agree a flag would be better.

The bit I'm struggling with is that the signal abort steps are main thread, but the *ongoingFetch* is created in the "in parallel" steps.

Maybe the answer is to find a way to initiate the ongoing fetch on the main thread (which would have the terminated flag), then start the actual fetching in the "in parallel" bit.

annevk: How about making the flag part of the underlying request that is passed to fetch? It basically acts like a signal of sorts.

See also #536 about restructuring how the fetch algorithm is invoked somewhat.

fetch.bs
```diff
@@ -4666,6 +4665,7 @@ interface Request {
   readonly attribute RequestRedirect redirect;
   readonly attribute DOMString integrity;
   readonly attribute boolean keepalive;
+  readonly attribute AbortSignal? signal;
```
annevk: You haven't defined this attribute in prose.

fetch.bs
```diff
@@ -4683,6 +4683,7 @@ dictionary RequestInit {
   DOMString integrity;
   boolean keepalive;
   any window; // can only be set to null
+  AbortSignal signal;
```
annevk: Please move this one up. So it stays aligned with keepalive.

jakearchibald: I'm not really sure what you mean by this. Do you mean I should move it just after the `boolean keepalive;` line? How does this make it aligned?

fetch.bs
jakearchibald: Oops. Fixed in 5d40518.

jakearchibald:
I think we should spec this by giving requests their own internal abort signal. If a signal object is provided to the request, the internal abort signal will listen to this and reflect changes. This internal signal will be the one that's posted to the service worker.

Aside from chaining to the two signals together, I don't think we should use "Add the following abort steps to the signal", and instead check the signal's flag step by step:

1. Do a small chunk of work.
1. If signal's abort flag is set, *abort somehow*.
1. Do another small chunk of work.
1. If signal's abort flag is set, *abort somehow*.
1. Another chunk of work.
1. If signal's abort flag is set, *abort somehow*.

Etc etc.

This means we can be super explicit about how abort is handled & when. However, this could lead to an overly verbose spec, so I wonder if we could do something like:

1. Let abortCheckpoint be the following steps:
    1. If signal's abort flag is not set, abort these substeps.
    1. Reject p with an "{{AbortError}}" {{DOMException}} and abort these steps.
1. Run abortCheckpoint then do a small chunk of work.
1. Run abortCheckpoint then do another small chunk of work.
1. Run abortCheckpoint then do yet another small chunk of work.

jakearchibald:
@domenic I've experimented with granular aborting over at https://github.com/whatwg/fetch/compare/cancelation...cancelation-idea, specifically for the "transmit body" steps. Does this seem like a completely broken way of speccing it?

domenic:
Hmm, that doesn't seem to work so great. I'd expect reading a chunk or transmitting _bs_ to be interruptible, whereas that doesn't seem possible with that spec.

Maybe we could start by listing all the places you anticipate being able to abort? Off the top of my head:

- Before taking any action
- During "transmit body"
  - During reading a chunk
  - During transmitting a chunk
  - Before reading the next chunk
- During service worker handling
- During foreign fetch handling
- During HTTP-network fetch's
  - During "making a HTTP request" (i.e. the step that gets the headers)
  - During the body byte receive loop
- All of the above, but during a CORS preflight fetch

It looks to me like several of these already have provisions for the fetch being terminated during them. IMO we should just make sure to handle the rest, and then make the abort steps terminate the fetch.

fetch.bs
domenic: It makes it match the order for `interface Request`

jakearchibald:
> Hmm, that doesn't seem to work so great. I'd expect reading a chunk or transmitting bs to be interruptible, whereas that doesn't seem possible with that spec.

I'm trying to avoid hand-waving cancelation. Maybe we could use language like:

1. Do some work that we can't really abort during.
1. Let *abortSteps* be the following sub steps:
    1. Cancel the stream etc etc
    1. Abort these steps (would this cancel the whole algorithm?)
1. If signal's abort flag is set, run *abortSteps*.
1. Add *abortSteps* to signal's abort steps.
1. Do some work
1. Do some work
1. Do some work
1. Remove *abortSteps* from signal's abort steps.
1. Do some work that we can't really abort during.

> Maybe we could start by listing all the places you anticipate being able to abort?

That's fair. I'll change my branch to add `Issue:` statements at meaningful abort points.

jakearchibald:
Done. Hopefully this will be useful.

domenic:
My point is I don't think it's hand-waving. In practice all this code is running in parallel and can be interrupted by other code. That's in fact made very explicit by saying it's "in parallel". I expect it'd be implemented exactly that way.

jakearchibald:
But surely the process for aborting is different if it's setting up a connection, vs sending a body, vs reading a body. We can't just wrap that up in a single "terminate the fetch" can we?

domenic:
That's fair, we should define "terminate the fetch" to do different things depending on which phase you're in.

jakearchibald:
I considered that, then thought it's easier to read if the abort steps are as close as possible to the thing being aborted.

As in:

1. Let abortSteps be the following sub steps and add them to signal:
    1. …how to abort during task A…
1. Do task A.
1. Remove abortSteps from signal.
1. Set abortSteps to the following sub steps and add them to signal:
    1. …how to abort during task B…
1. Do task B.
1. Remove abortSteps from signal.

Rather than:

To abort a fetch:

1. If currently performing task A:
  1. …how to abort during task A…
1. Else if currently performing task B:
  1. …how to abort during task B…

…then pages of scrolling away…

1. Do task A.
2. Do task B.

(I'm probably wrong here, just trying to figure out why)

domenic:
Well, the advantage of the latter is that it seems closer to how I'd guess it will be implemented, and it creates a reusable concept of "terminate the fetch" which any spec can call (e.g. HTML's navigation algorithm can terminate all ongoing non-keepalive fetches, no matter what stage they're in).

jakearchibald:
In terms of reusability, shouldn't it use signals? Fair enough on the implementation front, I didn't know that's how it worked.

domenic:
I don't think we want HTML's navigation algorithm to depend on JavaScript-exposed objects like signals, no. Signals are just how JavaScript signals that it wants to terminate the fetch; actual fetch termination is the primitive operation.

annevk:
As far as I know in implementations the caller can set some kind of flag that is shared across threads and is checked by the fetch algorithm at times where it can be terminated. Which roughly matches Jake's "internal signal" idea I think. @wanderview can correct me if wrong. I think that's how we want to define this, even if it's more verbose.

wanderview:
From an implementation perspective a simple "check a cancel flag" is not quite adequate.  There are places where you can be waiting for external input (connecting a socket, etc) and not performing any kind of active local code.  In these cases we would pro-actively take action to cancel like closing the socket instead of checking a flag.

I'm not sure if we really need to capture that kind of thing in the spec or not.

Talking with @annevk it sounds like we could probably immediately mark the DOM side of the fetch operation dead immediately.  There is probably no need to wait for the cancel operation to completely close the socket.  (Or is there?)

jakearchibald:
I tried to capture that in https://github.com/whatwg/fetch/pull/523#issuecomment-307340150.

The abort steps are different at different parts of a fetch. It will check a flag before it starts those steps, but it will also provide steps to run if cancellation happens during those steps. Once those steps are complete, it removes the abort steps since they no longer make sense for this phase of the fetch.

But I totally defer to whatever makes most sense for implementers. It's not something I have knowledge of. My thinking is very JavaScripty 😄 .

annevk:
I chatted a bit with Honza Bambas about Firefox's implementation. There's indeed phases. And any kind of work that would be considered "blocking" (I/O) is done during asynchronous calls (so the code can continue to receive cancel requests).

Refactoring Fetch to take that into account might be quite a bit of work though as it would effectively need its own event loop of sorts.

jakearchibald:
@annevk would the pattern in https://github.com/whatwg/fetch/pull/523#issuecomment-307404328 work?

annevk:
@jakearchibald I think that would create the false impression that the abort steps would run concurrently. However, I might not fully grasp the setup. Maybe if we only used that for the parts where Fetch is "blocking" (waiting for a network response for instance) it could work out okay? I suspect long term we'll hit some other wall though and have to refactor all the things anyway, but avoiding it for now would be nice.

annevk:
Gecko bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1378342.

fetch.bs
```diff
@@ -4942,10 +4997,19 @@ constructor must run these steps:
...
+ <li><p>If <var>init</var>'s <code>signal</code> member is present, set <var>signal</var> to it.
+
  <li><p>Let <var>r</var> be a new {{Request}} object associated with <var>request</var> and
  a new associated {{Headers}} object whose <a for=Headers>guard</a>
  is "<code>request</code>".
 
+ <li>If <var>signal</var> is not null, <a for=AbortSignal lt=add>add the following abort steps</a>
+ to <var>signal</var>:
+  <ol>
+   <li>Signal abort on <var>r</var>'s <a for=Request>signal</a>.
```
annevk: This needs to xref "signal abort" from DOM.

annevk:
FWIW, I think I'm generally on board with @jakearchibald's approach. I think the difficulty is in part with the "in parallel" language. In practice the fetch algorithm is more of a thread (at least from the point where we use "in parallel") that uses sub-threads for certain tasks that are currently blocking in the specification (such as sending or receiving a byte sequence).

However, until we have explicit threads there's not really a way to write it down as such. And rewriting the fetch algorithm as more of an event-loop based operation might also be somewhat tricky.

If @domenic wants to double down on JavaScript vs system layer separation I'd be okay with that (and might help implementers a bit), but then we should do the same for streams. That also doesn't really preclude using signals I think. We'd just need to be able to use them abstractly and make the JavaScript API a wrapper.

domenic:
> FWIW, I think I'm generally on board with [@jakearchibald](https://github.com/jakearchibald)'s approach.

Which is this, exactly? Do you mean https://github.com/whatwg/fetch/pull/523#issuecomment-307404328 ?

In general I don't care much about the non-streams aspects and am happy to let you all figure it out; I was just trying to help while you were away.

> If [@domenic](https://github.com/domenic) wants to double down on JavaScript vs system layer separation I'd be okay with that (and might help implementers a bit)

I do think we should keep things separate where possible.

> but then we should do the same for streams.

I've been thinking about how to do this for a long time. It sure would be nice. I'd like to do it later though.

> That also doesn't really preclude using signals I think. We'd just need to be able to use them abstractly and make the JavaScript API a wrapper.

My point of view is that most of the features of signals are designed around delegation of authority in JavaScript objects. I.e., you can give someone a promise or a Request or a Response or whatever, and if they don't have the AbortSignal, they cannot abort it. This is similar to how you can give someone a promise, but without the resolve/reject functions, they cannot change its fate.

But in spec land that separation just gets in the way. For promises, in spec land we happily say "resolve p with x" on implementation-created promises, with the understanding that this is theoretically translatable to tracking a (p, resolve) pair throughout the system, but we don't want to write that down explicitly. (And indeed at least Chrome implements it with direct access, not a (p, resolve) pair.)

So I think it's simpler to understand if we don't involve the signals at all in aborting a fetch, but instead simply define primitive operations like "abort the stream" or "abort the request" or similar and use those directly.

annevk:
The main problem with "abort the request" vs "abort the signal" is that you don't really have a handle on a request (that might be an underlying problem I suppose). As was suggested elsewhere the fetch endpoint is really such that you pass it some callbacks and wait for those to be called. So if you want to cancel fetch you'd need to signal that in some way, likely through an object or a thing you already passed.

domenic:
Hmm, can you give an example? All the specs I've seen set up a request, then feed it to the fetch algorithm.

annevk:
Sorry, you do have a handle on the request, but you don't have a handle on the fetch. I guess when you pass the request to fetch though it could set a pointer of sorts so you can then terminate the request (but that seems a little ugly). But the existing language is "terminate the fetch".

domenic:
Oh, I see, then it's not as obvious what to do. I think the easiest thing might be to have an operation "abort the fetch initiated by _request_", but maybe juggling abstract signals would not be that bad either in practice. You'd have to try it to find out. I'm happy to leave it up to you to decide or discover which is best.

jakearchibald:
@annevk to summarise from IRC, are you happy with:

1. Do some stuff
1. Run the following sub steps:
  1. If this fetch has terminated, or terminates during the current sub steps, run the following sub steps:
    1. Abort streams and/or reject promises etc etc.
    1. Abort these steps.
  1. Do some stuff.

…as a style for expressing fetch cancelation?

annevk:
Yeah, or maybe

1. Run these steps until fetch is terminated:
   1. ...
2. If fetch is terminated, then ...

jakearchibald:
I prefer that, as long as it works with the scoping of variables, which I suspect it will.

jakearchibald:
https://bugs.chromium.org/p/chromium/issues/detail?id=750599

annevk:
I like the former better as it's much clearer. With the second style it's not clear that step 3 and 4 can both be terminated.

fetch.bs
```diff
@@ -1212,8 +1213,17 @@ or "<code>worker</code>".
    is true, then <a>queue a fetch task</a> on <var>request</var> to
    <a>process request end-of-body</a> for <var>request</var> and abort these steps.
 
-   <li><p>Otherwise, <a lt=terminated for=fetch>terminate</a> the ongoing fetch with reason
-   <i>fatal</i> and abort these steps.
+   <li>
+    <p>Otherwise:
+
+    <ol>
+     <li><p>If the ongoing fetch has <a for=fetch>terminated</a>, abort these steps.
```
annevk: Newline before `<p>`.

fetch.bs
```diff
@@ -1485,21 +1495,29 @@ for each associated <a for="fetch group">fetch record</a> whose
 <var>origin</var> and <var>credentials</var>, run these steps:
 
 <ol>
- <li><p>If <a for=connection>connection pool</a> contains a
- <a>connection</a> whose <b>origin</b> is <var>origin</var> and
- <b>credentials</b> is <var>credentials</var>, return that
- <a>connection</a>.
+ <li><p>If <a for=connection>connection pool</a> contains a <a>connection</a> whose <b>origin</b> is
+ <var>origin</var> and <b>credentials</b> is <var>credentials</var>, return that <a>connection</a>.
```
annevk: then return*

fetch.bs
```diff
@@ -1485,21 +1495,29 @@ for each associated <a for="fetch group">fetch record</a> whose
...
+    <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
+
+    <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.
+
+    <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), return
+    failure.
+  </ol>
 
-  <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), return failure.
+ <li><p>If fetch is <a for=fetch>terminated</a>, close <var>connection</var> and return failure.
```
annevk: then close*

fetch.bs
```diff
@@ -1485,21 +1495,29 @@ for each associated <a for="fetch group">fetch record</a> whose
...
-  <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
+  <p>Run these steps until fetch is <a for=fetch>terminated</a>:</p>
 
-  <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.
+  <ol>
+   <li>
+    <p>Set <var>connection</var> to the result of establishing an HTTP connection to
+    <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
+
+    <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.
```
annevk: then do*

fetch.bs
```diff
@@ -1485,21 +1495,29 @@ for each associated <a for="fetch group">fetch record</a> whose
...
+
+    <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), return
```
annevk: then return*

fetch.bs
```diff
@@ -1485,21 +1495,29 @@ for each associated <a for="fetch group">fetch record</a> whose
...
  <li>
-  <p>Let <var>connection</var> be the result of establishing an HTTP connection to
-  <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
+  <p>Run these steps until fetch is <a for=fetch>terminated</a>:</p>
 
-  <p>If <var>credentials</var> is false, do <em>not</em> send a TLS client certificate.
+  <ol>
+   <li>
+    <p>Set <var>connection</var> to the result of establishing an HTTP connection to
+    <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
```
annevk: `[[!TLS]]` needs to be on a new line.

fetch.bs
```diff
@@ -1773,8 +1791,8 @@ from a {{ReadableStream}} object with <var>reader</var>, run these steps:
 </ol>
 
 <p>To <dfn export for=ReadableStream id=concept-cancel-readablestream>cancel</dfn> a
-{{ReadableStream}} object with <var>reader</var> and <var>reason</var>, return the result of calling
-<a abstract-op>ReadableStreamCancel</a>(<var>reader</var>, <var>reason</var>).
+{{ReadableStream}} object <var>stream</var> with <var>reason</var>, return the result of calling
+<a abstract-op>ReadableStreamCancel</a>(<var>stream</var>, <var>reason</var>).
 
 <p class="note no-backref">Because the reader grants exclusive access, the actual mechanism of how
 to read cannot be observed. Implementations could use more direct mechanism if convenient.
```
annevk: It seems the reference to a reader no longer makes sense given the change you made above.

fetch.bs
```diff
@@ -2293,131 +2311,139 @@ the request.
...
-   <a for=/>fetch</a> algorithm.
-
-   <li><p>Append <var>record</var> to <var>request</var>'s
-   <a for=request>client</a>'s
-   <a for=fetch>fetch group</a> list of
-   <a for="fetch group">fetch records</a>.
   </ol>
 
+ <li><p>If fetch is <a for=fetch>terminated</a>, return a <a>network error</a> with <a
+ for=response>termination reason</a> set to the termination <var>reason</var>.
```
annevk: You should probably introduce _reason_ in the conditional somehow. "If fetch is terminated with _reason_" maybe?

annevk: I also somewhat wonder if we need to wrap all these steps, since these are all synchronous operations in the place where they occur, no?

jakearchibald:
I still need to handle the service worker parts.

Also I need to write the following tests:

* The request body errors with the same object as the fetch promise.
* The response body errors with the same object as the trailers promise.
* Throws within the requests constructor have priority over the abort error (if the signal has already aborted).

fetch.bs
jakearchibald: fwiw there are a few places where the line limit was already being ignored (or in some cases it was wrapping much sooner). If that was the case I've continued to ignore it rather than creating even more of a diff to look through.

jakearchibald:
@annevk nits addressed! (although I realise it needs a fuller review)

fetch.bs
```diff
@@ -1196,7 +1196,8 @@ or "<code>worker</code>".
 
    <li>
     <p>If <var>read</var> is fulfilled with an object whose <code>done</code> property is false
-    and whose <code>value</code> property is a <code>Uint8Array</code> object, then run these steps:
+    and whose <code>value</code> property is a <code>Uint8Array</code> object, then run these steps
+    until fetch is <a for=fetch>terminated</a>:
```
annevk: The problem with this setup is that you don't reach the later "Otherwise" if the fetch is terminated because the initial "If" conditional is already met. So you probably need to create a new block after "then" where inside you first run steps until terminated and then handle the termination if terminated.

fetch.bs
```diff
@@ -1491,21 +1502,32 @@ for each associated <a for="fetch group">fetch record</a> whose
...
+ <li><p>If <a for=connection>connection pool</a> contains a <a>connection</a> whose <b>origin</b> is
+ <var>origin</var> and <b>credentials</b> is <var>credentials</var>, then return that
  <a>connection</a>.
 
+ <li><p>Let <var>connection</var> be null.
+
  <li>
-  <p>Let <var>connection</var> be the result of establishing an HTTP connection to
-  <var>origin</var>. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!TLS]]
+  <p>Run these steps until fetch is <a for=fetch>terminated</a>:</p>
```
annevk: As discussed we should rephrase this here and elsewhere. My suggestion is "Run these steps once and stop if fetch is terminated" but I'm open to something better since it's not exactly great.

fetch.bs
```diff
@@ -2299,131 +2318,139 @@ the request.
...
-   <var>request</var> and this instance of the
-   <a for=/>fetch</a> algorithm.
-
-   <li><p>Append <var>record</var> to <var>request</var>'s
-   <a for=request>client</a>'s
-   <a for=fetch>fetch group</a> list of
-   <a for="fetch group">fetch records</a>.
   </ol>
 
+ <li><p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, return a
```
annevk: then return*

fetch.bs
```diff
@@ -2440,81 +2467,90 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
...
+     <a href=https://tools.ietf.org/html/rfc6797#section-8.2>Known HSTS Host Domain Name
+     Matching</a> results in either a superdomain match with an asserted
+     <code>includeSubDomains</code> directive or a congruent match (with or without an asserted
+     <code>includeSubDomains</code> directive) [[!HSTS]]
+    </ul>
+   <!-- Per Mike West HSTS happens "probably after" Referrer -->
+
+  </ol>
+
+ <li><p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, return a
```
annevk: then return*

fetch.bs
```diff
@@ -3185,280 +3234,287 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+      <p class=note>See also the
+      "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
+      chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+
+      <li><p>Otherwise, if the <var>revalidatingFlag</var> is unset, then set <var>response</var> to
+      <var>storedResponse</var>.
+    </ol>
   </ol>
 
+ <li><p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, return a
```
annevk: then return*

fetch.bs
```diff
@@ -3569,6 +3625,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li class=XXX><p>Needs testing: multiple `<code>Proxy-Authenticate</code>` headers, missing,
    parsing issues.
 
+   <li><p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, return a
+   <a>network error</a> with <a for=response>termination reason</a> set to <var>reason</var>.
```
annevk: then return*

But also, this seems to be in a strange place. Why do we have this for 407, but not 401? There's also no enclosing step saying these steps can be terminated.

fetch.bs
```diff
@@ -3614,83 +3673,96 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+  <p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, then:
 
-  <p class="note no-backref">The exact determination here is up to user agents for the
-  time being. User agents are strongly encouraged to only succeed HTTPS connections with
-  strong security properties and return
-  <a>network errors</a> otherwise. Using the
-  "<code>deprecated</code>" state value ought to be a temporary and last resort kind
-  of option.
+  <ol>
+   <li><p>If <var>connection</var> is an HTTP/2 connection, transmit an "<code>RST_STREAM</code>" to
```
annevk: then transmit*

fetch.bs
```diff
@@ -3699,132 +3771,143 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+    href=https://tools.ietf.org/html/rfc6265#section-5.2>section 5.2</a> of [[!COOKIES]]) on the <a
+    for=header>value</a> of each <var>header</var> whose <a for=header>name</a> is a
+    <a>byte-case-insensitive</a> match for `<code>Set-Cookie</code>` in <var>response</var>'s <a
+    for=response>header list</a>, if any, and <var>request</var>'s <a for=request>current url</a>.
+
+    <p class=note>This is a fingerprinting vector.
+  </ol>
 
-  <p class=note>This is a fingerprinting vector.
+ <li><p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, set
```

fetch.bs
```diff
@@ -3699,132 +3771,143 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
  <li>
   <p>Run these substeps <a>in parallel</a>:
 
   <ol>
    <li>
-    <p>Whenever one or more bytes are transmitted from <var>response</var>'s message body, let
-    <var>bytes</var> be the transmitted bytes and run these subsubsteps:
-    <!-- XXX xref message body -->
+    <p>Loop the following steps, breaking immediately if fetch
+    <a for=fetch lt=terminated>terminates</a>, providing a <var>reason</var>:
```
annevk: "Loop the following steps" -> "While true".

I think we can also drop "immediately" since we always mean immediately (unless we state otherwise).

fetch.bs
```diff
@@ -3699,132 +3771,143 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
-     <li><p>If <var>stream</var> is <a for=ReadableStream>readable</a>,
-     <a abstract-op>error</a> <var>stream</var> with a
-     <code>TypeError</code>.
+     <li><p>Otherwise, if the bytes transmission for <var>response</var>'s message body is done
+     normally and <var>stream</var> is <a for=ReadableStream>readable</a>, then
+     <a abstract-op>close</a> <var>stream</var> and abort these steps.
     </ol>
+
+   <li><p>Set <var>response</var>'s <a lt="termination reason" for=response>termination reason</a>
+   to <var>reason</var>.
```
annevk: Shouldn't this and subsequent steps be wrapped in a conditional of the fetch being terminated?

fetch.bs
```diff
@@ -4076,6 +4159,8 @@ Entries may be removed before that moment arrives.
 <var>request</var> and <var>response</var>, run these steps:
 
 <ol>
+ <li><p>If <var>response</var> is a <a>network error</a>, return failure.
```
annevk: There's no need for this as a network error is a response without headers so will always fail this check.

fetch.bs
```diff
@@ -4387,7 +4472,6 @@ method, when invoked, must run these steps:
 <p>The <a>value pairs to iterate over</a> are the return value of running
 <a for="header list">sort and combine</a> with the <a for=Headers>header list</a>.
 
-
```
annevk: Please don't remove this newline.

fetch.bs
```diff
@@ -4948,9 +5040,19 @@ constructor must run these steps:
    to <var>method</var>.
   </ol>
 
- <li><p>Let <var>r</var> be a new {{Request}} object associated with <var>request</var> and
+ <li><p>If <var>init</var>'s <code>signal</code> member is present, set <var>signal</var> to it.
```

fetch.bs
```diff
@@ -4948,9 +5040,19 @@ constructor must run these steps:
...
+
+ <li><p>Let <var>r</var> be a new {{Request}} object associated with <var>request</var>,
  a new associated {{Headers}} object whose <a for=Headers>guard</a>
- is "<code>request</code>".
+ is "<code>request</code>", and a new {{AbortSignal}} object.
+
+ <li>
+  <p>If <var>signal</var> is not null, <a for=AbortSignal lt=add>add the following abort steps</a>
```
annevk: then add*

fetch.bs
```diff
@@ -5401,6 +5535,29 @@ method, must run these steps:
...
+<ol>
+ <li><p>Let <var>error</var> be an "<code><a exception>AbortError</a></code>" {{DOMException}}.
+
+ <li><p>Reject <var>promise</var> with <var>error</var>.
+
+ <li><p>If <var>request</var>'s <a for=request>body</a> is not null and is
+ <a for=ReadableStream>readable</a>, <a for=ReadableStream>cancel</a> <var>request</var>'s
+ <a for=request>body</a> with <var>error</var>.
+
+ <li><p>If <var>responseObject</var> is null, return.
```
annevk: then return*

fetch.bs
```diff
@@ -5401,6 +5535,29 @@ method, must run these steps:
...
+<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>, and
+<var>responseObject</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>error</var> be an "<code><a exception>AbortError</a></code>" {{DOMException}}.
+
+ <li><p>Reject <var>promise</var> with <var>error</var>.
+
+ <li><p>If <var>request</var>'s <a for=request>body</a> is not null and is
+ <a for=ReadableStream>readable</a>, <a for=ReadableStream>cancel</a> <var>request</var>'s
```
annevk: then cancel*

fetch.bs
```diff
@@ -5401,6 +5535,29 @@ method, must run these steps:
...
+
+ <li><p>If <var>responseObject</var> is null, return.
+
+ <li><p>Reject <var>responseObject</var>'s <a for=Response>trailer promise</a> with
+ <var>error</var>.
+
+ <li><p>Let <var>response</var> be <var>responseObject</var>'s <a for=Response>response</a>.
+
+ <li><p>If <var>response</var>'s <a for=response>body</a> is not null and is
+ <a for=ReadableStream>readable</a>, <a for=ReadableStream>cancel</a> <var>response</var>'s
```
annevk: then cancel*

fetch.bs
```diff
@@ -5401,6 +5535,29 @@ method, must run these steps:
...
+ <a for=response>body</a> with <var>error</var>.
+</ol>
```
annevk: Needs newline afterwards. Two newlines before the `<h3>`.

fetch.bs
```diff
@@ -2440,81 +2467,90 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
...
+     <a for=/>domain</a>
+     <li>Matching <var>request</var>'s <a for=request>current url</a>'s
+     <a for=url>host</a> per
+     <a href=https://tools.ietf.org/html/rfc6797#section-8.2>Known HSTS Host Domain Name
+     Matching</a> results in either a superdomain match with an asserted
+     <code>includeSubDomains</code> directive or a congruent match (with or without an asserted
+     <code>includeSubDomains</code> directive) [[!HSTS]]
+    </ul>
+   <!-- Per Mike West HSTS happens "probably after" Referrer -->
+
```
annevk: This newline can be removed.

fetch.bs
```diff
@@ -2944,9 +2994,8 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
        <a for=response>url list</a> has more than one item.
       </ul>
 
-     <li><p>Execute
-     <a href=https://w3c.github.io/webappsec-csp/#set-response-csp-list>set <var>response</var>'s CSP list</a>
-     on <var>actualResponse</var>. [[!CSP]]
+     <li><p>Execute <a href=https://w3c.github.io/webappsec-csp/#set-response-csp-list>set
```
annevk: No newlines inside `<a>`.

fetch.bs
```diff
@@ -3185,280 +3234,287 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+       <a for="header list">contains</a> `<code>Last-Modified</code>`, then
+       <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
+       <var>httpRequest</var>'s <a for=request>header list</a>.
+      </ol>
+
+      <p class=note>See also the
+      "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
+      chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+
+      <li><p>Otherwise, if the <var>revalidatingFlag</var> is unset, then set <var>response</var> to
```
annevk: This `<li>` needs one less indentation.

fetch.bs
```diff
@@ -3614,83 +3673,96 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
 
- <li><p>If <var>connection</var> is failure, return a
- <a>network error</a>.
 
- <li><p>If <var>connection</var> is not an HTTP/2 connection, <var>request</var>'s
- <a for=request>body</a> is non-null, and <var>request</var>'s <a for=request>body</a>'s
- <a for=body>source</a> is null, then <a for="header list">append</a>
- `<code>Transfer-Encoding</code>`/`<code>chunked</code>` to <var>request</var>'s
- <a for=request>header list</a>.
+  <li>
```
annevk: It seems this is indented one too few? (And then it gets worse.)

jakearchibald:
> Looks okay overall, couple of gotchas and many nits.

Apologies. I'll get used to the patterns eventually. Thanks for bearing with me.

> Do you plan to address the SW issues before landing this or do you want that as follow-up?

Yes, hoping to get a PR in for that today. It doesn't seem as complicated as I originally thought. The changes are entirely in the SW spec though, so I'll just remove the TODOs here.

Thanks for the review!

fetch.bs
jakearchibald: Ah balls. My editor tries to be "smart" but the lack of `</li>` caused its "smarts" to be "a pain in the arse".

fetch.bs
jakearchibald: I'll add this for 401 too – I just missed it.

I don't think we need enclosing steps here, it's just an extra check to avoid showing UI if the fetch has already terminated.

fetch.bs
jakearchibald: These steps can only be reached if the fetch has terminated. I'll add a note to that effect.

fetch.bs
jakearchibald: Mind if I add a note for that instead? It wasn't clear when I first read it.

fetch.bs
jakearchibald: lol, sorry about all these

jakearchibald:
Addressed feedback. Also, additional tests https://github.com/w3c/web-platform-tests/pull/6736.

Next up: the service worker stuff.

jakearchibald:
I'm happy with this now btw. Removed the WIP bit.

annevk:
@wanderview @domenic @yutakahirano care to do second review? See the Preview and Diff links at https://github.com/whatwg/fetch/pull/523#issue-220475854 to make it easier to read the changes.

fetch.bs
```diff
@@ -2796,44 +2828,54 @@ steps:
...
-   <li><p>Let <var>blob</var> be <var>request</var>'s
-   <a for=request>current url</a>'s
-   <a>object</a>.
-
    <li>
-    <p>If <var>request</var>'s <a for=request>method</a> is not
-    `<code>GET</code>` or <var>blob</var> is null, then return a
-    <a>network error</a>.
+    <p>Run these substeps, but abort the substeps if fetch is <a for=fetch>terminated</a>, providing
+    a <var>reason</var>:
```
domenic: This time the "providing" is up here, for unknown reasons... better to be consistent.

fetch.bs
```diff
@@ -3614,83 +3668,96 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
- <li><p>Let <var>pull</var> be an action that <a lt=resumed for=fetch>resumes</a> the
- ongoing fetch if it is <a lt=suspend for=fetch>suspended</a>.
+ <li><p>Let <var>pull</var> be an action that <a lt=resumed for=fetch>resumes</a> the ongoing fetch
+ if it is <a lt=suspend for=fetch>suspended</a>.
 
- <li><p>Let <var>cancel</var> be an action that
- <a lt=terminated for=fetch>terminates</a> the ongoing fetch with reason
- <i>end-user abort</i>.
+ <li><p>Let <var>cancel</var> be an action that <a lt=terminated for=fetch>terminates</a> the
+ ongoing fetch with reason <i>fatal</i>.
```
domenic: Why the change in reason?

fetch.bs
```diff
@@ -4948,9 +5039,20 @@ constructor must run these steps:
...
   </ol>
 
- <li><p>Let <var>r</var> be a new {{Request}} object associated with <var>request</var> and
+ <li><p>If <var>init</var>'s <code>signal</code> member is present, then set <var>signal</var> to
+ it.
+
+ <li><p>Let <var>r</var> be a new {{Request}} object associated with <var>request</var>,
  a new associated {{Headers}} object whose <a for=Headers>guard</a>
- is "<code>request</code>".
+ is "<code>request</code>", and a new {{AbortSignal}} object.
```
domenic: This sentence as exists is confusing, and you made it a bit worse. A band-aid is to maybe add "associated", but a proper fix would be to set _r_'s [=header=] and [=signal=] fields in explicit steps.

fetch.bs
```diff
@@ -5401,6 +5535,30 @@ method, must run these steps:
...
+<ol>
+ <li><p>Let <var>error</var> be an "<code><a exception>AbortError</a></code>" {{DOMException}}.
+
+ <li><p>Reject <var>promise</var> with <var>error</var>.
+
+ <li><p>If <var>request</var>'s <a for=request>body</a> is not null and is
+ <a for=ReadableStream>readable</a>, then <a for=ReadableStream>cancel</a> <var>request</var>'s
+ <a for=request>body</a> with <var>error</var>.
+
+ <li><p>If <var>responseObject</var> is null, then return.
```
domenic: _responseObject_ is never null

fetch.bs
```diff
@@ -5401,6 +5535,30 @@ method, must run these steps:
  <li><p>Return <var>p</var>.
 </ol>
 
+<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>, and
+<var>responseObject</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>error</var> be an "<code><a exception>AbortError</a></code>" {{DOMException}}.
+
+ <li><p>Reject <var>promise</var> with <var>error</var>.
```
domenic: It might be worth noting that this is a no-op if _promise_ is already fulfilled

fetch.bs
```diff
@@ -5401,6 +5535,30 @@ method, must run these steps:
...
+
+ <li><p>If <var>request</var>'s <a for=request>body</a> is not null and is
+ <a for=ReadableStream>readable</a>, then <a for=ReadableStream>cancel</a> <var>request</var>'s
+ <a for=request>body</a> with <var>error</var>.
+
+ <li><p>If <var>responseObject</var> is null, then return.
+
+ <li><p>Reject <var>responseObject</var>'s <a for=Response>trailer promise</a> with
+ <var>error</var>.
```
domenic: Again, might be a no-op.

fetch.bs
```diff
@@ -5401,6 +5535,30 @@ method, must run these steps:
...
+
+ <li><p>Let <var>response</var> be <var>responseObject</var>'s <a for=Response>response</a>.
+
+ <li><p>If <var>response</var>'s <a for=response>body</a> is not null and is
+ <a for=ReadableStream>readable</a>, then <a for=ReadableStream>cancel</a> <var>response</var>'s
+ <a for=response>body</a> with <var>error</var>.
```
domenic: I think this needs to error, not cancel, to match the tests.

fetch.bs
```diff
@@ -1192,38 +1192,45 @@ or "<code>worker</code>".
   <p><a>In parallel</a>, while true:
 
   <ol>
-   <li><p>Wait for <var>read</var> to be fulfilled or rejected.
-
    <li>
-    <p>If <var>read</var> is fulfilled with an object whose <code>done</code> property is false
-    and whose <code>value</code> property is a <code>Uint8Array</code> object, then run these steps:
+    <p>Run these substeps, but abort the substeps if fetch is <a for=fetch>terminated</a>:</p>
```
domenic: As noted in https://github.com/whatwg/infra/issues/143 I'm not sure there's any reason to use "substeps" over "steps" here and elsewhere.

annevk:
Certainly XMLHttpRequest branches on the termination reason: https://xhr.spec.whatwg.org/#handle-errors. It seems that `fetch()` currently doesn't define this at all though. I left that out because we hadn't really decided what the story was going to be for this class of "errors" and also for what should happen when an `iframe` element gets removed or the user terminates fetching through UI or some such. Unfortunately I also forgot about it while reviewing and working on this.

fetch.bs
jakearchibald: Yeah, I'll change those.

fetch.bs
```diff
@@ -2299,131 +2314,139 @@ the request.
...
-   <var>request</var> and this instance of the
-   <a for=/>fetch</a> algorithm.
-
-   <li><p>Append <var>record</var> to <var>request</var>'s
-   <a for=request>client</a>'s
-   <a for=fetch>fetch group</a> list of
-   <a for="fetch group">fetch records</a>.
   </ol>
 
+ <li><p>If fetch is <a for=fetch>terminated</a>, providing a <var>reason</var>, then return a
```
jakearchibald: I'm trying to avoid *reason* sounding like part of the conditional. As in "If the fetch is terminated, let the *reason* be the termination reason…" rather than "If the fetch is terminated and the termination reason is equal to *reason*"

fetch.bs
jakearchibald: I thought that "end-user abort" means that the user hit the X button, or closed the tab, but I'm struggling to find a reference for that (@annevk am I making this up?).

Since this would be called if I call .cancel() on the readable stream, "end-user abort" feels like the wrong kind of termination.

fetch.bs
jakearchibald: Ah yes, agreed.

fetch.bs
annevk: Yeah, "hit the X button" or some keyboard shortcut. Closing the tab is probably fatal due to how unloading documents works.

jakearchibald:
I also need to handle `request.clone()` in terms of signal. That'll need tests too.

jakearchibald:
Hold off on additional reviews. I have changes to push, but I'm basing them on top of https://github.com/whatwg/fetch/pull/575.

fetch.bs
jakearchibald: It is if it's called as part of `fetch()` step 4.1.

jakearchibald:
Ok, now up to date with #575.

jakearchibald:
Added tests for cloning https://github.com/w3c/web-platform-tests/pull/6736

jakearchibald:
@annevk I'm happy with this btw. Do you want to do a final review?

annevk:
@jakearchibald I guess the main thing that was unclear to me is whether we should resolve https://github.com/whatwg/fetch/pull/523#issuecomment-320494347. I think it's somewhat important though arguably we could make it a follow-up thing of sorts.

jakearchibald:
I think I've preserved XHR's use of termination reason. For the iframe case, I thought removing the iframe would terminate the fetch group, which terminates the fetches with "fatal".

annevk:
Yeah that sounds about right, but how does that affect `fetch()`? Rejected with a TypeError?

jakearchibald:
As spec'd now, yeah. I figured I'd leave `AbortError` for developer-aborted. I don't have strong opinions on that though.

annevk:
Okay, @domenic, could you have one last look? Once this lands I'll look into the Infra improvements to improve the wording a bit here and there.

annevk:
So rejecting the trailer promise is problematic. A readonly attribute shouldn't result in an exception. We should just use the empty header list in that case. (If that means you cannot detect a late abort, we should maybe figure something out, dunno.)

@bakulf noted that the `Request` constructor doesn't seem to consistently create a new `AbortSignal` object. It sometimes just sets the signal to what is given. That seems wrong?

jakearchibald:
> A readonly attribute shouldn't result in an exception. 

Why's that? FWIW the [web animation API](https://w3c.github.io/web-animations/#the-animation-interface) has a `.finished` property, the [font loading API](https://drafts.csswg.org/css-font-loading/#dom-fontface-loaded) has a `.loaded` property, and [streams](https://streams.spec.whatwg.org/#default-reader-closed) has a `.closed` property.

> @bakulf noted that the `Request` constructor doesn't seem to consistently create a new `AbortSignal` object. It sometimes just sets the signal to what is given. That seems wrong?

Yeah, that would be a bug, but I can't see it. `new Request()` always creates a new signal, and hooking up any provided signal happens in [step 31](https://fetch.spec.whatwg.org/branch-snapshots/cancelation/#dom-request).

annevk:
> Why's that?

I think there's a general principle that invoking a getter should be safe. And it would also mean that `obj.trailer != obj.trailer` which is also bad.

jakearchibald:
> I think there's a general principle that invoking a getter should be safe. And it would also mean that `obj.trailer != obj.trailer` which is also bad.

Hmm, that wasn't my intention. `.trailer` always returns the same promise, and rejecting rejects the existing promise.

Although, I think there's some existing weirdness here. Eg, the `Response` constructor always resolves the trailer promise ([step 11](https://fetch.spec.whatwg.org/#dom-response)), so it can't be later resolved with something else, as it tries to in the fetch method. As in, with the existing fetch spec, you'll never receive trailers.

annevk:
@jakearchibald the `fetch()` method doesn't invoke the `Response` constructor though, does it? (I don't think `obj.trailer != obj.trailer` is true after all here.)

jakearchibald:
Yeah, my bad. I got it into my head that the constructor steps ran, but they're clearly not.

fetch.bs
```diff
@@ -1782,11 +1806,8 @@ from a {{ReadableStream}} object with <var>reader</var>, run these steps:
 </ol>
 
 <p>To <dfn export for=ReadableStream id=concept-cancel-readablestream>cancel</dfn> a
-{{ReadableStream}} object with <var>reader</var> and <var>reason</var>, return the result of calling
-<a abstract-op>ReadableStreamCancel</a>(<var>reader</var>, <var>reason</var>).
-
-<p class="note no-backref">Because the reader grants exclusive access, the actual mechanism of how
```
domenic: Why remove this?

fetch.bs
domenic: Maybe this would be clearer by adding an additional step, "Let _reason_ be the provided termination reason". Or, in an inline version: "If fetch is terminated, then let _reason_ be the provided termination reason, and do X with _reason_."

fetch.bs
domenic: Closing a tab is not what's in play here. This is about the user calling .cancel() on a stream they are reading. Or, because it goes through the same JS API, the browser calling https://fetch.spec.whatwg.org/branch-snapshots/cancelation/#concept-cancel-readablestream .

That is called by https://fetch.spec.whatwg.org/branch-snapshots/cancelation/#abort-fetch , in fact. So this is saying that any fetch which is aborted through an AbortController will [manifest to XHR](https://xhr.spec.whatwg.org/#handle-errors) not as an abort, but instead as a network error with (I believe) no events fired.

I don't think this is right, in short.

fetch.bs
```diff
@@ -3702,132 +3776,145 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
-     <a for=response>body</a>'s
-     <a for=body>transmitted bytes</a> with <var>bytes</var>'
-     length.
+     <li>
+      <p>If one or more bytes have been transmitted from <var>response</var>'s message body, then:
 
-     <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
-     `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
+      <ol>
+       <li><p>let <var>bytes</var> be the transmitted bytes.
```
domenic: Uppercase

fetch.bs
```diff
@@ -3702,132 +3776,145 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+       data</a> and <var>request</var>'s <a>synchronous flag</a> is unset, ask the user agent to <a
+       for=fetch>suspend</a> the ongoing fetch.
+      </ol>
 
-     <li><p>If <var>stream</var> is <a for=ReadableStream>readable</a>,
-     <a abstract-op>error</a> <var>stream</var> with a
-     <code>TypeError</code>.
+     <li><p>Otherwise, if the bytes transmission for <var>response</var>'s message body is done
+     normally and <var>stream</var> is <a for=ReadableStream>readable</a>, then
+     <a abstract-op>close</a> <var>stream</var> and abort these steps.
```
domenic: abort these steps -> break

fetch.bs
```diff
@@ -3702,132 +3776,145 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
  <li>
   <p>Run these substeps <a>in parallel</a>:
 
   <ol>
    <li>
-    <p>Whenever one or more bytes are transmitted from <var>response</var>'s message body, let
-    <var>bytes</var> be the transmitted bytes and run these subsubsteps:
-    <!-- XXX xref message body -->
+    <p>While true, breaking if fetch <a for=fetch lt=terminated>terminates</a>, providing a
+    <var>reason</var>:
```
domenic: If this loop terminates due to reaching the end of the stream, the next step (step 2) still assumes reason exists, which it does not.

In fact, all of steps 2-5 seem wrong if the body successfully finishes uploading.

fetch.bs
```diff
@@ -4950,12 +5049,23 @@ constructor must run these steps:
    to <var>method</var>.
   </ol>
 
+ <li><p>If <var>init</var>'s <code>signal</code> member is present, then set <var>signal</var> to
+ it.
```
domenic: So, if you do `const r = new Request(otherRequest, { })` then otherRequest's signal will control aborting `r`. But if you do `const r = new Request(otherRequest, { signal })`, then `signal` will control aborting `r`, and `otherRequest` will have no impact at all. Is that correct? Is that what's tested?

annevk: This does make the most sense to me. Note that we need this behavior pretty much like this due to `fetch()` invoking the `Request` constructor.

fetch.bs
jakearchibald: I thought it was redundant since `ReadableStreamCancel` takes a stream, not a readable

fetch.bs
jakearchibald: Unless I'm reading things badly, XHR+fetch seems kinda broken as things stand.

Currently [`abort()`](https://xhr.spec.whatwg.org/#the-abort()-method) [terminates the request](https://xhr.spec.whatwg.org/#terminate-the-request) with reason *fatal*. Which surely means it won't [trigger the abort event](https://xhr.spec.whatwg.org/#handle-errors).

I think I'm going to have to PR XHR too ☹️.

The more I stare at it, the best way could be to remove termination reasons all together, and try and handle abort/timeout within XHR.

How does that sound?

fetch.bs
jakearchibald: I think I'm maybe getting the scope of "abort these steps" wrong.

Step 1.2 performs "abort these steps" if the stream reaches the end. My intent is for this to abort the whole "in parallel" block of steps. Is this not how it works?

fetch.bs
jakearchibald: Right, this is the root of the confusion. I don't want to break, I want to abort the whole "in parallel" block.

I thought "abort these steps" was scoped to the "in parallel" block, since it says "run these steps", whereas "while true" doesn't reference steps, so I didn't think it was a separate set of steps.

fetch.bs
domenic: I'm not sure how you'd do that, since IMO using the AbortSignal in a service worker should propagate an abort event to XHR, whereas closing the tab should not. So something very similar to the current spec is needed. But if you think you can make it work, please do so.

fetch.bs
domenic: I don't understand the connection between those two statements.

The thing you deleted is a general note about readable stream reading, and not related to canceling.

fetch.bs
domenic: Oh, well, "abort these steps" is generally pretty unclear, which is why we're trying to move away from it. So maybe "Abort the in parallel steps", or set a boolean and use that to guard subsequent steps.

Also, it seems you're forgetting to close the connection, if you truly do abort the steps.

fetch.bs
jakearchibald: Ah ok, I'll add it back

fetch.bs
jakearchibald: I guess we could see the `AbortError` coming from the service worker and turn that into a network error with some sort of abort flag set.

But things like "timeout" seem specific to XHR (right now), so having them in fetch isn't great.

jakearchibald:
I've removed termination reasons and replaced it with an "aborted flag". The next step is to PR XHR so it terminates fetch correctly and reads the aborted flag when it decides whether to fire the abort event or not.

I also noticed that the trailer promise never resolves if the response fails (without aborting). I may as well fix that while I'm here.

jakearchibald:
XHR PR is at https://github.com/whatwg/xhr/pull/152. I've also made sure the trailer promise settles.

jakearchibald:
Btw, I don't have anything else to add to this afaik.

fetch.bs
```diff
@@ -1311,6 +1327,10 @@ navigate algorithm. It ensures `<code>Location</code>` has
 <hr>
 
 <p>A <a for=/>response</a> whose
+<a for=response>type</a> is "<code>error</code>" and <a for=response>aborted flag</a> is set is
+known as a <dfn export id=concept-aborted-network-error>aborted network error</dfn>.
```
annevk: an*

fetch.bs
```diff
@@ -1488,21 +1508,38 @@ for each associated <a for="fetch group">fetch record</a> whose
...
+    failure.
+  </ol>
 
- <li><p>Add <var>connection</var> to the <a for=connection>connection pool</a>
- with <b>origin</b> being <var>origin</var> and <b>credentials</b> being <var>credentials</var>.
+ <li>
+  <p>If the ongoing fetch is <a for=fetch>terminated</a>, then:
+
+  <ol>
+   <li><p>If <var>connection</var> is not null, close <var>connection</var>.
```
annevk: then close*

fetch.bs
```diff
@@ -4695,6 +4861,7 @@ dictionary RequestInit {
   RequestRedirect redirect;
   DOMString integrity;
   boolean keepalive;
+  AbortSignal? signal;
```
annevk: Why is this nullable and the readonly attribute is not? What does null signify vs undefined?

jakearchibald:
```js
const request = new Request(otherRequest);
```

If the signal passed to `otherRequest` aborts, so will `request`.

```js
const request = new Request(otherRequest, {signal: null});
```

If the signal passed to `otherRequest` aborts, `request` will not abort.

annevk: And for the second scenario `request.signal` will still return an `AbortSignal`? That makes sense then, thanks.

jakearchibald: Yep!

jakearchibald:
Commit message:

Abortable fetch
Integrating `AbortSignal`.
Specifying abort points and how to handle termination.
Replacing termination reason with aborted flag.

annevk:
We need a little more, such as that this fixes #447.

Tests are at https://github.com/w3c/web-platform-tests/pull/6484.

Follow-up is https://github.com/w3c/ServiceWorker/pull/1178?

What's the status of #563? Is that follow-up?

jakearchibald:
Updated the message. Will update #563.

annevk:
Alright, thanks @jakearchibald!

jakearchibald:
Whoop! *collapses*

sideshowbarker:
For the record here, the following are the related browser feature bugs:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1378342<br>(resolved=fixed and shipped in Firefox 57)
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13009916/<br>(Status=“Fixed, not yet flighted”)
* https://bugs.chromium.org/p/chromium/issues/detail?id=750599
* https://bugs.webkit.org/show_bug.cgi?id=174980

PR #532, opened 2017-04-25
Queue "process request body" tasks during request body transmission

### LINKED ISSUE #495: "process request body" is gone
It seems that because of Streams we no longer need "process request body". However, that does mean we need to update XMLHttpRequest and Fetch so it's no longer used.

Found while investigating https://github.com/whatwg/xhr/issues/118.

yutakahirano:
With this change the "transmit body" algorithm queues "process request body" tasks so that the client can notice the body transmission progress.

Fixes #495.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/process-request/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/9289687...f523b19.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1187,16 +1187,19 @@ or "<code>worker</code>".
    <li>
     <p>When <var>read</var> is fulfilled with an object whose <code>done</code>
     property is false and whose <code>value</code> property is a
-    <code>Uint8Array</code> object, run these substeps:
+    <code>Uint8Array</code> object, run these substeps <a>in parallel</a>:
 
     <ol>
-     <li><p>Let <var>bytes</var> be the <a>byte sequence</a> represented by the
+     <li><p>Let <var>bs</var> be the <a>byte sequence</a> represented by the
      <code>Uint8Array</code> object.
 
-     <li><p>Transmit <var>bytes</var>.
+     <li>
+      <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
+      <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
+      <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
+      for <var>request</var>.
 
-     <li><p>Increase <var>body</var>'s <a for=body>transmitted bytes</a> by <var>bytes</var>'s
-     length.
+      <p class="note no-backref">This substep blocks until the entire <var>bs</var> is transmitted.
 
      <li><p>Run the above step again.
     </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -1170,43 +1170,6 @@ or "<code>worker</code>".
 
 <hr>
 
-<p>To <dfn export for=request id=concept-request-transmit-body-chunk>transmit a chunk</dfn> for a
-promise <var>p</var>, run these steps <a>in parallel</a>:
-
-<ol>
- <li><p>Wait for <var>p</var> to be fulfilled or rejected.
-
- <li>
-  <p>If <var>p</var> is fulfilled with an object whose <code>done</code> property is false
-  and whose <code>value</code> property is a <code>Uint8Array</code> object, run these steps:
-
-   <ol>
-    <li><p>Let <var>bs</var> be the <a>byte sequence</a> represented by the <code>Uint8Array</code>
-    object.
-
-    <li>
-     <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
-     <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
-     <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
-     for <var>request</var>.
-
-     <p class="note no-backref">This step blocks until <var>bs</var> is fully transmitted.
-
-    <li><p>Let <var>read</var> be the result of <a lt="read a chunk" for=ReadableStream>reading a
-    chunk</a> from <var>body</var>'s <a for=body>stream</a>.
-
-    <li><p><a for=request>Transmit a chunk</a> with <var>read</var>.
-   </ol>
-
- <li><p>If <var>p</var> is fulfilled with an object whose <code>done</code> property is true,
- <a>queue a fetch task</a> on <var>request</var> to <a>process request end-of-body</a> for
- <var>request</var>.
-
- <li><p>If <var>p</var> is fulfilled with a value that matches with neither of the above patterns,
- or <var>read</var> is rejected, <a lt=terminated for=fetch>terminate</a> the ongoing fetch with
- reason <i>fatal</i>.
-</ol>
-
 <p>To <dfn export for=request id=concept-request-transmit-body>transmit body</dfn> for a
 <a for=/>request</a> <var>request</var>, run these steps:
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1187,16 +1187,19 @@ or "<code>worker</code>".
...
+      <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
+      <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
+      <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
+      for <var>request</var>.
 
-     <li><p>Increase <var>body</var>'s <a for=body>transmitted bytes</a> by <var>bytes</var>'s
-     length.
+      <p class="note no-backref">This substep blocks until entire <var>bs</var> is transmitted.
 
      <li><p>Run the above step again.
```
annevk: Which step does this refer to?

fetch.bs
```diff
@@ -1187,16 +1187,19 @@ or "<code>worker</code>".
    <li>
     <p>When <var>read</var> is fulfilled with an object whose <code>done</code>
     property is false and whose <code>value</code> property is a
-    <code>Uint8Array</code> object, run these substeps:
+    <code>Uint8Array</code> object, run these substeps <a>in parallel</a>:
```
annevk: Can we use "steps" here rather than "substeps"?

fetch.bs
```diff
@@ -1187,16 +1187,19 @@ or "<code>worker</code>".
...
-     <li><p>Transmit <var>bytes</var>.
+     <li>
+      <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
+      <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
+      <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
+      for <var>request</var>.
 
-     <li><p>Increase <var>body</var>'s <a for=body>transmitted bytes</a> by <var>bytes</var>'s
-     length.
+      <p class="note no-backref">This substep blocks until entire <var>bs</var> is transmitted.
```
annevk: s/entire bs is/bs is fully/

fetch.bs
yutakahirano: I use the term "step" for outer steps (e.g., "Let _body_ be _request_'s body") and "substep" for inner steps (e.g., "Transmit _bs_"). The below "Run the above step again." relies on that.

fetch.bs
yutakahirano: "Let read be the result of reading a chunk from body’s stream."

I'm trying to describe an async loop here.

annevk:
I don't think it quite works for me and we try to get away from using "substep" since everyone uses it differently and it doesn't always carry specific meaning.

I don't have any good suggestions for asynchronous loops however. @domenic?

annevk:
Relevant algorithms: https://fetch.spec.whatwg.org/#concept-request-transmit-body and https://fetch.spec.whatwg.org/#concept-read-all-bytes-from-readablestream.

domenic:
This seems like a good way to describe the async loop (basically using recursion disguised as a goto). It indeed might help to label the step. You could alternately pull it out into a "function" to avoid the goto, but I think it's OK as-is.

domenic:
Oh, no, I meant the async loop part, getting rid of substep is good.

yutakahirano:
Fixed.

> You could alternately pull it out into a "function" to avoid the goto, but I think it's OK as-is.
I chose this option.

fetch.bs
```diff
@@ -1170,45 +1170,56 @@ or "<code>worker</code>".
...
+ <li><p>If <var>p</var> is fulfilled with an object whose <code>done</code> property is true,
+ <a>queue a fetch task</a> on <var>request</var> to <a>process request end-of-body</a> for
+ <var>request</var>.
 
-   <li><p>When <var>read</var> is fulfilled with a value that matches with neither of the
-   above patterns, or <var>read</var> is rejected, <a lt=terminated for=fetch>terminate</a> the
-   ongoing fetch with reason <i>fatal</i>.
-  </ul>
+ <li><p>If <var>p</var> is fulfilled with a value that matches with neither of the above patterns,
+ or <var>read</var> is rejected, <a lt=terminated for=fetch>terminate</a> the ongoing fetch with
```
annevk: "or read is rejected" isn't needed here, right? That already follows from p being fulfilled with a "wrong" value.

fetch.bs
```diff
@@ -1170,45 +1170,56 @@ or "<code>worker</code>".
...
-
- <li><p>If <var>body</var> is null, then <a>queue a fetch task</a> on <var>request</var> to
- <a>process request end-of-body</a> for <var>request</var> and abort these steps.
+ <li><p>Wait for <var>p</var> to be fulfilled or rejected.
 
  <li>
-  <p>Let <var>read</var> be the result of <a lt="read a chunk" for=ReadableStream>reading a
-  chunk</a> from <var>body</var>'s <a for=body>stream</a>.
+  <p>If <var>p</var> is fulfilled with an object whose <code>done</code> property is false
+  and whose <code>value</code> property is a <code>Uint8Array</code> object, run these steps:
```
annevk: then run*

fetch.bs
```diff
@@ -1170,45 +1170,56 @@ or "<code>worker</code>".
 
 <hr>
 
-<p>To <dfn export for=request id=concept-request-transmit-body>transmit body</dfn> for a
-<a for=/>request</a> <var>request</var>, run these steps:
+<p>To <dfn export for=request id=concept-request-transmit-body-chunk>transmit a chunk</dfn> for a
+promise <var>p</var>, run these steps <a>in parallel</a>:
```
annevk: It seems you still need to pass request or something as an argument here, since otherwise what do you end up doing the recursion on? _body_ below appears out of nowhere.

annevk: The original name "transmit body" might still be good therefore too. It just takes a promise as additional argument.

yutakahirano:
Hmm, after seeing https://github.com/whatwg/fetch/pull/532#discussion_r114273329, I came to think that giving a name to the internal operation ("transmit a chunk" is hard to understand). The newest change is written with almost synchronous style. What do you think about it?

fetch.bs
```diff
@@ -1179,36 +1179,42 @@ or "<code>worker</code>".
  <li><p>If <var>body</var> is null, then <a>queue a fetch task</a> on <var>request</var> to
  <a>process request end-of-body</a> for <var>request</var> and abort these steps.
 
+ <li><p>Let <var>read</var> be the result of <a lt="read a chunk" for=ReadableStream>reading a
+ chunk</a> from <var>body</var>'s <a for=body>stream</a>.
+
  <li>
-  <p>Let <var>read</var> be the result of <a lt="read a chunk" for=ReadableStream>reading a
-  chunk</a> from <var>body</var>'s <a for=body>stream</a>.
+  <p>Run these steps repeatedly, <a>in parallel</a>
```
annevk: I'd make this "In parallel, while true:" to make it clear you go in parallel once and then keep iterating.

fetch.bs
```diff
@@ -1179,36 +1179,42 @@ or "<code>worker</code>".
...
+   <li><p>Otherwise, if <var>read</var> is fulfilled with an object whose <code>done</code> property
+   is true, then <a>queue a fetch task</a> on <var>request</var> to
+   <a>process request end-of-body</a> for <var>request</var> and abort these steps.
 
-   <li><p>When <var>read</var> is fulfilled with a value that matches with neither of the
-   above patterns, or <var>read</var> is rejected, <a lt=terminated for=fetch>terminate</a> the
-   ongoing fetch with reason <i>fatal</i>.
-  </ul>
+   <li><p>Otherwise, <a lt=terminated for=fetch>terminate</a> the ongoing fetch with reason
+   <i>fatal</i> and abort these steps.
```
annevk: One day we should make it clearer what the "ongoing fetch" is, but this was an existing bug so this is fine.

fetch.bs

PR #537, opened 2017-05-03
Editorial: rename "basic fetch" to "scheme fetch"

jugglinmike:
As it relates to the fetching algorithm, the term "basic" is sub-optimal
for a number of reasons:

- It implies a relative level of simplicity. Because the algorithm
  includes an invocation of HTTP Fetch, it is conceptually more complex
  than most other fetching algorithms defined in the specification.
- It suggests a relationship to the "basic" request tainting value and
  "basic" filtered response. While the algorithm is invoked for requests
  whose tainting is "basic", it is also invoked for requests whose
  tainting is set to "opaque".

Re-name the algorithm to "scheme fetch" in order to more precisely
describe its behavior while avoiding association with any particular
request tainting value.

---

@annevk I'm probably only thinking along these lines because of my inexperience
with the specification, but I know that same inexperience makes it easier for
me to stomach vocabulary changes like this. These details certainly have
momentum for folks who have been working in this space for a while. My bias
notwithstanding, I do think this would make the text easier to follow. Other
alternatives that would address my concern are "scheme fetch" or "scheme-based
fetch", though those may be too technical/wordy.

***
[Preview](https://s3.amazonaws.com/pr-preview/bocoup/fetch/generic-fetch.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/ba50491...bocoup:ec0bb4f.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2516,7 +2516,7 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
      <li><p>Set <var>request</var>'s
      <a for=request>response tainting</a> to "<code>basic</code>".
 
-     <li><p>Return the result of performing a <a for=basic>basic fetch</a>
+     <li><p>Return the result of performing a <a>scheme fetch</a>
      using <var>request</var>.
     </ol>
 
@@ -2539,7 +2539,7 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
      <a for=request>response tainting</a> to
      "<code>opaque</code>".
 
-     <li><p>Return the result of performing a <a for=basic>basic fetch</a>
+     <li><p>Return the result of performing a <a>scheme fetch</a>
      using <var>request</var>.
      <!-- file URLs end up here as they are not same-origin typically. -->
     </ol>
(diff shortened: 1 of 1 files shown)


annevk:
I like "scheme fetch". We should probably not change the IDs though or at the very least preserve the old IDs. The `for` attribute used for this should also be removed throughout. If you could take care of that while you're there that would be great.

jugglinmike:
Sure thing. This required amending the commit message, so I've made the original version of this branch (including a fixup commit with the requested changes) available here: https://github.com/whatwg/fetch/compare/master...bocoup:generic-fetch-orig

fetch.bs
```diff
@@ -2737,10 +2737,12 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
  for <var>response</var>.
 </ol>
 
+<!-- Prior to May 2017, this algorithm was named "basic fetch." The
+     corresponding ID is maintained in order to preserve the functionality of
+     external references. -->
+<h3 id=scheme-fetch><span id=basic-fetch>Scheme fetch</span></h3>
```
annevk: `<h3 id=scheme-fetch oldids=basic-fetch>`.

fetch.bs
```diff
@@ -2737,10 +2737,12 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
...
 
-<h3 id=basic-fetch>Basic fetch</h3>
-
-<p>To perform a <dfn id=concept-basic-fetch for=basic>basic fetch</dfn> using
+<p>To perform a <dfn id=concept-scheme-fetch>scheme fetch</dfn> using
```
annevk: Add `oldids=concept-basic-fetch` here too. I suspect both of these to have external links.

jugglinmike:
Wow, Bikeshed is a great tool! I've shorted the accompanying comment as it is now clear that this practice is pretty common. Force-pushed to facilitate merging, but the individual commits are available on the branch I reference above.

annevk:
Thanks!

jugglinmike:
Sure thing :)

PR #545, opened 2017-05-18
Small cors typo

domfarolino:
Seems to make more sense to say "...to declare they can be shared with other **origins**" instead of **origin**. Linking still works with this change, but I'm curious as to why? Does bs just know how to link dfns to their pluralized references?

***
[Preview](https://s3.amazonaws.com/pr-preview/domfarolino/fetch/cors-typo.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/6e68ce9...domfarolino:1786dcf.html)


annevk:
Thanks!

Issue #555, opened 2017-06-19
What is expected effect of setting Request type and destination? For example, to "audio"?


guest271314:
What is expected effect of setting `Request` [`type`](https://fetch.spec.whatwg.org/#concept-request-type) and [`destination`](https://fetch.spec.whatwg.org/#concept-request-destination)? For example, to `"audio"` when requesting an audio resource? 

[5.3. Request class](https://fetch.spec.whatwg.org/#request-class)
```

  readonly attribute RequestType type;
  readonly attribute RequestDestination destination;

```

wanderview:
Javascript code cannot set the `type` or `destination` properties.  They are readonly and the `RequestInit` dictionary does not support them.

These properties are set by the browser for internally generated requests in order to apply various policies to the request.  For example, some types/destinations are allowed mixed content while others are not, etc.

guest271314:
@wanderview What do you mean by "set by the browser for internally generated requests"? The browser uses `Request` internally? 

The context is an attempt to use `fetch()` to request a media resource with media fragment uri concatenated to URL string, where expected result is for only that portion of the partial content to be returned at `.blob()` or `.arrayBuffer()` call. Probably not possible, though that is what tried, here, when read the specification where `type` and `destination` were mentioned. 

Related https://github.com/whatwg/fetch/issues/554

wanderview:
> What do you mean by "set by the browser for internally generated requests"? The browser uses Request internally?

The browser uses an internal `request` that is reflected to javascript as the `Request` DOM object.  The type and destination are set directly on the internal request.  The rules about what to set and when are summarized in the table here:

https://fetch.spec.whatwg.org/#concept-request-destination

guest271314:
@wanderview  Yes. Does `HTMLMediaElement` call internal `request` at 

`<audio controls src="https://upload.wikimedia.org/wikipedia/commons/b/be/Hidden_Tribe_-_Didgeridoo_1_Live.ogg#t=30,50"></audio>
`
?

How can we port the media fragment uri functionality to `Request`?

wanderview:
It should, but we don't implement `type` or `destination` yet in gecko.  I'm not sure about chrome.  This part of the spec went through some changes so implementation has lagged.  In the past it was called `context` and there was a lot of confusion about what values it should contain.

I'm not sure I know enough about media fragments to say, but it seems @jakearchibald has suggested we should just better support range requests.

guest271314:
[...]

Issue #554, opened 2017-06-15
Body.arrayBuffer([begin, end])


beaufortfrancois:
It would be very practical if `arrayBuffer()` method could take `begin` and `end` optional parameters in order to give the ability to reduce memory usage for cache responses whose content is "huge" but can still be useful in chunks (video and audio).

```javascript
const response = await caches.match('https://example.com/huge-file.mp4');
const slicedData = await response.arrayBuffer(0, 1024);
return new Response(slicedData);
```

R: @jakearchibald
FYI @paullewis

mkruisselbrink:
How is that better than getting the body as a blob, and then slicing and dicing that in whatever way you want? And would calling arrayBuffer like that still consume the whole body, meaning effectively you should have done a range request since you're never going to be able to read the rest of the body?

guest271314:
Are you trying to request and only receive response having `Content-Length` of `1024`? Or, read response in chunks of `1024`?

beaufortfrancois:
I've updated my code snippet to better reflect what I'm experiencing. A huge file is in my cache and I would like to return only a subset of it on a fetch event instead of bringing the entire file in memory.

Both samples are equivalent as far as I can test in my browser in term of memory usage:

```javascript
const response = await caches.match('https://example.com/huge-file.mp4');
const data = await response.arrayBuffer();
const sliceData = data.slice(0, 1024);
```
 
```javascript
const response = await caches.match('https://example.com/huge-file.mp4');
const blob = await response.blob();
const sliceData = blob.slice(0, 1024);
```

asutherland:
The blob() case is more efficient by far in terms of what the browser can and does optimize.  Speaking for Gecko, the blob() call only creates a handle to the underlying file on disk.  No reads of the file's contents need to be performed before returning the blob.  This handle can be given to the new Response and passed across processes, allowing the target process to perform just the 1024-byte read directly, and without needing to involve the ServiceWorker or its global.

In contrast, the arrayBuffer() call will result in the entirety of "huge-file.mp4" being read from disk and exposed to the ServiceWorker and its global.  The read needs to complete before the new Response can be created and returned, and its contents may then need to be streamed between processes (unless some kind of underlying shared memory/copy-on-write thing is done, which I'm pretty confident Gecko will not do at the current time).

jakearchibald:
I think https://github.com/w3c/ServiceWorker/issues/913 is the solution here.

guest271314:
[...]

PR #560, opened 2017-06-29
Allow range header to be set by APIs

jakearchibald:
This is part of #144.

The aim is to allow APIs to use the range header for no-cors requests, and allow them to pass through a service worker, but disallow modification of these requests, and disallow developers creating their own no-cors ranged requests.

***
<a href="https://whatpr.org/fetch/560.html" title="Last updated on May 29, 2018, 11:31 AM GMT (e1ad29d)">Preview</a> | <a href="https://whatpr.org/fetch/560/bd9b4e3...e1ad29d.html" title="Last updated on May 29, 2018, 11:31 AM GMT (e1ad29d)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1341,13 +1341,13 @@ run these steps:
 <ol>
  <li><p>Let <var>rangeValue</var> be `<code>bytes </code>`.
 
- <li><p><a lt="serialize an integer">Serialize</a> and <a>UTF-8 encode</a> <var>first</var>, and
- append the result to <var>rangeValue</var>.
+ <li><p><a lt="serialize an integer">Serialize</a> and <a>isomorphic encode</a> <var>first</var>,
+ and append the result to <var>rangeValue</var>.
 
  <li><p>Append 0x2D (-) to <var>rangeValue</var>.
 
  <li><p>If <var>last</var> is given, then <a lt="serialize an integer">serialize</a> and
- <a>UTF-8 encode</a> it, and append the result to <var>rangeValue</var>.
+ <a>isomorphic encode</a> it, and append the result to <var>rangeValue</var>.
 
  <li><p><a for="header list">Append</a> `<code>Range</code>`/<var>rangeValue</var> to
  <var>request</var>'s <a for=request>header list</a>.
@@ -3556,7 +3556,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <a for=request>body</a>'s <a for=body>source</a> is non-null, then set
    <var>contentLengthValue</var> to <var>httpRequest</var>'s <a for=request>body</a>'s
    <a for=body>total bytes</a>, <a lt="serialize an integer">serialized</a> and
-   <a>UTF-8 encoded</a>.
+   <a>isomorphic encoded</a>.
 
    <li><p>If <var>contentLengthValue</var> is non-null,
    <a for="header list">append</a>
--- a/fetch.bs
+++ b/fetch.bs
@@ -197,6 +197,12 @@ of:
 <a>queue a fetch task</a> on <var>request</var> to <a>process request end-of-body</a>
 for <var>request</var>.
 
+<p>To <dfn>serialize an integer</dfn>, represent it as a string of the shortest possible decimal
+number.
+
+<p class=XXX>This will be replaced by a more descriptive algorithm in Infra. See
+<a href="https://github.com/whatwg/infra/issues/201">infra/201</a>.
+
 
 <h3 id=url>URL</h3>
 
@@ -442,6 +448,24 @@ documented in <a href=#cors-protocol-exceptions>CORS protocol exceptions</a>.
 
 <p>A <dfn export>CORS non-wildcard request-header name</dfn> is a <a>byte-case-insensitive</a> match
 for `<code>Authorization</code>`.
+
+<p>A <dfn export>privileged no-cors request-header name</dfn> is a <a for=/>header</a>
+<a for=header>name</a> that is a <a>byte-case-insensitive</a> match for one of
+
+<ul class=brief>
+ <li>`<code>Range</code>`.
+</ul>
+
+<div class="note no-backref">
+ <p>These are headers that can be set by privileged APIs, and will be preserved if their associated
+ request object is copied, but will be removed if the request is modified by unprivilaged APIs.
+
+ <p>`<code>Range</code>` headers are commonly used by <a lt="downloads a hyperlink">downloads</a>
+ and <a lt="resource fetch algorithm">media fetches</a>, although neither of these currently specify
+ how. <a href=https://github.com/whatwg/html/pull/2814>html/2914</a> aims to solve this.
+
+ <p>A helper is provided to <a for=request>add a range header</a> to a particular request.
+</div>
 
 <p>A <dfn export>CORS-safelisted response-header name</dfn>, given a
 <a for=response>CORS-exposed header-name list</a> <var>list</var>, is a <a for=/>header</a>
@@ -1308,6 +1332,33 @@ run these steps:
   </ol>
 </ol>
 
+<hr>
+
+<p>To <dfn export for=request id=concept-request-add-range-header>add a range header</dfn> to a
+<a for=/>request</a> <var>request</var>, with an integer <var>first</var>, and an optional integer
+<var>last</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>rangeValue</var> be `<code>bytes </code>`.
+
+ <li><p><a lt="serialize an integer">Serialize</a> and <a>UTF-8 encode</a> <var>first</var>, and
+ append the result to <var>rangeValue</var>.
+
+ <li><p>Append 0x2D (-) to <var>rangeValue</var>.
+
+ <li><p>If <var>last</var> is given, then <a lt="serialize an integer">serialize</a> and
+ <a>UTF-8 encode</a> it, and append the result to <var>rangeValue</var>.
+
+ <li><p><a for="header list">Append</a> `<code>Range</code>`/<var>rangeValue</var> to
+ <var>request</var>'s <a for=request>header list</a>.
+</ol>
+
+<p class=note>A range header denotes an inclusive byte range. There a range header where
+<var>first</var> is 0 and <var>last</var> is 500, is a range of 501 bytes.
+
+<p class=note>Features that combine multiple responses into one logical resource are historically a
+source of security bugs. Please seek security review for features that deal with partial responses.
+
 
 <h4 id=responses>Responses</h4>
 
@@ -1402,6 +1453,14 @@ specified. [[!CSP]]
 <a for=response>CORS-exposed header-name list</a> set by <a>extracting header values</a> from the
 `<a http-header><code>Access-Control-Expose-Headers</code></a>` header. This list is used by a
 <a>CORS filtered response</a> to determine which headers to expose.
+
+<p>A <a for=/>response</a> has an associated
+<dfn for=response id=concept-response-range-requested-flag>range-requested flag</dfn>, which is
+initially unset.
+
+<p class=note>This is used to ensure to prevent a partial response from an earlier ranged request
+being provided to an API that didn't make a range request. See the flag's usage for a detailed
+description of the attack.
 
 <p>A <a for=/>response</a> can have an associated
 <dfn export for=response id=concept-response-location-url>location URL</dfn> (null, failure, or a
@@ -2930,6 +2989,31 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
   </ul>
 
  <li>
+  <p>If <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>",
+  <var>internalResponse</var>'s <a for=response>status</a> is <code>206</code>,
+  <var>internalResponse</var>'s <a for=response>range-requested flag</a> is set, and
+  <var>request</var>'s <a for=request>header list</a> does not <a for="header list">contain</a>
+  `<code>Range</code>`, then set <var>response</var> and <var>internalResponse</var> to a
+  <a>network error</a>.
+
+  <div class=note>
+   <p>Traditionally, APIs accept a ranged response even if a range was not requested. This prevents
+   a partial response from an earlier ranged request being provided to an API that did not make a
+   range request.
+
+   <details>
+    <summary>Further details</summary>
+
+    <p>The above steps prevent the following attack attack:
+
+    <p>A media element is used to request a range of a cross-origin HTML resource. Although this is
+    invalid media, a reference to a clone of the response can be retained in a service worker. This
+    can later be used as the response to a script element's fetch. If the partial response is valid
+    JavaScript (even though the whole resource is not), executing it would leak private data.
+   </details>
+  </div>
+
+ <li>
   <p>If <var>response</var> is not a <a>network error</a> and
   either <var>request</var>'s <a for=request>method</a> is
   `<code>HEAD</code>` or `<code>CONNECT</code>`, or <var>internalResponse</var>'s
@@ -3471,7 +3555,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li><p>If <var>httpRequest</var>'s <a for=request>body</a> is non-null and <var>httpRequest</var>'s
    <a for=request>body</a>'s <a for=body>source</a> is non-null, then set
    <var>contentLengthValue</var> to <var>httpRequest</var>'s <a for=request>body</a>'s
-   <a for=body>total bytes</a>, <a>UTF-8 encoded</a>.
+   <a for=body>total bytes</a>, <a lt="serialize an integer">serialized</a> and
+   <a>UTF-8 encoded</a>.
 
    <li><p>If <var>contentLengthValue</var> is non-null,
    <a for="header list">append</a>
@@ -3750,6 +3835,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
       the network error, which is sometimes known as "negative caching".
     </ol>
   </ol>
+
+ <li><p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">contains</a>
+ `<code>Range</code>`, then set <var>response</var>'s <a for=response>range-requested flag</a>.
 
  <li>
   <p>If <var>response</var>'s <a for=response>status</a> is <code>401</code>, <i>CORS flag</i>
@@ -4521,9 +4609,6 @@ objects.</span>
 "<code>response</code>" or
 "<code>none</code>".
 
-<p class="note no-backref">"<code>immutable</code>" exists for service workers.
-[[!SW]]
-
 <p>To <dfn export for=Headers id=concept-headers-append>append</dfn> a
 <a for=header>name</a>/<a for=header>value</a>
 (<var>name</var>/<var>value</var>) pair to a
@@ -4542,9 +4627,8 @@ objects.</span>
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/<var>value</var> is not a
- <a>CORS-safelisted request-header</a>, return.
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>" and
+ <var>name</var>/<var>value</var> is not a <a>CORS-safelisted request-header</a>, then return.
 
  <li><p>Otherwise, if <a for=Headers>guard</a> is
  "<code>response</code>" and <var>name</var> is a
@@ -4553,6 +4637,9 @@ objects.</span>
  <li><p><a for="header list">Append</a>
  <var>name</var>/<var>value</var> to
  <a for=Headers>header list</a>.
+
+ <li><p>If <var>headers</var>'s <a for=Headers>guard</a> is "<code>request-no-cors</code>", then
+ <a for=Headers>remove privileged no-cors request headers</a> from <var>headers</var>.
 </ol>
 
 <p>To <dfn export for=Headers id=concept-headers-fill>fill</dfn> a
@@ -4576,6 +4663,22 @@ run these steps:
  <var>key</var> → <var>value</var> in <var>object</var>, <a lt=append for=Headers>append</a>
  <var>key</var>/<var>value</var> to <var>headers</var>.
 </ol>
+
+<p>To
+<dfn for=Headers id=concept-headers-remove-privileged-no-cors-request-headers>remove privileged no-cors request headers</dfn>
+from a {{Headers}} object (<var>headers</var>), run these steps:
+
+<ol>
+ <li><p><a for="list">For each</a> <var>headerName</var> of
+ <a>privileged no-cors request-header names</a>:
+
+ <ol>
+  <li><p><a for="header list">Delete</a> <var>headerName</var> from <var>headers</var>'s
+  <a for=Headers>header list</a>.
+ </ol>
+</ol>
+
+<p class=note>This is called when headers are modified by unprivileged code.
 
 <p>The
 <dfn id=dom-headers export for=Headers constructor><code>Headers(<var>init</var>)</code></dfn>
@@ -4609,9 +4712,10 @@ method, when invoked, must run these steps:
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/`<code>invalid</code>` is
- not a <a>CORS-safelisted request-header</a>, then return.
+ <li>
+  <p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>",
+  <var>name</var>/`<code>invalid</code>` is not a <a>CORS-safelisted request-header</a>, and
+  <var>name</var> is not a <a>privileged no-cors request-header name</a>, then return.
 
   <p class=note>`<code>invalid</code>` is used because
   <a method for=Headers><code>delete()</code></a> is not passed a value as argument.
@@ -4620,8 +4724,14 @@ method, when invoked, must run these steps:
  "<code>response</code>" and <var>name</var> is a
  <a>forbidden response-header name</a>, return.
 
+ <li><p>If <a for=Headers>header list</a> does not <a for="header list">contain</a> <var>name</var>,
+ then return.
+
  <li><p><a for="header list">Delete</a> <var>name</var> from
  <a for=Headers>header list</a>.
+
+ <li><p>If <a for=Headers>guard</a> is "<code>request-no-cors</code>", then
+ <a for=Headers>remove privileged no-cors request headers</a> from the <a>context object</a>.
 </ol>
 
 <p>The <dfn export for=Headers method><code>get(<var>name</var>)</code></dfn> method, when
@@ -4666,9 +4776,8 @@ method, when invoked, must run these steps:
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/<var>value</var> is not a
- <a>CORS-safelisted request-header</a>, return.
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>" and
+ <var>name</var>/<var>value</var> is not a <a>CORS-safelisted request-header</a>, then return.
 
  <li><p>Otherwise, if <a for=Headers>guard</a> is
  "<code>response</code>" and <var>name</var> is a
@@ -4677,6 +4786,9 @@ method, when invoked, must run these steps:
  <li><p><a for="header list">Set</a>
  <var>name</var>/<var>value</var> in
  <a for=Headers>header list</a>.
+
+ <li><p>If <a for=Headers>guard</a> is "<code>request-no-cors</code>", then
+ <a for=Headers>remove privileged no-cors request headers</a> from the <a>context object</a>.
 </ol>
 
 <p>The <a>value pairs to iterate over</a> are the return value of running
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4219,9 +4251,6 @@ objects.</span>
 "<code>response</code>" or
 "<code>none</code>".
 
-<p class="note no-backref">"<code>immutable</code>" exists for service workers.
-[[!SW]]
```
annevk: Was this intentional?

fetch.bs
```diff
@@ -4240,6 +4269,17 @@ objects.</span>
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>" and
```
annevk: Newline after `<li>` as it contains multiple children.

fetch.bs
```diff
@@ -4839,14 +4892,17 @@ constructor must run these steps:
 
    <li><p>Set <var>request</var>'s
```
annevk: Newline due to multiple children again.

fetch.bs
```diff
@@ -4839,14 +4892,17 @@ constructor must run these steps:
...
    <a for=request>referrer policy</a> to the empty string.
+   <p class=note>This is done to ensure that when a service worker "redirects" a request, .e.g.,
```
annevk: No dot before e.g.

fetch.bs
jakearchibald: Yeah, "immutable" is also used for `fetch()` responses. I could add the comment back and include the other places "immutable" is used.

jakearchibald:
> The idea is that media APIs will set this flag?

And downloads, yeah.

I'm not quite sure how to spec that bit yet, I guess there'll be a section like "this API may issue range requests. To fetch a range, run the following steps…"

Those steps will detail how to handle ranges that don't match what was requested, and ensuring all parts come from the same source if any opaque response is involved. Also how to handle an HTTP 200 response.

For downloads I want to ensure that browsers restart the download if the length, etag or last-modified has changed. Firefox is the only browser doing a decent job here, the others happy merge the chunks together even though the content has changed.

guest271314:
@jakearchibald The `Range` header is set at `fetch()` headers. However, the issue here is that if the first byte is not `0` the media does not playback.

jakearchibald:
@guest271314 I don't think your analysis matches what the fetch spec is doing. See https://github.com/whatwg/fetch/issues/144

guest271314:
@jakearchibald You are probably correct. Have minimal experience with `ServiceWorker`. The naive expectation here was that a request for a media resource using `fetch()` - whether media fragment URI, or setting range header - would result in a `Blob` or `ArrayBuffer` comprising the only bytes, or time ranges as bytes specified. That is, `fetch("/media#t=30,60")` should have same result as `new Audio("/media#t=30,60")`. Not sure what the security risks associated with range requests are?

annevk:
@foolip mostly if these changes are compatible with eventual changes for the media API in HTML.

annevk:
@jakearchibald I guess we can't really test this change without defining some of the other parts? I wonder if we just want to land this (assuming it's editorially okay) and test later or wait until we have defined some of the integration bits.

jakearchibald:
@annevk I can write tests for the cases that must produce a network error. Eg returning a ranged response to a request that didn't have a range header.

Aside from that, given that the intent is to *prevent* script-created no-cors requests with range headers, it might not be possible to automate the tests. Seeking an audio element should produce enough range requests for testing, but the browser doesn't have to make range requests to be spec-compliant.

What's best for me to do next? Make a couple of tests for this PR, or start on the HTML changes?

jakearchibald:
@guest271314 media fragments are an instruction to the media element, which then makes the appropriate fetches with the correct `Range` headers. It may be many fetches too, as it'll need to download metadata so it can translate the temporal ranges into byte ranges.

`fetch("/media#t=30,60")` only fetches, it doesn't perform media decoding, so it'd be a bit weird if this call made multiple requests, decoded parts of the media, and chopped the response up.

annevk:
@jakearchibald if you're okay with doing HTML first before landing this I think that'd be ideal. HTML will require tests as well so you could also start there. I don't really have advice how to handle the case where a UA would opt not do range requests. Maybe we can make that non-conforming? Not sure there's a good reason for that.

foolip:
> @foolip mostly if these changes are compatible with eventual changes for the media API in HTML.

Probably. I think the non-obvious things that need to work for media elements are:
- Requesting the range 0- must be possible.
- If the response to that is 200, that's not an error, but it is a piece of information that's needed to avoid interrupting the fetch later, as resuming is impossible.
- If more data than requested is returned, that's probably OK, as long as it starts at the right place.
- Cache behavior is strange and probably inconsistent. At the very least, with no cache-related headers, I think implementations assume it is unchanging as long as the resource size doesn't change, or perhaps even a growing resource is OK. Needs testing.

guest271314:
> @guest271314 media fragments are an instruction to the media element, which then makes the appropriate fetches with the correct Range headers. It may be many fetches too, as it'll need to download metadata so it can translate the temporal ranges into byte ranges.
> 
> fetch("/media#t=30,60") only fetches, it doesn't perform media decoding, so it'd be a bit weird if this call made multiple requests, decoded parts of the media, and chopped the response up.

Well, from perspective here, that would not be a "weird" result. That precise application is what was attempting to describe at https://github.com/whatwg/fetch/issues/555. 

That is what initially read when browsing the concept of `type` and `destination` at the specification, though apparently that is not the case.

That is, how to set the context of a `fetch()`, `Request()` or `Response()` call at if the target was a `HTMLMediaElement`. In effect using `fetch()` as a proxy, where the network request would appear to be generated by, for example, `new Audio()`. Not sure if that degree of control is beyond the scope of present inquiry. 

We have `.blob()`, why cannot we have `.mediaFragment()`, i.e.g., https://github.com/whatwg/fetch/issues/554#issuecomment-312331259?

jakearchibald:
Like I said, the media fragment is used to generate the fetch, so if you're asking for a fragment after the fetch, it doesn't work.

This pull request is about allowing range requests to pass through fetch. If you aren't activity disagreeing with it, use other issues to discuss other things.

guest271314:
@jakearchibald Not actively disagreeing. Trying to fully gather what the linked code does https://plnkr.co/edit/HdeRlT25DX3etUMHYTi6?p=preview

foolip:
Media fragments `#t=30,60` as actually implemented doesn't actually interact with the fetch layer at all. Rather, you just try to load the file and then seek to time offset 30. In other words, there's no existing behavior beyond byte range requests that fetch needs to explain. (Disclaimer: I haven't read the whole discussion.)

jakearchibald:
@guest271314 the linked code allows me to observe the kinds of requests the browser makes for media, and how they respond to different types of responses. It helped me figure out the kind of requests the spec should define.

@foolip cheers – that matches my mental model too.

guest271314:
@foolip 

> Media fragments `#t=30,60` as actually implemented doesn't actually interact with the fetch layer at all. Rather, you just try to load the file and then seek to time offset 30. 

`new Audio("/path/to/media#t=30,60")` makes range request, yes? Being a novice at interacting with specification criteria, had the concept of implementing the feature 

```
fetch("/path/to/media#t=30,60"
, {context:HTMLMediaElement}
}) // consistent with new Audio() range request 
.then(response => response.blob())
.then(({destination, blob}) => {
  // do stuff with `Blob` having time slice of `30,60`, or full file with `destination` time slice set to `30,60`
  // `destination` : i.e.g., `<audio>`, `Audio`, `AudioContext`, `<video>`, `MediaSource`, `MediaStream`
})
```

foolip:
> `new Audio("/path/to/media#t=30,60")` makes range request, yes?

It does, but the requests should look roughly the same as when you do this:
```JavaScript
var audio = new Audio("/path/to/media")`
audio.currentTime = 30;
audio.play();
audio.ontimeupdate = function() {
  if (audio.currentTime >= 60)
    audio.pause();
}
```
That's actually fairly close to how this is implemented. The media fragment is handled in the media element code, not at any lower level.

If you want the server to do some of the work in slicing out 30-60 seconds, thenhttps://www.w3.org/TR/media-frags/#URIquery-media-fragments might be what you're looking for, but that then requires a clever server to interpret "/path/to/media?t=30,60".

guest271314:
@foolip No server involved. Something like https://github.com/whatwg/fetch/issues/554#issuecomment-312331259. Perhaps using `MediaRecorder` internally to only return the chunk of media as time slices. Will try to put together a proof of concept for either audio or video.

annevk:
@guest271314 it's a little bit unclear how your feedback thus far relates to this PR. I'd appreciate it if you put it someplace else. This is really only meant for review of the proposed changes.

guest271314:
@annevk Done. Carry on.

jakearchibald:
It seems like browsers will allow a 206 partial response to a `<script src>`. As in, it will execute the script, which this PR would prevent.

The current behaviour seems weird to me. The level of risk is unclear, but if a server could be tricked into thinking your request is a range request (via query string params), and produces a partial response, it could result in data leaking.

With a service worker involved, it means you could take an opaque partial response (from a request generated by a media element) and use it in response to a script fetch. Again, there's a potential for data leak.

Should I try to find a way to preserve browser behaviour here, or go ahead and change it?

mnot:
A server that generates a 206 based upon query parameters is clearly broken and intentionally so; I'd suggest trying to prevent brokenness in that case is a lost cause.

OTOH, a 206 response clearly has an application-visible semantic, as NOT being the whole response. 

AFAIK there isn't any existing markup that allows a `<script>` to specify a range. That says to me that it should specify that a partial response shouldn't be processed.

Relevant spec seems to be here:
  https://html.spec.whatwg.org/#prepare-a-script

... around step 21, which does not appear to talk about success or completeness of the result of fetching.

@annevk?

jakearchibald:
@mnot This PR says the fetch must fail if the request does not have a `Range` header, and the response is 206 partial. Are you saying you disagree with that?

jakearchibald:
Chrome's security team would prefer 206s to fail if a range wasn't requested. However, @ericlaw1979 says there are servers that do this, but the range covers the whole resource https://fiddler.ideas.aha.io/ideas/FID-I-191.

It'd be pretty easy to work around that case in the spec.

mnot:
@jakearchibald ah, sorry, I didn't realise this was a PR :)

That's fine by me, with one caveat; it's also legal to send a 206 in response to a request with `If-Range`.

jakearchibald:
@mnot oh wow, I totally forgot about `If-Range`. Thank you! I've never seen a browser make an `If-Range` request, are you aware of any cases where they do?

ericlaw:
Aren't we still okay here? "A client MUST NOT generate an If-Range header field in a request that does not contain a Range header field.  A server MUST ignore an If-Range header field received in a request that does not contain a Range header field."

jakearchibald:
@ericlaw correct! I thought that `If-Range` was an alternative to `Range`, but as you say, it isn't, so we're fine.

It would make sense for browsers to use `If-Range` when resuming downloads, but none do.

jakearchibald:
I need to take https://github.com/whatwg/fetch/issues/144#issuecomment-323453945 into account when I pick this up again.

wanderview:
I'm a little concerned that this may not match the gecko implementation at all.  Currently we add Range headers at the http cache level.  We do this because we use the http cache to known whether we are resuming a previously seen resource or if the resource has changed.  If the resource has changed we don't use a range request.

@mcmanus Is my reading of range requests in gecko correct here?  Are there other ways that we generate range requests?  For example, do media channels do something different here compared to resuming a download?

wanderview:
Oh, we do set it manually in other places:

https://searchfox.org/mozilla-central/source/dom/html/HTMLMediaElement.cpp#1264

jakearchibald:
Re-reading this, I'm not happy with:

```js
const request = new Request(event.request);
request.headers.set('accept', 'foo');
```

Now the request has the range header from `event.request`, and the modified `accept` header.

The range header will be ditched when it gets copied within `fetch()`, but it feels like privileged headers should be removed if the headers are tampered with, and the guard is `"request-no-cors"`.

fetch.bs
```diff
@@ -5111,42 +5190,47 @@ constructor must run these steps:
...
- <var>request</var>'s
- <a for=request>cache mode</a>,
- <a for=request>redirect mode</a> is
- <var>request</var>'s
- <a for=request>redirect mode</a>,
- <a for=request>integrity metadata</a> is
- <var>request</var>'s
- <a for=request>integrity metadata</a>, and
- <a>keepalive flag</a> is <var>request</var>'s <a>keepalive flag</a>.
+ <li>
```
jakearchibald: This is just a formatting change.

jakearchibald:
@annevk I don't need a spec review for this yet as I'm still working on it, but do you think the direction is right?

Headers *objects* now carry the "allow privileged no-cors headers flag".

`new Request(request)` and `request.clone()` will preserve this flag and the headers.

`new Request(request, { method: 'post' })` (or any other init member) will not preserve the flag. If the mode is no-cors the range header will not be preserved.

Any changes to mutable headers, such as `request.headers.set('foo', 'bar')` will unset the flag and remove the range header.

The "allow privileged no-cors headers flag" will be set in the service worker when it creates the request object for the fetch event. It will be set if the mode is no-cors. My assumption is, if the request has privileged headers at this point, they must have been set by a privileged API.

annevk:
That sounds somewhat reasonable for the request side. How is the flag different from checking for such a header and the mode being "no-cors"?

The response side also needs changes right, to indicate that it was range-requested and such?

jakearchibald:
> How is the flag different from checking for such a header and the mode being "no-cors"?

Right now the flag allows the headers to be carried over when calling `new Request(request)`, but I'll see if I can remove the need for it.

> The response side also needs changes right, to indicate that it was range-requested and such?

Yeah, I'm doing that next.

jakearchibald:
"request-no-cors" header objects can contain 'privileged' headers such as `Range`, if they're set by a privileged API. If the headers are mutated, the privileged headers are removed:

```js
headers.has('Range'); // true
headers.set('Accept-Language', 'Geordie');
headers.get('Accept-Language'); // Geordie
headers.has('Range'); // false
```

But I'm trying to figure out what happens if the developer adds/modifies a privileged header:

* `headers.append('Range', 'blah')` - No-op. Privileged headers remain.
* `headers.delete('Range')` - All privileged headers removed.
* `headers.set('Range', 'blah')` - This is the tricky one. I'm leaning towards removing all privileged headers.

fetch.bs
```diff
@@ -4442,9 +4501,8 @@ objects.</span>
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/<var>value</var> is not a
- <a>CORS-safelisted request-header</a>, return.
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>" and
+ <var>name</var>/<var>value</var> is not a <a>CORS-safelisted request-header</a>, return.
```
annevk: It seems you didn't change this line, either undo the changes or make it say "then return".

fetch.bs
```diff
@@ -4509,9 +4587,9 @@ method, when invoked, must run these steps:
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/`<code>invalid</code>` is
- not a <a>CORS-safelisted request-header</a>, then return.
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>" and
+ <var>name</var>/`<code>invalid</code>` is not a <a>CORS-safelisted request-header</a>, and
+ <var>name</var> is not a <a>privileged no-cors request-header name</a>, then return.
```
annevk: Just one "and" please.

fetch.bs
```diff
@@ -4522,6 +4600,9 @@ method, when invoked, must run these steps:
 
  <li><p><a for="header list">Delete</a> <var>name</var> from
  <a for=Headers>header list</a>.
+
+ <li><p>If <a for=Headers>guard</a> is "<code>request-no-cors</code>", then
+ <a for=Headers>remove privileged no-cors request headers</a> from <a>context object</a>.
```
annevk: I'm not sure how consistent Fetch is about it, but I've started using "the context object" elsewhere.

fetch.bs
```diff
@@ -4566,9 +4647,8 @@ method, when invoked, must run these steps:
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/<var>value</var> is not a
- <a>CORS-safelisted request-header</a>, return.
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>" and
+ <var>name</var>/<var>value</var> is not a <a>CORS-safelisted request-header</a>, return.
```
annevk: then return* or undo the formatting changes.

fetch.bs
```diff
@@ -4577,6 +4657,9 @@ method, when invoked, must run these steps:
  <li><p><a for="header list">Set</a>
  <var>name</var>/<var>value</var> in
  <a for=Headers>header list</a>.
+
+ <li><p>If <a for=Headers>guard</a> is "<code>request-no-cors</code>", then
+ <a for=Headers>remove privileged no-cors request headers</a> from <a>context object</a>.
```
annevk: the context object*

fetch.bs
```diff
@@ -5111,42 +5194,47 @@ constructor must run these steps:
...
- <a for=request>redirect mode</a>,
- <a for=request>integrity metadata</a> is
- <var>request</var>'s
- <a for=request>integrity metadata</a>, and
- <a>keepalive flag</a> is <var>request</var>'s <a>keepalive flag</a>.
+ <li>
+  <p>Set <var>request</var> to a new <a for=/>request</a> with the following properties:
+
+  <dl>
+   <dt><a for=request>url</a> <dd><var>request</var>'s <a for=request>current url</a>.
```
annevk: `<dd>` on a new line

fetch.bs
```diff
@@ -5156,17 +5244,19 @@ constructor must run these steps:
    <li><p>Set <var>request</var>'s <a for=request>referrer</a> to
    "<code>client</code>"
 
-   <li><p>Set <var>request</var>'s
-   <a for=request>referrer policy</a> to the empty string.
+   <li>
+    <p>Set <var>request</var>'s <a for=request>referrer policy</a> to the empty string.
+
+    <p class=note>This is done to ensure that when a service worker "redirects" a request, e.g.,
+    from an image in a cross-origin stylesheet, and makes modifications, it no longer appears to
```
annevk: style sheet* (throughout)

fetch.bs
```diff
@@ -5156,17 +5244,19 @@ constructor must run these steps:
...
+    come from the original source (i.e., the cross-origin stylesheet), but instead from the service
+    worker that "redirected" the request. This is important as the original source might not even be
+    able to generate the same kind of requests as the service worker. Services that trust the
+    original source could therefore be exploited were this not done, although that is somewhat
+    farfetched.
```
annevk: Why did you move this note by the way? It really applies to the whole "if any of init's member are present" clause.

fetch.bs
```diff
@@ -5451,10 +5550,18 @@ run these steps:
...
- <a for=Request>headers</a>' <a for=Headers>guard</a>.
+ <li>
+  <p>Set <var>clonedRequestObject</var>'s <a for=Request>headers</a> to a new {{Headers}} object
+  whose:
+
+  <ul>
+   <li><p><a for=Headers>header list</a> is set to <var>clonedRequest</var>'s
+   <a for=Headers>header list</a>.
+
+   <li><p><a for=Headers>guard</a> is <a>context object</a>'s <a for=Request>headers</a>'s
```
annevk: the context object*

fetch.bs
```diff
@@ -5451,10 +5550,18 @@ run these steps:
 
  <li><p>Set <var>clonedRequestObject</var>'s <a for=Request>request</a> to <var>clonedRequest</var>.
 
- <li><p>Set <var>clonedRequestObject</var>'s <a for=Request>headers</a> to a new {{Headers}} object
- whose <a for=Headers>header list</a> is set to <var>clonedRequest</var>'s
- <a for=Headers>header list</a>, and <a for=Headers>guard</a> is <a>context object</a>'s
- <a for=Request>headers</a>' <a for=Headers>guard</a>.
+ <li>
+  <p>Set <var>clonedRequestObject</var>'s <a for=Request>headers</a> to a new {{Headers}} object
+  whose:
```
annevk: It's a little weird we use `<ol>` here and for a similar thing above we use `<dl>`. I'm not really sure what style I prefer though.

fetch.bs
```diff
@@ -1360,6 +1394,10 @@ specified. [[!CSP]]
 `<a http-header><code>Access-Control-Expose-Headers</code></a>` header. This list is used by a
 <a>CORS filtered response</a> to determine which headers to expose.
 
+<p>A <a for=/>response</a> has an associated
+<dfn for=response id=concept-response-range-requested-flag>range requested flag</dfn>, which is
+initially unset.
```
annevk: It'd be good to add a note here about the purpose. Also, "range-requested" perhaps?

annevk: Also, do we need to export this or is it fine for just internal usage?

fetch.bs
```diff
@@ -4477,6 +4538,23 @@ run these steps:
  <var>key</var>/<var>value</var> to <var>headers</var>.
 </ol>
 
+<p>To
+<dfn for=Headers id=concept-headers-remove-privileged-no-cors-request-headers>remove privileged no-cors request headers</dfn>
+from a {{Headers}} object (<var>headers</var>), run these steps:
```
annevk: It seems this would be a little better as instead iterating over the privileged no-cors request-header names and then deleting (concept-header-list-delete) them.

fetch.bs
```diff
@@ -446,6 +446,17 @@ documented in <a href=#cors-protocol-exceptions>CORS protocol exceptions</a>.
...
 for `<code>Authorization</code>`.
 
+<p>A <dfn export>privileged no-cors request-header name</dfn> is a <a for=/>header</a>
+<a for=header>name</a> that is a <a>byte-case-insensitive</a> match for one of
+
+<ul class=brief>
+ <li>`<code>Range</code>`
+</ul>
+
+<p class="note no-backref">These are headers that <span class=allow-2119>may</span> be set by
```
annevk: Use can or could instead? No reason to use may I think. If you think it should be a normative statement then this should not be a note.

annevk: Perhaps this note should also point out the "add a range header" primitive.

fetch.bs
jakearchibald: I don't think we'll need this externally, but I guess we can export it if needed. I also filed https://github.com/tabatkins/bikeshed/issues/1216, which would make non-exported definitions more useful.

fetch.bs
```diff
@@ -3186,6 +3224,27 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
...
+  and <var>response</var>'s <a for=response>range requested flag</a> is set, and
+  <var>request</var>'s <a for=request>header list</a> does not <a for="header list">contain</a>
+  "<code>`Range`</code>", then return a <a>network error</a>.
+
+  <div class=note>
+   <p>Traditionally, APIs accept a ranged response even if a range wasn't requested. However, we
+   need to prevent a partial response from an earlier ranged request being provided to an API that
+   didn't make a range request.
+
+   <p>Example attack: A media element is used to request a range of a cross-origin HTML resource.
```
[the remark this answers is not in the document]
jakearchibald: Does example make sense? It's thing we're actively preventing.

fetch.bs
annevk: If we don't, then don't export it. I'd rather get requests when it's needed than find out at some point that a dependency on it was taken.

fetch.bs
```diff
@@ -3186,6 +3224,27 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
     <!-- not resetting actualResponse since it's no longer used anyway -->
   </ol>
 
+ <li>
+  <p>If <var>response</var>'s <a for=response>status</a> is <code>206</code>,
+  and <var>response</var>'s <a for=response>range requested flag</a> is set, and
+  <var>request</var>'s <a for=request>header list</a> does not <a for="header list">contain</a>
+  "<code>`Range`</code>", then return a <a>network error</a>.
```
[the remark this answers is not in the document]
annevk: Hmmmmm. How would that not be observable?

fetch.bs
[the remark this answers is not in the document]
annevk: Maybe in that case it's an example but you should just state it prevents that attack?

fetch.bs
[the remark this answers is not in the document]
jakearchibald: Ah, this should only happen for opaque responses. I'll fix that too.

fetch.bs
jakearchibald: I also need to change this so headers are only removed if an item was actually deleted

jakearchibald:
Comments should be addressed now.

Tests: https://github.com/w3c/web-platform-tests/pull/10348.

jakearchibald:
@annevk what are the next steps here? Do you want to hold off merging until I have the media stuff in the HTML spec done?

fetch.bs
```diff
@@ -446,6 +446,20 @@ documented in <a href=#cors-protocol-exceptions>CORS protocol exceptions</a>.
...
+<p>A <dfn export>privileged no-cors request-header name</dfn> is a <a for=/>header</a>
+<a for=header>name</a> that is a <a>byte-case-insensitive</a> match for one of
+
+<ul class=brief>
+ <li>`<code>Range</code>`
+</ul>
+
+<div class="note no-backref">
+ <p>These are headers that can be set by privileged APIs, and will be preserved if their associated
+ request object is copied, but will be removed if the request is modified in any way.
```
annevk: Should we enumerate some of the known suspects here?

fetch.bs
```diff
@@ -1277,6 +1291,29 @@ or "<code>worker</code>".
...
+<hr>
+
+<p>To <dfn export for=request id=concept-request-add-range-header>add a range header</dfn> to a <a
+for=/>request</a> <var>request</var>, with an integer <var>first</var>, and an optional integer
+<var>last</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>rangeValue</var> be `<code>bytes </code>`.
+
+ <li><p><a>UTF-8 encode</a> <var>first</var> and append the result to <var>rangeValue</var>.
```
annevk: You cannot encode an integer (same below).

fetch.bs
```diff
@@ -1277,6 +1291,29 @@ or "<code>worker</code>".
...
+
+ <li><p>Append `<code>-</code>` to <var>rangeValue</var>.
+
+ <li><p>If <var>last</var> is provided, then <a>UTF-8 encode</a> it and append the result to
+ <var>rangeValue</var>.
+
+ <li><p><a for="header list">Append</a> `<code>Range</code>`/<var>rangeValue</var> to
+ <var>request</var>'s <a for=request>header list</a>.
+</ol>
```
annevk: Maybe mention here that if you need this primitive you need to seek security review?

fetch.bs
annevk: This still needs a note about the purpose and be changed to "range-requested" (note the hyphen).

fetch.bs
```diff
@@ -2813,6 +2854,35 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
...
+
+   <p>A media element is used to request a range of a cross-origin HTML resource. Although this is
+   invalid media, a reference to a clone of the response can be retained in a service worker. This
+   can later be used as the response to a script element's fetch. If the partial response is valid
+   JavaScript (even though the whole resource is not), executing it would leak private data.
+
+   <p>The response is replaced with an "ok" response to avoid triggering error events or rejections
+   in APIs, as this would also leak information about the response.
+  </div>
+
```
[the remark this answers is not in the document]
annevk: I wonder if this should copy some fields as we established in the CORB PR, to keep CSP and such working. I guess maybe not since it happens at a later point?

fetch.bs
```diff
@@ -4509,9 +4597,9 @@ method, when invoked, must run these steps:
  "<code>request</code>" and <var>name</var> is a <a>forbidden header name</a>,
  return.
 
- <li><p>Otherwise, if <a for=Headers>guard</a> is
- "<code>request-no-cors</code>" and <var>name</var>/`<code>invalid</code>` is
- not a <a>CORS-safelisted request-header</a>, then return.
+ <li><p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>",
```
annevk: Given the note below we should put this `<p>` on a new line. (I realize this was already there. I can fix it therefore in a nit comment if you want.)

jakearchibald:
Note to self: I need to make sure my response sanitising matches https://github.com/whatwg/fetch/pull/686/files

annevk:
I suspect that'll merge first so you can rebase and create some kind of shared abstraction I suppose.

jakearchibald:
Rebased and addressed issues.

I didn't match the CORB sanitising in the end. As you said, it's kinda in a different place. CORB want to replace the response as soon as possible, but I'm doing it really late in the fetch as I need to check against the original request.

I'm a little unsure if I need to be setting *HTTPS state* or not.

jakearchibald:
https://github.com/w3c/web-platform-tests/pull/10348 is also rebased and ready for review.

fetch.bs
```diff
@@ -197,6 +197,11 @@ of:
 <a>queue a fetch task</a> on <var>request</var> to <a>process request end-of-body</a>
 for <var>request</var>.
 
+<p>To <dfn>serialize an integer</dfn>, represent it as a DOMString of the shortest possible decimal
```
annevk: s/DOMString/string/ please. I don't like using IDL types (and there's an issue against IDL about mapping DOMString to JavaScript string once it enters the specification side).

fetch.bs
```diff
@@ -197,6 +197,11 @@ of:
...
+number.
+
+<p class=note>This will be replaced by a more descriptive algorithm in the infra spec. See
```
annevk: class=XXX

fetch.bs
```diff
@@ -443,6 +448,26 @@ documented in <a href=#cors-protocol-exceptions>CORS protocol exceptions</a>.
 <p>A <dfn export>CORS non-wildcard request-header name</dfn> is a <a>byte-case-insensitive</a> match
 for `<code>Authorization</code>`.
 
+<p>A <dfn export>privileged no-cors request-header name</dfn> is a <a for=/>header</a>
+<a for=header>name</a> that is a <a>byte-case-insensitive</a> match for one of
+
+<ul class=brief>
+ <li>`<code>Range</code>`
```
annevk: There should be a dot after this. (This is also the case for the Set-Cookie list, which also has an incorrect colon leading it, but you don't have to fix that.)

fetch.bs
```diff
@@ -1291,6 +1316,34 @@ or "<code>worker</code>".
...
+<hr>
+
+<p>To <dfn export for=request id=concept-request-add-range-header>add a range header</dfn> to a
+<a for=/>request</a> <var>request</var>, with an integer <var>first</var>, and an optional integer
+<var>last</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>rangeValue</var> be `<code>bytes </code>`.
+
+ <li><p><a lt="serialize an integer">serialize</a> and <a>UTF-8 encode</a> <var>first</var>, and
```
annevk: Serialize* (uppercase)

fetch.bs
```diff
@@ -1291,6 +1316,34 @@ or "<code>worker</code>".
...
+ append the result to <var>rangeValue</var>.
+
+ <li><p>Append `<code>-</code>` to <var>rangeValue</var>.
+
+ <li><p>If <var>last</var> is provided, then <a lt="serialize an integer">serialize</a> and
```
annevk: is given*

fetch.bs
```diff
@@ -4598,19 +4711,27 @@ method, when invoked, must run these steps:
...
- "<code>request-no-cors</code>" and <var>name</var>/`<code>invalid</code>` is
- not a <a>CORS-safelisted request-header</a>, then return.
+ <li>
+  <p>Otherwise, if <a for=Headers>guard</a> is "<code>request-no-cors</code>",
+  <var>name</var>/`<code>invalid</code>` is not a <a>CORS-safelisted request-header</a>, and
+  <var>name</var> is not a <a>privileged no-cors request-header name</a>, then return.
 
   <p class=note>`<code>invalid</code>` is used because
   <a method for=Headers><code>delete()</code></a> is not passed a value as argument.
+ </li>
```
annevk: No need for closing `</li>` outside of the HTML Standard.

fetch.bs
```diff
@@ -2911,6 +2972,39 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
...
+  <p>If <var>internalResponse</var>'s <a for=response>status</a> is <code>206</code>,
+  <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>",
+  <var>internalResponse</var>'s <a for=response>range-requested flag</a> is set, and
+  <var>request</var>'s <a for=request>header list</a> does not <a for="header list">contain</a>
+  `<code>Range</code>`, then:
+
+  <ol>
+   <li><p>Set <var>internalResponse</var> to a new <a for=/>response</a>.
+
+   <li><p>Set <var>response</var>'s <a>internal response</a> to <var>internalResponse</var>.
```
annevk: Didn't that already happen in the above step? This resets all state though. Does that really work with the various algorithms?

jakearchibald: In the previous step, I've changed what *internalResponse* refers to, but *response's* *internal response* will still refer to the previous response.

Or would it be better to set *response* to a new opaque response, then set *internalResponse* to its *internal response*?

annevk: I think reusing the CORB strategy is probably better (by abstracting the way it replaces the response so it can be reused here), since I think we want them to be equivalent. But maybe I'm missing something?

jakearchibald:
Certainly no harm in it. I'll do that.

jakearchibald:
Ok, I've addressed the nits and abstracted creating a "sanitized response". There was a little bit of loop-jumping to do, since CORB deals exclusively in unfiltered responses, and in this PR I'm dealing exclusively with filtered responses.

Since I'm copying cover the CORS/MIX data over anyway, I've moved by range-check before those steps. This makes it a little closer to the CORB steps.

fetch.bs
```diff
@@ -1505,6 +1566,28 @@ is a <a>filtered response</a> whose
...
+ <li><p>Let <var>unfilteredResponse</var> be <var>response</var>'s <a>internal response</a> if
+ <var>response</var> is a <a>filtered response</a>, otherwise <var>response</var>.
+
+ <li><p>Let <var>sanitizedResponse</var> a new <a for=/>response</a> whose
+ <a for=response>status</a> is <var>unfilteredResponse</var>'s <a for=response>status</a>,
+ <a for=response>HTTPS state</a> is <var>unfilteredResponse</var>'s <a for=response>HTTPS state</a>,
+ and <a for=response>CSP list</a> is <var>unfilteredResponse</var>'s <a for=response>CSP list</a>.
+
+ <li><p>If <var>response</var> is a <a>filtered response</a>, set <var>sanitizedResponse</var> to
+ a new identical filtered response whose <a>internal response</a> is <var>sanitizedResponse</var>.
```
jakearchibald: "identical filtered response" feels a little hand-wavey, but it's what "clone" does above.

jakearchibald:
I need to update this to return a network error, and revert the changes to CORB. This should involve updating the tests too.

https://github.com/whatwg/fetch/issues/727#issuecomment-390917335

jakearchibald:
Spec updated following decision to return network error. Tests also updated.

annevk:
I rebased, squashed (to make rebasing going forward easier), and updated the commit message.

I guess the main thing outstanding here are the tests? Someone should probably ensure I didn't rebase incorrectly either.

jakearchibald:
The outstanding issue is the tests yeah. I'm hoping to get to that today.

jakearchibald:
Rebase looks good. Thanks for improving the byte sequence encoding.

annevk:
Unfortunately we'll have to rebase again. I guess the lesson here is that we shouldn't couple style changes and major new features.

jakearchibald:
I'll take care of that. But yeah, lesson learned. I should have split up the style changes so they could have been committed much sooner.

jakearchibald:
Rebased.

annevk:
The commit message is still good? (I'll squash my commit into yours when landing.)

annevk:
Oh, and we'll need browser bugs for anything broken, but I think you already filed those right?

jakearchibald:
Commit message is good. I'll file browser bugs now, and link to the attacks in https://github.com/whatwg/fetch/issues/144#issuecomment-368040980 in lieu of existing APIs being explicit about their usage of range requests.

jakearchibald:
Browser bugs:

https://bugs.chromium.org/p/chromium/issues/detail?id=847428
https://bugzilla.mozilla.org/show_bug.cgi?id=1465074
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17689699/
https://bugs.webkit.org/show_bug.cgi?id=186050

annevk:
Thanks @jakearchibald!

jakearchibald:
Whoop! Thanks for dragging me through this one. Maybe one day I'll get a PR through without 100 changes 😀

youennf:
Filed https://github.com/whatwg/fetch/issues/747

PR #564, opened 2017-07-13
Advise `Vary` even for non-CORS-request responses

sideshowbarker:
See https://stackoverflow.com/a/45081016/441757

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/sideshowbarker/vary-for-all-responses/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/f3bb219...6d74cb5.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5682,8 +5682,9 @@ from the previous non-<a>CORS request</a> that lacks
 <p>However, if `<a http-header><code>Access-Control-Allow-Origin</code></a>` is set to
 <code>*</code> or a static <a for=/>origin</a> for a particular resource, then configure the server
 to always send `<a http-header><code>Access-Control-Allow-Origin</code></a>` in responses for the
-resource — for non-<a lt="CORS request">CORS requests</a> as well as <a lt="CORS request">CORS
+resource — for non-<a>CORS requests</a> as well as <a>CORS
 requests</a> — and do not use `<code>Vary</code>`.
+
 
 <h2 id=acknowledgments class=no-num>Acknowledgments</h2>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -5679,6 +5679,11 @@ the user agent to <a for=/>fetch</a> a response that includes
 from the previous non-<a>CORS request</a> that lacks
 `<a http-header><code>Access-Control-Allow-Origin</code></a>`.
 
+<p>However, if `<a http-header><code>Access-Control-Allow-Origin</code></a>` is set to
+<code>*</code> or a static <a for=/>origin</a> for a particular resource, then configure the server
+to always send `<a http-header><code>Access-Control-Allow-Origin</code></a>` in responses for the
+resource — for non-<a lt="CORS request">CORS requests</a> as well as <a lt="CORS request">CORS
+requests</a> — and do not use `<code>Vary</code>`.
 
 <h2 id=acknowledgments class=no-num>Acknowledgments</h2>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -5664,6 +5664,20 @@ however, it is perfectly fine to do so.
 
 <pre id=example-vary-origin class=example>Vary: Origin</pre>
 
+<p>In particular, consider what happens if `<code>Vary</code>` is <em>not</em> used and a server is
+configured to send `<a http-header><code>Access-Control-Allow-Origin</code></a>` for a certain
+resource only in response to a <a>CORS request</a>. When a user agent receives a response to a
+non-<a>CORS request</a> for that resource (for example, as the result of a <a>navigation
+request</a>), the response will lack `<a http-header><code>Access-Control-Allow-Origin</code></a>`
+and the user agent will cache that response. Then, if the user agent subsequently encounters a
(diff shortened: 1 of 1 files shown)


tyoshino:
Thank you Michael. Looks good basically, but maybe for those who are not familiar with the issue, what "all" means is a little unclear? Can we just talk about the concrete story (a response to no-CORS request being cached and served for following CORS requests)?

Slight point, but I'd split the sentence, i.e. not "Vary is to be used for" but "Vary is to be used. Such a Vary header needs to be ...".

fetch.bs
```diff
@@ -5658,11 +5658,12 @@ however, it is perfectly fine to do so.
...
 <p>If <a>CORS protocol</a> requirements are more complicated than setting
 `<a http-header><code>Access-Control-Allow-Origin</code></a>` to <code>*</code> or a static
-<a for=/>origin</a>, `<code>Vary</code>` is to be used.
+<a for=/>origin</a>, `<code>Vary</code>` is to be used for <em>all</em> responses for a particular
+resource — not just for responses to <a lt="CORS request">CORS requests</a> for that resource.
 [[!HTML]]
 [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
 
-<pre id=example-vary-origin class=example>Vary: Origin</pre>
+<pre id=example-vary-origin class=example>Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method</pre>
```
annevk: I don't think it makes sense to include those additional headers, as they're only relevant for the preflight. They're not included in the "actual" request.

jakearchibald: Oh, yeah, good point. I guess you don't need `Access-Control-Allow-Credentials` in there as the credentials is part of the cache key.

annevk: Yeah, though we haven't written that down yet and last I heard Chrome doesn't believe in that. :/

sideshowbarker: > I don't think it makes sense to include those additional headers, as they're only relevant for the preflight. They're not included in the "actual" request.

OK, removed those

sideshowbarker:
> maybe for those who are not familiar with the issue, what "all" means is a little unclear? Can we just talk about the concrete story (a response to no-CORS request being cached and served for following CORS requests)?

Yes—I’ve now pushed b98d51b to here, with additional wording that attempts to describe the concrete scenario much more explicitly.

mnot:
Looks pretty good. Technically, `Vary` is only necessary when the response is cacheable, but that's often broader than most people realise.

tyoshino:
b98d51b looks good

annevk:
We should maybe also point out that if you intend to serve a static value such as a single origin or `*` you don't want to set the response header conditionally and don't want to use `Vary`.

annevk:
(Especially if we add comma-separated origins at some point as a possible value.)

sideshowbarker:
> We should maybe also point out that if you intend to serve a static value such as a single origin or \* you don't want to set the response header conditionally and don't want to use Vary.

OK, added in 5e33292

tyoshino:
Still looks good. Nice to mention the case.

fetch.bs
```diff
@@ -5664,7 +5664,26 @@ however, it is perfectly fine to do so.
...
+the user agent to <a for=/>fetch</a> a response that includes
+`<a http-header><code>Access-Control-Allow-Origin</code></a>`, rather than using the cached response
+from the previous non-<a>CORS request</a> that lacks
+`<a http-header><code>Access-Control-Allow-Origin</code></a>`.
+
+<p>However, if `<a http-header><code>Access-Control-Allow-Origin</code></a>` is set to
+<code>*</code> or a static <a for=/>origin</a> for a particular resource, then configure the server
+to always send `<a http-header><code>Access-Control-Allow-Origin</code></a>` in responses for the
+resource — for non-<a lt="CORS request">CORS requests</a> as well as <a lt="CORS request">CORS
+requests</a> — and do not use `<code>Vary</code>`.
```
annevk: Extra newline here.

fetch.bs
```diff
@@ -5664,7 +5664,26 @@ however, it is perfectly fine to do so.
...
+<p>But if `<code>Vary: Origin</code>` is used in the same scenario described above, it will cause
+the user agent to <a for=/>fetch</a> a response that includes
+`<a http-header><code>Access-Control-Allow-Origin</code></a>`, rather than using the cached response
+from the previous non-<a>CORS request</a> that lacks
+`<a http-header><code>Access-Control-Allow-Origin</code></a>`.
+
+<p>However, if `<a http-header><code>Access-Control-Allow-Origin</code></a>` is set to
+<code>*</code> or a static <a for=/>origin</a> for a particular resource, then configure the server
+to always send `<a http-header><code>Access-Control-Allow-Origin</code></a>` in responses for the
+resource — for non-<a lt="CORS request">CORS requests</a> as well as <a lt="CORS request">CORS
```
annevk: No need for the `lt` attribute here. Bikeshed knows about English plural.

annevk:
Thanks!

PR #578, opened 2017-08-12
Editorial: add note on event listeners & preflight

sideshowbarker:
We already have this in the XHR spec, but I think it would be helpful to
readers to also have a note in the Fetch spec at point of use, because the
Fetch spec is what gets cited at StackOverflow and such as the source for
requirements on what causes a preflight. So omitting mention of this case from
the Fetch spec risks making people (continue to) forget this is among the
things that could be causing a preflight they’re running into but that they
don’t understand the reason for or how to prevent it.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/sideshowbarker/event-listeners-and-preflight/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/07999e9...7bfea5c.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1003,10 +1003,10 @@ discouraged from using it for new features. It is rather unsafe. "<code>navigate
 <dfn id=use-cors-preflight-flag export for=request>use-CORS-preflight flag</dfn>. Unless stated
 otherwise, it is unset.
 
-<p class="note no-backref">The <a>use-CORS-preflight flag</a> being set is just one of several
-conditions that can result in a <a>CORS-preflight request</a>. The <a>use-CORS-preflight flag</a>
-will be set if either one or more event listeners are registered on an {{XMLHttpRequestUpload}}
-object, or else if a {{ReadableStream}} object is used for uploads.
+<p class="note no-backref">The <a>use-CORS-preflight flag</a> being set is one of several conditions
+that results in a <a>CORS-preflight request</a>. The <a>use-CORS-preflight flag</a> is set if either
+one or more event listeners are registered on an {{XMLHttpRequestUpload}} object or if a
+{{ReadableStream}} object is used in a request.
 
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-credentials-mode>credentials mode</dfn>,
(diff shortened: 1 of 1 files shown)


annevk:
I think that and using a `ReadableStream` object for uploads are the sole reasons. We should probably call out both.

sideshowbarker:
> I think that and using a `ReadableStream` object for uploads are the sole reasons. We should probably call out both.

OK, will add that. But where is the requirement defined about it happening when using a `ReadableStream` object for uploads?

I can’t seem to find anything either in the Fetch spec or Streams spec that states it…

sideshowbarker:
> OK, will add that. But where is the requirement defined about it happening when using a `ReadableStream` object for uploads?
> 
> I can’t seem to find anything either in the Fetch spec or Streams spec that states it…

Discussed on IRC at http://logs.glob.uno/?c=freenode%23whatwg&s=12%20Aug%202017&e=12%20Aug%202017#c1034349

fetch.bs
```diff
@@ -1003,6 +1003,11 @@ discouraged from using it for new features. It is rather unsafe. "<code>navigate
 <dfn id=use-cors-preflight-flag export for=request>use-CORS-preflight flag</dfn>. Unless stated
 otherwise, it is unset.
 
+<p class="note no-backref">The <a>use-CORS-preflight flag</a> being set is just one of several
+conditions that can result in a <a>CORS-preflight request</a>. The <a>use-CORS-preflight flag</a>
```
annevk: s/just//
s/can result/results/

fetch.bs
```diff
@@ -1003,6 +1003,11 @@ discouraged from using it for new features. It is rather unsafe. "<code>navigate
...
+will be set if either one or more event listeners are registered on an {{XMLHttpRequestUpload}}
```
annevk: s/will be set/is set/

fetch.bs
```diff
@@ -1003,6 +1003,11 @@ discouraged from using it for new features. It is rather unsafe. "<code>navigate
...
+object, or else if a {{ReadableStream}} object is used for uploads.
```
annevk: s/, or else/or if/ (I think that looks better)
s/for uploads/in a request/

PR #582, opened 2017-08-21
Fold request type into destination

### LINKED ISSUE #581: Fold request type into destination
Since ba175cf1100045da83ed0d7768629653aa979c20 merged there's no reason for type and destination to be different fields as far as I can tell. We'll need to change the following specifications to make this change:

* Fetch
* CSP, https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request (and maybe elsewhere?)
* Mixed Content, https://w3c.github.io/webappsec-mixed-content/#should-block-fetch
* HTML

In Fetch we should mention these specifications in a source comment below the definition of destination so we know where the dependencies are. SRI https://w3c.github.io/webappsec-subresource-integrity/ is another dependency but doesn't use type currently so doesn't need to be updated.

I can do Fetch and HTML. @yoavweiss are you willing to do CSP and Mixed Content?

cc @mikewest

annevk:
This was made possible due to ba175cf1100045da83ed0d7768629653aa979c20 making type a subset.

Fixes #581.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/remove-request-type/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/8a91018...25437fe.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -802,7 +802,8 @@ the empty string,
 "<code>worker</code>", or
 "<code>xslt</code>". Unless stated otherwise it is the empty string.
 
-<p>A <dfn for=export>script-like destination</dfn> is "<code>script</code>",
+<p>A <a for=/>request</a>'s <a for=request>destination</a> is
+<dfn export for=request/destination>script-like</dfn> if it is "<code>script</code>",
 "<code>serviceworker</code>", "<code>sharedworker</code>", or "<code>worker</code>".
 
 <!-- Dependencies:
@@ -1622,8 +1623,8 @@ run these steps:
  <li><p>Let <var>destination</var> be <var>request</var>'s <a for=request>destination</a>.
 
  <li>
-  <p>If <var>destination</var> is a <a>script-like destination</a> and one of the following is true,
-  then return <b>blocked</b>:
+  <p>If <var>destination</var> is <a for=request/destination>script-like</a> and one of the
+  following is true, then return <b>blocked</b>:
 
   <ul class=brief>
    <li><var>mimeType</var> starts with `<code>audio/</code>`, `<code>image/</code>`, or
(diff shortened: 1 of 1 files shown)


annevk:
Looking through the Preview I was reminded that what we have for CSP is no longer accurate. It might be worth fixing that as well, though perhaps best as a separate commit as it doesn't change anything normative.

fetch.bs
```diff
@@ -813,23 +802,27 @@ the empty string,
...
 "<code>xslt</code>". Unless stated otherwise it is the empty string.
 
+<p>A <dfn for=export>script-like destination</dfn> is "<code>script</code>",
+"<code>serviceworker</code>", "<code>sharedworker</code>", or "<code>worker</code>".
+
+<!-- Dependencies:
+     * CSP: https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request
+     * Mixed Content: https://w3c.github.io/webappsec-mixed-content/#should-block-fetch
+     * SRI: https://w3c.github.io/webappsec-subresource-integrity/#apply-algorithm-to-request
+     * HTML -->
```
yoavweiss: https://w3c.github.io/preload/ also relies on request destination (in the "appropriate times to obtain a resource" section)

fetch.bs
```diff
@@ -813,23 +802,28 @@ the empty string,
 "<code>worker</code>", or
 "<code>xslt</code>". Unless stated otherwise it is the empty string.
 
+<p>A <dfn for=export>script-like destination</dfn> is "<code>script</code>",
```
mikewest: `for=export`? `for=request export`, perhaps?

fetch.bs
```diff
@@ -813,23 +802,28 @@ the empty string,
...
+"<code>serviceworker</code>", "<code>sharedworker</code>", or "<code>worker</code>".
```
mikewest: Would this be more clear as `A [=request/destination=] is <dfn for="request/destination" export lt="script-like destination">script-like</dfn> if it is one of ..., ..., or ...`?

mikewest: Also, what about `XSLT`?

annevk: If we made "`xslt`" part of this, then [nosniff](https://fetch.spec.whatwg.org/#should-response-to-request-be-blocked-due-to-nosniff?) would fail.

And adding "`xslt`" to [MIME type blocking](https://fetch.spec.whatwg.org/#should-response-to-request-be-blocked-due-to-mime-type?) probably doesn't help either as XSLT already requires a correct MIME type?

mikewest: > If we made "xslt" part of this, then nosniff would fail.

I see. Hrm. Ok. :)

> And adding "xslt" to MIME type blocking probably doesn't help either as XSLT already requires a correct MIME type?

 I'm not familiar enough with the XSLT code in Chrome, so I'm not actually sure it does. But I agree it should.

So, yeah. I think you're right. Consider the suggestion dropped. :P

annevk:
I guess I'll add a note that "xslt" needs to be thought through in conjunction with script-like typically.

PR #584, opened 2017-08-24
Using integrity with "no-cors" is fine same-origin

### LINKED ISSUE #583: TypeError on Request.integrity with no-cors mode is a foot gun
It seems a lot of people trying to use SRI with service workers where Request.integrity is implemented are running into problems:

https://bugzilla.mozilla.org/show_bug.cgi?id=1393439

The issue is that its pretty common to do `<script integrity="hash">` in your document.  This creates a no-cors request by default.  If their service worker script passes through with `fetch(evt.request)` then they will get a TypeError.  This is required by step 32.3 here:

https://fetch.spec.whatwg.org/#dom-request

Perhaps we should move this check out of the `Request()` constructor and into the fetch algorithm at the point an opaque Response is created.  It seems its only really important if we are going to be returning an opaque Response and need to hide any information leakage about the contents.

annevk:
This exception also broke simple service worker passthrough handling.

Tests: ...

Fixes #583.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/no-cors-integrity/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/8a91018...334d67a.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4979,9 +4979,6 @@ constructor must run these steps:
    <li><p>If <var>r</var>'s <a for=Request>request</a>'s <a for=request>method</a> is not a
    <a>CORS-safelisted method</a>, then <a>throw</a> a <code>TypeError</code>.
 
-   <li><p>If <var>request</var>'s <a for=request>integrity metadata</a> is not the empty string,
-   then <a>throw</a> a <code>TypeError</code>.
-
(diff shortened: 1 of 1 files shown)


annevk:
Test changes landed as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1393439. Not landing this for now though as bikeshed is broken...

PR #585, opened 2017-08-25
Make fetch() use "same-origin" credentials by default

annevk:
Most features in the platform have this default and not having this default causes multiple connections in contemporary implementations. It also causes confusion as switching from XMLHttpRequest to fetch() is not as seamless as it could be.

Making this change will also make it easier for <script type=module> to have this default as per discussion in https://github.com/whatwg/html/issues/2557.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/fetch-api-credentials-mode/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/8a91018...5550bf9.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4774,7 +4774,7 @@ constructor must run these steps:
 
    <li><p>Set <var>fallbackMode</var> to "<code>cors</code>".
 
-   <li><p>Set <var>fallbackCredentials</var> to "<code>omit</code>".
+   <li><p>Set <var>fallbackCredentials</var> to "<code>same-origin</code>".
   </ol>
 
  <li>


annevk:
cc @youennf @wanderview @tyoshino @domenic @mikewest @jakearchibald @yutakahirano

Munter:
Even though I have followed the `fetch` spec development and implementation and have read multiple blog posts on how to use it, I still got tripped up by forgetting to manually add credentials on certain `fetch`calls. +1 from me on this change

tyoshino:
I agree that this is a more reasonable default, but also much concerned with possible break of existing apps, though your guess at https://github.com/whatwg/html/issues/2557#issuecomment-295299843 sounds correct.

Let me cc interop leads: @rbyers @foolip

jakearchibald:
I've reached out to devs on Twitter. Response is almost entirely positive.

One exception: https://twitter.com/rsjanabasis/status/901012433615020033, and I believe they're going to give us details.

tyoshino:
Thanks Jake for checking. It's good that it's supported. And, yeah, that looks really one example of the concern.

foolip:
IIUC, changing the default from "omit" to "same-origin" can only mean that credentials are sometimes sent when they otherwise would not be. For resources where it makes a difference at all, it seems rather unusual that the no-credentials response is the one that you want, but that the with-credentials response is some broken/unexpected resource. Certainly possible to create, and certain to appear in the wild given enough time, but are there cases we should worry about?

Overall, the compat risk here seems very low.

wanderview:
If we are going to do this it would be nice to try to coordinate shipping the change around the same time.

annevk:
Okay, I'll file four implementation bugs. My thinking is that we can update the tests as part of changing the implementation. Easiest to discover all the tests that need changing that way. Is that okay or should I try and prepare test updates?

foolip:
That seems like a good idea where testing without an implementation is error-prone, just leave an open "missing tests" issue and point to it from the impl bugs.

annevk:
Bugs:
* https://bugs.webkit.org/show_bug.cgi?id=176023
* https://bugzilla.mozilla.org/show_bug.cgi?id=1394399
* https://bugs.chromium.org/p/chromium/issues/detail?id=759543
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13474598/

annevk:
I was thinking I'd just leave this open until we have a patch somewhere.

foolip:
Yeah, that works too. If we have a buildup of "awaiting implementation" issues I guess labels can alleviate that.

tyoshino:
Thanks, Philip. I'd like to wait and hear the details of what Foggy said in the tweet quoted in https://github.com/whatwg/fetch/pull/585#issuecomment-324871935. But except for that, I share your analysis basically.

rsjanabasis:
I want to make sure that we can block sending credentials by appropriately filtering the request. 

For example, if we shadow "fetch" with safe_fetch:
function safe_fetch(url, init) { return fetch(arguments[0], { credentials: 'omit' } ); }
then this will always block sending credentials in every browser, correct?

tyoshino:
> function safe_fetch(url, init) { return fetch(arguments[0], { credentials: 'omit' } ); }
then this will always block sending credentials in every browser, correct?

Right. The topic is about changing the default which is used when the credentials parameter is not specified. You can still explicitly specify the credentials parameter with the value `omit` as you did to prevent credentials from being sent.

(Your example doesn't inherit values from `init`. Maybe you want to override credentials but inherit the other values from `init`.)

annevk:
I'd write `safe_fetch` as:
```
function safe_fetch(input, init) {
  const tempRequest = new Request(input, init);
  return fetch(tempRequest, { credentials: "omit" });
}
```

annevk:
Is there anything remaining that would block this from being implemented? I'd like this to be done so we can align JavaScript modules with this default and avoid a ton of confusion there.

rsjanabasis:
Yes, that's better.

tyoshino:
Thanks Foggy. So, I'm fine with the plan now.

annevk:
Okay, I'll merge this as soon as one of the bugs above gets marked fixed and web-platform-tests is updated.

domfarolino:
It seems like WPTs are one of the things holding this PR and Mozilla's implementation back (given https://bugs.chromium.org/p/chromium/issues/detail?id=759543#c7). I'm happy to work on this if nobody else has started.

wanderview:
> I wonder if we should simultaneously change module workers since that's the only other feature that exposes this default at the moment. @domfarolino what do you think?

Does anyone implement module workers yet?  Its not something shipped in firefox, so this change does not effect us there.

wanderview:
@domfarolino I'm looking to possibly ship this in FF61 around June 28.  Can chrome ship in a similar time frame?

annevk:
Yeah, I guess they are not implemented yet (and have some problems due to allowing CORS). We can work on them separately.

domenic:
Normal script type=module also exposes this default, no?

annevk:
Yeah, I realized that later. (Note that it's not tested if that actually works.) I started working on a fix in https://github.com/whatwg/html/issues/2557.

getify:
Just FYI: just ran into a bug with my cookies-setting (not working) via response from `fetch(..)`. It was working fine in FF62, but not in latest Chrome or recent Safari. I wrongly assumed that the bug was FF not matching spec. Finally figured out that FF was correct, and Chrome/Safari haven't updated on this yet. :(

domfarolino:
Yeah, this change hasn't landed in Chrome stable yet. What version are you using?

getify:
Chrome: 67.0.3396.99
Firefox: 63.0a1 (2018-07-08)

domfarolino:
Seems like Chrome stable and FF nightly or so? This change is definitely in Canary, unsure off the top of my head how far away from stable though.

yoavweiss:
@domfarolino's Chromium [change](https://chromium-review.googlesource.com/c/chromium/src/+/981512) landed on April 30th, so should be included in Chrome 68, [estimated](https://www.chromium.org/developers/calendar) to hit stable towards the end of July 2018.

@getify if you wish to verify, you can try it out in beta right now.

getify:
> Seems like Chrome stable and FF nightly or so? 

Yes, I happened to be checking in FF nightly when I responded earlier. But AFAICT it's already in FF stable -- v61 from a few weeks ago.

So apparently I just happened to get caught by this in the gap (about a month, it seems?) between FF and Chrome going stable with it.

Anyone know when Safari will be updating this (stable)?

yoavweiss:
Maybe @cdumez or @youennf would.

cdumez:
Definitely something for @youennf

Andreas-Hjortland:
I know this has already been closed, but shouldn't the sentence 
```
A request has an associated credentials mode, which is "omit", "same-origin", or "include". Unless stated otherwise, it is "omit". 
```
be 
```
A request has an associated credentials mode, which is "omit", "same-origin", or "include". Unless stated otherwise, it is "same-origin". 
```

Where we define the credentials mode concept? (https://github.com/whatwg/fetch/blob/master/fetch.bs#L1348-L1350)

annevk:
@Andreas-Hjortland thanks for taking the time to point that out, please see #901 for a discussion. In particular, it's indeed confusing, but is distinct from the API.

PR #587, opened 2017-08-28
Meta: use curl --verbose to see why it's failing

foolip:
See https://github.com/whatwg/whatwg.org/issues/65.


annevk:
Reading through https://travis-ci.org/whatwg/fetch/builds/269153466 doesn't really tell me what went wrong here. It seems to connect fine. Does running deploy.sh fail?

domenic:
I think the results here further confirm my https://github.com/whatwg/whatwg.org/issues/65#issuecomment-325331716 hypothesis.

PR #588, opened 2017-08-28
Editorial: correct CSP column

annevk:
***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/csp-column/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/d7052e2...d462491.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -873,15 +873,15 @@ not always relevant and might require different behavior.
    <td>HTML's <code>&lt;script></code>, <code>importScripts()</code>
   <tr>
    <td>"<code>serviceworker</code>"
-   <td>?
+   <td><code>child-src</code>, <code>script-src</code>, <code>worker-src</code>
    <td><code>navigator.serviceWorker.register()</code>
   <tr>
    <td>"<code>sharedworker</code>"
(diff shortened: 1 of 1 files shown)


annevk:
This is an attempt to make the table under https://fetch.spec.whatwg.org/#concept-request-destination more accurate. The directive for reports is still unclear to me.

I'm guessing downloads and normal navigation are both simply not a thing?

fetch.bs
```diff
@@ -873,15 +873,15 @@ not always relevant and might require different behavior.
    <td>HTML's <code>&lt;script></code>, <code>importScripts()</code>
   <tr>
    <td>"<code>serviceworker</code>"
-   <td>?
+   <td><code>child-src</code>, <code>script-src</code>, <code>worker-src</code>
```
mikewest: `child-src` is ostensibly deprecated, but until any browser other than Chrome implements `worker-src` with the new inheritance structure, I guess you're right that it's worth keeping it in this list.

annevk:
I have no great ideas. Sounds like it's worth revisiting once we have origin policies.

PR #592, opened 2017-08-30
Adjust wildcard handling slightly

### LINKED ISSUE #548: access-conrol-allow-origin: * can be interpreted in two ways
access-control-allow-origin = #field-name / wildcard

where

field-name = token
token = 1*tchar
tchar contains `*`
.

It means "*"  can be interpreted in two ways.

1. Allow all headers.
1. Allow a header whose name is "*".

> 1. Let headerNames be the result of extracting header list values given `Access-Control-Expose-Headers` and response’s header list.
> 1. If headerNames is `*` and request’s credentials mode is not "include", then set response’s CORS-exposed header-name list to all unique header names in response’s header list.
> 1. Otherwise, if headerNames is not null, failure, or `*`, then set response’s CORS-exposed header-name list to headerNames.

This sounds like

1. if _headerName_ is `*`, the symbol should be interpreted in the first way.
1. if _headerName_ contains `*` but _headerName_ is not `*`, the symbol should be interpreted in the second way.

I feel it confusing.

annevk:
Since `*` is a valid header name and method, only count it as a special value for requests without credentials.

Fixes #548.


annevk:
This will require some test changes, but overall seems slightly better and clearer.

fetch.bs
```diff
@@ -3916,21 +3912,40 @@ steps:
...
-   <a>network error</a>.
+   <li>
+    <p>If all of the following are true
+
+    <ul class=brief>
+     <li><var>request</var>'s <a for=request>method</a> is not in <var>methods</var>
+
+     <li><var>request</var>'s <a for=request>method</a> is not a <a>CORS-safelisted method</a>
+
+     <li><var>request</var>'s <a for=request>credentials mode</a> is not "<code>include</code>"
```
yutakahirano: The list is "and"-concatenated, so I think splitting this item into two is better.

fetch.bs
```diff
@@ -3916,21 +3912,40 @@ steps:
...
+    <p>If all of the following are true
+
+    <ul class=brief>
+     <li>one of <var>request</var>'s <a for=request>header list</a>'s <a for=header>names</a> is not
+     a <a>byte-case-insensitive</a> match for an <a for=list>item</a> in <var>headerNames</var>
+
+     <li>one of <var>request</var>'s <a for=request>header list</a>'s <a for=/>headers</a> is not a
+     <a>CORS-safelisted request-header</a>
+
+     <li><var>request</var>'s <a for=request>credentials mode</a> is not "<code>include</code>" and
```
yutakahirano: ditto

fetch.bs
annevk: I now think this is wrong. What if credentials mode is "include" and _methods_ does not contain request's method?

yutakahirano: Thanks, is the following correct?

 - the method is not listed,
 - the method is not CORS-safeliested, and
 - the list doesn't have a wildcard symbol, i.e., not (_credentials_ != "include" and  contains `*`)

annevk:
If credentials is "include" request's method needs to be in _methods_ or needs to be safelisted. If credentials is not "include" request's method needs to be in _methods_, needs to be safelisted, or _methods_ needs to contain *.

So I think the requirement is more like this:

> If request's method is not in _methods_, request's method is not safelisted, and if credentials is not "include", _methods_ does not contain *, then return a network error.

We could also word that as:

> If request's method is not in _methods_ and request's method is not safelisted, then:
> * If credentials is not "include" and _methods_ does not contain *, then return a network error.
> * If credentials is "include", then return a network error.

If the first variant is not clear enough. WDYT?

yutakahirano:
I like the second one.

annevk:
I used a variant of the second.

annevk:
(This isn't building because of https://github.com/tabatkins/bikeshed/issues/1085, but I think it's the correct fix now.)

annevk:
Wrote tests over at https://github.com/w3c/web-platform-tests/pull/7223.

fetch.bs
```diff
@@ -3913,39 +3913,29 @@ steps:
...
+    </ol>
 
    <li><p>If one of <var>request</var>'s <a for=request>header list</a>'s
    <a lt=name for=header>names</a> is a <a>CORS non-wildcard request-header name</a> and is not a
    <a>byte-case-insensitive</a> match for an <a for=list>item</a> in <var>headerNames</var>, then
    return a <a>network error</a>.
 
    <li>
-    <p>If all of the following are true
+    <p>If one of <var>request</var>'s <a for=request>header list</a>'s <a for=header>names</a> is
```
yutakahirano: Isn't this confusing? Imagine _request_ has two headers, "Accept" and "foobar", and "foobar" is listed in _headerNames_. But this branch is taken wrongly because

 - "Accept" is not a byte-case-insensitive match for an item in _headerNames_, and
 - "foobar" is not a CORS-safelisted request header.

annevk: Good catch, thanks! Posted a fix.

annevk:
Thanks @yutakahirano. Could you have a look at https://github.com/w3c/web-platform-tests/pull/7223 as well?

PR #594, opened 2017-08-31
Avoid using the CORS flag to reset request's origin in redirects

### LINKED ISSUE #593: "no-cors" POST and 307/308 redirects
We basically want the same behavior as we have with POST for the `Origin` header. From the moment you go across origins a flag is set and then if you cross origins again we set the origin to a unique identifier.

This isn't a problem for POST and 302 or some such as "no-cors" GET never includes an `Origin` header.

See also https://bugs.chromium.org/p/chromium/issues/detail?id=760487.

Related Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=446344.

There's also some issues with Safari here. If someone could write web-platform-tests that'd be great.

annevk:
Otherwise things go wrong for "no-cors" POST redirects.

Fixes #593.

***
<a href="https://whatpr.org/fetch/594.html" title="Last updated on May 24, 2018, 12:10 PM GMT (398ce19)">Preview</a> | <a href="https://whatpr.org/fetch/594/794dd54...398ce19.html" title="Last updated on May 24, 2018, 12:10 PM GMT (398ce19)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3336,9 +3336,10 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
  <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
  <a for=request>body</a>'s <a for=body>source</a> is null, then return a <a>network error</a>.
 
- <li><p>If <i>CORS flag</i> is set and <var>actualResponse</var>'s
- <a for=response>location URL</a>'s <a for=url>origin</a> is not <a>same origin</a> with
- <var>request</var>'s <a for=request>current url</a>'s <a for=url>origin</a>, then set
+ <li><p>If <var>actualResponse</var>'s <a for=response>location URL</a>'s <a for=url>origin</a> is
+ not <a>same origin</a> with <var>request</var>'s <a for=request>current url</a>'s
+ <a for=url>origin</a> and <var>request</var>'s <a for=request>origin</a> is not <a>same origin</a>
+ with <var>request</var>'s <a for=request>current url</a>'s <a for=url>origin</a>, then set
  <var>request</var>'s <a for=request>origin</a> to a unique <a>opaque origin</a>.
 
  <li><p>If either <var>actualResponse</var>'s <a for=response>status</a> is
@@ -6328,6 +6329,7 @@ Clement Pellerin,
 Collin Jackson,
 Daniel Robertson,
 Daniel Veditz,
+David Benjamin,
 David Håsäther,
 David Orchard,
 Dean Jackson,
--- a/fetch.bs
+++ b/fetch.bs
@@ -950,12 +950,9 @@ user-agent-defined object). Unless otherwise stated it is null.
 "<code>client</code>" or an <a for=/>origin</a>. Unless stated otherwise it is
 "<code>client</code>".
 
-<p class="note no-backref">"<code>client</code>" is changed to an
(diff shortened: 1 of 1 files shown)


annevk:
cc @davidben @yutakahirano @fmarier

annevk:
@davidben thanks, this also needs some tests still. Have you written any per chance?

yutakahirano:
This affects canvas tainting and friends, right?

```
request’s current url’s origin is same origin with request’s origin and CORS flag is unset
  => Set request’s response tainting to "basic".
```

annevk:
So the alternative I can think of here is that we always preserve request's origin to its initial value and instead set a tainted origin flag during redirects. We then use that flag for the `Origin` header as well as the CORS literal origin comparison.

annevk:
I pushed a commit that does that, what do you think @yutakahirano?

yutakahirano:
Sounds reasonable but don't we need to update the origin usage of preflight cache?

annevk:
I suspect the problem with the preflight cache dates back from when I added support redirects following a successful preflight. I don't think I fully considered how the preflight cache should behave in cases where the request's origin was opaque.

Most performant would probably be to change the origin key into a byte sequence and use `null` if the tainted origin flag is set.

annevk:
I tried to address the problem with the CORS-preflight cache by using a serialized origin as key. @yutakahirano @mikewest let me know what you all think.

annevk:
I also filed #735 on defining the CORS-preflight cache in terms of proper data structures.

fetch.bs
```diff
@@ -4140,8 +4149,10 @@ steps:
   <a for=request>initiator</a> is <var>request</var>'s <a for=request>initiator</a>,
   <a for=request>destination</a> is <var>request</var>'s <a for=request>destination</a>,
   <a for=request>origin</a> is <var>request</var>'s <a for=request>origin</a>,
-  <a for=request>referrer</a> is <var>request</var>'s <a for=request>referrer</a>, and
-  <a for=request>referrer policy</a> is <var>request</var>'s <a for=request>referrer policy</a>.
+  <a for=request>referrer</a> is <var>request</var>'s <a for=request>referrer</a>,
+  <a for=request>referrer policy</a> is <var>request</var>'s <a for=request>referrer policy</a>, and
+  <a for=request>tainted origin flag</a> is <var>request</var>'s
```
yutakahirano: Is this change needed?

yutakahirano: Never mind, I was wrong, it's needed.

annevk:
Thanks. I created a basic test at https://github.com/w3c/web-platform-tests/pull/11164 that can be expanded. It seems that Safari omits the `Origin` header upon redirects (rather than using the explicit `null` value) and Firefox retains the original value. Chrome does it correct.

annevk:
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1424076
Safari: https://bugs.webkit.org/show_bug.cgi?id=186030

annevk:
Hmm, should we change the first conditional of step 5 of main fetch to:

> request's current url's origin is same origin with request's origin, CORS flag is unset, and either request's tainted origin flag is unset or request's mode is not "cors"

I think setting the CORS flag would not work since that would make "no-cors" follow CORS code paths.

annevk:
Filed #756 to track this, thanks!

PR #596, opened 2017-09-04
Remove "foreign" service-workers mode

annevk:
Fixes part of https://github.com/w3c/ServiceWorker/issues/1188.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/foreign-fetch/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/9fa071b...2ff1550.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -745,26 +745,18 @@ explicitly set <a for=/>request</a>'s
 <code>navigator.sendBeacon</code> and the HTML <code>img</code> element set this flag. Requests with
 this flag set are subject to additional processing requirements.
 
-<p>A <a for=/>request</a> has an associated
-<dfn for=request export>service-workers mode</dfn>, that is "<code>all</code>",
-"<code>foreign</code>", or "<code>none</code>". Unless stated otherwise it is
-"<code>all</code>".
+<p>A <a for=/>request</a> has an associated <dfn for=request export>service-workers mode</dfn>, that
+is "<code>all</code>" or "<code>none</code>". Unless stated otherwise it is "<code>all</code>".
 
 <div class=note>
  <p>This determines which service workers will receive a {{fetch!!event}} event for this fetch.
 
  <dl>
   <dt>"<code>all</code>"
-  <dd>Relevant local and foreign service workers will get a {{fetch!!event}} or
-   {{foreignfetch!!event}} event for this fetch.
-
-  <dt>"<code>foreign</code>"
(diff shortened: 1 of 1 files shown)


annevk:
https://github.com/whatwg/fetch/commit/32411c753f46a10fd20d2ad3610244e7fdf44d4a seems useful even without foreign fetch so I think we'll keep that setup.

Service-workers mode also seems better than a flag so I'll keep that.

Looking at https://github.com/whatwg/fetch/commit/d41c2380dc828e7a23c6196a344b42b2d0e9beec perhaps the global object conditional can be removed altogether?

jakearchibald:
> perhaps the global object conditional can be removed altogether?

I'm not sure how. We need to prevent `fetch()`, when called from within a service worker, triggering a fetch event.

annevk:
Ah yeah, we still want service workers to intercept some requests that are created within a service worker so that still makes sense. Thanks!

sideshowbarker:
* Are test updates not needed for this change?
* Are browser bugs — or updates to existing browser bugs — not needed for this change?

annevk:
I'm not aware of any tests. @mkruisselbrink? The tests @jakearchibald added in one of the commits linked above are still relevant and didn't test foreign fetch.

https://bugzilla.mozilla.org/buglist.cgi?quicksearch=foreign%20fetch does suggest there are some tests that need to be removed and bugs to be closed. Hopefully that happens as part of the changes to service workers.

cc @jungkees

wanderview:
At least these tests should be removed:

https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/foreign-fetch-basics.https.html
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/foreign-fetch-cors.https.html
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/foreign-fetch-event.https.html
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/foreign-fetch-workers.https.html

There are also a bunch of link-element-register tests which were added with foreign fetch.  I guess we are keeping those?

annevk:
I updated https://github.com/w3c/ServiceWorker/issues/1188 to make sure this all gets taken into account.

rektide:
This greatly hobbles the ability for an offline capable web to ever happen. [I am deeply saddened to see this.](https://github.com/w3c/ServiceWorker/issues/1188#issuecomment-329549974)

PR #597, opened 2017-09-04
Add `Last-Event-ID` as CORS-safelisted request-header

### LINKED ISSUE #568: Safelist Last-Event-ID
Per https://github.com/whatwg/html/issues/689 it can contain any header value. And it seems that a page with a cooperating server can send that header anywhere due to redirects.

Basically, use `EventSource` to open a connection. Get the server to set the ID. Then trigger a reconnection. Then when the server gets the header, redirect that request to wherever. That request to wherever will include the `Last-Event-ID` header with a value that can be controlled by the server that did the redirect.

Given that this has been the case for forever, we might as well enshrine it. (Should probably write a test to confirm though.)

annevk:
It turns out that you can set the Last-Event-ID request header to arbitrary values and get it across origins. That seems like sufficient reason to safelist it and hopefully make it clear to server administrators to pay extra attention to this header.

Tests: ...

Fixes #568.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/last-event-id/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/9fa071b...aa394cf.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -426,6 +426,7 @@ whose <a for=header>name</a> is a <a>byte-case-insensitive</a> match for one of
  <a lt="extract header values">once extracted</a>, has a MIME type (ignoring parameters)
  that is `<code>application/x-www-form-urlencoded</code>`,
  `<code>multipart/form-data</code>`, or `<code>text/plain</code>`
+ <li>`<code>Last-Event-ID</code>`
 </ul>
 <!-- XXX * needs better xref
          * ignoring parameters has been the standard for a long time now


sideshowbarker:
Doesn’t this also need browser bugs?

annevk:
Yeah, it'll need browser bugs. And I suppose it does theoretically increase the attack surface a little bit by doing this, as you'll be able to control more than just `Last-Event-ID` once this is safelisted, but I don't think that's sufficient reason not to do it.

annevk:
This will need a new plan on top of #736. Given that I'm not sure this is still worth pursuing.

annevk:
Closing this, will leave the issue open for now.

PR #598, opened 2017-09-05
CORS preflights need a service-workers mode of "none"

### LINKED ISSUE #558: Should CORS preflight fetch also copy request's service-workers mode?
https://fetch.spec.whatwg.org/commit-snapshots/f3bb21991abdd335175fcc5d26a0d0b7b380d4fe/#cors-preflight-fetch

> Let preflight be a new request whose method is \`OPTIONS\`, url is request’s current url, initiator is request’s initiator, type is request’s type, destination is request’s destination, origin is request’s origin, referrer is request’s referrer, and referrer policy is request’s referrer policy\.

_preflight_ does not seem to have copied _request_'s service-workers mode. Should it do so?

annevk:
Fixes #558.

This probably does not matter as it's never going to be same-origin, but better safe than sorry.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/annevk/preflight-service-workers-mode/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/bc23529...7830c74.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3810,21 +3810,19 @@ number of these <a lt="CORS-preflight fetch">fetches</a>.
 steps:
 
 <ol>
- <li><p>Let <var>preflight</var> be a new <a for=/>request</a>
- whose <a for=request>method</a> is `<code>OPTIONS</code>`,
- <a for=request>url</a> is <var>request</var>'s
- <a for=request>current url</a>,
- <a for=request>initiator</a> is <var>request</var>'s
- <a for=request>initiator</a>,
- <a for=request>destination</a> is <var>request</var>'s
- <a for=request>destination</a>,
- <a for=request>origin</a> is <var>request</var>'s
- <a for=request>origin</a>,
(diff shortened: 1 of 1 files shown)


annevk:
It does actually matter I just realized. You can go cross-origin and then same-origin via a redirect.

jakearchibald:
I don't think this is technically needed because **CORS-preflight fetch** calls **HTTP-network-or-cache fetch**, whereas the service worker interception happens in **HTTP fetch**.

Redirects aren't an issue, because **HTTP-network-or-cache fetch** doesn't follow redirects. Well, it's an issue if CORS preflights are expected to follow redirects. Are they?

Setting this for safety seems fine, but maybe include a note to say it doesn't do anything right now?

annevk:
No, a CORS preflight itself doesn't follow redirects. But the actual request following it will (resulting in another CORS preflight for the new URL).

So given that, I think it would be better to point out your finding in a note than add this. That makes it much more clear what's going on.

annevk:
Okay, should be good now.

PR #600, opened 2017-09-07
Make status message default the empty byte sequence

### LINKED ISSUE #599: Response `statusText` is ambiguous and unclear for HTTP/2
The `statusText` property of `Response` is unclear for HTTP/2.

HTTP/2 does not include a status message / reason like HTTP/1.1: [RFC 7540](https://tools.ietf.org/html/rfc7540#section-8.1.2.4)

> 8.1.2.4.  Response Pseudo-Header Fields
>    ...
>    HTTP/2 does not define a way to carry the version or reason phrase
>    that is included in an HTTP/1.1 status line.

The Fetch spec simply defines `statusText` as the `status message` value, which unless otherwise specified is `OK`.

For HTTP/2 it seems status message could be interpreted any number of ways:

* empty string - because there is no reason phrase
* `"OK"` - because that is always the default (would be weird to have a 404 OK though)
* implementation defined - such as some default reason text for a status code

---

Different browser behaviors. Test on an HTTP/2 capable domain.

```js
fetch("/").then((r) => console.log(r.statusText)); // 200
```

* Chrome: `""`
* Firefox: `"OK"`
* Safari: `"HTTP/2.0 200"`

```js
fetch("/doesnotexistasdf").then((r) => console.log(r.statusText)); // 404
```

* Chrome: `""`
* Firefox: `"Not Found"`
* Safari: `"HTTP/2.0 404"`

annevk:
Fixes #599.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1252,8 +1252,11 @@ cannot be exposed to script. That would violate <a>atomic HTTP redirect handling
 <a for=/>status</a>. Unless stated otherwise it is <code>200</code>.
 
 <p>A <a for=/>response</a> has an associated
-<dfn export for=response id=concept-response-status-message>status message</dfn>.
-Unless stated otherwise it is `<code>OK</code>`.
+<dfn export for=response id=concept-response-status-message>status message</dfn>. Unless stated
+otherwise it is the empty byte sequence.
+
+<p class=note>This does not have a default value of `<code>OK</code>` as otherwise all HTTP/2
+<a for=/>responses</a> would have that value.
 
 <p>A <a for=/>response</a> has an associated
 <dfn export for=response id=concept-response-header-list>header list</dfn> (a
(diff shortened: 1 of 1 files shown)


annevk:
Tests: https://github.com/w3c/web-platform-tests/issues/7274 (note that this is an issue, not actual tests, due to lack of HTTP/2 support).

annevk:
@jakearchibald not sure who else to ask for review, but Chrome and Safari already ship this.

annevk:
It seems I should investigate #664 first. In particular, what is the status message for synthetic redirects.

annevk:
Note that the default for `new Response()` is still `OK`. It'd be nice to change that too, but should probably be done separately.

annevk:
I need to file a bug against Chrome, Edge, and Firefox for changing `statusText` for `Redirect.response`. I think this is once again ready.

annevk:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1453612
* https://bugs.chromium.org/p/chromium/issues/detail?id=832041
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16902841/

annevk:
Additional tests are https://github.com/w3c/web-platform-tests/pull/10442.

fetch.bs
```diff
@@ -1320,8 +1320,11 @@ cannot be exposed to script. That would violate <a>atomic HTTP redirect handling
 <a for=/>status</a>. Unless stated otherwise it is <code>200</code>.
 
 <p>A <a for=/>response</a> has an associated
-<dfn export for=response id=concept-response-status-message>status message</dfn>.
-Unless stated otherwise it is `<code>OK</code>`.
+<dfn export for=response id=concept-response-status-message>status message</dfn>. Unless stated
+otherwise it is the empty byte sequence.
+
+<p class=note>This does not have a default value of `<code>OK</code>` as otherwise all HTTP/2
+<a for=/>responses</a> would have that value.
```
jakearchibald: Maybe make it clear that HTTP/2 doesn't support status messages. Something like: "HTTP/2 doesn't support status messages, so this property will be an empty byte sequence for all responses from an HTTP/2 connection"

fetch.bs
```diff
@@ -2901,20 +2904,15 @@ steps:
...
-  if <var>request</var>'s <a for=request>client</a> is non-null.
+  <p>If <var>request</var>'s <a for=request>current url</a>'s <a>cannot-be-a-base-URL flag</a> is
+  set and <a for=url>path</a> contains a single string "<code>blank</code>", then return a new
+  <a for=/>response</a> whose <a for=response>status message</a> is `<code>OK</code>`,
+  <a for=response>header list</a> consist of a single <a for=/>header</a> whose
+  <a for=header>name</a> is `<code>Content-Type</code>` and <a for=header>value</a> is
+  `<code>text/html;charset=utf-8</code>`, <a for=response>body</a> is the empty byte sequence, and
+  <a for=response>HTTPS state</a> is <var>request</var>'s <a for=request>client</a>'s
+  <a for="environment settings object">HTTPS state</a> if <var>request</var>'s
+  <a for=request>client</a> is non-null.
```
jakearchibald: Given the number of properties, dl/ul would be clearer.

fetch.bs
```diff
@@ -2985,9 +2983,10 @@ steps:
 
    <li><p>If <var>dataURLStruct</var> is failure, then return a <a>network error</a>.
 
-   <li><p>Return a <a for=/>response</a> whose <a for=response>header list</a> consist of a single
-   <a for=/>header</a> whose <a for=header>name</a> is `<code>Content-Type</code>` and
-   <a for=header>value</a> is <var>dataURLStruct</var>'s <a for="data: URL struct">MIME type</a>,
+   <li><p>Return a <a for=/>response</a> whose <a for=response>status message</a> is
+   `<code>OK</code>`, whose <a for=response>header list</a> consist of a single <a for=/>header</a>
+   whose <a for=header>name</a> is `<code>Content-Type</code>` and <a for=header>value</a> is
+   <var>dataURLStruct</var>'s <a for="data: URL struct">MIME type</a>,
```
jakearchibald: I know it isn't something you changed in this PR, but there's a repetition of "whose" here which is inconsistent with the similar lines above.

Again, dl/ul could be useful here, especially when it comes to the header list.

PR #603, opened 2017-09-12
Remove "serviceworker" from RequestDestination 

### LINKED ISSUE #602: "serviceworker" in the RequestDestination enum
From the spec:

> "serviceworker" is omitted from RequestDestination as it cannot be observed from JavaScript. Implementations will still need to support it as a destination. 

But "serviceworker" is clearly listed in the IDL.

martinthomson:
This value isn't something that should appear in the API.  It's for internal use only.

Fixes #602

***
[Preview](https://s3.amazonaws.com/pr-preview/martinthomson/fetch/patch-1.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/358dbf5...martinthomson:913c0e6.html)


annevk:
Thanks! (And we'd like to acknowledge everyone that contributes, but it's not mandatory to be listed or anything.)

PR #616, opened 2017-10-13
Clarify that credentials include HTTP auth

### LINKED ISSUE #612: Doc: Definition of credentials unclear
Currently, the fetch spec defines **credentials** as follows:

> Credentials are HTTP cookies, TLS client certificates, and authentication entries.

with **authentication entries** further defined as follows:

> An authentication entry and a proxy-authentication entry are tuples of username, password, and realm, associated with one or more requests.

Two questions have been raised by my customers:

- From a real-world perspective, when server developers add `Access-Control-Allow-Credentials` they're thinking of cookies and maybe the `Authorization` request header. Is there a comprehensive list of other authentication entries?
- What is the process (if any) of specifying whether any new headers etc. should be added to the list of authentication entries?

When I checked the fetch spec example https://fetch.spec.whatwg.org/#example-cors-with-credentials (which I originally wrote, with editing by @annevk!) it uses the obvious example of Set-Cookie headers being ignored. Interested to know what other cases might apply. From a server-developer POV, what else would eb ignored in the response - everything?

sideshowbarker:
Fixes #612.

***
[Preview](https://fetch.spec.whatwg.org/branch-snapshots/sideshowbarker/credentials-clarification/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/34027b6...80231d2.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -166,10 +166,9 @@ Standards.
 
 <hr>
 
-<p><dfn id=credentials export>Credentials</dfn> are HTTP cookies, TLS client certificates, and, for
-HTTP authentication, <a lt="authentication entry">authentication entries</a> and <a
-lt="proxy-authentication entry">proxy-authentication entries</a>. [[!COOKIES]] [[!TLS]]
-[[!HTTP-AUTH]]
+<p><dfn id=credentials export>Credentials</dfn> are HTTP cookies, TLS client certificates, and <a
+lt="authentication entry">authentication entries</a> (for HTTP authentication). [[!COOKIES]]
+[[!TLS]] [[!HTTP-AUTH]]
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


annevk:
I think it would be better if instead of clarifying it here, we clarified in "authentication entries" that they are for HTTP authentication. Otherwise there's still the lack of clarity there.

annevk:
That used to be clearer since there was just a link to HTTP authentication I believe, but that got muddled somewhat at some point.

sideshowbarker:
> I think it would be better if instead of clarifying it here, we clarified in "authentication entries" that they are for HTTP authentication. Otherwise there's still the lack of clarity there.

OK I but it doesn’t need to be an either-or thing. I mean, I think it’s helpful to mention HTTP authentication as part of the *credentials* definition, at point of use — without requiring readers to need to go to the *authentication entries* definition to realize it’s about HTTP authentication.

But I can also update the *authentication entries* definition to make it more clear it’s about HTTP authentication.

And anyway I think to begin with the benefit of any changes here are for non-implementor readers, and not really for implementors. So I think if we’re going to add anything more here at all over what’s in the spec already now, it sort of should be optimized/targeted for non-implementor readers.

annevk:
Okay, I could live with a "(for HTTP authentication)" after "authentication entries" as well as the additional references which seem useful either way.

fetch.bs
```diff
@@ -1446,10 +1445,9 @@ is a <a>filtered response</a> whose
 
 <h3 id=authentication-entries>Authentication entries</h3>
 
-<p>An <dfn export>authentication entry</dfn> and a
-<dfn export>proxy-authentication entry</dfn> are tuples of username, password, and realm,
-associated with one or more <a for=/>requests</a>.
-
+<p>An <dfn export>authentication entry</dfn> and a <dfn export>proxy-authentication entry</dfn> are
+tuples of username, password, and realm, used for HTTP authentication, and associated with one or
+more <a for=/>requests</a>
```
annevk: Trailing dot went missing. Also, I'd make it "HTTP and HTTP proxy authentication" maybe?

PR #619, opened 2017-10-21
Address Domenic's Request domintro nits

domfarolino:
@domenic Addressing your Request domintro nits here, and will add the other `init` member info to the constructor entry in the domintro box in another PR soon. That should finish the Request stuff.

(see https://github.com/whatwg/fetch/pull/617#issuecomment-336970308)

***
[Preview](https://s3.amazonaws.com/pr-preview/domfarolino/fetch/request-domintro-nits.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/978cf9a...domfarolino:ff8fe74.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4909,7 +4909,7 @@ initially a new {{AbortSignal}} object.
  <dt><code><var>request</var> . <a attribute for=Request>integrity</a></code>
  <dd>Returns <var>request</var>'s subresource integrity metadata, which is a cryptographic hash of the
  resource being fetched. Its value may consist of multiple hashes separated by whitespace.
- [[!SRI]]
+ [[SRI]]
 
  <dt><code><var>request</var> . <a attribute for=Request>keepalive</a></code>
  <dd>Returns a boolean indicating whether or not <var>request</var> can outlive the global in which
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4880,35 +4879,35 @@ initially a new {{AbortSignal}} object.
...
- <dd>Returns the redirect mode associated with <var>request</var>, which is a string indicating
- how redirects for the request will be handled during fetching. A <a>request</a> will follow
- redirects by default.
+ <dd>Returns the <a>redirect mode</a> associated with <var>request</var>, which is a string
+ indicating how redirects for the request will be handled during fetching. A <a>request</a>
+ will follow redirects by default.
 
  <dt><code><var>request</var> . <a attribute for=Request>integrity</a></code>
- <dd>Returns <var>request</var>'s subresource integrity, which is a cryptographic hash of the
+ <dd>Returns <var>request</var>'s subresource integrity metadata, which is a cryptographic hash of the
```
domenic: Maybe this paragraph should get a [[SRI]] after it.

domfarolino: Cool I added the SRI after that `<dd>`

fetch.bs
```diff
@@ -4880,36 +4879,37 @@ initially a new {{AbortSignal}} object.
...
  resource being fetched. Its value may consist of multiple hashes separated by whitespace.
+ [[!SRI]]
```
annevk: You need to omit the ! as it's a non-normative reference in this case. I don't think it ends up mattering at the moment, but it's good to have it correct in case it ends up mattering some day.

Issue #567, opened 2017-07-19
What to do about requests that should (maybe) preflight but don't?


estark37:
Related issues:
https://github.com/whatwg/fetch/issues/530
https://github.com/bifurcation/expect-ct/issues/18
https://github.com/WICG/reporting/issues/41
https://github.com/httpwg/http-extensions/issues/356

(Feel free to close this if you feel it's duplicative with any of those issues.)

Various specs include various types of "special" requests without CORS preflights, even though they are triggered by web content, to a URL controlled by web content, and are not safe/simple requests.

Many of the examples (CSP, Reporting API, HPKP) are reporting requests, but not all (OCSP). In Expect-CT, the spec lets the UA decide whether a preflight is needed and doesn't take a position on it.

In some cases, like CSP, the lack of preflights is basically a bug, but realistically I doubt we'd be able to require preflights any time soon because of compatibility reasons. In other cases, like OCSP, it doesn't seem likely that implementations will ever want to preflight because of layering reasons.

Should Fetch handle these requests in some way?

annevk:
These are the bits you can potentially control in a request:

* URL
* method
* headers
* body

To make a request without a CORS preflight, you can only control a couple of headers and the method has to be either GET or POST. There's no restrictions on the body or URL in that case, but the `Content-Type` header can only be a couple of limited MIME types.

Now browsers sometimes go beyond those restrictions and don't issue a CORS preflight, but whenever this has happened thus far that comes with different restrictions. E.g., for reports the MIME type is fixed (and rather unique), the body is mostly controlled by the browser following the MIME type definition, and methods and headers are also fixed (and not interesting). That makes it highly unlikely those requests could be used for an attack.

It's a little unclean though as the browser effectively has some super powers that are not available (and cannot be made available) to content. And as long as it's not clearly documented it means that browsers end up issuing requests server administrators might not expect (this is also assuming there's no rogue extensions and such in play).

I think the simplest thing to do would be to embrace that it's unclean and clearly document these further same-origin policy exceptions in Fetch and encourage anyone that wants to introduce yet more to first discuss that and strongly consider abiding by the same-origin policy instead (and use CORS preflights if need be). But I'm happy to consider alternatives.

estark37:
I agree that actual security risk seems minimal. @annevk what would be the process for spec authors to discuss exceptions that they might want to introduce? Just file an issue against Fetch?

[...]

PR #621, opened 2017-10-26
Document CORS safelist exceptions

estark37:
As discussed in https://github.com/whatwg/fetch/issues/567, browsers have
allowed various cross-origin requests with non-safelisted Content-Type header
values to be sent without CORS preflights. These have occurred either by
accident (and now can't be reversed for compatibility reasons) or because of
design constraints (requests that are implemented outside of the web platform
layer). These CORS exceptions are believed to be safe, but the spec should
document them so that servers know to expect them.

I've added a note about the Content-Type exceptions, but haven't added them to
the safelist, because doing so would imply that web content can trigger
requests with these Content-Type headers and arbitrary bodies. We don't want to
allow fully attacker-controlled requests with these headers, but rather just
want to document the current state where web content can trigger the requests
but not control the headers or bodies.

***
[Preview](https://s3.amazonaws.com/pr-preview/estark37/fetch/cors-exceptions.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/fetch/e5abf0f...estark37:3d17f9f.html)

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -445,6 +445,14 @@ whose <a for=header>name</a> is a <a>byte-case-insensitive</a> match for one of
 <p>and whose <a for=header>value</a>, <a lt="extract header values">once extracted</a>, is not
 failure.
 
+<p class="note">Browsers allow limited exceptions to the CORS safelist for non-safelisted
+`<code>Content-Type</code>` values. These exceptions are made for requests that can be triggered by
+web content but whose headers and bodies can be only minimally controlled by the web content.
+Therefore, servers should expect cross-origin web content to be allowed to trigger non-preflighted
+requests with the following non-safelisted `<code>Content-Type</code>` values:
+`<code>application/csp-report</code>`, `<code>application/report</code>`,
+`<code>application/expect-ct-report+json</code>`, `<code>application/ocsp-request</code>`.
+
 <p>A <dfn export>CORS non-wildcard request-header name</dfn> is a <a>byte-case-insensitive</a> match
 for `<code>Authorization</code>`.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -2253,9 +2253,14 @@ triggered by web content but whose headers and bodies can be only minimally cont
 content. Therefore, servers should expect cross-origin web content to be allowed to trigger
 non-preflighted requests with the following non-safelisted `<code>Content-Type</code>` header
 values:
-`<code>application/csp-report</code>` [[CSP]], `<code>application/report</code>` [[REPORTING]],
-`<code>application/expect-ct-report+json</code>` [[EXPECT-CT]],
-`<code>application/xss-auditor-report</code>`, and `<code>application/ocsp-request</code>` [[OCSP]].
+
+<ul class=brief>
+ <li>`<code>application/csp-report</code>` [[CSP]]
+ <li>`<code>application/report</code>` [[REPORTING]]
+ <li>`<code>application/expect-ct-report+json</code>` [[EXPECT-CT]]
+ <li>`<code>application/xss-auditor-report</code>`
+ <li>`<code>application/ocsp-request</code>` [[OCSP]]
+</ul>
 
 <p>Specifications should avoid introducing new exceptions and should only do so with careful
 consideration for the security consequences. New exceptions can be proposed by
--- a/fetch.bs
+++ b/fetch.bs
@@ -97,6 +97,22 @@ url: https://tools.ietf.org/html/rfc7234#section-1.2.1;text:delta-seconds;type:d
         "publisher": "US-CERT",
         "href": "https://www.kb.cert.org/vuls/id/150227",
         "title": "HTTP proxy default configurations allow arbitrary TCP connections."
+    },
+    "REPORTING": {
+        "authors": ["Ilya Grigorik", "Mike West"],
+        "href": "https://wicg.github.io/reporting/",
+        "title": "Reporting API"
+    },
+    "EXPECT-CT": {
+        "authors": [
+            "Emily Stark"
+        ],
+        "href": "https://tools.ietf.org/html/draft-ietf-httpbis-expect-ct-02",
+        "publisher": "IETF",
+        "title": "Expect-CT Extension for HTTP"
+    },
+    "OCSP": {
+        "aliasOf": "RFC2560"
     }
 }
 </pre>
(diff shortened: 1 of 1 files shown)


estark37:
@annevk I don't know (a) if we are ready to take this approach yet (discussion on #567 is not totally satisfyingly resolved) and (b) if this is the right way to document the existing CORS exceptions. But I thought I'd throw something together as a strawman for discussion.

annevk:
Thanks for starting this. I agree that the discussion isn't quite resolved, but we might as well document this meanwhile and discourage further such deviations.

I think this text is in part normative due to the requirement put on servers (mostly to be aware of it) and therefore I'd suggest we create a new "CORS protocol exceptions" section within the "CORS protocol" section after the "Examples" section, so as 3.2.7., with this text. And perhaps the note you have now could become a simple pointer to that section.

Also, instead of framing it on behalf of browsers, I would suggest saying it's specifications that caused this. And also caution new specifications to not go further down this path (or at least not do so without discussion).

estark37:
@annevk addressed your suggestions, PTAL. Do we want to expand more on what should be involved in introducing a new exception (e.g. consideration for how much the attacker controls the request body and other headers, the uniqueness of the Content-Type value, etc.)?

annevk:
Perhaps we should simply encourage them to file an issue at https://github.com/whatwg/fetch/issues/new?

annevk:
Would be nice if @whatwg/security could have a quick look at this as well.

fetch.bs
```diff
@@ -2226,6 +2229,19 @@ Access-Control-Allow-Credentials: true</pre>
...
 
+<h4 id=cors-protocol-exceptions>CORS protocol exceptions</h4>
+
+<p>Specifications have allowed limited exceptions to the CORS safelist for non-safelisted
+`<code>Content-Type</code>` values. These exceptions are made for requests that can be triggered by
+web content but whose headers and bodies can be only minimally controlled by the web content.
+Therefore, servers should expect cross-origin web content to be allowed to trigger non-preflighted
+requests with the following non-safelisted `<code>Content-Type</code>` values:
+`<code>application/csp-report</code>`, `<code>application/report</code>`,
+`<code>application/expect-ct-report+json</code>`, `<code>application/ocsp-request</code>`.
```
annevk: Please add "and" here.

estark37:
Added "header" to "Content-Type", added an "and", and added a pointer to file an issue for new exceptions. Also wanted to give a heads-up to @igrigorik and @juliatuttle that I'm including the Reporting API content-type in this.

annevk:
Thanks, I'd really like review from one more person before landing this. Another thing that might be good here is to have non-normative reference for each of the mentioned MIME types, in case folks want to read up on what they can expect.

annevk:
Also, I saw on Twitter that the motivation for adding this might be to remove CORS preflights from all reporting infrastructure?

estark37:
> Also, I saw on Twitter that the motivation for adding this might be to remove CORS preflights from all reporting infrastructure?

The motivation for adding this is to document the existing exceptions. The only thing which currently preflights (in Chrome) but I'd like to remove is Expect-CT, because it's done as part of certificate verification and in my mind is similar to OCSP requests. But I can take that Content-Type out of this commit and save it for a separate discussion if you'd like.

annevk:
I personally think it's fine, but I wanted it to be clear for other people who might follow along.

fetch.bs
```diff
@@ -2226,6 +2229,20 @@ Access-Control-Allow-Credentials: true</pre>
...
 
+<h4 id=cors-protocol-exceptions>CORS protocol exceptions</h4>
+
+<p>Specifications have allowed limited exceptions to the CORS safelist for non-safelisted
+`<code>Content-Type</code>` header values. These exceptions are made for requests that can be
+triggered by web content but whose headers and bodies can be only minimally controlled by the web
+content. Therefore, servers should expect cross-origin web content to be allowed to trigger
+non-preflighted requests with the following non-safelisted `<code>Content-Type</code>` header
+values: `<code>application/csp-report</code>`, `<code>application/report</code>`,
+`<code>application/expect-ct-report+json</code>`, and `<code>application/ocsp-request</code>`.
```
mikewest: Chrome also sends [`application/xss-auditor-report`](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/loader/PingLoader.cpp?rcl=898669974e9ce0867e3dc80fe9385de8983a77ce&l=298).

annevk:
(I have a somewhat unrelated question. Are all of these reports JSON? Should we consider updating https://html.spec.whatwg.org/#json-mime-type so user agents can show more of these as JSON when you'd browse to them directly (admittedly a rather unlikely scenario). I'm somewhat surprised they don't all use the `+json` convention.)

mikewest:
I think all of these reports are JSON. OCSP request bodies are not.

I think I'd be fine with adding something like `+json` to the MIME type if that would be helpful. My main concern is to distinguish these from other kinds of JSON requests that the server might be expecting in order to mitigate the risk of CSRF, which meant getting them off `application/json` directly. But `application/report+json` seems less likely to be a problem, even for less well-behaved applications.

mikewest:
We talked about this in the WebAppSec meeting at TPAC last week. My sense of the room was that folks were fine with adding it to Fetch. @estark37, WDYT?

I asked folks in the room from other vendors (@dveditz, @johnwilander, @patrickkettner, et al) to skim through their codebases to see if they had similar mechanisms (and/or to align with these MIME types for existing mechanisms).

annevk:
@mikewest if we can still change these I think using `+json` would be a lot better. Otherwise your generic JSON end point has to know about all kinds of MIME types.

estark37:
Sorry for dropping this on the floor! I added references and application/xss-auditor-report (though I couldn't find anything explanatory for it, @mikewest do you know of anything?) I'm not sure how to make these into informative references though?

mikewest:
> @mikewest if we can still change these I think using +json would be a lot better. Otherwise your generic JSON end point has to know about all kinds of MIME types.

I'd suggest we land this patch as-is to reflect the status quo, and come back and update it if/when we are able to update the MIME types browsers are sending.

> application/xss-auditor-report (though I couldn't find anything explanatory for it, @mikewest do you know of anything?)

Nope. We've done a bad job documenting this. When we move it into the reporting API, we'll need to write up the format (or add some metrics and look at usage to see if anyone cares about it at all). I'd suggest we come back and link it then.

annevk:
@mikewest is there a follow-up issue for changing the MIME type that I can subscribe to? And make sure that results in a change to Fetch if made?

@estark37 you made them non-normative already (by omitting a ! in front of the letters). However, the dominant style is to put all references at the end of the paragraph after the period. Perhaps that would read weird as the connection between the value and the reference is no longer as clear though. What do you think about making this a list?

estark37:
> What do you think about making this a list?

Sure, done.

annevk:
Thanks a lot!

PR #670, opened 2018-02-09
Following signals. Fixes #668.

### LINKED ISSUE #668: "If signal is not null, then add the following a..."
https://fetch.spec.whatwg.org/commit-snapshots/36ef3c8aef34ff77ebf713b6498d008fe853034f/#request-class

> If signal is not null, then add the following abort steps to signal:​
> 
> Signal abort on r’s signal\.

I may be reading this wrong, but it appears to me that this implies that r's signal will not have the aborted flag set even if |signal| does.

This doesn't appear to the correct behaviour. For example, test "Signal on request object" https://github.com/w3c/web-platform-tests/blob/0d311189506d0268f62198fde499a7a32fc58c87/fetch/api/abort/general.any.js#L71 expects that a fetch will be aborted if created from a Request that was passed an already-aborted signal.

I think it should be

31. If _signal_ is not null, then
    1. If _signal_’s [aborted flag](https://dom.spec.whatwg.org/#abortsignal-aborted-flag) is set, set _r_'s [signal](https://fetch.spec.whatwg.org/#request-signal)'s [aborted flag](https://dom.spec.whatwg.org/#abortsignal-aborted-flag).
    1. Otherwise, [add the following abort steps](https://dom.spec.whatwg.org/#abortsignal-add) to _signal_:
       1. [Signal abort](https://dom.spec.whatwg.org/#abortsignal-signal-abort) on _r_’s [signal](https://fetch.spec.whatwg.org/#request-signal).

jakearchibald:
Handles cases where the signal's aborted flag is already set.

***
<a href="https://whatpr.org/fetch/670.html" title="Last updated on Feb 12, 2018, 8:41 AM GMT (1a4c685)">Preview</a> | <a href="https://whatpr.org/fetch/670/36ef3c8...1a4c685.html" title="Last updated on Feb 12, 2018, 8:41 AM GMT (1a4c685)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5273,8 +5273,8 @@ constructor must run these steps:
  <a for=Headers>header list</a> is <var>request</var>'s <a for=request>header list</a>, and
  <a for=Headers>guard</a> is "<code>request</code>".
 
- <li><p>If <var>signal</var> is not null, then <var>r</var>'s <a for=Request>signal</a>
- <a for=AbortSignal>follows</a> <var>signal</var>.
+ <li><p>If <var>signal</var> is not null, then make <var>r</var>'s <a for=Request>signal</a>
+ <a for=AbortSignal>follow</a> <var>signal</var>.
 
  <li><p>Let <var>headers</var> be a copy of <var>r</var>'s <a for=Request>headers</a> and its
  associated <a for=Headers>header list</a>.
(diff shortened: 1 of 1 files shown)


domenic:
As with https://github.com/whatwg/dom/pull/569, I'm happy to merge in @annevk's absence to mitigate implementer confusion on an area being actively implemented.

However, can you check that these two cases are already tested somewhere?

jakearchibald:
The tests are ok.

https://github.com/w3c/web-platform-tests/blob/master/fetch/api/abort/general.any.js#L71 will only pass if the new signal copies the current state of the signal passed to the request constructor. However, I added an explicit check: https://github.com/w3c/web-platform-tests/pull/9515.

Cloning is already tested, and this was fine in the spec before these changes anyway. https://github.com/w3c/web-platform-tests/blob/master/fetch/api/abort/general.any.js#L491

Issue #463, opened 2017-01-24
Specify body transfer in Request constructor more precisely


yutakahirano:
This is a follow-up for https://github.com/whatwg/fetch/pull/458. Transform streams and piping need to be specified more precisely in the Streams spec.

annevk:
@ricea defined transform streams, so is this fixable now?

PR #675, opened 2018-02-28
Allow used body replacement in Request constructor

### LINKED ISSUE #674: Inconsistency in Request constructor when rewriting a used body
Script authors sometimes have a need to rewrite a Request's body, e.g. a service worker that transforms POST messages. There is currently no easy way to do this if the original body must first be read. Currently, the following code will throw:

```javascript
let request = new Request("https://example.com/", {
  method: "POST",
  body: "an old body"
})

// Disturb the request's body.
let body = await request.text()

// Throws in Request ctor, step 7
let newRequest = new Request(request, {
  body: "a new body"
})
```

The last line throws, because Request's constructor is supposed to check that its first parameter is not disturbed (step 7) *before* it checks if its initialization dictionary contains a `body` property that will override the old body (steps 38-44).

This seems like a bug in the spec, because if we were to comment out the `await request.text()` line, the code snippet would work as intended: `newRequest` would have `"a new body"` as its body. That tells me that the intent was probably to allow authors to concisely rewrite bodies in just this way.

Note that even if step 7's check were modified or moved after the body extraction steps, there remains some ambiguity: if `Request()`'s first parameter is another Request object, step 43 unconditionally creates a new ReadableStream object from the old request's body, which I would expect to throw if the body were disturbed. Perhaps #463 will resolve that ambiguity.

/cc @kentonv

harrishancock:
Currently, the following code snippet replaces `request`'s body as one would expect:

```javascript
let request = new Request(url, { method: "POST", body: "foo" })
request = new Request(request, { body: "bar" })
```

But this snippet throws a TypeError early in Request's constructor:

```javascript
let request = new Request(url, { method: "POST", body: "foo" })
await request.text()  // disturb the body
request = new Request(request, { body: "bar" })  // throws
```

This commit's changes allows the latter code snippet to work like the first one.

Fixes #674.

***
<a href="https://whatpr.org/fetch/675.html" title="Last updated on Mar 15, 2018, 8:10 AM GMT (03a8ac0)">Preview</a> | <a href="https://whatpr.org/fetch/675/ae71682...03a8ac0.html" title="Last updated on Mar 15, 2018, 8:10 AM GMT (03a8ac0)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5088,9 +5088,6 @@ constructor must run these steps:
   <p>Otherwise (<var>input</var> is a {{Request}} object), run these substeps:
 
   <ol>
-   <li><p>If <var>input</var> is <a for=Body>disturbed</a> or <a for=Body>locked</a>, then
-   <a>throw</a> a <code>TypeError</code>.
-
    <li><p>Set <var>request</var> to <var>input</var>'s
    <a for=Request>request</a>.
 
@@ -5348,10 +5345,14 @@ constructor must run these steps:
    <a for=request>use-CORS-preflight flag</a>.
   </ol>
 
+ <li><p>If <var>inputBody</var> is <var>body</var> and <var>input</var> is <a for=Body>disturbed</a>
+ or <a for=Body>locked</a> then <a>throw</a> a <code>TypeError</code>.
+
  <!-- Any steps after this must not throw. -->
 
  <li>
-  <p>If <var>inputBody</var> is non-null, then run these substeps:
+  <p>If <var>inputBody</var> is non-null and <var>inputBody</var> is <var>body</var>, then run these
+  substeps:
 
   <ol>
    <li>
@@ -5366,10 +5367,9 @@ constructor must run these steps:
     <p class="note no-backref">This makes <var>inputBody</var>'s <a for=body>stream</a>
     <a for=ReadableStream>locked</a> and <a for=ReadableStream>disturbed</a> immediately.
 
-   <li><p>If <var>inputBody</var> is <var>body</var>, then set <var>body</var> to a new
-   <a for=/>body</a> whose <a for=body>stream</a> is <var>rs</var>, whose <a for=body>source</a>
-   is <var>inputBody</var>'s <a for=body>source</a> and whose <a for=body>total bytes</a> is
-   <var>inputBody</var>'s <a for=body>total bytes</a>.
+   <li><p>Set <var>body</var> to a new <a for=/>body</a> whose <a for=body>stream</a> is
+   <var>rs</var>, whose <a for=body>source</a> is <var>inputBody</var>'s <a for=body>source</a> and
+   whose <a for=body>total bytes</a> is <var>inputBody</var>'s <a for=body>total bytes</a>.
   </ol>
 
  <li><p>Set <var>r</var>'s <a for=Request>request</a>'s <a for=request>body</a> to <var>body</var>.
@@ -6251,6 +6251,7 @@ Glenn Maynard,
 Graham Klyne,
 Hal Lockhart,
 Hallvord R. M. Steen,
+Harris Hancock,
 Henri Sivonen,
 Henry Story,
 Hiroshige Hayashizaki,
--- a/fetch.bs
+++ b/fetch.bs
@@ -5346,13 +5346,12 @@ constructor must run these steps:
   </ol>
 
  <li><p>If <var>inputBody</var> is <var>body</var> and <var>input</var> is <a for=Body>disturbed</a>
- or <a for=Body>locked</a> then <a>throw</a> a <code>TypeError</code>.
+ or <a for=Body>locked</a>, then <a>throw</a> a <code>TypeError</code>.
 
  <!-- Any steps after this must not throw. -->
 
  <li>
-  <p>If <var>inputBody</var> is non-null and <var>inputBody</var> is <var>body</var>, then run these
-  substeps:
+  <p>If <var>inputBody</var> is non-null and <var>inputBody</var> is <var>body</var>, then:
 
   <ol>
    <li>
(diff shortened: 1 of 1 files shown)


annevk:
@terinjokes could you add @harrishancock as a public member of the cloudflare-whatwg organization?

annevk:
I wonder if we should move step 7.1 down to the other body checks (and I also wonder if we could simplify those somewhat).

terinjokes:
@annevk This is in progress.

terinjokes:
@annevk Done. Sorry for the delay.

harrishancock:
> I wonder if we should move step 7.1 down to the other body checks (and I also wonder if we could simplify those somewhat).

Agreed about moving the check down with the others. I'll find a nice spot for it and amend the commit.

I'm not sure there's a better factorization of the body setup steps (38-44). :/ It seems largely complicated by the fact that nothing after the *input* Request object body stealing (setting up *rs* and piping to it) is allowed to throw (to enforce the strong exception safety guarantee, I suppose). Maybe the best thing to do would be to defer the piping action until after all the checks are complete?

1. Let *action* be null.
2. Let *body* be null.
3. *[Step 41]* If *init*'s `body` member is present, run these substeps:
   1. Let *Content-Type* be null.
   2. If *init*'s `keepalive` member is present and is true, then set *body*
      and *Content-Type* to the result of extracting *init*'s `body` member,
      with *keepalive flag* set.
   3. Otherwise, set *body* and *Content-Type* to the result of extracting
      *init*'s `body` member.
   4. If *Content-Type* is non-null and *r*'s headers's header list does not
      contain '`Content-Type`', then append '`Content-Type`'/*Content-Type* to
      *r*'s headers.
4. Otherwise, if *input* is a `Request` object, run these substeps:
   1. *[Step 7.1]* If *input* is disturbed or locked, throw a `TypeError`.
   2. Set *body* to *input*'s request's body.
   3. Set *action* to an action that runs the following substeps:
      1. *[Step 43.1]* Let *rs* be a `ReadableStream` object from which one can read the exact
         same data as one could read from *body*'s stream.
      2. *[Step 43.2]* Set *body* to a new body whose stream is *rs*, whose source is *body*'s
         source and whose total bytes is *body*'s total bytes.
5. *[Step 39]* If *body* is non-null and *request*'s method is `'GET'` or `'HEAD'`, then
   throw a `TypeError`.
6. *[Step 42]* If *body* is non-null and *body*'s source is null, then run these substeps:
   1. If *r*'s request's mode is neither "`same-origin`" nor "`cors`", then
      throw a `TypeError`.
   2. Set *r*'s request's use-CORS-preflight flag.
7. If *action* is non-null, run *action* **immediately**.
8. *[Step 44]* Set *r*'s request's body to *body*.

I'm not really sure that's "better". :) @yutakahirano?

harrishancock:
> I wonder if we should move step 7.1 down to the other body checks

Actually, is the check in 7.1 even necessary? The logical place for it is to add it to the beginning of step 43's substeps, but it seems like it's already implicit in step 43.1 -- piping to `rs`.

annevk:
Good question, I guess it depends on if it's observably different. Maybe not, though we'll need to fix https://github.com/whatwg/fetch/issues/463 to be sure.

yutakahirano:
My intention has been they are same.

yutakahirano:
Whether we should keep step 7.1 depends on whether we want to separate the read-only part and state-modifying part (currently we don't throw an exception from this part).

harrishancock:
Amended to move the disturbed or locked check closer to where it's relevant: right before the *inputBody* piping. To me, it would make most sense to either put the check in the *inputBody* piping substeps, or make it implicit in the piping itself, as I mentioned. Until that decision is finalized, this seems like the best we can do: the "critical line" between the read-only part of the constructor and the state-modifying part remains in the top-level steps (as opposed to the piping substeps, which are conditional); the piping operation is guaranteed not to throw; and the check and the piping operation are written as close to each other as possible to make future maintenance easier.

yutakahirano:
Can you add a web platform test?

harrishancock:
> Can you add a web platform test?

Will do. I had hoped to get to it today, but ran out of time. Soon!

harrishancock:
Ping. :) WPT PR here: https://github.com/w3c/web-platform-tests/pull/9931

fetch.bs
```diff
@@ -5348,10 +5345,14 @@ constructor must run these steps:
    <a for=request>use-CORS-preflight flag</a>.
   </ol>
 
+ <li><p>If <var>inputBody</var> is <var>body</var> and <var>input</var> is <a for=Body>disturbed</a>
+ or <a for=Body>locked</a> then <a>throw</a> a <code>TypeError</code>.
```
annevk: Needs a comma before then.

fetch.bs
```diff
@@ -5348,10 +5345,14 @@ constructor must run these steps:
...
+
  <!-- Any steps after this must not throw. -->
 
  <li>
-  <p>If <var>inputBody</var> is non-null, then run these substeps:
+  <p>If <var>inputBody</var> is non-null and <var>inputBody</var> is <var>body</var>, then run these
+  substeps:
```
annevk: This should be just ", then:"

fetch.bs
```diff
@@ -5366,10 +5367,9 @@ constructor must run these steps:
     <p class="note no-backref">This makes <var>inputBody</var>'s <a for=body>stream</a>
     <a for=ReadableStream>locked</a> and <a for=ReadableStream>disturbed</a> immediately.
 
-   <li><p>If <var>inputBody</var> is <var>body</var>, then set <var>body</var> to a new
-   <a for=/>body</a> whose <a for=body>stream</a> is <var>rs</var>, whose <a for=body>source</a>
-   is <var>inputBody</var>'s <a for=body>source</a> and whose <a for=body>total bytes</a> is
-   <var>inputBody</var>'s <a for=body>total bytes</a>.
+   <li><p>Set <var>body</var> to a new <a for=/>body</a> whose <a for=body>stream</a> is
+   <var>rs</var>, whose <a for=body>source</a> is <var>inputBody</var>'s <a for=body>source</a> and
+   whose <a for=body>total bytes</a> is <var>inputBody</var>'s <a for=body>total bytes</a>.
```
annevk: This needs an Oxford comma. (Also a problem in the original it seems.)

annevk:
It seems this feature isn't implemented by browsers yet so no impl bugs to file.

annevk:
If someone could confirm the above I think we're good to go.

ricea:
@annevk Safari tech preview has it, does that count? It throws TypeError: "Request input is disturbed or locked.".

annevk:
That certainly counts, thanks! @harrishancock could you file a bug at https://bugs.webkit.org/enter_bug.cgi?product=WebKit&component=HTML%20DOM?

harrishancock:
@annevk, thank you for fixing the nits, I appreciate it!

I've submitted a WebKit bug here: https://bugs.webkit.org/show_bug.cgi?id=183703

annevk:
Thanks, all merged now!

Issue #572, opened 2017-08-01
Body should not be transmitted if service worker intercepts request


jakearchibald:
https://fetch.spec.whatwg.org/#http-fetch 3.3.1 "Transmit body for request".

If the service worker has provided a response, it should have transmitted the body itself in a separate call to fetch, or transmitting is unnecessary since the SW provided a response from the cache API or similar.

annevk:
cc @yutakahirano 

So this relates to upload streams. Main question I guess is whether a stream gets drained if the service worker doesn't touch it? (If no, this means that a page can observe a service worker in a new way, maybe unexpected.)

annevk:
cc @domenic

wanderview:
Do we care about sites observing there is a service worker?  We break that in other places.

For example, this code:

```
let response = await fetch(crossOriginURL, { mode: 'no-cors' });
let text = await response.text();
```

Without a service worker the response will always be opaque and the text will be empty string.  A service worker can return a basic Response, though, and the spec currently says to expose that basic response to the outer page.  So in that case the body could be observed.   (There was a spec issue about this case, but I don't have it handy at the moment.)

jakearchibald:
Seems fine that it's observable. Ideally it would cancel the body stream if it doesn't need it.

annevk:
Yeah, I wonder how it works for responses created in a service worker today. Since that also somewhat depends on how we cross the boundaries between Document<>Fetch<>Service Worker<>Network.

domenic:
I'm having a bit of trouble understanding that step (3.3.1) in the first place, but I think it is probably unnecessary, yeah... I mean, the body needs to get from the main thread to the service worker, but it sounds like that step is not how it happens? How does it happen?

wanderview:
It effectively needs the transfer thing we've talked about for postMessage.

annevk:
Before streams the way it happened is that requests and responses (as concepts) can just be transferred across threads. But JavaScript streams were never backed by internal streams so at that point that strategy started falling apart and we haven't really made an effort to patch it back together again.

yutakahirano:
Sorry for the delay. Probably I added the sentence but it doesn't make sense at all. I'll create a PR which removes the item.

annevk:
Note that just removing it doesn't really work either, since it results in XMLHttpRequest not getting the tasks queued that it needs for uploads. Unless I'm missing something.

annevk:
[...]

PR #678, opened 2018-03-06
Remove unnecessary body transmission for service worker trapped requests

yutakahirano:
As the request is trapped by service worker, there is no need to transmit the body to network.

See #572 for more info.

***
<a href="https://whatpr.org/fetch/678.html" title="Last updated on Mar 6, 2018, 8:59 AM GMT (a572aed)">Preview</a> | <a href="https://whatpr.org/fetch/678/ae71682...a572aed.html" title="Last updated on Mar 6, 2018, 8:59 AM GMT (a572aed)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3044,8 +3044,6 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
     <p>If <var>response</var> is not null, then run these substeps:
 
     <ol>
-     <li><p><a for=request>Transmit body</a> for <var>request</var>.
-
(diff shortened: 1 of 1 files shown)


annevk:
Doesn't this cause XMLHttpRequest upload progress events to not dispatch?

yutakahirano:
This is the service worker intercept case, and transmitting the body to network makes no sense at all. Can we fix the upload progress issue separately?

annevk:
Transmit body doesn't necessarily mean a network transmission. It can be from one process to another. If you run blame on this line I think you'll find we added this for XHR.

PR #685, opened 2018-03-26
Introduce {reload, history, submission}-navigation flags

yutakahirano:
This change introduces {reload, history, submission}-navigation flags and
is{Reload, History, Submission}Navigation attributes on Request class.

This is for https://github.com/w3c/ServiceWorker/issues/1167.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1162,6 +1162,18 @@ Unless stated otherwise, it is "<code>basic</code>".
 <p>A <a for=/>request</a> has an associated <dfn export for=request id=done-flag>done flag</dfn>.
 Unless stated otherwise, it is unset.
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=reload-navigation-flag>reload-navigation flag</dfn>. Unless stated
+otherwise, it is unset.
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=history-navigation-flag>history-navigation flag</dfn>. Unless stated
+otherwise, it is unset.
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=submission-navigation-flag>submission-navigation flag</dfn>. Unless stated
+otherwise, it is unset.
+
 <p class="note no-backref">A <a for=/>request</a>'s
 <a for=request>url list</a>,
 <a for=request>current url</a>,
@@ -4882,6 +4894,9 @@ interface Request {
   readonly attribute RequestRedirect redirect;
   readonly attribute DOMString integrity;
   readonly attribute boolean keepalive;
+  readonly attribute boolean isReloadNavigation;
+  readonly attribute boolean isHistoryNavigation;
+  readonly attribute boolean isSubmissionNavigation;
   readonly attribute AbortSignal signal;
 
   [NewObject] Request clone();
@@ -5039,6 +5054,16 @@ initially a new {{AbortSignal}} object.
  <dd>Returns a boolean indicating whether or not <var>request</var> can outlive the global in which
  it was created.
 
+ <dt><code><var>request</var> . <a attribute for=Request>isReloadNavigation</a></code>
+ <dd>Returns a boolean indicating whether or not <var>request</var> is for a reload navigation.
+
+ <dt><code><var>request</var> . <a attribute for=Request>isHistoryNavigation</a></code>
+ <dd>Returns a boolean indicating whether or not <var>request</var> is for a back or forward
+ navigation.
+
+ <dt><code><var>request</var> . <a attribute for=Request>isSubmissionNavigation</a></code>
+ <dd>Returns a boolean indicating whether or not <var>request</var> is for a submission navigation.
+
  <dt><code><var>request</var> . <a attribute for=Request>signal</a></code>
  <dd>Returns the signal associated with <var>request</var>, which is an
  {{AbortSignal}} object indicating whether or not <var>request</var> has been aborted, and its abort
@@ -5141,8 +5166,11 @@ constructor must run these steps:
  <a for=request>redirect mode</a>,
  <a for=request>integrity metadata</a> is
  <var>request</var>'s
- <a for=request>integrity metadata</a>, and
- <a>keepalive flag</a> is <var>request</var>'s <a>keepalive flag</a>.
+ <a for=request>integrity metadata</a>,
+ <a>keepalive flag</a> is <var>request</var>'s <a>keepalive flag</a>,
+ <a>reload-navigation flag</a> is  <var>request</var>'s <a>reload-navigation flag</a>,
+ <a>history-navigation flag</a> is  <var>request</var>'s <a>history-navigation flag</a>, and
+ <a>submission-navigation flag</a> is  <var>request</var>'s <a>submission-navigation flag</a>.
 
 
  <li>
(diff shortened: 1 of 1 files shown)


yutakahirano:
I'll work on the html spec to set the flags.

fetch.bs
```diff
@@ -1162,6 +1162,18 @@ Unless stated otherwise, it is "<code>basic</code>".
...
+<dfn export for=request id=reload-navigation-flag>reload-navigation flag</dfn>. Unless stated
+otherwise, it is unset.
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=history-navigation-flag>history-navigation flag</dfn>. Unless stated
+otherwise, it is unset.
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=submission-navigation-flag>submission-navigation flag</dfn>. Unless stated
+otherwise, it is unset.
```
annevk: It would be good to add a note here why we need these. I also wonder if they should be tri-state since they don't make sense for the Cache API.

jakearchibald: Tri-state so it can have an 'unset' value? Seems like it'd be less confusing to just store these values in the cache API, even though it isn't particularly useful.

fetch.bs
```diff
@@ -5141,8 +5166,11 @@ constructor must run these steps:
  <a for=request>redirect mode</a>,
  <a for=request>integrity metadata</a> is
  <var>request</var>'s
- <a for=request>integrity metadata</a>, and
- <a>keepalive flag</a> is <var>request</var>'s <a>keepalive flag</a>.
+ <a for=request>integrity metadata</a>,
+ <a>keepalive flag</a> is <var>request</var>'s <a>keepalive flag</a>,
+ <a>reload-navigation flag</a> is  <var>request</var>'s <a>reload-navigation flag</a>,
+ <a>history-navigation flag</a> is  <var>request</var>'s <a>history-navigation flag</a>, and
+ <a>submission-navigation flag</a> is  <var>request</var>'s <a>submission-navigation flag</a>.
```
jakearchibald: @annevk do you think this should become an unordered list? It's becoming a bit tough to read.

fetch.bs
annevk: Yeah, to indicate it's not relevant/known. I guess I'm fine with storing more bits though.

fetch.bs
annevk: We could do that, though if we do we should perhaps also clean up the other 2-3 "a new request" places to make them all similar.

Also, if we don't, we should remove the double spaces here.

jakearchibald:
LGTM if the double spaces issue is resolved.

fetch.bs
wanderview: In general I think any field that is added to `Request` must be stored in cache API.  At least that is what we currently do and I think it would be unexpected for `cache.keys()` to return something that differs from what was passed to `cache.put()`.

domenic:
For anyone wondering, I figured out from around https://github.com/w3c/ServiceWorker/issues/1167#issuecomment-343209882 and nearby comments that form submission and reload can both be true. Maybe a non-normative note explaining the design would be helpful.

fetch.bs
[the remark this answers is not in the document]
yutakahirano: I think `r2.mode` will be "navigation"... or are you talking about "If any of init’s members are present,..."?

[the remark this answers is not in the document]
wanderview: Right.  I was thinking of that clause.  Should we reset this property in the same case?

yutakahirano:
> For anyone wondering, I figured out from around w3c/ServiceWorker#1167 (comment) and nearby comments that form submission and reload can both be true. (That's why it's three booleans instead of an enum.) Maybe a non-normative note explaining the design would be helpful.

Added some texts.

fetch.bs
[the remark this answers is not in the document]
yutakahirano: Nice catch, thank you. Fixed.

wanderview:
Sorry, I suck at github review interface.

annevk:
Can't isHistoryNavigation and isSubmissionNavigation be true at the same time as well? If you navigate back to a page that was the result of a form submission? What happens in that case?

yutakahirano:
I come to think that narrowing down this change to reload-only is good. That way it will be aligned with other PRs(https://github.com/w3c/web-platform-tests/pull/10192 and https://github.com/whatwg/html/pull/3592) and we can land the changes without waiting for other features / tests. @annevk, @domenic , what do you think about the idea?

annevk:
That sounds like a good idea; a lot less complexity to think through.

yutakahirano:
Done.

fetch.bs
```diff
@@ -1162,6 +1162,10 @@ Unless stated otherwise, it is "<code>basic</code>".
 <p>A <a for=/>request</a> has an associated <dfn export for=request id=done-flag>done flag</dfn>.
 Unless stated otherwise, it is unset.
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-reload-navigation-flag>reload-navigation flag</dfn>.
+Unless stated otherwise, it is unset.
```
annevk: Let's add a note here that this is for the exclusive use of HTML's navigate algorithm.

fetch.bs
```diff
@@ -5153,6 +5162,8 @@ constructor must run these steps:
    "<code>navigate</code>", then set it to "<code>same-origin</code>".
    <!-- This works because we have reset request's client too. -->
 
+   <li><p>Unset <var>request</var>'s <a for=request>reload-navigation flag</a>.
```
annevk: Nice! Is this tested?

annevk:
Is this flag stored in the Cache API?

wanderview:
> Is this flag stored in the Cache API?

It should be, IMO.

annevk:
Okay, then you can observe the "Nice!" bit I spotted above. By storing the request on the service worker side in the Cache API, then passing it to `fetch()` on the client side, and then checking on the service worker side whether the flag got reset or not based on the surrounding conditions. (I wonder if we sufficiently considered the implications of what I describe here for other `Request` bits by the way.)

wanderview:
> Okay, then you can observe the "Nice!" bit I spotted above. By storing the request on the service worker side in the Cache API, then passing it to fetch() on the client side, and then checking on the service worker side whether the flag got reset or not based on the surrounding conditions.

In theory we could use this still test to that validate fetch() correctly handles all the different conditions in step 14 of Request() algorithm.

fetch.bs
yutakahirano: Added a test in https://github.com/w3c/web-platform-tests/pull/10192 (fetch/api/request/request-reset-attributes.https.html)

fetch.bs

yutakahirano:
Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17186383/
WebKit: https://bugs.webkit.org/show_bug.cgi?id=184957
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1456709
Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=836697

wanderview:
For Firefox please see this bug instead:

https://bugzilla.mozilla.org/show_bug.cgi?id=1456479

Sorry I filed it before, but forgot to link it here.

PR #696, opened 2018-04-12
Align redirect fragment handling with HTTP

### LINKED ISSUE #505: Browsers preserve fragments on redirects
Via https://github.com/tmpvar/jsdom/pull/1762, I learned something new about browser behavior. I was wondering if this was something fetch needed to cover better? Ctrl+Fing "fragment" doesn't turn up much.

Maybe it is already handled by HTTP. But I wanted to file an issue to check.

annevk:
This clarifies how fragment conflicts between the request's current URL
and the Location header are resolved (the latter wins if non-null).
This is the same for documents and subresources.

Tests: web-platform-tests/wpt#27575.

HTML change: whatwg/html#6391.

Follow-ups: #1167 and #1175.

Fixes #505.

   * Firefox
   * Chrome
   * https://github.com/web-platform-tests/wpt/pull/27575
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1178467
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1690286
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=221897

***
<a href="https://whatpr.org/fetch/696.html" title="Last updated on Feb 19, 2021, 10:07 AM UTC (99e0110)">Preview</a> | <a href="https://whatpr.org/fetch/696/c64f074...99e0110.html" title="Last updated on Feb 19, 2021, 10:07 AM UTC (99e0110)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -218,13 +218,6 @@ number.
 <p class="note no-backref"><a>HTTP(S) scheme</a> and <a>fetch scheme</a> are also used by
 <cite>HTML</cite>. [[HTML]]
 
-<hr>
-
-<p id=fetch-url>A <dfn>response URL</dfn> is a <a for=/>URL</a> for which implementations need not
-store the <a for=url>fragment</a> as it is never exposed. When
-<a lt="URL serializer">serialized</a>, <a for="URL serializer"><i>exclude fragment</i></a> is set to
-true, meaning implementations can store the <a for=url>fragment</a> nonetheless.
-
 
 <h3 id=http>HTTP</h3>
 
@@ -1849,15 +1842,15 @@ end-user.
 
 <p>A <a for=/>response</a> has an associated
 <dfn export for=response id=concept-response-url>URL</dfn>. It is a pointer to the last
-<a>response URL</a> in <a for=/>response</a>'s <a for=response>URL list</a> and null if
-<a for=/>response</a>'s <a for=response>URL list</a> is the empty list.
+<a for=/>URL</a> in <a for=/>response</a>'s <a for=response>URL list</a> and null if
+<a for=/>response</a>'s <a for=response>URL list</a> <a for=list>is empty</a>.
 
 <p>A <a for=/>response</a> has an associated
 <dfn export for=response id=concept-response-url-list>URL list</dfn> (a <a for=/>list</a> of zero or
-more <a>response URLs</a>). Unless stated otherwise, it is the empty list.
+more <a for=/>URLs</a>). Unless stated otherwise, it is the empty list.
 
-<p class="note no-backref">Except for the last <a>response URL</a>, if any, a
-<a for=/>response</a>'s <a for=response>URL list</a> cannot be exposed to script. That would violate
+<p class="note no-backref">Except for the last <a for=/>URL</a>, if any, a <a for=/>response</a>'s
+<a for=response>URL list</a> is not exposed to script as that would violate
 <a>atomic HTTP redirect handling</a>.
 
 <p>A <a for=/>response</a> has an associated
@@ -2037,8 +2030,9 @@ not a <a>fresh response</a> or a <a>stale-while-revalidate response</a>.
 
 <hr>
 
-<p>The <dfn export for=response id=concept-response-location-url>location URL</dfn> algorithm of
-given a <a for=/>response</a> <var>response</var> is the following steps. They return null, failure,
+<p>The <dfn export for=response id=concept-response-location-url>location URL</dfn> of a
+<a for=/>response</a> <var>response</var>, given null or an <a for=/>ASCII string</a>
+<var>requestFragment</var>, is the value returned by the following steps. They return null, failure,
 or a <a for=/>URL</a>.
 
 <ol>
@@ -2057,6 +2051,13 @@ or a <a for=/>URL</a>.
   <p class=note>If <var>response</var> was constructed through the {{Response}} constructor,
   <var>response</var>'s <a for=response>URL</a> will be null, meaning that <var>location</var> will
   only parse successfully if it is an <a>absolute-URL-with-fragment string</a>.
+
+ <li>
+  <p>If <var>locationURL</var>'s <a for=url>fragment</a> is null, then set <var>locationURL</var>'s
+  <a for=url>fragment</a> to <var>requestFragment</var>.
+
+  <p class=note>This ensures that synthetic (indeed, all) responses follow the processing model for
+  redirects defined by HTTP. [[HTTP-SEMANTICS]]
 
  <li><p>Return <var>location</var>.
 </ol>
@@ -3965,7 +3966,8 @@ run these steps:
  <a>filtered response</a>, and <var>response</var>'s
  <a for="filtered response">internal response</a> otherwise.
 
- <li><p>Let <var>locationURL</var> be <var>actualResponse</var>'s <a for=response>location URL</a>.
+ <li><p>Let <var>locationURL</var> be <var>actualResponse</var>'s <a for=response>location URL</a>
+ given <var>request</var>'s <a for=request>current URL</a>'s <a for=url>fragment</a>.
 
  <li><p>If <var>locationURL</var> is null, then return <var>response</var>.
 


annevk:
It seems fragment handling was removed from HTML in https://github.com/whatwg/html/commit/8b630f5e4fa2ec8b0999470d09490bffe6e9a1e3 (by me!) unintentionally. However, that's all good given that HTTP handles it now.

wanderview:
FWIW, the bit I implemented for propagating the request fragment to the final response in firefox is here:

https://searchfox.org/mozilla-central/rev/f2ac80ab7dbde5400a3400d463e07331194dec94/dom/serviceworkers/ServiceWorkerEvents.cpp#720-731

annevk:
Tests: 

* https://github.com/web-platform-tests/wpt/pull/7261
* https://github.com/web-platform-tests/wpt/pull/12674

Still need to write tests for the more involved SW scenarios.

annevk:
Some SW tests: https://github.com/web-platform-tests/wpt/pull/12680.

We should also test navigation though, as there the request URL always wins in terms of what is displayed, so we should probably not propagate a service worker response URL fragment. Right?

annevk:
I updated this PR so Fetch handles fragment propagation for all redirects. This also handles the HTML navigation case.

I'm wondering if we want to do what you proposed above, whereby a request can forward its fragment to a response without a redirect:

* This will always happen for synthetic responses as they do not have a URL list of their own so they will get the URL list of the request.
* It currently does not happen when a service worker returns something like `fetch("test")` to a request for `hi#test`. I.e., the resulting URL is not defined to be `test#test`, it's defined to be `test`.

@wanderview proposes above that for the second case we apply the same kind of forwarding. If response's URL's fragment is null, then set it to request's current URL's fragment.

If we do that it should probably happen inside Fetch after "handle fetch" so it's clear it doesn't affect any `Response` objects in a service worker. However, should we be doing that? It's not really a redirect.

annevk:
I also raised this at https://twitter.com/annevk/status/1034409399702487040.

fetch.bs
```diff
@@ -3402,6 +3395,13 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
       </ul>
 
       <p>then return a <a>network error</a>.
+
+     <li>
+      <p>If <var>actualResponse</var>'s <a for=response>URL</a>'s <a for=url>fragment</a> is null,
+      then set <var>actualResponse</var>'s <a for=response>URL</a>'s <a for=url>fragment</a> to
+      <var>request</var>'s <a for=request>current URL</a>'s <a for=url>fragment</a>.
+
+      <p class=note>Without this step, the fragment would end up getting lost due to the Cache API.
```
annevk: In particular, a scenario @wanderview pointed out that I'm worried about is that `demo.svg` gets stored in the Cache API. A request for `demo.svg#viewBox...` comes in, `demo.svg` gets returned and the fragment identifier is lost.

This does suggest you have to be quite careful with how you store things in the cache if we keep the fragment around, or perhaps the Cache API should always drop the fragment beforehand?

@jakearchibald @jungkees @wanderview it'd be good to resolve that question and create a PR for the outcome against service workers.

wanderview: How does http cache handle fragments?  I assume it ignores fragments in terms of its matching?  I think it would be reasonable to match that behavior (assuming that is what it does...).

annevk: I'm pretty sure that's what it does, yes. @mnot can confirm.

jungkees: Cache API stores a response without stripping its fragment. It ignores fragments in terms of matching. So, if a newly fetched response has a different fragment than the one in Cache, that new response replace the old record.

Shouldn't we serve the fragment as-is in the response's url even if its served from the cache? Response's url should always win even if fetch didn't send out the request to the network. Using a request's url's fragment for cached responses seems odd to me.

HTTP Cache doesn't store fragments?

jungkees:
> It currently does not happen when a service worker returns something like fetch("test") to a request for hi#test. I.e., the resulting URL is not defined to be test#test, it's defined to be test.

I think that looks okay. I don't think we should apply the request's fragment forwarding in this case. The request is intercepted and programmed to return a response in many different ways. Devs might even want to intentionally set a different fragment in the fetch event handler.

annevk:
@jungkees so the PR does do forwarding now, per the latest commit, to address the Cache API use case. I guess an alternative solution here would be that only the Cache API does the fragment forwarding. Hmm.

wanderview:
Personally I'd prefer to strip response url fragments in Cache API by default.  If there is a use case for preserving the fragment there we could add a `preserveFragment:true` option or something.

Also, I don't think we can get rid of propagating the request url fragment if the response url fragment is missing, regardless.  Consider the case where an install event handler adds a resource to the cache with no fragment, but every FetchEvent request does have a fragment.  The fragment should not be lost on those FetchEvents if you pull the response out of the cache.

jungkees:
Correct me if I'm not understanding it correctly. The basic rule is the request url fragment propagates. But fragments given in Location urls will replace if it redirects. If response url's fragment is not null, that's the final value. Else, the request url's fragment propagates.

I think a response returned from SW to fetch is just a response. So, following the rule, if it has a fragment, that should be the last value. Else, the response url should get the request url's fragment.

E.g.
```js
e.request.url: 'foo#bar'
onfetch = e => { e.respondWidh(fetch('foo#baz')) };
```
The client should get response.url === 'foo#baz' (am I right?)

For the Cache behavior, whether we should strip response url's fragment before storing it by default, I think we should be consistent with what HTTP Cache does.

annevk:
@jungkees your prose and examples in that last comment match the proposed semantics of this PR. Hope that helps.

https://tools.ietf.org/html/rfc7234 does not talk about fragments at all. As I understand HTTP, the request going to the server does not contain a fragment and a response coming from the server will never have a fragment. Therefore the HTTP cache does not concern itself with fragments. It probably makes sense for the Cache API to follow those semantics.

jungkees:
@annevk, thanks for checking in on those points. I'm fine with the behavior this PR proposes. Ignore my comment in https://github.com/whatwg/fetch/pull/696#issuecomment-416775370. I'll change the Cache API to strip the fragment before storing responses.

annevk:
I would like to pick this up again. I have rebased this PR, updated OP, and updated the tests. I left some questions in #505 that will need to be addressed as it's still not completely clear what to do about service workers and how to best integrate it with redirects (in particular `Location` processing currently doesn't have access to a request) in a way that works well for Fetch and HTML.

annevk:
Tentative commit message:
```
Align fragment handling with HTTP

This clarifies how fragment conflicts between the request's current URL 
and the Location header are resolved (the latter wins if non-null). And
how fragment conflicts between the request's current URL and service 
worker response URLs are handled (the former wins). Both are 
irrespective of whether it's a request for a document or subresource.

Tests: https://github.com/web-platform-tests/wpt/pull/27575, 
https://github.com/web-platform-tests/wpt/pull/9604, and 
https://github.com/web-platform-tests/wpt/pull/12680.

HTML change: https://github.com/whatwg/html/pull/6391.

Follow-up: #1167.

Fixes #505.
```

annevk:
@davidben if you could review as well that'd be great.

fetch.bs
```diff
@@ -2035,9 +2028,9 @@ not a <a>fresh response</a> or a <a>stale-while-revalidate response</a>.
 
 <hr>
 
-<p>The <dfn export for=response id=concept-response-location-url>location URL</dfn> algorithm of
-given a <a for=/>response</a> <var>response</var> is the following steps. They return null, failure,
-or a <a for=/>URL</a>.
+<p>The <dfn export for=response id=concept-response-location-url>location URL</dfn> algorithm, given
+a <a for=/>response</a> <var>response</var> and null or an <a for=/>ASCII string</a>
+<var>requestFragment</var> is the following steps. They return null, failure, or a <a for=/>URL</a>.
```
domenic, line 2033: I'd suggest:

> The **location URL** of a response _response_, given null or an ASCII string _requestFragment_, is the value returned by the following steps. It will be either null, failure, or a URL.

domenic (approved):

fetch.bs
annevk: "They return" is language from Infra. Agreed that what's before the dot is awkward dot and needs improving.

fetch.bs
```diff
@@ -3835,6 +3836,13 @@ these steps:
       </ul>
 
       <p>then return a <a>network error</a>.
+
+     <li>
+      <p>Set <var>actualResponse</var>'s <a for=response>URL</a>'s <a for=url>fragment</a> to
+      <var>request</var>'s <a for=request>current URL</a>'s <a for=url>fragment</a>.
+
+      <p class=note>There is currently no mechanism for the service worker to control
+      <var>actualResponse</var>'s <a for=response>URL</a>'s <a for=url>fragment</a>.
```
davidben, line 3845: If I'm understanding this right, it means:

* If a SW-less document contains ``, we'll render `colors.svg#green`, since the location URL algorithm prefers the `Location` header's fragment.
* If that same document has a service worker which simply does `ev.respondWith(fetch(ev.request))`, the FetchEvent will be resolved with a Response containing `colors.svg#green` (since redirects were followed), but then this logic will replace the fragment with `#red`.

Am I understanding this right? Is that intentional?

That seems kinda weird, though I don't see clear alternatives given service worker caches should not be sensitive to fragments, and service workers take over URLs. One possibility, which is janky in a different way (but has a nice symmetry with redirects in general), would be to say:

* When a SW gets a FetchEvent, remove the fragment from the request URL. The SW, like an HTTP intermediary, should not get to react differently on request fragments.
* The response URL preserves the fragment. But, due to step 1, any fragments would come from redirects or the service worker itself. We haven't yet incorporated the request fragment.
* When we consume a SW-forwarded response, apply the request fragment similar to redirects: if the response URL had a fragment, use that. If not, use the request fragment.

(By the way, are there subresource contexts other than SVGs where the fragments matter?)

davidben: And I guess both formulations have another weird corner: the response's URL isn't a field per se. It's a pointer to the last entry in the URL list. In which case, what does it mean to update it? Do you write into the last entry of the URL list but leave the others alone? What if the URL list is empty because this was script-constructed?

https://fetch.spec.whatwg.org/#concept-response-url

wanderview: I agree It would be nice if respondWith() functioned similar to redirects when it came to fragments.

annevk: Thanks @davidben, you're right, this isn't great. I think your model is better but it would change `fetchEvent.request.url`. Is Chrome willing to experiment with that? I suppose synthetic responses would always end up with the fragment of the request from the document, which seems about right, but the current wording indeed isn't adequate.

Perhaps I should split out the service worker change into its own issue and pull request (it already has its own tests) as that clearly warrants some further consideration.

davidben: (@wanderview is probably a better person to answer what we're willing to experiment with. I don't actually work on service worker stuff. I'm just really distractible and like to stick my nose into everything. Although I see I'm rapidly becoming "the redirect person" which is... a life decision. 😄)

I was pondering other ideas, and they broke on this case, so probably something to include in tests. Suppose the server responds to requests as follows:

GET /a => 302; Location: /b#foo
GET /b => 302; Location: /c
GET /c => 200

We'd want `fetch("/a")` to get a redirect path of /a -> /b#foo -> /c#foo. That implies that the fragment fixup logic needs to be part of `{redirect: "follow"}` and not something we only apply to the final response. However, `fetch("/a#foo")` would then go /a#foo -> /b#foo -> /c#foo and there isn't enough information to know whether c's fragment was intrinsic or came from the request fragment. That means any caching fetch intermediary must store the result of a fragment-less fetch. Otherwise you lose information and cannot correctly share cache entries between /a#foo and /a#bar.

Also, all this relies on it being impossible for the server to redirect and *remove* a fragment. (If the server sends a fragment-less location header, we fill in the request URL's fragment.) If it were possible, we'd need to store extra information in the response. That is, if the server says:

GET /a => 302; Location: /b#:magic-syntax-to-remove-fragment

Then we have /a -> /b, but also /a#foo -> /b, but that conflicts with the "When we consume a SW-forwarded response, [...]" logic. We'd need responses to store an extra copyFragmentFromRequestURL boolean.

To that end, that does suggest one more possible model which doesn't need to change FetchEvent:

* Responses have a "copy fragment from request URL" boolean property
* SW still sees the fragment and are allowed to fetch as before
* Fetch sets the new property to true if every Location header we followed was fragment-agnostic
* When you get a response from a SW, clobber the fragment if that property is true.

But this one seems even more fussy and probably isn't what we want. Yet another possibility would be to move the fragment fixup logic from FetchEvent.respondWith() to CacheStorage. Something like:

* Responses track whether fragments came from the request URL or not as above.
* When you serialize a response, don't serialize fragments that came from the request URL, only intrinsic ones.
* When you get a response out from CacheStorage.match(), apply the fragment fixup logic.

annevk:
I've reduced the scope of this PR (for the second time) to be about redirects only and filed a second follow-up issue, this time for service workers. Who knew fragments would be this much fun.

annevk:
@davidben @wanderview any concerns with merging this now that the service worker bits have been moved to #1175?

davidben (approved):

triple-underscore:
In the [location URL](https://fetch.spec.whatwg.org/#concept-response-location-url) algorithm:

> If locationURL’s fragment is null, then set locationURL’s fragment to requestFragment. 

The variable `locationURL` is not declared; it may be `location`.
Assuming `location`, it can be failure. Thus, the if clause needs to include for checking it.

annevk:
Thanks, I created https://github.com/whatwg/fetch/pull/1180 to address that.

PR #699, opened 2018-04-13
Return network error for scheme-fetch for ftp/file

mozfreddyb:
Currently, it's defined as "exercise for the reader".
I suggest changing the wording to default to network error.

Is this acceptable?
If yes, is this patch OK or do we want to add a note?

CCing @evilpie

***
<a href="https://whatpr.org/fetch/699.html" title="Last updated on Apr 13, 2018, 10:28 AM GMT (5ea9140)">Preview</a> | <a href="https://whatpr.org/fetch/699/5b7dae0...5ea9140.html" title="Last updated on Apr 13, 2018, 10:28 AM GMT (5ea9140)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2998,10 +2998,8 @@ steps:
  <dt>"<code>file</code>"
  <dt>"<code>ftp</code>"
  <dd>
-  <p>For now, unfortunate as it is, <code>file</code> and <code>ftp</code> <a for=/>URLs</a> are
-  left as an exercise for the reader.
-
-  <p>When in doubt, return a <a>network error</a>.
+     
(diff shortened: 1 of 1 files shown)


annevk:
I know I said Monday, but... This would break navigating to `ftp:` URLs. (Also, if we did this we could remove the other `ftp:` URL check and remove these two schemes here altogether as the default is a network error for unlisted schemes.)

annevk:
Closing per above. FTP restrictions are handled by step 2.8 of https://fetch.spec.whatwg.org/#concept-main-fetch and we cannot restrict more I think until we remove it completely.

PR #325, opened 2016-06-20
Update Fetch to support Token Binding.

vanupam:
***
<a href="https://whatpr.org/fetch/325.html" title="Last updated on Apr 14, 2018, 3:43 AM GMT (a0c5c1f)">Preview</a> | <a href="https://whatpr.org/fetch/325/5b7dae0...a0c5c1f.html" title="Last updated on Apr 14, 2018, 3:43 AM GMT (a0c5c1f)">Diff</a>

PR #715, opened 2018-04-19
Update Fetch to support Token Binding

### LINKED ISSUE #30: Add support for federated token binding to fetch
I'm proposing to add support for (federated) [Token Binding](https://github.com/TokenBinding/) to fetch.

I've written a requirements/problem-description document for a Token Binding DOM API: https://docs.google.com/document/d/1x_3KI_v35NQFEsAbGGjNcyTw3Waoy1mL3YfTaQbyyuM/view?rm=minimal&pli=1

While I have included a strawman proposal in the doc, I'd appreciate help from the fetch editors/implementers to figure out what the right approach is.

vanupam:
This PR obsoletes a prior one:

https://github.com/whatwg/fetch/pull/325

[This fixes #30.]

***
<a href="https://whatpr.org/fetch/715.html" title="Last updated on Jul 25, 2018, 11:44 PM GMT (a823d27)">Preview</a> | <a href="https://whatpr.org/fetch/715/7ab665e...a823d27.html" title="Last updated on Jul 25, 2018, 11:44 PM GMT (a823d27)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1069,8 +1069,7 @@ e.g., bind HTTP cookies or OAuth tokens that it issues to the user agent.
 
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-referred-token-binding-origin>referred-token-binding origin</dfn>,
-which is an <a for=/>origin</a>.
-Unless stated otherwise, it is null.
+which is an <a for=/>origin</a> or null. Unless stated otherwise, it is null.
 
 <p class="note no-backref">When <a for=request>Request</a>'s
 <a for=request>referred-token-binding origin</a> is not null,
@@ -1298,7 +1297,6 @@ or "<code>worker</code>".
     </ol>
 
    <li><p>If the ongoing fetch is <a for=fetch>terminated</a>, then abort these in-parallel steps.
-
   </ol>
 </ol>
 
@@ -1565,20 +1563,30 @@ for each associated <a for="fetch group">fetch record</a> whose
 
 <h3 id=connections>Connections</h3>
 
+<p>A <dfn>connection</dfn> is a communication channel between the user agent and an origin.
+
+<p>A <a>connection</a> has an associated <dfn for=connection>origin</dfn>, which is an <a for=/>origin</a>.
+
+<p>A <a>connection</a> has an associated <dfn for=connection>credential</dfn>, which is a boolean. Its value is `false` unless otherwise specified.
+
+<p>A <a>connection</a> has an associated <a for=connection>token-binding protocol version</a>,
+which is a byte sequence.
+Its value is null unless otherwise specified.
+
+<p>A <a>connection</a> has an associated <a for=connection>token-binding key parameters</a>, which is a byte.
+Its value is null unless otherwise specified.
+
+<p>A <a>connection</a> has an associated <a for=connection>token-binding Exported Keying Material</a>,
+which is a byte sequence.
+Its value is null unless otherwise specified.
+
+
 <p>A user agent has an associated
-<dfn export id=concept-connection-pool for=connection>connection pool</dfn>. A
-<a for=connection>connection pool</a> consists of zero or more
+<dfn export id=concept-connection-pool for=/>connection pool</dfn>. A
+<a for=/>connection pool</a> consists of zero or more
 <dfn lt=connection export id=concept-connection>connections</dfn>. Each
 <a>connection</a> is identified by an <b>origin</b> (an
 <a for=/>origin</a>) and <b>credentials</b> (a boolean).
-
-<p>A <a>connection</a> has associated
-<dfn export for=connection id=concept-tls-connection-metadata>TLS connection metadata</dfn> that consists of:
-<ul class=brief>
- <li><a for=/>token-binding protocol version</a>, initially null.
- <li><a for=/>token-binding key parameters</a>, initially null.
- <li><a for=/>token-binding Exported Keying Material</a>, initially null.
-</ul>
 
 <h4 id=obtain-connection>Obtaining a Connection</h4>
 
@@ -1586,7 +1594,7 @@ for each associated <a for="fetch group">fetch record</a> whose
 <var>origin</var> and <var>credentials</var>, run these steps:
 
 <ol>
- <li><p>If <a for=connection>connection pool</a> contains a <a>connection</a> whose <b>origin</b> is
+ <li><p>If <a for=/>connection pool</a> contains a <a>connection</a> whose <b>origin</b> is
  <var>origin</var> and <b>credentials</b> is <var>credentials</var>, then return that
  <a>connection</a>.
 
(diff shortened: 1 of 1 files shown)


mikewest:
Thanks for reuploading this, @vanupam! I'll take another pass this week!

fetch.bs
```diff
@@ -2353,7 +2435,374 @@ X-Content-Type-Options           = "nosniff" ; case-insensitive</pre>
...
+and <var>tlsConnection</var>'s
+<a for=connection>token-binding Exported Keying Material</a>.
+
+ <li><p>Append <var>providedTokenBinding</var> to
+ <var>tokenBindingMessage</var>'s <code>TokenBindingMessage.tokenbindings</code> field.
+
+ <li><p>If <var>httpRequest</var>'s <a for=request>referred-token-binding origin</a> is not null,
+ then run these substeps:
+  <ol>
+   <li><p>Let <var>referredTlsConnection</var> be the result of
```
nharper: Continuing a comment thread from the previous PR:

I still don't like that referredTlsConnection is created, when all it's needed for is the key type. For the case of a referred Token Binding via an HTTP header, I think the most sensible thing to do is use the same key type/key as was used for the request that received the HTTP response header. This is behavior that doesn't require opening a new connection.

Supporting the fetch() API's mechanism for referred Token Bindings might have some rule that's this simple, like "use the last used key/key type for that origin", but I don't know if that's the rule we'd actually want to write down for this spec. I think we can write a much simpler "Computing a Token Binding Header Value" that doesn't rely on creating this referredTlsConnection if we limit this PR to exclude changes to the fetch() API.

(On the previous PR, the rationale for why the referredTlsConnection is needed was that the server might change the key type (key parameters) it wants to use. Servers changing the Token Binding key parameters they support is fraught with peril and most likely broken if the server wants to have any chance of keeping tokens bound to some Token Binding key.)

fetch.bs
```diff
@@ -2353,7 +2435,374 @@ X-Content-Type-Options           = "nosniff" ; case-insensitive</pre>
...
+<h4 id=negotiating-token-binding>Negotiating Token Binding</h4>
+
+The user agent proposes and agrees to the use of Token Binding as described in
+<a href="https://tools.ietf.org/html/draft-ietf-tokbind-negotiation#section-2">section 2</a>
+of the Token Binding Negotiation spec [[!TOKBIND-NEGOTIATION]].
+While setting up a TLS <a>connection</a>, it sends its highest supported
+<dfn export id=concept-token-binding-protocol-version for=connection>token-binding protocol version</dfn>
+and all supported cryptographic algorithms and parameters (the
+<dfn export id=concept-token-binding-key-parameters for=connection>token-binding key parameters</dfn>),
+in order of preference,
```
nharper: > in order of preference,

If we want to make it possible for a server to try to migrate users from one key type to another, this probably needs to be explained in a lot more detail what "order of preference" means for a client. It probably depends on what keys the client already has in its key store for the domain it's trying to connect to. (There are more complications than just that here.)

fetch.bs
```diff
@@ -2353,7 +2435,374 @@ X-Content-Type-Options           = "nosniff" ; case-insensitive</pre>
...
+(e.g., an identity provider) are presented by the user agent to a different server
+(e.g., a relying party).
+[[TOKBIND-PROTOCOL]] and [[TOKBIND-HTTPS]] describe support for Federated Token Binding.
+
+<p>A <dfn export id=referred-token-binding-id>referred-token-binding ID</dfn> is the
+<a for=/>token binding ID</a> for a server other than the server to which the
+<a for=/>request</a> is being sent.
+
+Script code from an <a for=/>origin</a> can set a <a for=/>request</a>'s
+<var>useReferredTokenBinding</var> attribute to ask the user agent to send
```
nharper: Either "request" or "useReferredTokenBinding" is wrong here - a request doesn't have a useReferredTokenBinding attribute, but it does have a "referred-token-binding origin". A Request class/object does have a useReferredTokenBinding.

fetch.bs
```diff
@@ -5227,6 +5731,10 @@ constructor must run these steps:
  <a for=request>credentials mode</a> to
  <var>credentials</var>.
 
+ <li><p>If <var>useReferredTokenBinding</var> is true, set <var>request</var>'s
+ <a for=request>referred-token-binding origin</a> to
+ <var>origin</var>.
```
nharper: This makes it possible for a request's referred-token-binding origin to be set even if no Token Binding key exists for that origin. This seems a bit weird to me.

fetch.bs
vanupam: I removed fetch() API related changes.

For referred Token Binding using redirects, the referredTlsConnection will already exist - it will just be looked up.

fetch.bs
vanupam: The intent here is to allow scripts to easily indicate that they want to use referred token binding, without letting them set the referred-token-binding origin. This origin is relevant only when TB is being used and there is a TB key.

(Moot now: API changes have been removed.)

nharper:
The only thing left on my list of comments is what to do with referredTlsConnection. I don't like the aesthetics of needing to go through the step of "obtaining a connection" (even if there most likely is already connection is in the connection pool, and a new connection isn't made), especially since it's the only other place (besides HTTP-network fetch) that calls that step. However, I'll defer judgement of this to those who know the Fetch spec better than I.

On the previous PR, there was a conversation about whether Token Binding should be mentioned in section 2 (Infrastructure) with the definition of Credentials, which I don't think reached a conclusion. I am of the opinion that a Sec-Token-Binding header should be treated as a credential (both in the CORS withCredentials context and in a third-party cookie blocking context): A site could choose to set up and operate its infrastructure so that a Token Binding ID is the only information stored on the client for the client's identity (and the Token Binding ID serves as a key into a server-side session state table).

equalsJeffH:
I merged-from-master and fixed conflicts.  I think I did it correctly, but someone should probably review, especially in S 4.5. HTTP-network-or-cache fetch, substep 5.17.2.5, thx.

equalsJeffH:
@nharper 
> I am of the opinion that a Sec-Token-Binding header should be treated as a credential...

submit that as an issue once this PR lands?  or do you consider it a blocker for landing this PR?

fetch.bs
```diff
@@ -4058,6 +4511,22 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li><p>If <var>connection</var> is failure, return a
    <a>network error</a>.
 
+   <li><p>If <var>request</var>'s <a for=request>use-token-binding flag</a>
+   is set, build and add a `<a http-header><code>Sec-Token-Binding</code></a>`
```
equalsJeffH: nit: this spec seems to use the term "construct" rather than "build"

fetch.bs
```diff
@@ -4058,6 +4511,22 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+   request <a for=/>header</a> by running the following substeps:
+
+    <ol>
+     <li><p>Let <var>tokenBindingMessage</var> be the result of
+     <a abstract-op lt="Compute a Token Binding header value">computing a Token Binding header value</a>
+     for <var>request</var> and <var>connection</var>,
+     using the user agent's <a for=/>token-binding key store</a> and
```
equalsJeffH: I do not think we need the phrase beginning with "using" because that info is baked into the definition of the "Compute a Token Binding header value" alg

fetch.bs
```diff
@@ -4058,6 +4511,22 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+     <a>connection pool</a>.
+
+     <li><p>If <var>tokenBindingMessage</var> is not null, append
+     `<a http-header><code>Sec-Token-Binding</code></a>`/<var>tokenBindingMessage</var>
```
equalsJeffH: nit: put a space in `/<var>`  eg: `/ <var>`

zandbelt:
For implementer's interest see the thread on the intent for Chrome to remove token binding support...:
https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/OkdLUyYmY1E

annevk:
I'm going to close this given that there's not sufficient implementer interest, no activity, and a significant number of outstanding issues.

PR #718, opened 2018-04-27
Add Request's isHistoryNavigation

yutakahirano:
And also a member on the request concept (history-navigation flag) to support this API. See w3c/ServiceWorker#1167 for the discussion that led to this.

Tests: TBD

Corresponding HTML change: TBD

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1165,6 +1165,12 @@ Unless stated otherwise, it is unset.
 
 <p class="note no-backref">This flag is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-history-navigation-flag>history-navigation flag</dfn>.
+Unless stated otherwise, it is unset.
+
+<p class="note no-backref">This flag is for exclusive use by HTML's navigate algorithm. [[!HTML]]
+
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1165,6 +1165,12 @@ Unless stated otherwise, it is unset.
 
 <p class="note no-backref">This flag is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-history-navigation-flag>history-navigation flag</dfn>.
+Unless stated otherwise, it is unset.
+
+<p class="note no-backref">This flag is for exclusive use by HTML's navigate algorithm. [[!HTML]]
```
annevk: We should move this and the flag before it, up a bit. See #732. Sorry for not noticing that earlier. I can probably take care of this as a fixup commit once everything is close to landing, but I'm also happy for you to take care of it.

yutakahirano:
Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17677800/
WebKit: https://bugs.webkit.org/show_bug.cgi?id=186028
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1464739
Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=842551

fetch.bs

Issue #652, opened 2018-01-04
what is the Request.destination for modules and es `import()`?


wanderview:
I think the `Request.destination` table here needs to be updated for modules:

https://fetch.spec.whatwg.org/#concept-request-destination

I assume they have a destination of "script".

wanderview:
@domenic does "script" destination sound reasonable?  cc @yoavweiss

domenic:
Yes, this is already specified in HTML. It uses script/audioworklet/paintworklet/serviceworker/sharedworker/worker as appropriate.

I'm not sure the table needs any updates since it's just an alternate entry point into all the same APIs?

wanderview:
Ah, I'm not used to the other specs being updated!  I guess that's fine, although the table is nice for checking test coverage, etc.

wanderview:
I guess I worry about people looking at the table and thinking it covers all cases.  Could we either:

1. Make a note that its not exhaustive.
2. Or make a note outside the table itself that summarizes how modules work and points at the html spec.

I assume things like `import()` would be added on the same line as `importScripts()` in the table, etc.

PR #724, opened 2018-05-14
Associate import() with the "script" destination

nhiroki:
Before this patch, the Fetch spec says nothing about the destination of
dynamic/static import(). This patch clarifies it.

importScripts() that is a similar feature of import() is associated with the
"script" destinaton, so import() should also be associated with the "script"
destination.

cc: @nyaxt

***
<a href="https://whatpr.org/fetch/724.html" title="Last updated on May 14, 2018, 4:39 AM GMT (ac10f0a)">Preview</a> | <a href="https://whatpr.org/fetch/724/3a896ef...ac10f0a.html" title="Last updated on May 14, 2018, 4:39 AM GMT (ac10f0a)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -873,7 +873,7 @@ not always relevant and might require different behavior.
   <tr>
    <td>"<code>script</code>"
    <td><code>script-src</code>
-   <td>HTML's <code>&lt;script></code>, <code>importScripts()</code>
+   <td>HTML's <code>&lt;script></code>, <code>importScripts()</code>, <code>import()</code>
   <tr>
    <td>"<code>serviceworker</code>"
    <td><code>child-src</code>, <code>script-src</code>, <code>worker-src</code>


annevk:
That table is non-normative. Fetch is not the source of truth on this. Perhaps I should remove the table since it seems to continue to confuse people.

Having said that, per https://html.spec.whatwg.org/#hostimportmoduledynamically(referencingscriptormodule,-specifier,-promisecapability) it does always use "script" so we could make this change.

nhiroki:
Thank you for the clarification. I missed it's specified in the import() side. I should have carefully read this text above the table :p

> The following table illustrates the relationship between a request’s initiator, destination, CSP directives, and features. It is not exhaustive with respect to features. Features need to have the relevant values defined in their respective standards.

I have no strong opinion on deletion of the table, but I feel it's still useful for knowing the overview.

I'll close this PR.

domenic:
This was previously discussed in https://github.com/whatwg/fetch/issues/652 and the conclusion was to add the spec's current text:

> It is not exhaustive with respect to features. Features need to have the relevant values defined in their respective standards.

PR #733, opened 2018-05-24
Define Cross-Origin-Resource-Policy response header

### LINKED ISSUE #687: Cross-Origin-Resource-Policy (was: From-Origin)
Websites should have an explicit way to restrict any kind of cross-origin load to protect themselves against Spectre attacks. Content such as images, video, and audio may be sensitive and websites may be protected solely by virtue of their network position (inside a firewall), relying on the same-origin policy to protect against exfiltration.

There's a previous proposal from 2012 called the [From-Origin header](https://www.w3.org/TR/from-origin/) that we'd like to resurrect. With it, a server can send a `From-Origin : same` header on responses it wants to protect from Spectre attacks. Here's a snippet from the currently inactive proposal:

> The From-Origin header can be used to restrict embedding of a resource to only certain origins. When used it must match the following ABNF:
> 
> `From-Origin = "From-Origin" ":" #(serialized-origin | "same")`

[Cross-Origin Read Blocking (CORB)](https://github.com/whatwg/fetch/issues/681) automatically protects against Spectre attacks that load cross-origin, cross-type HTML, XML, and JSON resources, and is based on the browser’s ability to distinguish resource types. We think CORB is a good idea. From-Origin would offer servers an opt-in protection beyond CORB.

In addition to the original proposal, we might want to offer servers a way to say cross-origin requests are OK within the same eTLD+1, e.g. the server may want to say that cross-origin subresources from cdn.example.com may be loaded from *.example.com without listing all those origins.

annevk:
This header makes it easier for sites to block unwanted "no-cors" cross-origin requests.

Tests: ...

Fixes #687.

***
<a href="https://whatpr.org/fetch/733.html" title="Last updated on Jun 18, 2018, 12:20 PM GMT (ed96a17)">Preview</a> | <a href="https://whatpr.org/fetch/733/2f3d04d...ed96a17.html" title="Last updated on Jun 18, 2018, 12:20 PM GMT (ed96a17)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2517,6 +2517,58 @@ run these steps:</p>
 </ol>
 
 
+<h3 id=cross-origin-resource-policy-header>`<code>Cross-Origin-Resource-Policy</code>` header</h3>
+
+<p>The
+`<dfn export http-header id=http-cross-origin-resource-policy><code>Cross-Origin-Resource-Policy</code></dfn>`
+response <a for=/>header</a> can be used to require checking a <a for=/>request</a>'s
+<a for=request>current url</a>'s <a for=url>origin</a> against a <a for=/>request</a>'s
+<a for=request>origin</a> when <a for=/>request</a>'s <a for=request>mode</a> is
+"<code>no-cors</code>".
+
+<p>Its <a for=header>value</a> <a>ABNF</a>:
+
+<pre>
+Cross-Origin-Resource-Policy     = %x73.61.6D.65.2D.6F.72.69.67.69.6E / %x73.61.6D.65.2D.73.69.74.65 ; "same-origin" / "same-site", case-sensitive</pre>
+
+<p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <var>request</var> and
+<var>response</var>, run these steps:</p>
+
+<ol>
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not "<code>no-cors</code>", then return
+ <b>allowed</b>.
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
+  <var>request</var>'s <a for=request>current url</a>'s <a for=url>origin</a>, then return
+  <b>allowed</b>.
+
+  <p class="note no-backref">A cross-origin response redirecting to a response that is
+  <a>same origin</a> or <a>same site</a> with the initial request and has a
+  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header specified, does not affect
+  anything. I.e., <var>request</var>'s <a for=request>tainted origin flag</a> is not checked.
+
+ <li>
+  <p>Let <var>policy</var> be the <a>combined value</a> with
+  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` and <var>response</var>'s
+  <a for=response>header list</a>.
+
+  <p class=note>This means that `<code>Cross-Origin-Resource-Policy: same-site, same-origin</code>`
+  ends up as <b>allowed</b> below as it will never match anything. Two or more
+  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` headers will have the same effect.
+
+ <li><p>If <var>policy</var> is `<code>same-origin</code>`, then return <b>blocked</b>.
+
+ <li><p>If <var>request</var>'s <a for=request>origin</a>'s <a for=url>host</a> is <a>same site</a>
+ with <var>request</var>'s <a for=request>current url</a>'s <a for=url>host</a>, then return
+ <b>allowed</b>.
+
+ <li><p>If <var>policy</var> is `<code>same-site</code>`, then return <b>blocked</b>.
+
+ <li><p>Return <b>allowed</b>.
+</ol>
+
+
 
 <h2 id=fetching>Fetching</h2>
 
@@ -3849,6 +3901,10 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
  <li><p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">contains</a>
  `<code>Range</code>`, then set <var>response</var>'s <a for=response>range-requested flag</a>.
+
+ <li><p>If the <i>CORS flag</i> is unset and the <a>cross-origin resource policy check</a> with
+ <var>request</var> and <var>response</var> returns <b>blocked</b>, then return a
+ <a>network error</a>.
 
  <li>
   <p>If <var>response</var>'s <a for=response>status</a> is <code>401</code>, <i>CORS flag</i>
@@ -6548,6 +6604,7 @@ Jeffrey Yasskin,
 Jesse M. Heines,
 Jinho Bang,
 Jochen Eisinger,
+John Wilander,
 Jonas Sicking,
 Jonathan Kingston,
 Jonathan Watt,
--- a/fetch.bs
+++ b/fetch.bs
@@ -2543,10 +2543,10 @@ Cross-Origin-Resource-Policy     = %x73.61.6D.65.2D.6F.72.69.67.69.6E / %x73.61.
   <var>request</var>'s <a for=request>current url</a>'s <a for=url>origin</a>, then return
   <b>allowed</b>.
 
-  <p class="note no-backref">A cross-origin response redirecting to a response that is
-  <a>same origin</a> or <a>same site</a> with the initial request and has a
-  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header specified, does not affect
-  anything. I.e., <var>request</var>'s <a for=request>tainted origin flag</a> is not checked.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2437,6 +2437,64 @@ run these steps:</p>
...
+`<dfn export http-header id=http-cross-origin-resource-policy><code>Cross-Origin-Resource-Policy</code></dfn>`
+response <a for=/>header</a> can be used to require checking a <a for=/>request</a>'s
+<a for=request>current url</a>'s <a for=url>origin</a> against a <a for=/>request</a>'s
+<a for=request>origin</a> when <a for=/>request</a>'s <a for=request>mode</a> is
+"<code>no-cors</code>".
+
+<p>Its <a for=header>value</a> <a>ABNF</a>:
+
+<pre>
+Cross-Origin-Resource-Policy     = %x73.61.6D.65 / %x73.61.6D.65.2D.73.69.74.65 ; "same" / "same-site"; case-sensitive</pre>
```
mikewest: Nit: The last semicolon should be a comma (similar to the `origin-or-null` definition).

fetch.bs
```diff
@@ -2437,6 +2437,64 @@ run these steps:</p>
...
+<ol>
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not "<code>no-cors</code>", then return
+ <b>allowed</b>.
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
+  <var>request</var>'s <a for=request>current url</a>'s <a for=url>origin</a>, then return
+  <b>allowed</b>.
+
+  <p class=note>A cross-origin response redirecting to a same or same-site resource with the
```
mikewest: Nit: This is a little unclear. Is the redirect target same-site with the request's origin, or with the cross-origin response?

fetch.bs
```diff
@@ -2437,6 +2437,64 @@ run these steps:</p>
...
+  <p>If <var>policy</var> is `<code>same-site</code>` and neither of the following is true
+
+  <ul class=brief>
+   <li><p><var>request</var>'s <a for=request>origin</a>'s <a for=origin>host</a>
+   <a>is a registrable domain suffix of or is equal to</a> <var>request</var>'s
+   <a for=request>current url</a>'s <a for=url>host</a>
+
+   <li><p><var>request</var>'s <a for=request>current url</a>'s <a for=url>host</a>
+   <a>is a registrable domain suffix of or is equal to</a> <var>request</var>'s
+   <a for=request>origin</a>'s <a for=origin>host</a>
```
mikewest: I think this isn't broad enough. The bidirectional `is a registrable domain suffix of or is equal to` comparison would return false for a request whose origin's host is `subdomain.subdomain.example.com` and whose current url is `notsubdomain.example.com`.

Instead, I think we need to calculate the relevant registrable domain (perhaps by pointing to the (not terribly clear) algorithm on https://publicsuffix.org/list/) for both the request's origin and the request's current URL, and compare those.

mikewest: (And, actually, looking at that algorithm again, I'm not sure it does what it needs to do for `document.domain` either... It doesn't calculate the registrable domain at all, but the public suffix. I'll file a separate bug.)

mikewest: https://github.com/whatwg/html/issues/3711

fetch.bs
```diff
@@ -2437,6 +2437,64 @@ run these steps:</p>
...
+  <p class=note>A cross-origin response redirecting to a same or same-site resource with the
+  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header specified does not affect
+  anything.
+  <!-- We could make this have an effect if we fix https://github.com/whatwg/fetch/pull/594 first,
+       but even then we normally do not let this have any effect for "no-cors" so it would be
+       somewhat inconsistent if it did here, but might still be better... -->
+
+ <li><p>Let <var>policy</var> be the <a>combined value</a> with
+ `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` and <var>response</var>'s
+ <a for=response>header list</a>.
```
mikewest: It might be worth adding a note here about the meaning of `same, same` or `same, same-site`, as I believe this means that both will be ignored.

mikewest: I don't think it matters in this patch, but I agree that that's probably what we'd want to do in a future patch (e.g. define this as a [structured header list](https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-04#section-4.2) whose members are strings that we parse as origins or something.

fetch.bs
mikewest: Yeah. Ok, assuming I can dig myself out from under everything else, would you prefer those concepts be defined in HTML, Fetch, or URL?

fetch.bs
annevk: Hmm, wouldn't that be incompatible with `same` and `same-site` as tokens? And origins cannot be used as tokens, at least not unless https://github.com/httpwg/http-extensions/issues/629 is fixed somehow.

mikewest: I don't think there's any requirement that all the items in a list are the same type. @mnot can confirm, but my understanding is that we could have `same` and `same-site` as "labels" and some other type ("string", I suppose) for origins.

fetch.bs
```diff
@@ -3736,6 +3794,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     </ol>
   </ol>
 
+ <li><p>If the <a>cross-origin resource policy check</a> with <var>request</var> and
```
annevk: @youennf if we fixed #741 and skipped this step if the `CORS flag` is set, we'd address the CORS-preflight issue I think.

fetch.bs
annevk: Note: this is now tracked by https://github.com/whatwg/url/pull/391.

annevk:
I pushed a commit that should address the review feedback.

It builds upon https://github.com/whatwg/url/pull/391 (nearly done) and https://github.com/whatwg/fetch/pull/743 (needs review).

@mnot can you confirm that structured headers will allow mixing of identifiers and strings?

annevk:
FYI: I rebased and renamed `same` to `same-origin`.

annevk:
Dependencies have been merged, plan of action is at https://github.com/whatwg/fetch/issues/687#issuecomment-395366076. TL;DR: this will merge next week barring any objections.

fetch.bs
```diff
@@ -2517,6 +2517,71 @@ run these steps:</p>
...
+
+ <li>
+  <p>If the following are true
+
+  <ul class=brief>
+   <li><var>request</var>'s <a for=request>origin</a>'s <a for=url>host</a> is <a>same site</a> with
+   <var>request</var>'s <a for=request>current url</a>'s <a for=url>host</a>
+   <li><var>request</var>'s <a for=request>origin</a>'s <a for=url>scheme</a> is
+   "<code>https</code>" or <var>response</var>'s <a for=response>HTTPS state</a> is
+   "<code>none</code>"
```
annevk: @mikewest is this too cute and should I be checking the scheme of request's current url instead or is this fine/preferred?

mikewest: I feel like I had to think about what this means more than I'd like to. An explicit comparison seems simpler. Or maybe just skipping the check around? Is something like this equivalent?

> If response's HTTPS state is `modern`, request's origin's scheme is `https`.

fetch.bs
```diff
@@ -2517,6 +2517,71 @@ run these steps:</p>
...
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
+  <var>request</var>'s <a for=request>current url</a>'s <a for=url>origin</a>, then return
+  <b>allowed</b>.
+
+  <p class="note no-backref">A cross-origin response redirecting to a response that is
+  <a>same origin</a> or <a>same site</a> with the initial request and has a
+  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header specified, does not affect
+  anything. I.e., <var>request</var>'s <a for=request>tainted origin flag</a> is not checked.
```
mikewest: Nit: I'd rephrase this a little bit. Perhaps something like "The cross-origin resource policy check does not consider a request's redirect chain when processing a given response's `Cross-Origin-Resource-Policy` header, but only the relationship between the requesting origin and the response which asserts the header."?

fetch.bs
annevk: That doesn't seem equivalent as it doesn't clearly evaluate to a boolean.

fetch.bs
annevk: Rewrote, but didn't use "redirect chain" as that's not really defined.

annevk:
I filed #767 on adding advice.

annevk:
Bugs:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1459573
* https://bugs.chromium.org/p/chromium/issues/detail?id=853723
* https://bugs.webkit.org/show_bug.cgi?id=186761
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17934554/

PR #738, opened 2018-05-25
Block AFP ports (427, 548)

### LINKED ISSUE #694: Add { 548, 'afp' } to the blocked bad ports
Chrome team is discussing to disallow port 548.
Can we have the 548 in the bad port list, https://fetch.spec.whatwg.org/#port-blocking ?

Also any other reason not to disallow other well-known < 1024 ports except for http/https ?
I guess it's just for possible compatibility breakage, but it would be great if we can disallow them eventually.

annevk:
Also remove some obsolete comments.

Fixes #694.

***
<a href="https://whatpr.org/fetch/738.html" title="Last updated on May 30, 2018, 1:40 PM GMT (fbcaa6f)">Preview</a> | <a href="https://whatpr.org/fetch/738/78a8dcd...fbcaa6f.html" title="Last updated on May 30, 2018, 1:40 PM GMT (fbcaa6f)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1690,7 +1690,8 @@ run these steps:
  <tr><td>143<td>imap2
  <tr><td>179<td>bgp
  <tr><td>389<td>ldap
- <tr><td>465<td>smtp+ssl
+ <tr><td>427<td>afp (alternate)
+ <tr><td>465<td>smtp (alternate)
  <tr><td>512<td>print / exec
  <tr><td>513<td>login
  <tr><td>514<td>shell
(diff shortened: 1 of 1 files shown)


annevk:
I wonder if there's some formal way to represent all these protocols. This is https://www-archive.mozilla.org/projects/netlib/PortBanning.html lowercased, but that's not necessarily great.

annevk:
(Note to self: scan web-platform-tests for this list to ensure it's kept up-to-date if it exists.)

annevk:
Tests: https://github.com/web-platform-tests/wpt/pull/11249.

annevk:
Note: I added a second commit that will block a port that Chrome has been blocking for quite a while.

annevk:
Bugs:
* https://bugs.webkit.org/show_bug.cgi?id=186092
* https://bugzilla.mozilla.org/show_bug.cgi?id=1465465
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17705091/
* https://bugs.chromium.org/p/chromium/issues/detail?id=847842

foolip:
I've also filed https://bugs.chromium.org/p/chromium/issues/detail?id=872252 about the fact that all of the tests in https://github.com/web-platform-tests/wpt/pull/11249 are failing because the bad port check happens synchronously in the constructor. @annevk, given that all browsers seem to do this, do you think we should change the spec instead?

foolip:
I was wrong, Edge passes these tests: https://wpt.fyi/results/websockets/Create-blocked-port.any.html?sha=edb72456b7

PR #749, opened 2018-05-30
Fixing variable definitions and scoping. Fixes #706.

### LINKED ISSUE #706: Abort refactoring led to variable errors
While working on #705 I noticed that because large swaths of code are now inside a conditional, sometimes variables get defined within that conditional that later on get referenced, except that doesn't work due to block scoping.

E.g., in https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch _storedResponse_ is now defined in a substep of step 2 and then referenced in a substep of step 4. That doesn't work.

@jakearchibald I'd appreciate if you could sort this, but let me know if you're too busy.

jakearchibald:
I used [this horrible piece of code](https://gist.github.com/jakearchibald/ceb666e0a9271b55cd48d07dac2b74b6) (run in the console) to identify variables that weren't correctly defined.

***
<a href="https://whatpr.org/fetch/749.html" title="Last updated on May 30, 2018, 2:48 PM GMT (8ce0a4c)">Preview</a> | <a href="https://whatpr.org/fetch/749/ffbaefb...8ce0a4c.html" title="Last updated on May 30, 2018, 2:48 PM GMT (8ce0a4c)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3501,7 +3501,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
  <li><p>Let <var>storedResponse</var> be null.
 
- <li><p>Let <var>revalidatingFlag</var> be an unset flag.
+ <li><p>Let the <var>revalidatingFlag</var> be unset.
 
  <li>
   <p>Run these steps, but <a>abort when</a> the ongoing fetch is <a for=fetch>terminated</a>:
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2857,16 +2857,14 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
      <var>noCorsResponse</var>.
 
      <li>
-      <p>Set <var>corbSanitizedResponse</var> to a new <a for=/>response</a> whose
+      <p>Return a new <a for=/>response</a> whose
```
jakearchibald: Let me fix the wrapping here

annevk:
Thanks, I hope that works out!

PR #751, opened 2018-06-01
Specify identity encoding for range requests. Fixes #747.

### LINKED ISSUE #747: Some servers seem to expect 'Accept-Encoding : identity' to serve Range requests
As part of https://bugs.webkit.org/show_bug.cgi?id=184447, it seems  https://test.shhnjk.com/media_test.html server is requiring 'Accept-Encoding : identity' to be in a request to serve range responses.
That might create some issues with multimedia loads intercepted by service workers.

jakearchibald:


jakearchibald:
Tests: https://github.com/web-platform-tests/wpt/pull/11291

jakearchibald:
This is different to what browser appear to implement. Chrome and Safari appear to special-case media, and they add the `Accept-Encoding` header before the service worker (although Safari then filters it before the service worker).

This PR adds `Accept-Encoding: identity` for all requests with a `Range` header, regardless of source, and adds the header after the service worker (where `Accept-Encoding` is usually added).

youennf:
The PR looks good to me.

I wonder what is the fetch spec policy regarding external links like https://jakearchibald.github.io/accept-encoding-range-test/.
It does not seem to work in Safari for instance and my locally installed Chrome bails out because of TransformStream not being present.
Maybe a snapshot of the results could be hosted somewhere (whatwg owned maybe) and would link to that particular URL for fresher results. 

> This PR adds Accept-Encoding: identity for all requests with a Range header, regardless of source, and adds the header after the service worker (where Accept-Encoding is usually added).

That makes sense to me. This is simple to implement, will simplify range requests authoring, including when intercepted by service workers. There seems to be no justification for making Accept-Encoding a special privileged header at the moment.

jakearchibald:
@youennf 

> It does not seem to work in Safari for instance and my locally installed Chrome bails out because of TransformStream not being present.

I reckon I can make it work in the latest versions of Safari, Chrome, Firefox, & Edge. The TransformStream stuff is just a performance enhancement, and mostly just me using an excuse to play.

> could be hosted somewhere (whatwg owned maybe)

I've got no problem with moving it. @annevk, any preferences?

fetch.bs
```diff
@@ -3651,6 +3651,16 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
     </ol>
 
+   <li>
+    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">contains</a>
+    `<code>Range</code>`, then <a for="header list">append</a>
+    `<code>Accept-Encoding</code>`/`<code>identity</code>` to <var>httpRequest</var>'s
+    <a for=request>header list</a>.
+
+    <p class="note no-backref">
+    <a href="https://jakearchibald.github.io/accept-encoding-range-test/">Many servers</a>
```
annevk: You cannot have a newline before `<a` here. We only allow whitespace at the end of paragraphs.

fetch.bs
```diff
@@ -3661,7 +3671,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     `<code>Connection</code>`,
     `<code>DNT</code>`, and
     `<code>Host</code>`,
-    are to be <a for="header list">appended</a> if necessary.
+    are to be <a for="header list">appended</a> if necessary, unless already specified.
```
annevk: This should be part of the normative requirement, no?

> Modify httpRequest's header list per HTTP, except for headers already appended prior to this step.

jakearchibald:
@annevk I've tried to make the test normative, although it's a bit clunky.

fetch.bs
```diff
@@ -3652,8 +3652,20 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
    <li>
-    <p>Modify <var>httpRequest</var>'s
-    <a for=request>header list</a> per HTTP.
+    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">contains</a>
+    `<code>Range</code>`, then <a for="header list">append</a>
+    `<code>Accept-Encoding</code>`/`<code>identity</code>` to <var>httpRequest</var>'s
+    <a for=request>header list</a>.
+
+    <p class="note no-backref"><a href="https://jakearchibald.github.io/accept-encoding-range-test/">Many
+    servers</a> mistakenly ignore `<code>Range</code>` headers if a non-identity encoding is
```
annevk: No newline here. Only HTML allows newlines in inlines. I can fix this up for you before landing though.

jakearchibald: Ah, I thought the fetch rule was only for references. Fixed.

jakearchibald:
Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=849952
WebKit: https://bugs.webkit.org/show_bug.cgi?id=186335 (@youennf I couldn't find a general 'fetch' component for this to go in, so it may need triaged).
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1467010 (@annevk can you triage? The 'component' drop down seems to only contain irrelevant stuff these days).
Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17784386/

Issue #565, opened 2017-07-13
WebSocket: "HTTP authentication will not function" is not correct


ricea:
https://fetch.spec.whatwg.org/commit-snapshots/f3bb21991abdd335175fcc5d26a0d0b7b380d4fe/#websocket-opening-handshake

> HTTP authentication will not function

Actually HTTP authentication functions with WebSockets in Chrome, Firefox, IE, and Edge. It doesn't work in Safari, but that's a bug: https://bugs.webkit.org/show_bug.cgi?id=80362

Firefox will also give you a password prompt if it gets a 401 response on a WebSocket. Chrome and Edge will only use cached credentials. In Chrome's case, this is the policy for all subresources, rather than being a WebSocket-specific restriction.

annevk:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1380973 to remove the prompting in Firefox.

annevk:
@ricea do you wait for a challenge or just send credentials if you have them?

annevk:
Also, are there tests for this?

ricea:
As far as I know, Chrome always sends credentials in the initial request if it has them, but from the limited testing I've done it appears that Firefox doesn't. In Chrome we share the logic with ordinary HTTP.

There don't appear to be any web-platform-tests testing this. We have Chrome-specific tests, but not of a sort that could be straightforwardly ported.

ricea:
I can't actually find any normative text that prohibits HTTP authentication. I will create a PR to remove the text that says it doesn't.

ricea:
Simple wpt at https://github.com/web-platform-tests/wpt/pull/11433.

dveditz:
> As far as I know, Chrome always sends credentials in the initial request if it has them, 

Without the 401 that could in theory be sending auth for the wrong realm.

ricea:
@dveditz wrote:
> Without the 401 that could in theory be sending auth for the wrong realm.

Thanks for the clarification! If I catch Chrome doing the wrong thing I will file a bug.

PR #761, opened 2018-06-08
Stop saying WebSocket auth is disallowed

ricea:
Remove the warning that HTTP authentication is prohibited for
WebSockets.

See #565 for background.

***
<a href="https://whatpr.org/fetch/761.html" title="Last updated on Jun 15, 2018, 2:21 PM GMT (21e683d)">Preview</a> | <a href="https://whatpr.org/fetch/761/2f3d04d...21e683d.html" title="Last updated on Jun 15, 2018, 2:21 PM GMT (21e683d)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6301,12 +6301,11 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <p><dfn>Fail the WebSocket connection</dfn> and <dfn>the WebSocket connection is established</dfn>
 are defined by The WebSocket Protocol. [[!WSP]]
 
-<p class=warning>The reason redirects are not followed, HTTP authentication will not function, and
-this handshake is generally restricted is because that could introduce serious security problems in
-a web browser context. For example, consider a host with a WebSocket server at one path and an open
-HTTP redirector at another. Suddenly, any script that can be given a particular WebSocket URL can be
-tricked into communicating to (and potentially sharing secrets with) any host on the internet, even
-if the script checks that the URL has the right hostname.
+<p class=warning>The reason redirects are not followed is because it could introduce serious
+security problems in a web browser context. For example, consider a host with a WebSocket server at
+one path and an open HTTP redirector at another. Suddenly, any script that can be given a particular
+WebSocket URL can be tricked into communicating to (and potentially sharing secrets with) any host
+on the internet, even if the script checks that the URL has the right hostname.
 <!-- https://www.ietf.org/mail-archive/web/hybi/current/msg06951.html -->
 
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -6301,11 +6301,12 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <p><dfn>Fail the WebSocket connection</dfn> and <dfn>the WebSocket connection is established</dfn>
 are defined by The WebSocket Protocol. [[!WSP]]
 
-<p class=warning>The reason redirects are not followed is because it could introduce serious
-security problems in a web browser context. For example, consider a host with a WebSocket server at
-one path and an open HTTP redirector at another. Suddenly, any script that can be given a particular
-WebSocket URL can be tricked into communicating to (and potentially sharing secrets with) any host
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -6301,12 +6301,11 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <p><dfn>Fail the WebSocket connection</dfn> and <dfn>the WebSocket connection is established</dfn>
 are defined by The WebSocket Protocol. [[!WSP]]
 
-<p class=warning>The reason redirects are not followed, HTTP authentication will not function, and
-this handshake is generally restricted is because that could introduce serious security problems in
```
annevk: It seems keeping "handshake is generally restricted" would be good. We should probably define how HTTP authentication is supposed to work though. I.e., only works if there's an authentication entry.

ricea: I took out the "generally restricted" part because I couldn't describe the nature of the restriction.

In Chrome's implementation we're pretty restrictive: we error on anything other than 101, 401 or 407 response, and UI is not shown for authentication or TLS failures. Do you think there's a way we could codify some or all of these restrictions in the standard?

annevk: Doesn't it codify those restrictions already? This is just warning you about why you should follow them.

ricea: > Doesn't it codify those restrictions already? 

I'm planning to write up my analysis of what I think the existing specified restrictions are, but I'm buried in a pile of other stuff. Sorry for the delay.

ricea: Okay, I've done my analysis of the restrictions. It does match pretty closely to Chrome's implementation.I will add back the "generally restricted" text. The 401 behaviour described by the standard is odd, but maybe we should discuss that elsewhere.

--

> Let request be a new [request](https://fetch.spec.whatwg.org/#concept-request), whose [url](https://fetch.spec.whatwg.org/#concept-request-url) is requestURL, [client](https://fetch.spec.whatwg.org/#concept-request-client) is client, [service-workers mode](https://fetch.spec.whatwg.org/#request-service-workers-mode) is "none", [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer) is "no-referrer", [synchronous flag](https://fetch.spec.whatwg.org/#synchronous-flag) is set, [mode](https://fetch.spec.whatwg.org/#concept-request-mode) is "websocket", [credentials mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) is "include", [cache mode](https://fetch.spec.whatwg.org/#concept-request-cache-mode) is "no-store", and [redirect mode](https://fetch.spec.whatwg.org/#concept-request-redirect-mode) is "error".

[service-workers mode](https://fetch.spec.whatwg.org/#request-service-workers-mode) is "none" -- No service workers will get events for this fetch.

 [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer) is "no-referrer", -- No "Referer:" header is sent in the request

 [mode](https://fetch.spec.whatwg.org/#concept-request-mode) is "websocket" -- No preflight. "Origin:" header is included. "Obtaining a WebSocket connection" is used.

[credentials mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) is "include" -- cookies are sent. Authorization header is set if known to be needed. On 401 response we maybe prompt for credentials? Or not, because "request's [window](https://fetch.spec.whatwg.org/#concept-request-window) is an [environment settings object](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object)," is not true?

[cache mode](https://fetch.spec.whatwg.org/#concept-request-cache-mode) is "no-store", adding "Pragma: no-cache", and "Cache-Control: no-cache" to header list.

> If response is a [network error](https://fetch.spec.whatwg.org/#concept-network-error) or its [status](https://fetch.spec.whatwg.org/#concept-response-status) is not 101, [fail the WebSocket connection](https://fetch.spec.whatwg.org/#fail-the-websocket-connection).

Effectively bans any status except for 401, 407 and 101.

annevk: Thanks, filed #766 as a follow-up.

PR #764, opened 2018-06-14
Additional reasoning for sending Accept-Encoding: identity

jakearchibald:
I hadn't really appreciated this with the initial PR, but preventing encoding solves a pretty big gotcha if developers tried to fetch ranges with `fetch()`.

***
<a href="https://whatpr.org/fetch/764.html" title="Last updated on Jun 14, 2018, 12:54 PM GMT (571aa0f)">Preview</a> | <a href="https://whatpr.org/fetch/764/2f3d04d...571aa0f.html" title="Last updated on Jun 14, 2018, 12:54 PM GMT (571aa0f)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3657,8 +3657,14 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     `<code>Accept-Encoding</code>`/`<code>identity</code>` to <var>httpRequest</var>'s
     <a for=request>header list</a>.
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
I hadn't realised that if a resource is encoded, a range is taken from that encoded resource. So if the range is `bytes=1-`, and the content is encoded, `fetch()` will always reject when it tries to decode the content.

If the server returns unencoded data, this failure won't occur.

Issue #726, opened 2018-05-17
Integrate latest Client Hints updates


igrigorik:
Consolidating the tasks into a meta bug, so we can track progress..

### Update list of hints

- [ ] Add `RTT`, `ECT`, `Downlink`, and `Device-Memory` headers to the [CORS-safelisted list](https://fetch.spec.whatwg.org/#cors-safelisted-request-header), [Client Hints list](https://fetch.spec.whatwg.org/#client-hints-list), and [processing (step 6)](https://fetch.spec.whatwg.org/#fetching).
    - Covered by https://github.com/whatwg/fetch/pull/725

### Integrate `Accept-CH-Lifetime` / opt-in logic

- [ ] Add a step after step 10 in [Main Fetch](https://fetch.spec.whatwg.org/#main-fetch), which reads `Accept-CH-Lifetime` and adds it to some CH-opt-in cache
  - [ ] Add definition of said cache
  - Covered by https://github.com/whatwg/html/pull/3774
- [ ] Change steps 6-7 in [Fetching](https://fetch.spec.whatwg.org/#fetching) to take `Accept-CH` and/or CH-opt-in cache into consideration.
  - [ ] Limit above steps to secure origins
  - [ ] Limit above steps to same-origin - unless the origin is on some list of allowed origins
  - Covered by https://github.com/whatwg/fetch/pull/773
- [ ] Update HTML spec to support `<meta http-equiv>` opt-in
   - Also covered by https://github.com/whatwg/html/pull/3774
- [ ] Add WPT tests for opt-in logic
   - Covered by https://wpt.fyi/results/client-hints

### Integrate Feature-Policy + Client Hints hooks
- [ ] Hook up Feature Policy and https://github.com/WICG/feature-policy/issues/129 to said list of allowed origins
- [ ] Add WPT tests for FP opt-in

@tarunban @yoavweiss @annevk anything missing?

yoavweiss:
I started working on the "CH opt-in" cache part at https://github.com/whatwg/fetch/pull/729, but following an [IRC discussion](https://freenode.logbot.info/whatwg/20180522#c1556544) with @annevk I'm now wondering what's the right way forward.

The above PR adds that infrastructure to Fetch, but it was suggested that HTML might be a more appropriate place. Among other considerations, it depends if the response header handling is limited to navigation responses (which are also [handled in HTML](https://html.spec.whatwg.org/#process-a-navigate-response)) or also cover subresource responses (handled only in Fetch AFAICT).

[...]

PR #725, opened 2018-05-15
Add device-memory and network quality client hints to the list

tarunban:
annevk: please take a look. This is the first PR where the new client hints are added to the spec.

***
<a href="https://whatpr.org/fetch/725.html" title="Last updated on May 30, 2018, 11:54 PM GMT (43eea2f)">Preview</a> | <a href="https://whatpr.org/fetch/725/3055f13...43eea2f.html" title="Last updated on May 30, 2018, 11:54 PM GMT (43eea2f)">Diff</a>

PR #773, opened 2018-06-26
Align CH processing model with implementation and HTML#3774

yoavweiss:
This PR tackles a part of https://github.com/whatwg/fetch/issues/726, by copying the request's client-hints list from the environment settings object's one (as defined in https://github.com/whatwg/html/pull/3774). It also aligns the addition of CH headers to current implementations, and adds the Device-Memory, RTT, Downlink and ECT hints.

***
<a href="https://whatpr.org/fetch/773.html" title="Last updated on Mar 26, 2019, 9:25 AM UTC (5067615)">Preview</a> | <a href="https://whatpr.org/fetch/773/6a644c6...5067615.html" title="Last updated on Mar 26, 2019, 9:25 AM UTC (5067615)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -644,7 +644,9 @@ production as
 
 <ol>
  <li><p>Let <var>value</var> be <var>header</var>'s <a for=header>value</a>.
- <li><p>If <var>value</var> starts with "Sec-", return true.
+
+ <li><p>If <var>header</var>'s <a for=header>name</a> starts with a <a>byte-case-insensitive</a>
+ match to `Sec-`, return true.
 
  <li>
   <p><a>Byte-lowercase</a> <var>header</var>'s <a for=header>name</a> and switch on the result:
--- a/fetch.bs
+++ b/fetch.bs
@@ -3187,8 +3187,7 @@ the request.
         <dd>a suitable <a href=https://wicg.github.io/netinfo/#ect-request-header-field>ECT value</a>
        </dl>
        
-      <li><p>If <var>request</var> is a cross-origin <a>subresource request</a> and the result of
-       running <a
+      <li><p>If <var>request</var> is a <a>subresource request</a> and the result of running <a
        href="https://wicg.github.io/feature-policy/#should-request-be-allowed-to-use-feature">Should
        <var>request</var> be allowed to use <var>feature</var>?</a>,
        given <var>request</var> and <var>hintName</var>’s
--- a/fetch.bs
+++ b/fetch.bs
@@ -644,6 +644,7 @@ production as
 
 <ol>
  <li><p>Let <var>value</var> be <var>header</var>'s <a for=header>value</a>.
+ <li><p>If <var>value</var> starts with "Sec-", return true.
 
  <li>
   <p><a>Byte-lowercase</a> <var>header</var>'s <a for=header>name</a> and switch on the result:
@@ -1762,6 +1763,10 @@ specified. [[!CSP]]
 <p>A <a for=/>response</a> has an associated
 <dfn for=response id=concept-response-range-requested-flag>range-requested flag</dfn>, which is
 initially unset.
+
+<p>A <a for=/>response</a> has an associated
+<dfn for=response id=concept-response-image-density>image density</dfn>, which is initially set to
+zero.
 
 <p class=note>This is used to ensure to prevent a partial response from an earlier ranged request
 being provided to an API that didn't make a range request. See the flag's usage for a detailed
@@ -3191,6 +3196,7 @@ the request.
 
        <!-- TODO(yoav): Fix the above Feature policy link -->
        <!-- TODO(yoav): Fix the Client hints reference to the HTML one, once it lands -->
+      <li><p>Set <var>hintName</var> to "Sec-" concatenated with <var>hintName</var>.
 
       <li><p>If <var>request</var>'s <a for=request>header list</a> <a for="header list">does not
       contain</a> <var>hintName</var>, a user agent should <a for="header list">append</a>
@@ -3583,6 +3589,12 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
    <a>queue a fetch-request-done task</a> for <var>request</var>.
   </ol>
 
+ <li><p>If <var>request</var>'s destination is "image" and <var>response</var>'s
+ <a for=response>header list</a> <a for="header list">contains</a>
+  `<code>Content-DPR</code>`, set <var>response</var>'s <a for=response>density</a> value to the
+ result of parsing the header value as float.
+ <!-- TODO: propoerly link parse as float -->
+
  <li><p><a>Queue a fetch task</a> on <var>request</var> to
  <a>process response</a> for <var>response</var>.
 
@@ -3945,6 +3957,20 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
  <a for=request>tainted origin flag</a> is set or <var>request</var>'s <a for=request>origin</a> is
  not <a>same origin</a> with <var>actualResponse</var>'s <a for=response>location URL</a>'s
  <a for=url>origin</a>, then return a <a>network error</a>.
+
+ <li><p><a for=list>For each</a> <var>hintName</var> of <var>request</var>'s
+ <a for=request>client-hints set</a>:
+ <ol>
+  <li><p>Set <var>hintName</var> to "Sec-" concatenated with <var>hintName</var>.
+  <li><p>If <var>request</var>'s <a for=request>header list</a> <a for="header list"> contains</a>
+  <var>hintName</var> and if the result of running <a
+  href="https://wicg.github.io/feature-policy/#should-request-be-allowed-to-use-feature">Should
+  request be allowed to use feature?</a>, given <var>request</var> and <var>hintName</var>’s
+  <a href="http://httpwg.org/http-extensions/client-hints.html#opt-in-via-feature-policy">associated
+  policy-controlled feature</a>, returns <code>false</code>, then remove <var>hintName</var> from
+  <a for=request>header list</a>.
+  [[!FEATURE-POLICY]] [[!CLIENT-HINTS]]
+ </ol>
 
  <li>
   <p>If <i>CORS flag</i> is set and <var>actualResponse</var>'s
--- a/fetch.bs
+++ b/fetch.bs
@@ -3186,23 +3186,24 @@ the request.
         <dt>`<code>ECT</code>`
         <dd>a suitable <a href=https://wicg.github.io/netinfo/#ect-request-header-field>ECT value</a>
        </dl>
-
-       <li><p>If the result of running
-       <a href="https://wicg.github.io/feature-policy/#should-request-be-allowed-to-use-feature">Should
+       
+      <li><p>If <var>request</var> is a cross-origin <a>subresource request</a> and the result of
+       running <a
+       href="https://wicg.github.io/feature-policy/#should-request-be-allowed-to-use-feature">Should
        <var>request</var> be allowed to use <var>feature</var>?</a>,
        given <var>request</var> and <var>hintName</var>’s
        <a href="http://httpwg.org/http-extensions/client-hints.html#opt-in-via-feature-policy">associated
-       policy-controlled feature</a>, returns <code>false</code>, then continue.
+       policy-controlled feature</a>, returns <code>false</code>, then skip the next steps and
+       continue to the next <var>hintName</var>.
        [[!FEATURE-POLICY]] [[!CLIENT-HINTS]]
 
+       <!-- TODO(yoav): Fix the above Feature policy link -->
+       <!-- TODO(yoav): Fix the Client hints reference to the HTML one, once it lands -->
+       
       <li><p>If <var>request</var>'s <a for=request>header list</a> <a for="header list">does not
       contain</a> <var>hintName</var>, a user agent should <a for="header list">append</a>
       <var>hintName</var>/<var>value</var> to <var>request</var>'s <a for=request>header list</a>.
      </ol>
-
-     <p class=XXX>The above step should be applicable only for same-origin requests. See
-     <a href="https://github.com/whatwg/fetch/issues/800">issue</a> for more details.</p>
-
 
     <li><p>If <var>request</var> is a <a>subresource request</a>, then:
      <ol>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -996,11 +996,11 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
 
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-client-hints-list>client hints list</dfn>,
-which is a <a lt="client hints list" for=client>client-hints list</a>. Unless stated
-otherwise, it is the empty list.
+<dfn export for=request id=concept-request-client-hints-set>client hints set</dfn>,
```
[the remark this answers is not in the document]
yoavweiss: ok

fetch.bs
```diff
@@ -996,11 +996,11 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
...
+which is a <a lt="client hints set" for=client>client-hints set</a>. Unless stated
```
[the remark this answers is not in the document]
yoavweiss: Removed "for=client".

Agreed regarding consistency. Added hyphen everywhere.

fetch.bs
```diff
@@ -1825,14 +1825,14 @@ run these steps:
 </ol>
 
 
-<h3 id=client-hints-list>Client hints list</h3>
+<h3 id=client-hints-set>Client hints set</h3>
```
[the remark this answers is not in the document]
yoavweiss: ok

fetch.bs
```diff
@@ -1825,14 +1825,14 @@ run these steps:
...
 
 <p class=note>This section will be integrated into HTTP Client Hints.
 [[!CLIENT-HINTS]]
 <!-- XXX -->
 
-<p>A <dfn export id=concept-client-hints-list for=client>client hints list</dfn> is a
-<a for=/>list</a> of
+<p>A <dfn export id=concept-client-hints-set for=client>client hints set</dfn> is a
```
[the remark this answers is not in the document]
yoavweiss: removed

fetch.bs
```diff
@@ -2697,27 +2701,24 @@ the request.
...
+    <p>If <var>request</var>'s <a for=request>current url</a>'s
+    <a for=url>origin</a> is <a>same origin</a> with <var>request</var>'s
+    <a for=request>origin</a>, then:
 
     <ol>
      <li>
+
       <p><a for=list>For each</a> <var>hintName</var> of <var>request</var>'s
-      <a for=client>client hints list</a>:
+      <a for=client>client-hints set</a>:
```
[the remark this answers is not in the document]
yoavweiss: ack

fetch.bs
```diff
@@ -2697,27 +2701,24 @@ the request.
...
 
       <ol>
-       <li>
+        <li>
```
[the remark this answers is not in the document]
yoavweiss: ok

fetch.bs
```diff
@@ -2729,23 +2730,35 @@ the request.
...
-     <a for=/>fetch</a> algorithm.
-
-     <li><p>Append <var>record</var> to <var>request</var>'s
-     <a for=request>client</a>'s
-     <a for=fetch>fetch group</a> list of
-     <a for="fetch group">fetch records</a>.
     </ol>
-  </ol>
+
+    <li><p>If <var>request</var> is a <a>subresource request</a>, then:
```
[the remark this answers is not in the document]
yoavweiss: ack

fetch.bs
```diff
@@ -2697,27 +2701,24 @@ the request.
...
-     <tr>
-      <td>`<code>Viewport-Width</code>`
-      <td>a suitable <a href=http://httpwg.org/http-extensions/client-hints.html#viewport-width>viewport-width value</a>
     </table>
 
    <li>
-    <p>If <var>request</var> is a <a>subresource request</a>, then:
+    <p>If <var>request</var>'s <a for=request>current url</a>'s
+    <a for=url>origin</a> is <a>same origin</a> with <var>request</var>'s
+    <a for=request>origin</a>, then:
```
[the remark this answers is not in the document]
yoavweiss: > This means that a cross-origin URL inbetween doesn't influence this

Do you mean a cross-origin redirection between two same origin requests? or something else?

fetch.bs
```diff
@@ -2640,6 +2640,10 @@ the request.
    <a for=request>origin</a> to  <var>request</var>'s
    <a for=request>client</a>'s <a for="environment settings object">origin</a>.
 
+   <li><p>Set <var>request</var>'s <a for=request>client-hints set</a> to be a copy of the <a
```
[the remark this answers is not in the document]
yoavweiss: Added a direct link to the definition, but not sure that's the best way to do that. Let me know if there's a better way

annevk:
@yoavweiss same-origin request -> redirect to cross-origin URL -> redirect to same-origin URL.

fetch.bs
```diff
@@ -2640,6 +2640,11 @@ the request.
    <a for=request>origin</a> to  <var>request</var>'s
    <a for=request>client</a>'s <a for="environment settings object">origin</a>.
 
+   <li><p>Set <var>request</var>'s <a for=request>client-hints set</a> to be a <a
+   href="https://infra.spec.whatwg.org/#list-clone">clone</a> of the <a
```
annevk: You can use `<a for=list>clone</a>`.

annevk: Also, no newlines inside tags or inline elements (slightly different rules from HTML, sorry).

yoavweiss: cool

fetch.bs
```diff
@@ -1825,14 +1825,14 @@ run these steps:
 </ol>
 
 
-<h3 id=client-hints-list>Client hints list</h3>
+<h3 id=client-hints-list>Client-hints set</h3>
 
-<p class=note>This section will be integrated into HTTP Client Hints.
+<p class=note>This section will be integrated into HTTP Client-Hints.
```
annevk: Is that still the plan? In any event, this section should have a `<p class=XXX>` explaining that currently the integration is not well defined.

fetch.bs
```diff
@@ -2697,55 +2701,60 @@ the request.
...
-     <li><p>Append <var>record</var> to <var>request</var>'s
-     <a for=request>client</a>'s
-     <a for=fetch>fetch group</a> list of
-     <a for="fetch group">fetch records</a>.
-    </ol>
-  </ol>
+    <li>
+
+     <p class=note>The following step should be applicable only for same-origin requests. See
+     <a href="https://github.com/whatwg/fetch/issues/800">issue</a> for more details.</p>
```
annevk: This should be class=XXX and should be placed at the end of the step per convention.

yoavweiss:
> It also seems this adds more Client-Hints headers that are not CORS-safelisted. Is that intentional?

those are added to the list at https://github.com/whatwg/fetch/pull/725

yoavweiss:
> A comment message detailing the changes made here would help

I'll squash and add a meaningful commit message

fetch.bs
```diff
@@ -1825,14 +1825,14 @@ run these steps:
 </ol>
 
 
-<h3 id=client-hints-list>Client hints list</h3>
+<h3 id=client-hints-list>Client-hints set</h3>
 
-<p class=note>This section will be integrated into HTTP Client Hints.
+<p class=XXX>This section will be integrated into HTTP Client-Hints.
```
annevk: Sorry for being unclear, this issue marker should indicate that Client Hints integration with Fetch is far from finished, similar to the issue marker you added below. I was thinking it would be good to stress that in multiple places.

annevk:
The commit message needs to explain the removal of DPR and Viewport-Width from the first table as well.

fetch.bs
yoavweiss: Changed to point at https://github.com/whatwg/fetch/issues/726
Let me know if that works

yoavweiss:
> The commit message needs to explain the removal of DPR and Viewport-Width from the first table as well.

added

annevk:
Having considered this further I'm not sure how to accept this as it has the same problems as #725:

1. It's unclear what the level of implementer support is.
2. There are no tests.
3. Adequate tests cannot be written as the infrastructure is still broken (per #726 and #800).

yoavweiss:
> It's unclear what the level of implementer support is.

This is currently implemented by Chromium, but not elsewhere. What's the typical process for integrating such features before they gain wider adoption?

> There are no tests.

That is not true. See https://wpt.fyi/results/client-hints?label=experimental

> Adequate tests cannot be written as the infrastructure is still broken (per #726 and #800).

Are you suggesting that we need the feature policy (https://github.com/WICG/feature-policy/issues/129) and redirection logic (#800) pieces in place before any of this can land?

annevk:
@yoavweiss typically we don't merge until there's at least two implementers committed to some extent (no need for two implementations).

As for tests, I meant tests covering the problematic bits.

And since the redirection logic is vital to how this ends up being defined, yes. It would be less problematic if that were a simple change, but basically all the existing infrastructure is in the wrong place for it, and we might need new primitives as well depending on how we deal with UA-set vs developer-set headers.

annevk:
https://whatwg.org/working-mode#changes has a more formal description of WHATWG's changes policy.

yoavweiss:
> And since the redirection logic is vital to how this ends up being defined, yes. It would be less problematic if that were a simple change, but basically all the existing infrastructure is in the wrong place for it, and we might need new primitives as well depending on how we deal with UA-set vs developer-set headers.

I agree that the redirection logic is critical once we start doing same-origin checks. This PR (as well as current Chromium implementation) does not do that. And indeed, things may move once that logic is settled and lands. 

At the same time, I thought it'd be worthwhile to define the current behavior and then iterate on that. (as the current spec definition doesn't match the current implementation)

fetch.bs
```diff
@@ -3208,23 +3208,23 @@ the request.
         <dd>a suitable <a href=https://wicg.github.io/netinfo/#ect-request-header-field>ECT value</a>
        </dl>
        
-      <li><p>If the result of running
-       <a href="https://wicg.github.io/feature-policy/#should-request-be-allowed-to-use-feature">Should
+      <li><p>If <var>request</var> is a cross-origin <a>subresource request</a> and the result of
```
eeeps: I don't think we want to limit this check to cross-origin requests. The default policies for these features should certainly include 'self', and I can't really think of a use case for disabling them for 'self', but the FP spec/syntax gives authors full control over all origins, including 'self,' and I don't want this to be a weird exception to that.

fetch.bs
```diff
@@ -2146,14 +2154,15 @@ run these steps:
...
 
-<p class=note>This section will be integrated into HTTP Client Hints.
+<p class=XXX>HTTP Client-Hints requires further integration. More details on the
+<a href="https://github.com/whatwg/fetch/issues/726">issue</a>.
 [[!CLIENT-HINTS]]
 <!-- XXX -->
 
-<p>A <dfn export id=concept-client-hints-list for=client>client hints list</dfn> is a
-<a for=/>list</a> of
+<p>A <dfn export id=concept-client-hints-list>client-hints set</dfn> is a
```
Ms2ger: Try this?

```suggestion
<p>A <dfn export id=concept-client-hints-list for=/>client-hints set</dfn> is a
```

yoavweiss: Travis still [doesn't like me](https://travis-ci.org/whatwg/fetch/builds/481359853?utm_source=github_status&utm_medium=notification)

fetch.bs
```diff
@@ -1777,6 +1777,10 @@ specified. [[!CSP]]
 <dfn for=response id=concept-response-range-requested-flag>range-requested flag</dfn>, which is
 initially unset.
 
+<p>A <a for=/>response</a> has an associated
```
eeeps: This should move below the note, below, which relates to the paragraph about range-requested flags, above.

fetch.bs
```diff
@@ -2146,19 +2150,14 @@ run these steps:
...
-<!-- XXX -->
+<h3 id=client-hints-list>Client-hints set</h3>
 
-<p>A <dfn export id=concept-client-hints-list for=client>client hints list</dfn> is a
-<a for=/>list</a> of
+<p>A <dfn export id=concept-client-hints-list>client-hints set</dfn> is a
+<a for=/>set</a> of
 <a href=http://httpwg.org/http-extensions/client-hints.html#accept-ch>Client hint tokens</a>, each
 of which is one of `<code>DPR</code>`, `<code>Save-Data</code>`, `<code>Viewport-Width</code>`, or
 `<code>Width</code>`.
```
eeeps: Add the other new hints here?

fetch.bs
```diff
@@ -3162,55 +3165,69 @@ the request.
...
+        <dt>`<code>Downlink</code>`
+        <dd>a suitable <a href=https://wicg.github.io/netinfo/#downlink-request-header-field>Downlink value</a>
+        <dt>`<code>ECT</code>`
+        <dd>a suitable <a href=https://wicg.github.io/netinfo/#ect-request-header-field>ECT value</a>
+       </dl>
+
+      <li><p>If <var>request</var> is a <a>subresource request</a> and the result of running <a
+       href="https://wicg.github.io/feature-policy/#should-request-be-allowed-to-use-feature">Should
+       request be allowed to use feature?</a>, given <var>request</var> and <var>hintName</var>’s
+       <a href="http://httpwg.org/http-extensions/client-hints.html#opt-in-via-feature-policy">associated
```
eeeps: Reading the discussion on the relevant PR over in the Client Hints spec (https://github.com/httpwg/http-extensions/pull/696), I don't think the definitions of the policy-controlled features that we're linking to here will be landing there. We need to find a new place for those definitions to live, and change this link.

fetch.bs
```diff
@@ -3584,6 +3601,12 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
    <a>queue a fetch-request-done task</a> for <var>request</var>.
   </ol>
 
+ <li><p>If <var>request</var>'s destination is "image" and <var>response</var>'s
+ <a for=response>header list</a> <a for="header list">contains</a>
+  `<code>Content-DPR</code>`, set <var>response</var>'s <a for=response>density</a> value to the
```
eeeps: This should be “*image* density", rather than just "density", to match up with the definition.

yoavweiss: good catch, thanks!

fetch.bs
yoavweiss: ack

fetch.bs
yoavweiss: yup

yoavweiss:
I've moved most of the processing defined in this PR to https://yoavweiss.github.io/client-hints-infrastructure/
Regarding the rest, it's PRed in https://github.com/whatwg/fetch/pull/971

I'm therefore closing this PR.

Issue #267, opened 2016-03-31
Use a ReadableStream with byte source (formerly called ReadableByteStream) for .body


tyoshino:
/cc @yutakahirano

ReadableByteStream has been merged into ReadableStream interface. It's time to update the ReadableStream section of the Fetch spec (https://fetch.spec.whatwg.org/#readablestream) to adopt the updated ReadableStream and construct a ReadableStream with the byte handling feature (i.e. getReader() returns a ReadableStreamBYOBReader when a certain option is passed).

In the first step, we don't need to fully use the BYOB responding API but may just call the same enqueue() method.

youennf:
Currently the fetch spec allows constructing a Response with any ReadableStream, which can thus be enqueued with ArrayBuffers, Uint8Arrays or potatoes.

Restricting the Response body to a ReadableStream with a byte source would remove the possibility to enqueue potatoes, leaving space for ArrayBuffer and ArrayBufferView.

But enqueuing something else than Uint8Array would still lead to TypeError when consuming the body, since "read all bytes" requires Uint8Array objects.
Can we remove that restriction?
That would make it consistent with the possibility to create a Response with whatever ArrayBuffer/ArrayBufferView.

annevk:
I thought we wanted to require Uint8Array?

domenic:
Yeah, a while ago we discussed whether we should perform some auto-conversion from array buffer views or blobs or strings, and decided to settle on being conservative and just allowing Uint8Array. I think the discussion was at https://github.com/yutakahirano/fetch-with-streams/issues/53.

youennf:
"read all bytes" is not really doing any conversion, it merely appends a Uint8Array to a byte sequence. Any BufferSource would be as easily appendable there.

I do not see what is gained with this restriction.
I can see how inconsistent it might be if Response gets restricted to a byte ReadableStream.

https://github.com/yutakahirano/fetch-with-streams/issues/53 seems mostly about upload.

annevk:
Yeah, though a `Response` object created in a service worker feeding a document/worker API is very similar to upload. I think we should try to keep them consistent.

annevk:
@yutakahirano is this fixed?

yutakahirano:
Not yet.

ricea:
This might not happen quickly, as we want to run an experiment in Chrome to ensure it is web-compatible to use a byte stream. In theory there is no problem, but AFAIK it's never been tested and it's not something we can afford to break.

[...]

PR #781, opened 2018-08-01
Be clearer about how Fetch creates ReadableStreams.

jyasskin:
It could be that this should call https://streams.spec.whatwg.org/#create-readable-stream or https://streams.spec.whatwg.org/#create-readable-byte-stream (#267?) instead of trying to go through the `ReadableStream()` constructor, with all the ambiguity about how modifying `Object`'s prototype should affect things.

@domenic

***
<a href="https://whatpr.org/fetch/781.html" title="Last updated on Aug 9, 2018, 1:25 PM GMT (f514b21)">Preview</a> | <a href="https://whatpr.org/fetch/781/b3492ec...f514b21.html" title="Last updated on Aug 9, 2018, 1:25 PM GMT (f514b21)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1872,49 +1872,33 @@ we define common operations for {{ReadableStream}} objects. [[!STREAMS]]
 
 <p>To
 <dfn export for=ReadableStream id=concept-construct-readablestream>construct a <code>ReadableStream</code> object</dfn>
-with given <var>strategy</var>, <var>pull</var> action and <var>cancel</var> action, all of which
-are optional, run these steps:
+with given <var>highWaterMark</var>, <var>sizeAlgorithm</var> algorithm, <var>pull</var> action and
+<var>cancel</var> action, all of which are optional, run these steps:
+
+<p class=note>This algorithm used to take a <var ignore>strategy</var> parameter, whose
+<code>highWaterMark</code> and <code>sizeAlgorithm</code> members were extracted to provide what are
+now separate parameters. If another specification still passes that <var ignore>strategy</var>
+parameter, please update it.
 
 <ol>
- <li><p>Let <var>startAlg</var> be an algorithm that returns <code>undefined</code>.
+ <li><p>Let <var>startAlgorithm</var> be an algorithm that returns <code>undefined</code>.
 
- <li><p>Let <var>pullAlg</var> be an algorithm that returns:
+ <li><p>If <var>pull</var> is absent, set it to an action that returns <code>undefined</code>.
 
- <dl class=switch>
- <dt>If <var>pull</var> is given
- <dd>the result of [=promise-calling=] <var>pull</var>().
- <dt>Otherwise
- <dd>[=a promise resolved with=] <code>undefined</code>.
- </dl>
+ <li><p>Let <var>pullAlgorithm</var> be an algorithm that returns the result of [=promise-calling=]
+ <var>pull</var>().
 
- <li><p>Let <var>cancelAlg</var> be an algorithm that takes a <var>reason</var> and returns:
- <dl class=switch>
- <dt>If <var>cancel</var> is given
- <dd>the result of [=promise-calling=] <var>cancel</var>(<var>reason</var>).
- <dt>Otherwise
- <dd>[=a promise resolved with=] <code>undefined</code>.
- </dl>
+ <li><p>If <var>cancel</var> is absent, set it to an action that returns <code>undefined</code>.
 
- <li><p>Let <var>highWaterMark</var> be:
- <dl class=switch>
- <dt>If <var>strategy</var> is given and <var>strategy</var>.<a spec=streams for="queuing
- strategy">highWaterMark</a> exists
- <dd><var>strategy</var>.<a spec=streams for="queuing strategy">highWaterMark</a>.
- <dt>Otherwise
- <dd><code>1</code>.
- </dl>
+ <li><p>Let <var>cancelAlgorithm</var> be an algorithm that takes a <var>reason</var> and returns
+ the result of [=promise-calling=] <var>cancel</var>(<var>reason</var>).
 
- <li><p>Let <var>sizeAlg</var> be:
- <dl class=switch>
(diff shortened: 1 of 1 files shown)


jyasskin:
@ricea Here's a terrible attempt at calling CreateReadableStream(). With some more work, it's probably possible to simplify that wording.

fetch.bs
```diff
@@ -1876,18 +1876,45 @@ with given <var>strategy</var>, <var>pull</var> action and <var>cancel</var> act
 are optional, run these steps:
 
 <ol>
- <li><p>Let <var>init</var> be a new object.
+ <li><p>Let <var>startAlg</var> be an algorithm that returns <code>undefined</code>.
```
ricea: Personally I find it clearer just to write out "startAlgorithm" in full.

fetch.bs
```diff
@@ -1876,18 +1876,45 @@ with given <var>strategy</var>, <var>pull</var> action and <var>cancel</var> act
...
+ <dt>If <var>cancel</var> is given
+ <dd>the result of [=promise-calling=] <var>cancel</var>(<var>reason</var>).
+ <dt>Otherwise
+ <dd>[=a promise resolved with=] <code>undefined</code>.
+ </dl>
 
- <li><p>Let <var>stream</var> be the result of calling the initial value of {{ReadableStream}} as
- constructor with <var>init</var> and <var>strategy</var> if given.
+ <li><p>Let <var>highWaterMark</var> be:
+ <dl class=switch>
```
ricea: The need to unpack the strategy argument is unfortunate. If the [reference in HTTP-Network-Fetch](https://fetch.spec.whatwg.org/#ref-for-concept-construct-readablestream%E2%91%A0) is the only usage, I would prefer to change that from the (scary) phrasing:

> Let _strategy_ be an object. The user agent may choose any object.

to something like

> Let _highWaterMark_ be a be a non-negative, non-NaN number, chosen by the user agent.
> Let _sizeAlgorithm_ be an algorithm that accepts chunk objects and returns a number, chosen by the user agent.

fetch.bs
```diff
@@ -1876,18 +1876,45 @@ with given <var>strategy</var>, <var>pull</var> action and <var>cancel</var> act
...
 
- <li><p>Set <var>init</var>["pull"] to a function that runs <var>pull</var> if <var>pull</var> is
- given.
+ <li><p>Let <var>pullAlg</var> be an algorithm that returns:
 
- <li><p>Set <var>init</var>["cancel"] to a function that runs <var>cancel</var> if
- <var>cancel</var> is given.
+ <dl class=switch>
+ <dt>If <var>pull</var> is given
+ <dd>the result of [=promise-calling=] <var>pull</var>().
```
ricea: @annevk usually objects to using the promises guide. However, I don't see much alternative except for copying the phrasing that the promises guide uses.

ricea:
This is not so terrible. It seems pretty close to usable to me.

Some more bad news: the difference is actually observable if `Object.prototype.start` or `Object.prototype.type` is set. This is one motivation for moving to `CreateReadableStream()`. Someone needs to write a web platform test for this. I can do it if there are no other volunteers.

fetch.bs
jyasskin: Yep. I didn't do that because I'm nervous about other callers outside Fetch, but I could leave a note here about the previous state and expect them to update.

jyasskin: I've done this and added the note.

ricea:
Okay, I will write the web platform test tomorrow.

This structurally looks good to me. I'm not sure whether we can leave a note in there. Last time I checked, the only spec relying on this was the service worker one. Maybe we can just file an issue there to update?

I think the markup will need some nitpicking, which I will leave to @annevk.

ricea:
PR for the web platform test: https://github.com/web-platform-tests/wpt/pull/12350

fetch.bs
domfarolino: Very out-of-context question: What is the reasoning for objection to using the promises guide? Haven't read all of it, just curious (sorry this is not the best place).

ricea: @domfarolino Possibly https://github.com/w3ctag/promises-guide/issues/52.

I know of one trap in there: because [aggregating promises](https://www.w3.org/2001/tag/doc/promises-guide#aggregating-promises) uses `Promise.all()` it calls the current (_not_ original) value of `Promise.reject()` which may have been replaced. This makes it unsafe to use in algorithms that are not re-entrant or are otherwise vulnerable to unexpected JS execution.There may be other traps I don't know about.

Having said that, we use it in the streams standard and I think it is fine if you use it with care.

ricea:
PR to update the ServiceWorker standard to match: https://github.com/w3c/ServiceWorker/pull/1338.

domenic:
Hmm yeah I didn't quite get the named arguments thing before. I dunno, this is OK-ish for now... Definitely a big improvement over the previous situation. I think the whole area of how streams interfaces with other specs needs a re-think, but I've been putting that off forever.

Also worth noting that, it seems like the idea of this wrapper, versus just using CreateReadableStream directly, is:

- Omits startAlgorithm
- Uses named arguments so that pullAlgorithm, and cancelAlgorithm become optional
- Allows pullAlgorithm and cancelAlgorithm to return non-promises

annevk:
Thanks @ricea, that looks good (modulo Oxford comma). Thoughts on my nit commit?

I guess I'm okay with merging this given @domenic's feedback.

jakearchibald:
I'm making use of streams in background fetch, eg step 7 of https://wicg.github.io/background-fetch/#create-settled-fetches-algorithm, which is streaming an item from storage that may still be writing from another algorithm.

It felt like I wanted to kind of spec-level stream concept, but I haven't thought too much about it.

Anyway, the changes here don't disrupt background-fetch. The change is pretty easy.

ricea:
The named arguments don't bother me. I think this is a clear improvement on the status quo.

As @domenic said, longer term we have to figure out a better way for other standards to talk about streams. I think this will tide us over in the meantime.

PR #785, opened 2018-08-10
Copy priority so service worker pass-through requests don't lose it

jakearchibald:
H2 priorities are lost when passing through a service worker. https://medium.baqend.com/chromes-service-workers-break-http-2-priorities-649c4e0fa930

Although priorities are hand-wavy right now, it seems right to copy them from one request to another.

***
<a href="https://whatpr.org/fetch/785.html" title="Last updated on Aug 15, 2018, 3:06 PM GMT (f477930)">Preview</a> | <a href="https://whatpr.org/fetch/785/28f06ba...f477930.html" title="Last updated on Aug 15, 2018, 3:06 PM GMT (f477930)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5451,11 +5451,11 @@ constructor must run these steps:
    <dt><a for=request>window</a>
    <dd><var>window</var>.
 
-   <dt><a for=request>origin</a>
-   <dd>"<code>client</code>".
-
    <dt><a for=request>priority</a>
    <dd><var>request</var>'s <a for=request>priority</a>.
+
+   <dt><a for=request>origin</a>
+   <dd>"<code>client</code>".
 
    <dt><a for=request>referrer</a>
    <dd><var>request</var>'s <a for=request>referrer</a>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -5454,6 +5454,9 @@ constructor must run these steps:
    <dt><a for=request>origin</a>
    <dd>"<code>client</code>".
 
+   <dt><a for=request>priority</a>
+   <dd><var>request</var>'s <a for=request>priority</a>.
+
    <dt><a for=request>referrer</a>
    <dd><var>request</var>'s <a for=request>referrer</a>.
 
@@ -5483,9 +5486,6 @@ constructor must run these steps:
 
    <dt><a for=request>history-navigation flag</a>
    <dd><var>request</var>'s <a for=request>history-navigation flag</a>.
-
-   <dt><a for=request>priority</a>
-   <dd><var>request</var>'s <a for=request>priority</a>.
   </dl>
 
  <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -5484,8 +5484,8 @@ constructor must run these steps:
    <dt><a for=request>history-navigation flag</a>
    <dd><var>request</var>'s <a for=request>history-navigation flag</a>.
 
-    <dt><a for=request>priority</a>
-     <dd><var>request</var>'s <a for=request>priority</a>.
+   <dt><a for=request>priority</a>
+   <dd><var>request</var>'s <a for=request>priority</a>.
   </dl>
 
  <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -5483,6 +5483,9 @@ constructor must run these steps:
 
    <dt><a for=request>history-navigation flag</a>
    <dd><var>request</var>'s <a for=request>history-navigation flag</a>.
+
+    <dt><a for=request>priority</a>
+     <dd><var>request</var>'s <a for=request>priority</a>.
   </dl>
 
  <li>


ErikWitt:
It would be great if priorities could be copied from one request to another even when modifying the URL of the request in the Service Worker.

Maybe pass the original request in the init object of Request() and copy its priority or something like this?

```
addEventListener('fetch', event => {
  const newURL = new URL(event.request.url);
  newURL.searchParams.append('fromSW', 1);
  
  const newRequest = new Request(newURL, { request: event.request });

  event.respondWith(fetch(newRequest))
});
```

Maybe considering the request destination when assigning a new priority could help also. Unfortunately, the destination is read-only :(

wanderview:
Note, there is an existing gecko bug about this from when we discussed it before:

https://bugzilla.mozilla.org/show_bug.cgi?id=1415003

Do we have adequate WPT infrastructure to write a test for this?

domfarolino:
I don’t think so. Right now a UA can prioritize requests however it would like, and since priorities aren’t currently exposed via any object, it seems impossible to “detect” priority in action without making non-normative assumptions.

wanderview:
Well, if we could inspect the h2 priority in WPT we might be able to write a test that shows that it doesn't change whether you go through a service worker or not.  That would test this PR and the issue being observed by web developers, but not require an exact priority value.

domfarolino:
Ah good call, sorry for the noise. Not entirely sure who the best person to ask would be. @foolip?

foolip:
https://github.com/web-platform-tests/wpt/issues/12236 is about HTTP/2.0 support in WPT. Feedback there on your testing needs might be good, and perhaps @dhdavvie can take a look at this issue?

domfarolino:
After taking a closer look at Chrome's implementation and the spec, I'm wondering if copying over [priority](https://fetch.spec.whatwg.org/#concept-request-priority) is enough? We never explicitly set the priority concept, instead we just tell the implementation to generate it, given [destination](https://fetch.spec.whatwg.org/#concept-request-destination) and [initiator](https://fetch.spec.whatwg.org/#concept-request-initiator), in [concept fetch](https://fetch.spec.whatwg.org/#concept-fetch) > 1 > 5.

With this, I'm thinking we might not solve the problem at hand by copying over priority, because the destination and initiator will not be preserved, and the priority will just be re-generated in the later concept fetch steps to what it would have been if we never preserved priority. I'm wondering if we'd be better off copying over the destination and initiator, but we should also confirm that this doesn't break anything.

----
With the current PR, imagine the steps below in action:

1. We [obtain](https://html.spec.whatwg.org/multipage/semantics.html#obtaining-a-resource-from-a-link-element) a `<link rel=preload as=style href=dom.css>` (which explicitly sets the request's destination)
2. The request is fetched, and its priority is generated in [fetch](https://fetch.spec.whatwg.org/#concept-fetch) > 1 > 5
3. The request goes through a Service Worker that does `event.respondWith(fetch(event.request))`
4. Since [`fetch()`](https://fetch.spec.whatwg.org/#fetch-method) will call the Request() constructor, we'll lose our destination and initiator concepts, but preserve our priority
5. [`fetch()`](https://fetch.spec.whatwg.org/#fetch-method) calls [fetch](https://fetch.spec.whatwg.org/#concept-fetch) (ikr) and goes back through steps 1 > 5, resetting our preserved priority, given the default destination ("") and initiator ("")
6. It is as if we never saved our priority

Disclaimer: entirely possible I'm missing something here

fetch.bs
```diff
@@ -5483,6 +5483,9 @@ constructor must run these steps:
 
    <dt><a for=request>history-navigation flag</a>
    <dd><var>request</var>'s <a for=request>history-navigation flag</a>.
+
+   <dt><a for=request>priority</a>
+   <dd><var>request</var>'s <a for=request>priority</a>.
```
annevk: Ideally we keep these in a similar order as the order they are introduced in, so this should be higher up. Otherwise I think this change looks fine.

dhdavvie:
Priority checking/inspection would definitely be valuable functionality in the WPT HTTP/2.0 server. The library that the server uses does not currently expose this easily, but it is present! I will do some digging and see if I can modify the library a bit to expose this to us :) Thanks for the tag @foolip

jakearchibald:
I've moved the definition up a bit, although the current order doesn't really match up to the request fields.

annevk:
It should be before origin, as far as I can tell, but you're right that it doesn't entirely match up. Sigh.

annevk:
I've landed without tests as the test infrastructure for this is not in place yet. It seems we have sufficient implementation bugs filed as well here?

wanderview:
> It seems we have sufficient implementation bugs filed as well here?

I think we still need bugs for edge and webkit.

annevk:
Ah, yes:

* https://bugs.webkit.org/show_bug.cgi?id=188650
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18695705/

PR #788, opened 2018-08-16
Don't append controls for upstream caches.

### LINKED ISSUE #722: Cache-mode and upstream Cache-Control
[HTTP-network-or-cache-fetch](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch) says:

> 12. If httpRequest’s cache mode is "default" and httpRequest’s header list contains `If-Modified-Since`, `If-None-Match`, `If-Unmodified-Since`, `If-Match`, or `If-Range`, then set httpRequest’s cache mode to "no-store".

The following text was added to fix #201, #231 and #232 (AFAICT):

> 13. If httpRequest’s cache mode is "no-cache" and httpRequest’s header list does not contain `Cache-Control`, then append `Cache-Control`/`max-age=0` to httpRequest’s header list.
>
> 14. If httpRequest’s cache mode is "no-store" or "reload", then:
> 
> >  1. If httpRequest’s header list does not contain `Pragma`, then append `Pragma`/`no-cache` to httpRequest’s header list.
> >
> >  2. If httpRequest’s header list does not contain `Cache-Control`, then append `Cache-Control`/`no-cache` to httpRequest’s header list.

This means that in several situations when the developer is trying to influence the local cache behaviour, it triggers request `Pragma` and `Cache-Control` headers that effectively disable upstream caches, even though this might not be desired.

In particular, setting `If-Modified-Since`, `If-None-Match` and other validating headers doesn't need to disable upstream caches; it's an entirely reasonable expectation for the upstream cache to be able to serve the request as normal. As specified, this is only an incentive for upstream caches to ignore these request directives, which isn't good for interop.

Even without validation, a developer might wish to bypass the browser cache with `no-store` but still use upstream caches (perhaps because they have a reliable out-of-band way to invalidate those caches, e.g., in a CDN).

I think the best solution here would be to remove (13) and (14), or to provide a separate control over these behaviours (e.g., an init flag to turn it off), so that developers have as much flexibility as they need; as specified, fetch() is making a decision for them. 

If not, at a minimum the [documentation for cache mode](https://fetch.spec.whatwg.org/#concept-request-cache-mode) needs to be updated to indicate that it is intended to influence not only the local cache, but also upstream caches.

P.S. I've tested to see if these headers are actually added, by making `no-store` requests. From what I can see:

* Chrome doesn't append `Pragma: no-cache` or `Cache-Control: no-cache`
* Firefox only appends `Pragma: no-cache`, and ignores the "If httpRequest's header list does not contain `Pragma`" condition (i.e, it appends it even if `Pragma` is present)
* Safari appends both if they aren't already present

mnot:
The developer may want to control the local cache separately from those
upstream; if they want to control both, they can append their own
request headers.

Fixes #722.

***
<a href="https://whatpr.org/fetch/788.html" title="Last updated on Jan 15, 2021, 7:39 AM UTC (7cb2a7b)">Preview</a> | <a href="https://whatpr.org/fetch/788/7e088f6...7cb2a7b.html" title="Last updated on Jan 15, 2021, 7:39 AM UTC (7cb2a7b)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3699,29 +3699,6 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    `<code>If-Range</code>`, then set <var>httpRequest</var>'s
    <a for=request>cache mode</a> to "<code>no-store</code>".
 
-   <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>no-cache</code>" and
-   <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">does not contain</a>
-   `<code>Cache-Control</code>`, then <a for="header list">append</a>
-   `<code>Cache-Control</code>`/`<code>max-age=0</code>` to
-   <var>httpRequest</var>'s <a for=request>header list</a>.
-
-   <li>
-    <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>no-store</code>" or
-    "<code>reload</code>", then:
-
-    <ol>
-     <li><p>If <var>httpRequest</var>'s <a for=request>header list</a>
-     <a for="header list">does not contain</a> `<code>Pragma</code>`, then
-     <a for="header list">append</a> `<code>Pragma</code>`/`<code>no-cache</code>` to
-     <var>httpRequest</var>'s <a for=request>header list</a>.
-
-     <li><p>If <var>httpRequest</var>'s <a for=request>header list</a>
-     <a for="header list">does not contain</a> `<code>Cache-Control</code>`, then
-     <a for="header list">append</a> `<code>Cache-Control</code>`/`<code>no-cache</code>` to
-     <var>httpRequest</var>'s <a for=request>header list</a>.
-     <!-- Technically this only applies to HTTP/1.1 and up -->
-    </ol>
-
    <li>
     <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">contains</a>
     `<code>Range</code>`, then <a for="header list">append</a>


annevk:
@mnot you or your employer will have to sign the agreement: https://participate.whatwg.org/agreement. I'm guessing Fastly would have to sign it since it's in the business of web technologies.

It'd be good to have test coverage for this as well and implementation bugs.

mnot:
Working on both, thanks.

mayhemer:
Just to make this request clear - the request.cache modes should influence only the local cache now?

Note that request Cache-control headers influence local cache as well in Gecko:
https://searchfox.org/mozilla-central/rev/847b64cc28b74b44c379f9bff4f415b97da1c6d7/netwerk/protocol/http/nsHttp.cpp#436-454

mnot:
@annevk seems like the verification finally got recognised by the tools; take another look?

annevk:
Apologies for the late reply, I think the PR looks okay, but the point @mayhemer raises above is important. These headers are the way Firefox deals with the local cache. Other implementations might have a similar setup. That's why we need a set of tests and implementer agreement to implement the division as you would like to see it.

mnot:
So, I think to address that, we'd need to include text that explicitly bypasses the local cache (etc.) when these flags are set, no?

annevk:
I think that's mostly covered already as https://fetch.spec.whatwg.org/#concept-http-network-fetch is pretty explicit when we finally get to making a request: "Set response to the result of making an HTTP request over connection using request with the following caveats". I suppose we could further clarify there though that the HTTP cache is not involved and only consulted during https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch. But we could do that independently from this change I think.

To restate my comment above, I think the PR as-is is okay; the problem is with existing implementations in browsers, inadequate test coverage for the change, and lack of clarity whether implementations are prepared to align on this. (Reading #722 again it sounds like both Chrome and Firefox would have difficulty adopting this change and are not necessarily prepared to go through the trouble.)

Richyraymond (commented):
Can anyone please test this

Richyraymond (approved):

lastthyme (approved):

PR #791, opened 2018-08-16
Clarify that the `ping` attribute is for <a>

tomayac:
***
<a href="https://whatpr.org/fetch/791.html" title="Last updated on Aug 17, 2018, 9:22 AM GMT (ffcb30f)">Preview</a> | <a href="https://whatpr.org/fetch/791/e2d62ff...ffcb30f.html" title="Last updated on Aug 17, 2018, 9:22 AM GMT (ffcb30f)">Diff</a>


annevk:
Well, it's also for `area`, right?

tomayac:
#TIL, it's not mentioned in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area), but it appears in the [spec](https://html.spec.whatwg.org/multipage/image-maps.html#the-area-element). Interestingly, among the [other Links](https://html.spec.whatwg.org/multipage/links.html#links), it [cannot](https://html.spec.whatwg.org/multipage/semantics.html#the-link-element) occur in `<link>` (which might be an oversight?).

So to summarize, `a[ping]` and `area[ping]` are a thing, but `link[ping]` isn't. Is this correct?

annevk:
I don't think that was an oversight, but I'd have to find all the old discussions. And yes, you're correct.

tomayac:
Cool, I have added `<area>` to the Pull Request.

tomayac:
(Fixed in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area$compare?locale=en-US&to=1405208&from=1405127), added `area[ping]`.)

annevk:
I was about to merge this, but Thomas Steiner isn't yet in the Acknowledgments section. Would you like to add yourself? (Please pull first to take my nit into account.)

tomayac:
Done. https://github.com/whatwg/fetch/pull/793.

annevk:
@tomayac you are still planning on adding a commit here, right?

annevk:
Thanks!

PR #801, opened 2018-08-21
Make extract throw for a disturbed/locked ReadableStream

### LINKED ISSUE #792: It is possible to construct a Request with a used body
Given some disturbed but unlocked ReadableStream `rs`, the following spec-conformant code will construct a Request object with a disturbed body:

```javascript
let rs = // a disturbed, unlocked ReadableStream
let request = new Request(url, { method: "POST", body: rs })
```

This works because Request's constructor, `Request(input, init)`, only throws if `input.body` is disturbed, but does not throw if `init.body` is disturbed.

This seems a little strange, because it means iteratively applying Request's constructor could throw upon the second invocation. An example would be in `fetch()`, which is required by the spec to pass its arguments to Request's constructor:

```javascript
// This will NOT throw, because fetch() invokes new Request(input, init).
await fetch(url, { method: "POST", body: rs })

// This WILL throw, because fetch() invokes new Request(input).
let request = new Request(url, { method: "POST", body: rs })
await fetch(request)
```

Should this be considered a bug?

annevk:
This also introduces "safely extract" which mainly serves to make it explicit which extraction operations cannot throw.

Tests: ...

Fixes #792.

***
<a href="https://whatpr.org/fetch/801.html" title="Last updated on Aug 23, 2018, 6:47 AM GMT (eeadb52)">Preview</a> | <a href="https://whatpr.org/fetch/801/5088fce...eeadb52.html" title="Last updated on Aug 23, 2018, 6:47 AM GMT (eeadb52)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5764,10 +5764,13 @@ constructor must run these steps:
    <a for=request>use-CORS-preflight flag</a>.
   </ol>
 
+ <li><p>If <var>inputBody</var> is <var>body</var> and <var>input</var> is <a for=Body>disturbed</a>
+ or <a for=Body>locked</a>, then <a>throw</a> a {{TypeError}}.
+
  <!-- Any steps after this must not throw. -->
 
  <li>
-  <p>If <var>inputBody</var> is non-null and <var>inputBody</var> is <var>body</var>, then:
+  <p>If <var>inputBody</var> is <var>body</var> and <var>inputBody</var> is non-null, then:
 
   <ol>
    <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -3585,11 +3585,11 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
 
  <li>
   <p>If <var>request</var>'s <a for=request>body</a> is non-null, then set <var>request</var>'s
-  <a for=request>body</a> to the first part of <a lt=extract for=BodyInit>extracting</a>
+  <a for=request>body</a> to the first part of <a for=BodyInit>safely extracting</a>
   <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
 
   <p class="note no-backref"><var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>'s
-  nullity has already been checked. The <a lt=extract for=BodyInit>extracting</a> operation cannot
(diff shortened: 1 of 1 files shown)


annevk:
As @harrishancock noted this will also make `new Response()` throw. Hopefully this is not an issue.

I'm a little surprised that the notes for the two (now called) "safely extract" operations already anticipated throwing behavior and explain why that won't happen. Perhaps "extract" should have thrown all along after all? Reading through #425 doesn't provide much clues. Any ideas @yutakahirano?

In any event, review appreciated.

fetch.bs
```diff
@@ -5746,9 +5764,6 @@ constructor must run these steps:
    <a for=request>use-CORS-preflight flag</a>.
   </ol>
 
- <li><p>If <var>inputBody</var> is <var>body</var> and <var>input</var> is <a for=Body>disturbed</a>
- or <a for=Body>locked</a>, then <a>throw</a> a {{TypeError}}.
```
harrishancock: Don't we still need this? `input.body` is not itself extracted, only `init.body`, so the change would only affect `init.body`.

annevk: Oh, I blindly trusted https://github.com/whatwg/fetch/issues/792#issuecomment-413898585. I will double check.

harrishancock: Sorry about that! I should have double checked before writing that.

yutakahirano:
LGTM. I think it's still valuable because now it is clear that adding logic which makes "safe extract" throw to extract is wrong. Perhaps removing the notes from https://fetch.spec.whatwg.org/#concept-http-redirect-fetch and  https://fetch.spec.whatwg.org/#http-network-or-cache-fetch, and adding one to "safe extract" would be good.

annevk:
Thank you, I've adjusted the notes.

annevk:
I created tests, review of those and the final text here would be appreciated.

annevk:
@yutakahirano just to be sure, this all looks good to you?

yutakahirano:
Still LGTM.

annevk:
Bugs:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1486698
* https://bugs.webkit.org/show_bug.cgi?id=189042
* https://bugs.chromium.org/p/chromium/issues/detail?id=878281
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18818191/

Issue #752, opened 2018-06-01
What does "combined value" return for a name not in the header list?


jyasskin:
> A [combined value](https://fetch.spec.whatwg.org/#concept-header-value-combined), given a name (name) and header list (list), is the values of all headers in list whose name is a byte-case-insensitive match for name, separated from each other by 0x2C 0x20, in order.

* The call in https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine only calls it for names in the header list. That'd match an "Assert: There are some values."
* The call in https://fetch.spec.whatwg.org/#dom-headers-get explicitly returns null for a non-contained name. That'd match having "combined value" return null.
* If we interpret the specification as the Javascript `[values].join(", ")`, we'd get `""` for no-values.

I'd appreciate specifying one of these so I don't accidentally write a use that's ambiguous.

annevk:
The current setup requires you to do a contains check (which maybe we should rename to has?).

I want to add "get" to https://fetch.spec.whatwg.org/#concept-header-list that gives you null or a byte sequence, but I'm not a 100% sure if the current "combined value" definition is good enough.

In particular, I think @mnot's structured headers don't have the additional space. So:
```
Some-String-Header: "test
Some-String-Header: test"
```
ends up with `"test,test"` and not `"test, test"` as Fetch (and believe most implementations) do.

annevk:
@domenic was also interested in this.

mnot:
SH doesn't specify how to combine headers; HTTP does. Currently it doesn't have advice about this, but if we did enough testing, perhaps we could get some in.

SH *could* specify how to combine headers when the actual SH processor is doing it; do you think that would be useful?

annevk:
See also #757 about us currently not excluding empty values whereas HTTP does do that. (I don't think browsers do that universally though.)

annevk:
@mnot I don't think it makes sense for SH to define it. SH should just operate on combined values and leave combining to HTTP (or Fetch).

annevk:
#818 has a fix for this in the form of a "get" operation.

PR #813, opened 2018-10-11
Change combined values

### LINKED ISSUE #757: "combined value" is wrong
It needs to exclude values that are the empty byte sequence.

annevk:
This marks XMLHttpRequest's approach in setRequestHeader() for combining values as naïve (it doesn't follow the spirit of HTTP, but is technically allowed) and changes "combined value" to more closely match HTTP semantics.

Tests: https://github.com/web-platform-tests/wpt/pull/13471.

Fixes #757. (Addresses #752 to some extent, but leaving that open for a better API.)

***
<a href="https://whatpr.org/fetch/813.html" title="Last updated on Oct 11, 2018, 3:23 PM GMT (59e13ed)">Preview</a> | <a href="https://whatpr.org/fetch/813/daca6a8...59e13ed.html" title="Last updated on Oct 11, 2018, 3:23 PM GMT (59e13ed)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -335,22 +335,6 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
  <var>name</var> and <a for=header>value</a> is <var>value</var> to <var>list</var>.
 </ol>
 
-<p>To <dfn export for="header list" id=concept-header-list-combine>combine</dfn> a
-<a for=header>name</a>/<a for=header>value</a> (<var>name</var>/<var>value</var>) pair in a
-<a for=/>header list</a> (<var>list</var>), run these steps:
-
-<ol>
- <li><p>If <var>list</var> <a for="header list">contains</a> <var>name</var>, then set the
- <a for=header>value</a> of the first such <a for=/>header</a> to its <a for=header>value</a>,
- followed by 0x2C 0x20, followed by <var>value</var>.
-
- <li><p>Otherwise, <a for=list>append</a> a new <a for=/>header</a> whose <a for=header>name</a> is
- <var>name</var> and <a for=header>value</a> is <var>value</var> to <var>list</var>.
-</ol>
-
-<p class="note no-backref"><a for="header list">Combine</a> is used by {{XMLHttpRequest}} and the
-<a lt="establish a WebSocket connection">WebSocket protocol handshake</a>.
-
 <p>To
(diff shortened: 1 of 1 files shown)


annevk:
The other thing I'm interested in is if we should use this same definition when defining parsers for particular headers, such as `Content-Type`, or if we should use a version that does not include the space. This ends up mattering a bit in edge cases unfortunately (double quotes split across headers). I guess ideally we'd fix the API here, but ugh.

@wisniewskit I suppose you might have thoughts on this too.

MattMenke2:
I'm concerned about ignoring empty response header fields.  Something like:

Content-Length:
Content-Length: 10

Seems more like something that should be rejected than treated as the logical equivalent to "Content-Length: 10".  Seems like allowing it would potentially allow for more shenanigans around deliberately tricking (Or accidentally confusing) proxies and security software.

annevk:
An intermediary that combines would make that `Content-Length: 10` and you wouldn't be wiser. Unless we change the HTTP standard I'm not sure there's much we can do there?

Note also that `Content-Length: ,10` (not the combined value, I know, but interesting) ends up being treated as 10. And `10,` too! (Arguably we could/should fix these.)

Given that HTTP combining is a thing, I'm not sure what alternative strategy we have. Not ignore empty/whitespace values and contradict the HTTP specification?

annevk:
Sigh, I got misled by https://tools.ietf.org/html/rfc7230#section-7 and the discussion at https://github.com/httpwg/http-extensions/issues/596#issuecomment-394300296. But that should happen at the level of parsing individual headers, it shouldn't happen at this level. My bad, this is a bug in Firefox.

Thanks a lot for catching this!

annevk:
> Under what pre-existing spec is "Content-Length: ,10" the same as "Content-Length: 10"? Am I missing something?

No spec, but this is what browsers do at the moment. (Whereas two separate headers does result in a network error.)

MattMenke2:
> > Under what pre-existing spec is "Content-Length: ,10" the same as "Content-Length: 10"? Am I missing something?
> 
> No spec, but this is what browsers do at the moment. (Whereas two separate headers does result in a network error.)

I'm surprised that Chrome does this, but you're absolutely correct - deep within Chrome's parser is code that just throws away empty HTTP response headers.

Issue #814, opened 2018-10-12
Redesign "extract header values" and "extract header list values"


annevk:
These APIs had the wrong design and are also incorrectly used at times (#559). "Extract a MIME type" also needs a new design, likely on top of this as a new `Content-Type` parser: #529.

In particular, other than for `Set-Cookie` the ideal way to parse a header is to combine first, and then parse, which isn't really what the current operations are doing. How to combine headers is nearly settled: #813. Once that's done my tentative plan is to introduce a new "get" for header lists to solve #752 which would return null or the combined value for a given header name from a header list.

Another problem with the current setup is that the actual parsing is handwavy by referencing ABNF defined "somewhere", which in practice leads to it being poorly tested. We can continue to allow ABNF for headers where that is actually how they are implemented, but we need to be more strict about referencing the ABNF and writing the tests and not allowing as much indirection as is currently allowed for.

Unfortunately this will impact downstream, but I think it's the right thing to do long term and will lead to better considered header parsers (and encourage more reuse as well). I'll also start filing downstream issues now linking this issue to ensure we actually design something that works for everyone.

So I think what we want is basically "get" and then per header name we'll need to define its parser (which could just be an ANBF reference plus tests) or parsers (`Content-Type` differs between request and response, yay). Thoughts on the details here appreciated as well as on the overall plan of course.

annevk:
In #818 @mikewest suggested that perhaps it'd be useful if there was a way to always get the "string" value from a header. So basically "get" on a header list and then return the result of isomorphic decoding that (if non-null). We could add "get as string" for this provided there's more than one consumer, which seems likely if we'll continue to define most parsers on top of strings.

annevk:
Things to define anew within Fetch (and test) to get rid of "extract header values":

- [ ] request Content-Type (CORS-safelisted request-header)
- [ ] request DPR/Downlink/Save-Data/Viewport-Width/Width (CORS-safelisted request-header)
- [ ] X-Content-Type-Options (see #818 for a patch)

For "extract header list values":

- [ ] response Content-Type (see https://github.com/whatwg/mimesniff/issues/30)
- [ ] Access-Control-Expose-Headers
- [ ] Location
- [ ] Content-Encoding (see #816 for removing a quirk, #716 for the remainder)
- [ ] Access-Control-Allow-Methods
- [ ] Access-Control-Allow-Headers
- [ ] Access-Control-Max-Age
[...]

Issue #817, opened 2018-10-17
Editorial: drop parenthesis around variables


annevk:
The header list section does this and it looks wrong now. Type followed by variable name is fine and there's no need for any parenthesis.

PR #818, opened 2018-10-17
Define parsing for X-Content-Type-Options in detail

### LINKED ISSUE #752: What does "combined value" return for a name not in the header list?

annevk:
And add some of the infrastructure needed to define parsing better for all headers going forward (needed for #814). Fixes #752.

This also fixes an issue with CORB as it simply assumed an X-Content-Type-Options was present.

Tests: https://github.com/web-platform-tests/wpt/pull/13559.

***
<a href="https://whatpr.org/fetch/818.html" title="Last updated on Oct 31, 2018, 9:14 AM GMT (3515d9b)">Preview</a> | <a href="https://whatpr.org/fetch/818/daca6a8...3515d9b.html" title="Last updated on Oct 31, 2018, 9:14 AM GMT (3515d9b)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -236,12 +236,14 @@ implementations can store the <a for=url>fragment</a> nonetheless.
 borrows a number of concepts from HTTP and applies these to resources obtained via other
 means (e.g., <code>data</code> URLs).
 
-<p><dfn export>HTTP whitespace</dfn> is U+0009 TAB, U+000A LF, U+000D CR, or U+0020 SPACE.
+<p>An <dfn export>HTTP tab or space</dfn> is U+0009 TAB or U+0020 SPACE.
 
-<p class=note>As opposed to <a>ASCII whitespace</a> this excludes U+000C FF. It is generally
-preferable to use <a>ASCII whitespace</a>, except when defining parsers for HTTP header values.
+<p>An <dfn export>HTTP newline byte</dfn> is 0x0A (LF) or 0x0D (CR).
 
-<p>The <dfn export>HTTP whitespace bytes</dfn> are 0x09, 0x0A, 0x0D, and 0x20.
+<p>An <dfn export>HTTP tab or space byte</dfn> is 0x09 (HT) or 0x20 (SP).
+
+<p>An <dfn export>HTTP whitespace byte</dfn> is an <a>HTTP newline byte</a> or
+<a>HTTP tab or space byte</a>.
 
 <p>An <dfn export id=concept-https-state-value>HTTPS state value</dfn> is "<code>none</code>",
 "<code>deprecated</code>", or "<code>modern</code>".
@@ -404,8 +406,8 @@ token production.
 <p>A <a for=header>value</a> is a <a>byte sequence</a> that matches the following conditions:
 
 <ul class=brief>
- <li><p>Has no leading or trailing <a>HTTP whitespace bytes</a>.
- <li><p>Contains no 0x00, 0x0A or 0x0D bytes.
+ <li><p>Has no leading or trailing <a>HTTP tab or space bytes</a>.
+ <li><p>Contains no 0x00 (NUL) or <a>HTTP newline bytes</a>.
 </ul>
 
 <p class=note>The definition of <a for=header>value</a> is not defined in terms of an HTTP token
@@ -2527,7 +2529,7 @@ steps:
  <li><p>Let <var>tokens</var> be the result of
  <a lt="strictly split a string">strictly splitting</a> <var>stringValue</var> on U+002C (,).
 
- <li><p>Let <var>firstToken</var> be the result of removing all <a>HTTP whitespace</a> from the
+ <li><p>Let <var>firstToken</var> be the result of removing all <a>HTTP tab or space</a> from the
  start and end of <var>tokens</var>[0].
 
  <li><p>If <var>firstToken</var> is an <a>ASCII case-insensitive</a> match for
@@ -6659,6 +6661,7 @@ Anssi Kostiainen,
 Arkadiusz Michalski,
 Arne Johannessen,
 Arthur Barstow,
+Asanka Herath,
 Axel Rauschmayer,
 Ben Kelly,
 Benjamin Gruenbaum,
--- a/fetch.bs
+++ b/fetch.bs
@@ -236,6 +236,11 @@ implementations can store the <a for=url>fragment</a> nonetheless.
 borrows a number of concepts from HTTP and applies these to resources obtained via other
 means (e.g., <code>data</code> URLs).
 
+<p><dfn export>HTTP whitespace</dfn> is U+0009 TAB, U+000A LF, U+000D CR, or U+0020 SPACE.
+
+<p class=note>As opposed to <a>ASCII whitespace</a> this excludes U+000C FF. It is generally
+preferable to use <a>ASCII whitespace</a>, except when defining parsers for HTTP header values.
+
 <p>The <dfn export>HTTP whitespace bytes</dfn> are 0x09, 0x0A, 0x0D, and 0x20.
 
 <p>An <dfn export id=concept-https-state-value>HTTPS state value</dfn> is "<code>none</code>",
@@ -2519,14 +2524,13 @@ steps:
 
  <li><p>Let <var>stringValue</var> be the <a>isomorphic decode</a> of <var>value</var>.
 
- <li>
-  <p>Let <var>tokens</var> be the result of
-  <a lt="split on commas">spliting <var>stringValue</var> on commas</a>.
+ <li><p>Let <var>tokens</var> be the result of
+ <a lt="strictly split a string">strictly splitting</a> <var>stringValue</var> on U+002C (,).
 
-  <p class="note">This intentionally strips U+000C FORM FEED, despite 0x0C not being being a
-  <a>HTTP whitespace byte</a>.
+ <li><p>Let <var>firstToken</var> be the result of removing all <a>HTTP whitespace</a> from the
+ start and end of <var>tokens</var>[0].
 
- <li><p>If <var>tokens</var>[0] is an <a>ASCII case-insensitive</a> match for
+ <li><p>If <var>firstToken</var> is an <a>ASCII case-insensitive</a> match for
  "<code>nosniff</code>", then return true.
 
  <li><p>Return false.
--- a/fetch.bs
+++ b/fetch.bs
@@ -2517,11 +2517,11 @@ steps:
 
  <li><p>If <var>value</var> is null, then return false.
 
- <li><p>Let <var>stringValue</var> be the <a>isomorphic encode</a> of <var>value</var>.
+ <li><p>Let <var>stringValue</var> be the <a>isomorphic decode</a> of <var>value</var>.
 
  <li>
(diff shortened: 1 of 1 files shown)


annevk:
The only somewhat icky bit here in my opinion is using "isomorphic encode" to avoid the lack of byte sequence manipulation infrastructure. It seems fine for now, but worth keeping an eye on.

I also filed a follow-up, unrelated to that: #817.

cc @MattMenke2

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
...
+ <li><p>Let <var>value</var> be the result of <a for="header list">getting</a>
+ `<a http-header><code>X-Content-Type-Options</code></a>` from <var>list</var>.
+
+ <li><p>If <var>value</var> is null, then return false.
+
+ <li><p>Let <var>stringValue</var> be the <a>isomorphic encode</a> of <var>value</var>.
+
+ <li>
+  <p>Let <var>tokens</var> be the result of
+  <a lt="split on commas">Spliting <var>stringValue</var> on commas</a>.
```
domenic: lowercase "s"

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
...
+ "<code>nosniff</code>", then return true.
+
+ <li><p>Return false.
+</ol>
+
+<p>Web developers and conformance checkers must use the following <a for=header>value</a>
+<a>ABNF</a> for `<a http-header><code>X-Content-Type-Options</code></a>`:
 
 <pre>
 X-Content-Type-Options           = "nosniff" ; case-insensitive</pre>
```
domenic: Why are there so many spaces here? (Preexisting problem)

fetch.bs
```diff
@@ -2578,14 +2605,10 @@ run these steps:</p>
...
- <li><p>Let <var>nosniff</var> be the result of <a>extracting header values</a> from the
- <em>first</em> <a for=/>header</a> whose <a for=header>name</a> is a <a>byte-case-insensitive</a>
- match for `<a http-header><code>X-Content-Type-Options</code></a>` in <var>response</var>'s
- <a for=response>header list</a>.
-
  <li>
-  <p>If <var>nosniff</var> is not failure and <var>mimeType</var> (ignoring parameters) is a
-  <a>CORB-protected MIME type</a> or <code>text/plain</code>, then return <b>blocked</b>.
+  <p>If <a>determine nosniff</a> with <var>response</var>'s <a for=response>header list</a> is true
+  and <var>mimeType</var> (ignoring parameters) is a <a>CORB-protected MIME type</a> or
```
domenic: "(ignoring parameters)" is technically redundant here as the definition of each of these MIME types is checking the essence. But, maybe best to leave in for clarity.

fetch.bs
annevk: I think it's to align with other ABNF pieces in Fetch, but not sure.

fetch.bs
annevk: Once we have `Content-Type` parsing defined (request and response) I'm hoping to clean a lot of this up.

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
...
+<p>To <dfn>determine nosniff</dfn>, given a <a for=/>header list</a> <var>list</var>, run these
+steps:
+
+<ol>
+ <li><p>Let <var>value</var> be the result of <a for="header list">getting</a>
+ `<a http-header><code>X-Content-Type-Options</code></a>` from <var>list</var>.
+
+ <li><p>If <var>value</var> is null, then return false.
+
+ <li><p>Let <var>stringValue</var> be the <a>isomorphic encode</a> of <var>value</var>.
```
mikewest: I wonder if this should be part of the generic `get` algorithm?

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
...
+
+ <li>
+  <p>Let <var>tokens</var> be the result of
+  <a lt="split on commas">Spliting <var>stringValue</var> on commas</a>.
+
+  <p class="note">This intentionally strips U+000C FORM FEED, despite 0x0C not being being a
+  <a>HTTP whitespace byte</a>.
+
+ <li><p>If <var>tokens</var>[0] is an <a>ASCII case-insensitive</a> match for
+ "<code>nosniff</code>", then return true.
```
mikewest: I think this is accurate: Chrome only looks at the first item (see https://cs.chromium.org/chromium/src/third_party/blink/renderer/platform/network/http_parsers.cc?rcl=80ce30280a48850aad465138b84fe209474733e8&l=384), and it looks like that's what's going on in Firefox as well (https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpChannel.cpp#1444).

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
 `<code>Content-Type</code>` <a for=/>header</a> against the <a for=request>destination</a> of a
 <a for=/>request</a>.
 
-<p>Its <a for=header>value</a> <a>ABNF</a>:
+<p>To <dfn>determine nosniff</dfn>, given a <a for=/>header list</a> <var>list</var>, run these
+steps:
```
mikewest: The usage below ("If <a>determine nosniff</a> with <var>response</var>'s <a for=response>header list</a> is
 false") reads a bit strangely. Perhaps we could reformulate this as "A given header list <dfn>enforces strict MIME type checking</dfn> if the following steps return `true`:", which would lead to "If <var>response</var>'s <a for=response>header list</a> <a>enforces strict MIME type checking</a>...", which reads more naturally to me.

fetch.bs
annevk: It should be isomorphic decode. I think for some header values we want to always decode using UTF-8?

annevk: I'd support adding a "get as string|getting as string" for header lists though to make this easier for simple cases.

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
...
+ <li><p>If <var>value</var> is null, then return false.
+
+ <li><p>Let <var>stringValue</var> be the <a>isomorphic decode</a> of <var>value</var>.
+
+ <li>
+  <p>Let <var>tokens</var> be the result of
+  <a lt="split on commas">spliting <var>stringValue</var> on commas</a>.
+
+  <p class="note">This intentionally strips U+000C FORM FEED, despite 0x0C not being being a
+  <a>HTTP whitespace byte</a>.
```
MattMenke2: Sorry, hadn't really focused on this.  What's the motivation here?  Historical reasons?  Looks like Chrome manages to not treat this byte as special, at least in the HTTP header parser.  If we don't absolutely need this, I'd rather get rid of it, just from a keeping things consistent standpoint.

annevk: Implementations are a little divergent here. I guess we can try to push for not stripping U+000C here and elsewhere.

MattMenke2: I'm rather reluctant to add this to Chrome, since it's inconsistent with handling of other headers, and we've gotten away without it for all these years.

I have a CL in progress to remover single quote handling in our general parser for things like 'foo=bar; baz="Jumbo, shrimp"', which was configurable so it could just recognize double quotes, or double-quotes or single quotes, which will make parsing of headers more consistent.  I think it would be unfortunate if we had to add a way of configuring it to treat U+000C as white space in some, but not all header-related contexts, taking a step in the opposite direction.  I also think treating it as whitespace everywhere is a bigger change to the web platform than allowing it as whitespace nowhere.

annevk: Yeah, you're probably right. I guess I should just add a second whitespace concept. Thanks!

fetch.bs
```diff
@@ -2498,7 +2508,32 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
...
+ <li><p>Let <var>value</var> be the result of <a for="header list">getting</a>
+ `<a http-header><code>X-Content-Type-Options</code></a>` from <var>list</var>.
+
+ <li><p>If <var>value</var> is null, then return false.
+
+ <li><p>Let <var>stringValue</var> be the <a>isomorphic decode</a> of <var>value</var>.
+
+ <li>
+  <p>Let <var>tokens</var> be the result of
+  <a lt="split on commas">spliting <var>stringValue</var> on commas</a>.
```
GPHemsley:
```suggestion
  <a lt="split on commas">splitting <var>stringValue</var> on commas</a>.
```

annevk:
@MattMenke2 so I updated the tests and ran them again and Chrome and Safari both treat `nosniff` immediately followed by 0x0B or 0x0C as a "valid value". Firefox does so for 0x0C only. I think that's why I went with treating 0x0C as valid and 0x0B as invalid. I'm happy for them to both be restricted, but are you equally sure about wanting this after knowing these results?

annevk:
That seems somewhat bad… I guess fingers crossed that we can align on not treating 0x0B and 0x0C as whitespace across all code bases?

MattMenke2:
I agree.  Actually, it seems very bad to me.

I still suspect we won't see much breakage here (I mean, really, who uses that character?), but it does mean this has the potential to break pages.  I'm much less gung ho about removing support for it now, but I still think doing so is best for the long term health of the platform, if we can manage it.  I've made enough changes to Chrome's header parsing (One more is in the pipeline) in this version that I'm not comfortable making more until the current set has made it to stable or at least been on beta for a while, but we are going to have to change Chrome's behavior here, regardless, at some point.

MattMenke2:
This bifurcated space handling logic also applies to a number of other response headers in blink.  Cache headers, Accept headers, and quite possibly others.  Blink doesn't have an HTTP-specific whitespace checking method, so it will be difficult to locate them all.

fetch.bs
annevk: That was it.

annevk:
Browser bugs:
* https://bugs.chromium.org/p/chromium/issues/detail?id=900223
* https://bugzilla.mozilla.org/show_bug.cgi?id=1503517
* https://bugs.webkit.org/show_bug.cgi?id=191107
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19470277/

annevk:
As a heads up, I plan to merge this tomorrow or Friday, as it sets out new infrastructure for header parsing it's kind of nice to have it landed so the other bits can start to be converted.

PR #822, opened 2018-10-31
Deduplicate 'the'

saschanaz:


annevk:
Thanks, looks great. Would you like to add yourself to the Acknowledgments section? Either via a new commit or commit rewrite is fine.

Issue #529, opened 2017-04-20
"Extract a MIME type" algorithm should pick the first entry?


tyoshino:
`extract header list values` algorithm returns a list. `extract a MIME type` and sites calling `extract a MIME type` look using the result as a byte sequence.

Should `extract a MIME type` pick the first entry of the result of `extract header list values` if it's not null?

annevk:
Yeah, although we should probably have tests for multiple `Content-Type` headers. I haven't researched that much at all.

tyoshino:
I see. Considering the following, `extract header list values` would return failure when there are multiple entries. So, the fix is not what I said in the OP but pick the single entry while asserting it's a single element list if it's not either null or failure.
- https://tools.ietf.org/html/rfc7230#section-3.2.2
- https://tools.ietf.org/html/rfc7231#section-3.1.1.5
- https://tools.ietf.org/html/rfc7231#section-3.1.1.1

annevk:
Is that what browsers implement? What if you navigate to a URL with two Content-Type headers. Do you get a network error? We should probably have a bunch of web-platform-tests. The HTTP WG generally does a bad job of making sure their requirements can actually be implemented.

tyoshino:
Chrome's fetch() is currently broken (Blink's ExtractMIMEType is using combined value http://crbug.com/713634), but looks Chrome has been not erroring for multiple Content-Types but using the last Content-Type header before the break.

Navigating to a resource with multiple Content-Type also works. Haven't checked which one is used for navigation.

annevk:
Another angle to test would be a single `Content-Type` header with a comma-separated value. They ought to be treated identically, but browsers...

Relates to https://github.com/whatwg/mimesniff/issues/30.

annevk:
https://github.com/whatwg/mimesniff/issues/30#issuecomment-382721437 has my findings on this. Input there much appreciated, even though we'll likely define it here.

annevk:
For anyone subscribed to this issue, please note that #831 is pretty close towards solving this. Tests are linked from there as well.

PR #829, opened 2018-11-05
Be strict on request's Content-Type

annevk:
Tests: https://github.com/web-platform-tests/wpt/pull/13921.

Helps with #814.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -438,18 +438,7 @@ each other by 0x2C 0x20, in order.
   <dl class=switch>
    <dt>`<code>accept</code>`
    <dd>
-    <p>If one of the following is true
-
-    <ul class=brief>
-     <li><p><var>value</var> contains a byte less than 0x20 that is not 0x09 HT
-     <li><p><var>value</var> contains 0x22 ("), 0x28 (left parenthesis), 0x29 (right parenthesis),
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -460,12 +449,20 @@ each other by 0x2C 0x20, in order.
...
+
+     <li><p>Let <var>mimeType</var> be the result of <a lt="parse a MIME type">parsing</a>
+     <var>value</var>.
+
+     <li><p>If <var>mimeType</var> is falure, then return false.
+
+     <li><p>If <var>mimeType</var>'s <a for="MIME type">essence</a> is not
+     "<code>application/x-www-form-urlencoded</code>", "<code>multipart/form-data</code>", or
+     "<code>text/plain</code>", then return false.
+    </ol>
```
annevk: It would be good to add a note here illustrating that this is very different from how response Content-Type is parsed, but maybe that should wait until that's actually defined properly.

annevk:
Bugs:
* https://bugs.webkit.org/show_bug.cgi?id=191356
* https://bugs.chromium.org/p/chromium/issues/detail?id=902681
* https://bugzilla.mozilla.org/show_bug.cgi?id=1483815
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18694370/

I plan on merging this soonish, unless someone volunteers to review.

fetch.bs
annevk: I'll do this as part of #529.

PR #831, opened 2018-11-09
Define the Content-Type header parser

### LINKED ISSUE #529: "Extract a MIME type" algorithm should pick the first entry?

annevk:
Also known as "extract a MIME type" down right.

Tests: https://github.com/web-platform-tests/wpt/pull/10525.

Helps with #814.

Fixes #529. Closes https://github.com/whatwg/mimesniff/issues/30.

***
<a href="https://whatpr.org/fetch/831.html" title="Last updated on Nov 22, 2018, 1:38 PM GMT (eed035b)">Preview</a> | <a href="https://whatpr.org/fetch/831/a243f1f...eed035b.html" title="Last updated on Nov 22, 2018, 1:38 PM GMT (eed035b)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -90,6 +90,12 @@ url:https://tools.ietf.org/html/rfc7234#section-1.2.1;text:delta-seconds;type:df
 }
 </pre>
 
+<pre class=link-defaults>
+spec:infra; type:dfn; text:string
+</pre>
+
+
+
 <h2 id=goals class="no-num short">Goals</h2>
 
 <p>To unify fetching across the web platform this specification supplants a number of algorithms and
@@ -323,6 +329,111 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
 
  <li><p>Return the <a for="header">combined value</a> with <var>name</var> and <var>list</var>.
 </ol>
+
+<p>To
+<dfn export for="header list" lt="get, decode, and split|getting, decoding, and splitting" id=concept-header-list-get-decode-split>get, decode, and split</dfn>
+a <a for=header>name</a> <var>name</var> from <a for=/>header list</a> <var>list</var>, run these
+steps:
+
+<ol>
+ <li><p>Let <var>initialValue</var> be the result of <a for="header list">getting</a>
+ <var>name</var> from <var>list</var>.
+
+ <li><p>If <var>initialValue</var> is null, then return null.
+
+ <li><p>Let <var>input</var> be the result of <a>isomorphic decoding</a> <var>initialValue</var>.
+
+ <li><p>Let <var>position</var> be a <a for=string>position variable</a> for <var>input</var>,
+ initially pointing at the start of <var>input</var>.
+
+ <li><p>Let <var>values</var> be a <a for=/>list</a> of <a for=/>strings</a>, initially empty.
+
+ <li><p>Let <var>value</var> be null.
+
+ <li>
+  <p>While <var>position</var> is not past the end of <var>input</var>:
+
+  <ol>
+   <li>
+    <p>Let <var>temporaryValue</var> be the result of <a>collecting a sequence of code points</a>
+    that are not U+0022 (") or U+002C (,) from <var>input</var>, given <var>position</var>.
+
+    <p class=note><var>temporaryValue</var> might be the empty string.
+
+   <li><p>If <var>value</var> is null, then set <var>value</var> to <var>temporaryValue</var>, and
+   append <var>temporaryValue</var> to <var>value</var> otherwise.
+
+   <li>
+    <p>If <var>position</var> is not past the end of <var>input</var>, then:
+
+    <ol>
+     <li>
+      <p>If the <a for=/>code point</a> at <var>position</var> within <var>input</var> is
+      U+0022 ("), then:
+
+      <ol>
+       <li><p>Append U+0022 (") to <var>value</var>.
+
+       <li><p>Append to <var>value</var> the result of <a>collecting a sequence of code points</a>
+       that are not U+0022 (") from <var>input</var>, given <var>position</var>.
+
+       <li>
+        <p>If <var>position</var> is not past the end of <var>input</var>, then:
+
+        <ol>
+         <li><p>Assert: the <a for=/>code point</a> at <var>position</var> within <var>input</var>
+         is U+0022 (").
+
+         <li><p>Advance <var>position</var> by 1.
+
+         <li><p>Append U+0022 (") to <var>value</var>.
+
+         <li><p><a for=iteration>Continue</a>.
+        </ol>
+      </ol>
+
+     <li>
+      <p>Otherwise:
+
+      <ol>
+       <li><p>Assert: the <a for=/>code point</a> at <var>position</var> within <var>input</var> is
+       U+002C (,).
+
+       <li><p>Advance <var>position</var> by 1.
+      </ol>
+    </ol>
+
+   <li><p>Remove all <a>HTTP tab or space</a> from the start and end of <var>value</var>.
+
+   <li><p><a for=list>Append</a> <var>value</var> to <var>values</var>.
+
+   <li><p>Set <var>value</var> to null.
+  </ol>
+
+ <li><p>Return <var>values</var>.
+</ol>
+
+<div class=example id=example-header-list-get-decode-split>
+ <p>This is how <a>get, decode, and split</a> functions in practice:
+
+ <table>
+  <tr>
+   <th>Input
+   <th>Output
+  <tr>
+   <td>`<code>nosniff,</code>`
+   <td>"<code>nosniff</code>" and the empty string
+  <tr>
+   <td>`<code>text/html;", x/x</code>`
+   <td>"<code>text/html;", x/x</code>"
+  <tr>
+   <td>`<code>x/x;test="hi",y/y</code>`
+   <td>"<code>x/x;test="hi"</code>" and "<code>y/y</code>"
+  <tr>
+   <td>`<code>x / x,,,1</code>`
+   <td>"<code>x / x</code>", the empty string, the empty string, and "<code>1</code>"
+ </table>
+</div>
 
 <p>To <dfn export for="header list" id=concept-header-list-append>append</dfn> a
 <a for=header>name</a>/<a for=header>value</a> <var>name</var>/<var>value</var> pair to a
@@ -690,12 +801,48 @@ run these steps:
 from a <a for=/>header list</a> <var>headers</var>, run these steps:
 
 <ol>
- <li><p>Let <var>mimeType</var> be the result of <a>extracting header list values</a> given
- `<code>Content-Type</code>` and <var>headers</var>.
+ <li><p>Let <var>charset</var> be null.
 
- <li><p>If <var>mimeType</var> is null or failure, then return the empty byte sequence.
+ <li><p>Let <var>essence</var> be null.
 
- <li><p>Return <var>mimeType</var>, <a lt="byte-lowercase">byte-lowercased</a>.
+ <li><p>Let <var>mimeType</var> be null.
+
+ <li><p>Let <var>values</var> be the result of
+ <a for="header list">getting, decoding, and splitting</a> `<code>Content-Type</code>` from
+ <var>headers</var>.
+
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>values</var>:
+
+  <ol>
+   <li><p>Set <var>mimeType</var> to the result of <a lt="parse a MIME type">parsing</a>
+   <var>value</var>.
+
+   <li>
+    <p>If <var>mimeType</var> is not failure and <var>mimeType</var>'s
+    <a for="MIME type">essence</a> is not "<code>*/*</code>" or <var>essence</var>, then:
+
+    <ol>
+     <li><p>Set <var>charset</var> to null.
+
+     <li><p>If <var>mimeType</var>'s <a for="MIME type">parameters</a>["<code>charset</code>"]
+     <a for=map>exists</a>, then set <var>charset</var> to <var>mimeType</var>'s
+     <a for="MIME type">parameters</a>["<code>charset</code>"].
+
+     <li><p>Set <var>essence</var> to <var>mimeType</var>'s <a for="MIME type">essence</a>.
+    </ol>
+
+   <li>
+    <p>Otherwise, if <var>mimeType</var>'s <a for="MIME type">essence</a> is <var>essence</var>:
+
+    <ol>
+     <li><p>If <var>mimeType</var>'s <a for="MIME type">parameters</a>["<code>charset</code>"] does
+     not <a for=map>exist</a> and <var>charset</var> is non-null, then set <var>mimeType</var>'s
+     <a for="MIME type">parameters</a>["<code>charset</code>"] to <var>charset</var>.
+    </ol>
+  </ol>
+
+ <li><p>Return <var>mimeType</var>.
 </ol>
 
 <hr>
@@ -2532,20 +2679,13 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
 steps:
 
 <ol>
- <li><p>Let <var>value</var> be the result of <a for="header list">getting</a>
+ <li><p>Let <var>values</var> be the result of
+ <a for="header list">getting, decoding, and splitting</a>
  `<a http-header><code>X-Content-Type-Options</code></a>` from <var>list</var>.
 
- <li><p>If <var>value</var> is null, then return false.
+ <li><p>If <var>values</var> is null, then return false.
 
- <li><p>Let <var>stringValue</var> be the <a>isomorphic decode</a> of <var>value</var>.
-
- <li><p>Let <var>tokens</var> be the result of
- <a lt="strictly split a string">strictly splitting</a> <var>stringValue</var> on U+002C (,).
-
- <li><p>Let <var>firstToken</var> be the result of removing all <a>HTTP tab or space</a> from the
- start and end of <var>tokens</var>[0].
-
- <li><p>If <var>firstToken</var> is an <a>ASCII case-insensitive</a> match for
+ <li><p>If <var>values</var>[0] is an <a>ASCII case-insensitive</a> match for
  "<code>nosniff</code>", then return true.
 
  <li><p>Return false.
--- a/fetch.bs
+++ b/fetch.bs
@@ -682,6 +682,23 @@ each other by 0x2C 0x20, in order.
     <p class=warning>This intentionally does not use <a>extract a MIME type</a> as that algorithm is
     rather forgiving and servers are not expected to implement it.
 
+    <div class="example no-backref" id=example-cors-safelisted-request-header-content-type>
+     <p>If <a>extract a MIME type</a> were used the following request would not result in a CORS
+     preflight and a naïve parser on the server might treat the request body as JSON:
+
+     <pre><code>
+fetch("https://victim.example/naïve-endpoint", {
+  method: "POST",
+  headers: [
+    ["Content-Type", "application/json"],
+    ["Content-Type", "text/plain"]
+  ],
+  credentials: "include",
+  body: JSON.stringify(exerciseForTheReader)
+});
+</code></pre>
+    </div>
+
    <dt>`<code><a href=http://httpwg.org/http-extensions/client-hints.html#dpr>DPR</a></code>`
    <dt>`<code><a href=http://httpwg.org/http-extensions/client-hints.html#downlink>Downlink</a></code>`
    <dt>`<code><a href=http://httpwg.org/http-extensions/client-hints.html#save-data>Save-Data</a></code>`
--- a/fetch.bs
+++ b/fetch.bs
@@ -291,36 +291,35 @@ run these steps:
    <li><p>Append the result of <a>collecting a sequence of code points</a> that are not U+0022 (")
    or U+005C (\) from <var>input</var>, given <var>position</var>, to <var>value</var>.
 
+   <li><p>If <var>position</var> is past the end of <var>input</var>, then
+   <a for=iteration>break</a>.
+
+   <li><p>Let <var>quoteOrBackslash</var> be the <a>code point</a> at <var>position</var> within
+   <var>input</var>.
+
+   <li><p>Advance <var>position</var> by 1.
+
    <li>
-    <p>If <var>position</var> is not past the end of <var>input</var>, then:
+    <p>If <var>quoteOrBackslash</var> is U+005C (\), then:
 
     <ol>
-     <li><p>Let <var>quoteOrBackslash</var> be the <a>code point</a> at <var>position</var> within
-     <var>input</var>.
+     <li><p>If <var>position</var> is past the end of <var>input</var>, then append U+005C (\) to
+     <var>value</var> and <a for=iteration>break</a>.
+
+     <li><p>Append the <a>code point</a> at <var>position</var> within <var>input</var> to
+     <var>value</var>.
 
      <li><p>Advance <var>position</var> by 1.
-
-     <li>
-      <p>If <var>quoteOrBackslash</var> is U+005C (\), then:
-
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -324,6 +330,111 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+
+ <li><p>Return <var>values</var>.
+</ol>
+
+<div class=example id=example-header-list-get-decode-split>
+ <p>This is how <a>get, decode, and split</a> functions in practice:
+
+ <table>
+  <tr>
+   <th>Input
```
annevk: Since input is a header list, this should probably be reworked somewhat.

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>values</var>:
+
+  <ol>
+   <li><p>Set <var>mimeType</var> to the result of <a lt="parse a MIME type">parsing</a>
+   <var>value</var>.
+
+   <li>
+    <p>If <var>mimeType</var> is not failure and <var>mimeType</var>'s
+    <a for="MIME type">essence</a> is not "<code>*/*</code>" or <var>essence</var>, then:
```
annevk: Still need to test the empty string here. It'd also be good if this were consistent with MIME sniffing which does consider application/unknown and co...

fetch.bs
```diff
@@ -324,6 +330,113 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+
+  <ol>
+   <li>
+    <p>Let <var>temporaryValue</var> be the result of <a>collecting a sequence of code points</a>
+    that are not U+0022 (") or U+002C (,) from <var>input</var>, given <var>position</var>.
+
+    <p class=note><var>temporaryValue</var> might be the empty string.
+
+   <li><p>If <var>value</var> is null, then set <var>value</var> to <var>temporaryValue</var>, and
+   append <var>temporaryValue</var> to <var>value</var> otherwise.
```
domenic: I find this structure a bit jarring to read an prefer "If X, then Y; otherwise Z". (Or as two separate sentences.)

fetch.bs
```diff
@@ -324,6 +330,113 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+   <li><p>Set <var>value</var> to null.
+  </ol>
+
+ <li><p>Return <var>values</var>.
+</ol>
+
+<div class=example id=example-header-list-get-decode-split>
+ <p>This is how <a>get, decode, and split</a> functions in practice:
+
+ <table>
```
domenic: It'd be good to have an example of a comma inside quotes.

fetch.bs
```diff
@@ -324,6 +330,113 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+  <tr>
+   <th>Input
+   <th>Output
+  <tr>
+   <td>`<code>nosniff,</code>`
```
domenic: You could use list notation on the right hand side.

fetch.bs
```diff
@@ -471,6 +584,9 @@ each other by 0x2C 0x20, in order.
      "<code>text/plain</code>", then return false.
     </ol>
 
+    <p class=warning>This intentionally does not use <a>extract a MIME type</a> as it is rather
+    forgiving and servers are not expected to implement it.
```
domenic: Which is "it"?

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+
+ <li><p>Let <var>essence</var> be null.
+
+ <li><p>Let <var>mimeType</var> be null.
+
+ <li><p>Let <var>values</var> be the result of
+ <a for="header list">getting, decoding, and splitting</a> `<code>Content-Type</code>` from
+ <var>headers</var>.
+
+ <li><p>If <var>values</var> is null, then return failure.
```
domenic: I wonder if getting/decoding/splitting should return failure too?

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>values</var>:
+
+  <ol>
+   <li><p>Set <var>mimeType</var> to the result of <a lt="parse a MIME type">parsing</a>
+   <var>value</var>.
```
domenic: It might be clearer to immediately "continue" if mimeType is failure.

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+
+ <li><p>Return <var>mimeType</var>.
+</ol>
+
+<p class=warning>Treat <a>extract a MIME type</a> returning failure or anything but the needed
+<a for=/>MIME type</a>'s <a for="MIME type">essence</a> as a fatal error. Existing web platform
+features not following this has been a major source of security vulnerabilities over the years.
+
+<p class=note>A <a for=/>MIME type</a>'s <a for="MIME type">parameters</a> are typically ignored and
+this practice has not been a problem.
```
domenic: "this has not been a problem in practice"?

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+   <var>charset</var> is non-null, set <var>mimeType</var>'s
+   <a for="MIME type">parameters</a>["<code>charset</code>"] to <var>charset</var>.
+  </ol>
+
+ <li><p>Return <var>mimeType</var>.
+</ol>
+
+<p class=warning>Treat <a>extract a MIME type</a> returning failure or anything but the needed
+<a for=/>MIME type</a>'s <a for="MIME type">essence</a> as a fatal error. Existing web platform
+features not following this has been a major source of security vulnerabilities over the years.
```
domenic: s/has/have

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+   <a for="MIME type">parameters</a>["<code>charset</code>"] does not <a for=map>exist</a>, and
+   <var>charset</var> is non-null, set <var>mimeType</var>'s
+   <a for="MIME type">parameters</a>["<code>charset</code>"] to <var>charset</var>.
+  </ol>
+
+ <li><p>Return <var>mimeType</var>.
+</ol>
+
+<p class=warning>Treat <a>extract a MIME type</a> returning failure or anything but the needed
+<a for=/>MIME type</a>'s <a for="MIME type">essence</a> as a fatal error. Existing web platform
```
domenic: This is a bit unclear. It returns a whole MIME type, not an essence, so strictly speaking according to this sentence, its return value should always be treated as fatal error.

fetch.bs
```diff
@@ -2520,6 +2625,68 @@ consideration for the security consequences. New exceptions can be proposed by
...
+
+   <li><p>Otherwise, if <var>mimeType</var> is not failure, <var>mimeType</var>'s
+   <a for="MIME type">essence</a> is <var>essence</var>, <var>mimeType</var>'s
+   <a for="MIME type">parameters</a>["<code>charset</code>"] does not <a for=map>exist</a>, and
+   <var>charset</var> is non-null, set <var>mimeType</var>'s
+   <a for="MIME type">parameters</a>["<code>charset</code>"] to <var>charset</var>.
+  </ol>
+
+ <li><p>Return <var>mimeType</var>.
+</ol>
```
domenic: This algorithm would probably also benefit from a few examples. The main thing to communicate is how later values win, and how this charset overriding business works.

domenic:
Oh, also, should the new `Content-Type` header section mention that "extract a MIME type" is meant only for use on responses? Maybe the algorithm name should be more specific, e.g. "parse the response Content-Type header"?

fetch.bs
annevk: extract a MIME type

fetch.bs
annevk: A heading that is missing is typically not a failure case.

annevk:
@MattMenke2 @bzbarsky any final thoughts here? I realize this doesn't match browsers 100% due to eager combining, but it feels like the only sensible model to specify given that intermediaries are allowed to combine. Hopefully given enough time browsers can converge on this model. And if they can't, we can revisit…

fetch.bs
```diff
@@ -324,6 +330,165 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+
+       <li><p>Append the result of <a>collecting a sequence of code points</a> that are not
+       U+0022 (") from <var>input</var>, given <var>position</var>, to <var>value</var>.
+
+       <li>
+        <p>If <var>position</var> is not past the end of <var>input</var>, then:
+
+        <ol>
+         <li><p>Assert: the <a for=/>code point</a> at <var>position</var> within <var>input</var>
+         is U+0022 (").
```
MattMenke2: What about escaped quotes?

fetch.bs
```diff
@@ -324,6 +330,165 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+</code></pre>
+   <td rowspan=2>« "<code>nosniff</code>", "" »
+  <tr>
+   <td>
+    <pre><code>
+A: nosniff
+B: sniff
+A:
+</code></pre>
+  <tr>
```
MattMenke2: Maybe add blank lines before each "real" row.  The rowspan and <pre> tags make it rather difficult to follow this table.  Or just add HTML comments between entries.

annevk: The Preview and Diff links in OP should help here, that way you don't have to read source code but can look at rendered HTML.

fetch.bs
annevk: Wow, yes, back to the drawing board. I guess I should see if this can reuse some code from the MIME type parser.

bzbarsky:
I don't really understand this space well enough at this point to comment intelligently...

annevk:
There's now additional tests for escaped quotes as well.

fetch.bs
```diff
@@ -324,6 +393,155 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+ <li><p>If <var>initialValue</var> is null, then return null.
+
+ <li><p>Let <var>input</var> be the result of <a>isomorphic decoding</a> <var>initialValue</var>.
+
+ <li><p>Let <var>position</var> be a <a for=string>position variable</a> for <var>input</var>,
+ initially pointing at the start of <var>input</var>.
+
+ <li><p>Let <var>values</var> be a <a for=/>list</a> of <a for=/>strings</a>, initially empty.
+
+ <li><p>Let <var>value</var> be null.
```
MattMenke2: Do we get anything by starting with value as null instead of the empty string?  Seems like if we had it as the empty string, it saves a line of "code".  We do end up with values being an empty string at the end of the code, but I don't think that hurts?  The code is correct, either way, just think that less complexity is better, though don't have strong feelings here.

fetch.bs
```diff
@@ -324,6 +393,155 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+    <pre><code>
+A: text/html;"
+A: x/x
+</code></pre>
+  <tr>
+   <td>
+    <pre><code>
+A: x/x;test="hi",y/y
+</code></pre>
+   <td rowspan=2>« "<code>x/x;test="hi"</code>", "<code>y/y</code>" »
```
MattMenke2: This is correct, but distinguishing the code and non-code quotes can be a bit confusing.  Maybe remove the extra delimiters and put each value on a separate line instead?  Or use something other than quotes?

<"x/x;test="hi">, <y/y> (With or without the surrounding «»)

fetch.bs
```diff
@@ -2520,6 +2730,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+  <tr>
+   <td>
+    <pre><code>
+Content-Type: text/plain;charset=gbk, text/html
+</code></pre>
+   <td><code>text/html</code>
+  <tr>
+   <td>
+    <pre><code>
+Content-Type: text/html;charset=gbk, text/html;x=y
```
MattMenke2: Maybe add an ";a=b" to the first set, which shouldn't carry over?

fetch.bs
```diff
@@ -2520,6 +2730,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+</code></pre>
+   <td><code>text/html;x=y;charset=gbk</code>
```
MattMenke2: The output is an essense and a parameters array, not a string, right?  i.e., there's nothing to say that this is more like:

text/html;x=y;charset=gbk

instead of:

text/html;charset=gbk;x=y

fetch.bs
annevk: Using a different convention would require changes to Infra. @domenic what do you think?

fetch.bs
annevk: The output is a MIME type of which the parameters are ordered (an ordered map).

MattMenke2: Thanks!  Missed that.

annevk:
Yeah, I agree https://whatpr.org/fetch/831.html#example-header-list-get-decode-split doesn't look good. I filed https://github.com/whatwg/infra/issues/228 to discuss it.

I'll let @domenic do a review pass as well and will attempt to bring the formatting issue to a satisfactory resolution before landing.

fetch.bs
```diff
@@ -2520,6 +2726,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+   <li><p>If <var>mimeType</var> is failure, then <a for=iteration>continue</a>.
+
+   <li>
+    <p>If <var>mimeType</var>'s <a for="MIME type">essence</a> is not "<code>*/*</code>" or
+    <var>essence</var>, then:
+
+    <ol>
+     <li><p>Set <var>charset</var> to null.
+
+     <li><p>If <var>mimeType</var>'s <a for="MIME type">parameters</a>["<code>charset</code>"]
```
MattMenke2: One other question:  Should we be taking the first MIME type rather than the last?  This would change behavior, but it would match the fact we prefer earlier charsets.  I had thought this was discussed as a potential change elsewhere.  I'm not pushing strongly for a change here, as I think all browsers currently prefer later MIME types, but thought I'd bring it up.

fetch.bs
```diff
@@ -264,6 +270,69 @@ use "<code>deprecated</code>" is not defined by this specification. An
...
+
+ <li>
+  <p>While true:
+
+  <ol>
+   <li><p>Append the result of <a>collecting a sequence of code points</a> that are not U+0022 (")
+   or U+005C (\) from <var>input</var>, given <var>position</var>, to <var>value</var>.
+
+   <li>
+    <p>If <var>position</var> is not past the end of <var>input</var>, then:
```
domenic: It would be simpler to flip this so that you break if position is past the end of input. That would probably also allow you to remove the "continue" step later.

fetch.bs
```diff
@@ -324,6 +393,151 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+
+   <li><p>Set <var>value</var> to the empty string.
+  </ol>
+
+ <li><p>Return <var>values</var>.
+</ol>
+
+<div class=example id=example-header-list-get-decode-split>
+ <p>This is how <a>get, decode, and split</a> functions in practice with `<code>A</code>` as
+ <var>name</var> argument:
```
domenic: "as" -> "as the"

fetch.bs
```diff
@@ -324,6 +393,151 @@ specialized multimap. An ordered list of key-value pairs with potentially duplic
...
+ <table>
+  <tr>
+   <th>Headers (as on the network)
+   <th>Output
+  <tr>
+   <td>
+    <pre><code>
+A: nosniff,
+</code></pre>
+   <td rowspan=2>« "<code>nosniff</code>", "" »
```
domenic: So you could do the following formatting, not sure if it'd be better:

```html
«<br>
&nbsp;&nbsp;"<code>nosniff</code>",<br>
&nbsp;&nbsp;""<br>
»
```

(or maybe some variant with white-space: pre on the rows). Example:

image

fetch.bs
```diff
@@ -471,6 +685,9 @@ each other by 0x2C 0x20, in order.
      "<code>text/plain</code>", then return false.
     </ol>
 
+    <p class=warning>This intentionally does not use <a>extract a MIME type</a> as that algorithm is
+    rather forgiving and servers are not expected to implement it.
```
domenic: I find this warning confusing. Why are we talking about what servers should implement here? Isn't this part of the processing model for the browser?

fetch.bs
```diff
@@ -2520,6 +2726,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+   <a for="MIME type">parameters</a>["<code>charset</code>"] to <var>charset</var>.
+  </ol>
+
+ <li><p>Return <var>mimeType</var>.
+</ol>
+
+<p class=warning>When <a>extract a MIME type</a> returns failure or a <a for=/>MIME type</a> whose
+<a for="MIME type">essence</a> is incorrect for a given format, treat this as a fatal error.
+Existing web platform features not always following this pattern has been a major source of security
+vulnerabilities over the years.
```
domenic: Maybe "Existing web platform features have not always followed this pattern, which has been a major source of security vulnerabilities in those features over the years".

fetch.bs
```diff
@@ -2520,6 +2726,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+
+<p class=note>A <a for=/>MIME type</a>'s <a for="MIME type">parameters</a> are typically ignored and
+this has not been a problem in practice.
```
domenic: This note seems out of place right after an algorithm that preserves the parameters. Maybe it is supposed to relate to the previous warning? If so I'd put it inside the warning, instead of as a separate note.

fetch.bs
```diff
@@ -2520,6 +2726,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+
+<div class=example id=example-extract-a-mime-type>
+ <p>This is how <a>extract a MIME type</a> functions in practice:
+
+ <table>
+  <tr>
+   <th>Headers (as on the network)
+   <th>Output
```
domenic: Maybe "Output (serialized)", optionally linking "serialized" to the MIME type serializer.

fetch.bs
```diff
@@ -2520,6 +2726,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+  <tr>
+   <td>
+    <pre><code>
+Content-Type: text/plain;charset=gbk, text/html
+</code></pre>
+   <td><code>text/html</code>
+  <tr>
+   <td>
+    <pre><code>
+Content-Type: text/html;charset=gbk;a=b, text/html;x=y
```
domenic: Can you do a rowspan with this example having two Content-Type lines? It's an especially weird feature that "later" Content-Type lines can get a charset "added to them" by earlier ones, so showing that would be good. (I know a comma version is equivalent, but making it explicit is nice too.)

fetch.bs
```diff
@@ -2520,6 +2726,100 @@ consideration for the security consequences. New exceptions can be proposed by
...
+Existing web platform features not always following this pattern has been a major source of security
+vulnerabilities over the years.
+
+<p class=note>A <a for=/>MIME type</a>'s <a for="MIME type">parameters</a> are typically ignored and
+this has not been a problem in practice.
+
+<div class=example id=example-extract-a-mime-type>
+ <p>This is how <a>extract a MIME type</a> functions in practice:
+
+ <table>
```
domenic: Maybe add a `*/*` example?

fetch.bs
annevk: So yeah, if you think that's doable it's certainly worth considering. But what if it's `*/*` or fails to parse, would we then still go to the next? If so, I'm not sure what it would buy us in terms of security.

fetch.bs
annevk: Yeah, but the security problem is a mismatch between the browser and a server and therefore this parser is more conservative.

fetch.bs
annevk: I added a commit that does this, but note that you'll need an additional break due to having to deal with ". But there's less nesting, so probably acceptable.

fetch.bs
domenic: I still don't really understand. Why do we have a second parser on the client, to solve a client/server mismatch?

Maybe a better way to state this (either to help me understand here, or in the spec) would be with an example of what would go wrong if we used extract a MIME type.

annevk: If we use the "liberal" parser for this the client might end up with text/plain but the server with application/json or some such. This would end up rejecting such input.

domenic: So the mismatch scenario is: client uses the Headers class to add content-type: someHardToParseThingWhichTheyIntendToBeTextPlain, and does a fetch() post to the server, which uses a different parser and ends up with application/json?

If so, I'll note that's not really an attack (the client could just send application/json directly). So maybe this is more of a note than a warning. In either case, adding such an example would be helpful.

annevk: How could the client send application/json without a preflight?

domenic: I see, text/plain is explicitly allowed! This example would indeed be great to add.

annevk: Added.

fetch.bs
MattMenke2: Not sure that skipping over invalid MIME types really costs us much in terms of security, but I'd be perfectly fine with just taking the first line, and ignoring the rest.  Downside of that is, as with most every decision here, we just don't know how much of the web that breaks.

annevk: So to be clear, we'd combine, split, and then parse? As opposed to combine and then parse (which we do on the request side for CORS)?

MattMenke2: We're already combining, splitting, and then parsing, actually.  The "For each value of values" would just be removed, and we'd use value of [0] everywhere.  So it's one less step in the code, which I'm not sure can really be described as splitting, if what we were doing before isn't splitting.

Sorry, I'm not all that familiar with CORS.  Which section are you referring to as the request side of CORS that works differently?

annevk: To be specific, the `Content-Type` parser defined at https://whatpr.org/fetch/831.html#cors-safelisted-request-header for use on requests fails on `text/plain,application/json` as it doesn't split. (And this is intentional for security reasons.)

Anyway, it would indeed be a rather big simplification if we'd simply use the first value as that would also remove the need for `charset` moving and such. I'm rather skeptical this'll work out, but if you want to try it I guess I'm not opposed.

MattMenke2: The more I think about it, the more I think you're probably right that just taking the first would break too much.

Seems to me like the safest way to CORS-safelist-check check the Content-Type request header would be to either check all MIME types, not just the first one, or refuse to send any but the first.  That way, we wouldn't have to rely on servers implementing a fully compliant Content-Type parser.  But that's perhaps getting a bit far afield.

annevk: https://github.com/whatwg/fetch/issues/838

fetch.bs
```diff
@@ -471,6 +679,26 @@ each other by 0x2C 0x20, in order.
...
     </ol>
 
+    <p class=warning>This intentionally does not use <a>extract a MIME type</a> as that algorithm is
+    rather forgiving and servers are not expected to implement it.
+
+    <div class="example no-backref" id=example-cors-safelisted-request-header-content-type>
+     <p>If <a>extract a MIME type</a> were used the following request would not result in a CORS
+     preflight and a naïve parser on the server might treat the request body as JSON:
+
+     <pre><code>
```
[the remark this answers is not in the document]
annevk: https://github.com/whatwg/fetch/issues/837

annevk:
Bugs:
* https://bugs.webkit.org/show_bug.cgi?id=192000
* https://bugzilla.mozilla.org/show_bug.cgi?id=1510180
* https://bugs.chromium.org/p/chromium/issues/detail?id=908765
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19732802/

Issue #737, opened 2018-05-25
fetch() "no-cors": cross-origin to same-origin redirect taints response


annevk:
Although per step 5 of main fetch https://fetch.spec.whatwg.org/#main-fetch if you're on a same-origin URL again after a redirect the tainting gets reset to "basic", that doesn't appear to happen in implementations.

At least for `fetch()` they'll taint once you go cross-origin.

I don't think they're consistent in that however as cross-origin to same-origin for images will allow image data extraction.

I guess this requires more research.

domfarolino:
Interesting; Chrome is indeed one of the violators of not resetting tainting back to "basic" after a `fetch()` lands on same-origin after cross-origin redirect, but this is a simple change. Are you considering a spec change here depending on what implementations are doing?

annevk:
I was initially, but I'm wondering if what browsers do here is preferable, since even though you can read such responses as the result of navigation (`<iframe>`) or image loads, that doesn't allow you to read the bytes of all such responses (e.g., a PDF).

So I think we probably want to keep this restriction intact where we can (and break it for navigation, images, scripts?, ?).

wanderview:
I think we should avoid reseting back to basic tainting where we can.  This is consistent with how cors requests can never become same-origin again once they cross an origin boundary.  It would be a bit unexpected to me to make no-cors less restrictive.

domfarolino:
(Whoops, this slipped through the cracks in my email). I agree with this as well here, and think we should avoid resetting back to basic tainting.

youennf:
+1 to not resetting tainting.
I haven't tested it on WebKit but if we are resetting tainting in some cases like canvas, this is probably a bug.

yutakahirano:
Hi, I'm going to add a test because I was about to change the behavior without noticing this issue. I don't have a strong opinion whether we should reset tainting, but I strongly want to have a consistent behavior across fetch() and other initiators.

yutakahirano:
The test is here. https://github.com/web-platform-tests/wpt/pull/12566

annevk:
I created a fix for this for Fetch that I think we should land unless it has a bug I didn't see: #834.

However, A -> B -> A is considered same-origin for `<img>`, `<script>`, and similar such contexts and we'll need to make sure that's defined properly there (by having HTML poke into the opaque response) and tested.

I'll leave this issue open until that's fully taken care of.

Hope that seems reasonable to everyone.

youennf:
[...]

PR #834, opened 2018-11-16
Take tainted origin flag into account for the same origin check

### LINKED ISSUE #756: Request's tainted origin flag fallout
@yutakahirano noticed one more place that relied request's origin becoming an opaque origin: step 5 of main fetch.

I incorrectly assumed this wouldn't be a problem due to the CORS flag, but the CORS flag is not set if the initial request was same-origin.

I think making the first conditional of step 5 instead read

> request's current url's origin is same origin with request's origin, CORS flag is unset, and either request's tainted origin flag is unset or request's mode is not "cors"

would fix this.

This is somewhat related to #737.

annevk:
This also addresses #737 in that now A -> B -> A would be considered cross-origin even for "no-cors", but leaving that open for further plumbing in HTML et al to override that in select cases (e.g., ``).

Fixes #756.

***
<a href="https://whatpr.org/fetch/834.html" title="Last updated on Nov 19, 2018, 10:02 AM GMT (dca2295)">Preview</a> | <a href="https://whatpr.org/fetch/834/ba2fb9c...dca2295.html" title="Last updated on Nov 19, 2018, 10:02 AM GMT (dca2295)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2997,8 +2997,8 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
 
   <dl class=switch>
    <dt><var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> is
-   <a>same origin</a> with <var>request</var>'s <a for=request>origin</a> and <i>CORS flag</i> is
-   unset
+   <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, <var>request</var>'s
+   <a for=request>tainted origin flag</a> is unset, and the <i>CORS flag</i> is unset
    <dt><var>request</var>'s <a for=request>current URL</a>'s <a for=url>scheme</a> is
    "<code>data</code>"
    <dt><var>request</var>'s <a for=request>mode</a> is


yutakahirano:
> but leaving that open for further plumbing in HTML et al to override that in select cases (e.g., ).

Can you elaborate on this part?

annevk:
My thinking was that for `` A -> B -> A (without CORS) results in a "same origin" image that can be read through `<canvas>`. HTML would have to poke through the opaque response filter Fetch adds to enable that. (However, given https://github.com/whatwg/fetch/issues/737#issuecomment-439592046 we might be able to change things around here if we wanted.)

domfarolino:
Ok so my understanding is that this spec change will make A -> B -> A "no-cors" `` responses "opaque", thus asserting the behavior in https://github.com/web-platform-tests/wpt/pull/14112. Having a way for HTML to poke through would allow browsers to avoid that if they wanted. FWIW Chrome is OK with this change, but will collect some metrics before implementation to see if it might break things.

So I guess if we can get all browsers on board then we wouldn't need to add prose allowing HTML to effectively get around the opaqueness of the response?

annevk:
Yes, I'm leaving #737 open to study whether or not to change HTML around this, but I agree it would be nice not to change it, even if you can still workaround some cases with `<iframe>`.

PR #836, opened 2018-11-20
Change Response's statusText's default

### LINKED ISSUE #698: Change new Response() statusText default?
In #600 I'm changing the default for network responses (and `Response.redirect()`, which Safari already aligns with) since HTTP/2 doesn't have a status message. It would make sense to change the default for the API too, but that would be a "breaking" change across all implementations.

It would make sense though.

Concretely, this would mean changing the dictionary default from "OK" to "".

cc @youennf @wanderview @yutakahirano

annevk:
The empty string is a better default as the field is optional in HTTP and rather meaningless. This way if you change the status and forget about statusText it won't give the mistaken impression things are OK.

Tests: ...

Fixes #698.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5922,7 +5922,7 @@ Response includes Body;
 
 dictionary ResponseInit {
   unsigned short status = 200;
-  ByteString statusText = "OK";
+  ByteString statusText = "";
   HeadersInit headers;
 };
 


annevk:
@youennf @yutakahirano what do you think? With browser bugs this is reasonable to do? It's an API change, but extremely minor.

domfarolino:
FWIW I also think this makes sense, but do wonder how much potentially older JS out there is relying on "OK".

annevk:
It seems unlikely given this only affects synthetic responses and you'd generally not check this field (or set it), but yes.

domfarolino:
Oh that's right, good point.

jdm:
https://github.com/servo/servo/issues/22238

PR #840, opened 2018-11-27
Fix: Appending a header to headers with "request-no-cors" guard

yutakahirano:
If the headers doesn't contain entries for |name|, |temporaryValue|
should be |value|, not the combined value followed by 0x2c 0x20
followed by |value|.

***
<a href="https://whatpr.org/fetch/840.html" title="Last updated on Nov 28, 2018, 1:14 AM GMT (991adf3)">Preview</a> | <a href="https://whatpr.org/fetch/840/0b2bc05...991adf3.html" title="Last updated on Nov 28, 2018, 1:14 AM GMT (991adf3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -602,11 +602,10 @@ a <a for=/>header list</a> <var>list</var>, run these steps:
   <p><a for=list>For each</a> <var>name</var> in <var>names</var>:
 
   <ol>
-   <li>
-    <p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var>
-    from <var>list</var>.
+   <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var>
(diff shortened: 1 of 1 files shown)


yutakahirano:
> So I think this should use https://fetch.spec.whatwg.org/#concept-header-list-get rather than "combined value" and then we need to branch on null. If the existing value happens to be the empty string, we should still append ", " and the given value to it.

Done.

> I think the right fix here would be to remove "combined value" entirely in favor of "get". This would require patching XHR as well. Would you be willing to take that on?

Fine, but let me do that separately.

Also I think this doesn't need tests. This is needed to pass existing tests in https://github.com/web-platform-tests/wpt/tree/master/fetch/api/headers/headers-no-cors.window.js.

yutakahirano:
Created #842.

PR #851, opened 2018-12-12
Export "determine nosniff".

jyasskin:
I'm not certain it's right to scope this under `header list`, but "determine nosniff" doesn't seem like it scopes itself well enough to leave it global.

I'm hoping to call this (or anything better you might suggest) from https://github.com/WICG/webpackage/pull/348, where @mikewest encouraged me to require the new format to always be served with the nosniff header.

***
<a href="https://whatpr.org/fetch/851.html" title="Last updated on Dec 19, 2018, 10:28 PM UTC (4bba761)">Preview</a> | <a href="https://whatpr.org/fetch/851/c6b3a75...4bba761.html" title="Last updated on Dec 19, 2018, 10:28 PM UTC (4bba761)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2874,8 +2874,8 @@ response <a for=/>header</a> can be used to require checking of a <a for=/>respo
 `<code>Content-Type</code>` <a for=/>header</a> against the <a for=request>destination</a> of a
 <a for=/>request</a>.
 
-<p>To <dfn export for="header list">determine nosniff</dfn>, given a <a for=/>header list</a>
-<var>list</var>, run these steps:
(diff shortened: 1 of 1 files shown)


annevk:
I left some questions there for now. If we still need to export this, I'd suggest we do it without `for`. I would expect this as a function that comes separate from the header list class.

jyasskin:
The `for` is removed now.

annevk:
I'm sorry for leaving this hanging, I guess this would still be useful at this point?

jyasskin:
I'm currently referring to it with [`<a spec="fetch">`](https://github.com/WICG/webpackage/blob/master/loading.bs#L460) to get around the fact that it wasn't exported yet. Probably more honest to get it exported. :)

PR #853, opened 2018-12-21
Add Stale While Revalidate Handling

### LINKED ISSUE #852: Add support for stale-while-revalidate handling
Add support for handling stale-while-revalidate handling.

Context is Chrome is wanting to [ship handling](
https://groups.google.com/a/chromium.org/d/msg/blink-dev/rspPrQHfFkI/rrhGyO4DBwAJ)

[RFC](https://tools.ietf.org/html/rfc5861)

dtapuska:
Define a new cache mode "async-stale-revalidated" to indicate that the
fetch stack should return a stale response if it can and aysnchronously
revalidate the response.

Fixes #852

***
<a href="https://whatpr.org/fetch/853.html" title="Last updated on May 3, 2019, 12:53 PM UTC (e17297b)">Preview</a> | <a href="https://whatpr.org/fetch/853/6a644c6...e17297b.html" title="Last updated on May 3, 2019, 12:53 PM UTC (e17297b)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1881,11 +1881,13 @@ is a <a>filtered response</a> whose
  <li><p>Return <var>newResponse</var>.
 </ol>
 
-<p>A <dfn id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose age is
+<p>A <dfn id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose
+<a href=https://tools.ietf.org/html/rfc7234#section-4.2.3>current age</a> is
 within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
 
 <p>A <dfn id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
-<a for=/>response</a> that is not a <a>fresh response</a>, and its current age is within
+<a for=/>response</a> that is not a <a>fresh response</a>, and its
+<a href=https://tools.ietf.org/html/rfc7234#section-4.2.3>current age</a> is within
 the <a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
 
 <p>A <dfn export id=concept-stale-response>stale response</dfn> is a <a for=/>response</a> is a
--- a/fetch.bs
+++ b/fetch.bs
@@ -41,6 +41,9 @@ url:https://tools.ietf.org/html/rfc7234#section-1.2.1;text:delta-seconds;type:df
     "REFERRER": {
         "aliasOf": "referrer-policy"
     },
+    "STALE-WHILE-REVALIDATE": {
+        "aliasOf": "RFC5861"
+    },
     "SW": {
         "aliasOf": "service-workers"
     },
@@ -1378,10 +1381,14 @@ Unless stated otherwise, it is unset.
 <div class="note no-backref">
  <dl>
   <dt>"<code>default</code>"
-  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network.
-  If there is a fresh response it will be used. If there is a stale response a conditional request
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network. If the HTTP cache
+  contains a matching <a>fresh response</a> it will be returned. If the HTTP cache contains a
+  matching <a>stale-while-revalidate response</a> it will be returned, and a conditional network
+  fetch will be made to update the entry in the HTTP cache. If the HTTP cache contains a matching
+  <a>stale response</a>, a conditional network fetch will be returned to update the entry in
+  the HTTP cache. Otherwise, a non-conditional network fetch will be returned to update the entry
+  in the HTTP cache.
+  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!STALE-WHILE-REVALIDATE]]
 
   <dt>"<code>no-store</code>"
   <dd>Fetch behaves as if there is no HTTP cache at all.
@@ -1494,6 +1501,10 @@ Unless stated otherwise, it is zero.
 <dfn export for=request id=concept-request-response-tainting>response tainting</dfn>,
 which is "<code>basic</code>", "<code>cors</code>", or "<code>opaque</code>".
 Unless stated otherwise, it is "<code>basic</code>".
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=no-cache-prevent-cache-control>prevent no-cache cache-control header modification flag</dfn>.
+Unless stated otherwise, it is unset.
 
 <p>A <a for=/>request</a> has an associated <dfn export for=request id=done-flag>done flag</dfn>.
 Unless stated otherwise, it is unset.
@@ -1869,6 +1880,17 @@ is a <a>filtered response</a> whose
 
  <li><p>Return <var>newResponse</var>.
 </ol>
+
+<p>A <dfn export id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose
+age is within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
+
+<p>A <dfn export id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
+<a for=/>response</a> that is not a <a>fresh response</a>, and its current age is within
+the <a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
+
+<p>A <dfn export id=concept-stale-response>stale response</dfn> is
+a <a for=/>response</a> is a response that is not a <a>fresh response</a> or a
+<a>stale-while-revalidate response</a>.
 
 
 <h4 id=miscellaneous>Miscellaneous</h4>
@@ -4065,10 +4087,11 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    `<code>If-Range</code>`, then set <var>httpRequest</var>'s
    <a for=request>cache mode</a> to "<code>no-store</code>".
 
-   <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>no-cache</code>" and
-   <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">does not contain</a>
-   `<code>Cache-Control</code>`, then <a for="header list">append</a>
-   `<code>Cache-Control</code>`/`<code>max-age=0</code>` to
+   <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>no-cache</code>",
+   <var>httpRequest</var>'s <a for=request>prevent no-cache cache-control header modification flag</a>
+   is unset, and <var>httpRequest</var>'s <a for=request>header list</a>
+   <a for="header list">does not contain</a> `<code>Cache-Control</code>`, then
+   <a for="header list">append</a> `<code>Cache-Control</code>`/`<code>max-age=0</code>` to
    <var>httpRequest</var>'s <a for=request>header list</a>.
 
    <li>
@@ -4197,32 +4220,57 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
       <!-- cache hit -->
       <ol>
-       <li><p>If <var>storedResponse</var> requires validation (i.e., it is not fresh), then set the
-       <var>revalidatingFlag</var>.
-
        <li>
-        <p>If the <var>revalidatingFlag</var> is set and <var>httpRequest</var>'s
-        <a for=request>cache mode</a> is neither "<code>force-cache</code>" nor
-        "<code>only-if-cached</code>", then:
-
+        <p>If <a for=request>cache mode</a> is "<code>default</code>" and
+        <var>storedResponse</var> is a <a>stale-while-revalidate response</a>, then:
         <ol>
-         <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
-         <a for="header list">contains</a> `<code>ETag</code>`, then <a for="header list">append</a>
-         `<code>If-None-Match</code>` with its value to <var>httpRequest</var>'s
-         <a for=request>header list</a>.
+         <li><p>Set <var>response</var> to <var>storedResponse</var>.
+         <li><p>Set <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
+         <li><p>Let <var>revalidateRequest</var> be a <a for=request>clone</a> of <var>request</var>.
+         <li><p>Set <var>revalidateRequest</var>'s <a for=request>cache mode</a> set to
+         "<code>no-cache</code>".
+         <li><p>Set <var>revalidateRequest</var>'s <a for=request>prevent no-cache cache-control header modification flag</a>.
+         <li><p>Set <var>revalidateRequest</var>'s <a for=request>service-workers mode</a> set to
+         "<code>none</code>".
+         <li>
+          <p><a>In parallel</a> perform <a for=main>main fetch</a> using <var>revalidateRequest</var>.
 
-         <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
-         <a for="header list">contains</a> `<code>Last-Modified</code>`, then
-         <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
-         <var>httpRequest</var>'s <a for=request>header list</a>.
+          <p class=note>Note the parallel fetch is only meant to update the state of the HTTP cache
+          and the response will be unused until another cache access. The stale response will be used
+          as the response to current request. The parallel fetch will be issued in the context of
+          the browsing context so if it is destroyed the request will be aborted.
+
         </ol>
+       <li>
+        <p>Otherwise:
+        <ol>
+         <li><p>If <var>storedResponse</var> is a <a>stale response</a>, then set the
+         <var>revalidatingFlag</var>.
 
-        <p class=note>See also the
-        "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
-        chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+         <li>
+          <p>If the <var>revalidatingFlag</var> is set and <var>httpRequest</var>'s
+          <a for=request>cache mode</a> is neither "<code>force-cache</code>" nor
+          "<code>only-if-cached</code>", then:
 
-       <li><p>Otherwise, set <var>response</var> to <var>storedResponse</var> and set
-       <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
+          <ol>
+           <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
+           <a for="header list">contains</a> `<code>ETag</code>`, then <a for="header list">append</a>
+           `<code>If-None-Match</code>` with its value to <var>httpRequest</var>'s
+           <a for=request>header list</a>.
+
+           <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
+           <a for="header list">contains</a> `<code>Last-Modified</code>`, then
+           <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
+           <var>httpRequest</var>'s <a for=request>header list</a>.
+          </ol>
+
+          <p class=note>See also the
+          "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
+          chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+
+         <li><p>Otherwise, set <var>response</var> to <var>storedResponse</var> and set
+         <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
+        </ol>
       </ol>
     </ol>
   </ol>
@@ -6942,6 +6990,7 @@ Clement Pellerin,
 Collin Jackson,
 Daniel Robertson,
 Daniel Veditz,
+Dave Tapuska,
 David Benjamin,
 David Håsäther,
 David Orchard,
--- a/fetch.bs
+++ b/fetch.bs
@@ -1882,13 +1882,13 @@ is a <a>filtered response</a> whose
 </ol>
 
 <p>A <dfn id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose
-<a href=https://tools.ietf.org/html/rfc7234#section-4.2.3>current age</a> is
-within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
+<a href=https://tools.ietf.org/html/rfc7234#section-4.2.3>current age</a> is within its
+<a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
 
 <p>A <dfn id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
-<a for=/>response</a> that is not a <a>fresh response</a>, and its
-<a href=https://tools.ietf.org/html/rfc7234#section-4.2.3>current age</a> is within
-the <a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
+<a for=/>response</a> that is not a <a>fresh response</a> and whose
+<a href=https://tools.ietf.org/html/rfc7234#section-4.2.3>current age</a> is within the
+<a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
 
 <p>A <dfn export id=concept-stale-response>stale response</dfn> is a <a for=/>response</a> is a
 <a for=/>response</a> that is not a <a>fresh response</a> or a
@@ -4247,10 +4247,10 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
           <p><a>In parallel</a>, perform <a for=main>main fetch</a> using
           <var>revalidateRequest</var>.
 
-          <p class=note>Note the parallel fetch is only meant to update the state of the HTTP cache
-          and the response will be unused until another cache access. The stale response will be used
-          as the response to current request. The parallel fetch will be issued in the context of
-          the browsing context so if it is destroyed the request will be aborted.
+          <p class=note>This fetch is only meant to update the state of the HTTP cache and the
+          response will be unused until another cache access. The stale response will be used as the
+          response to current request. This fetch is issued in the context of a client so if it goes
+          away the request will be terminated.
         </ol>
 
        <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4223,8 +4223,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
       <!-- cache hit -->
       <ol>
        <li>
-        <p>If <a for=request>cache mode</a> is "<code>default</code>" and <var>storedResponse</var>
-        is a <a>stale-while-revalidate response</a>, then:
+        <p>If <a for=request>cache mode</a> is "<code>default</code>", <var>storedResponse</var>
+        is a <a>stale-while-revalidate response</a>, and <var>httpRequest</var>'s
+        <a for=request>client</a> is not null, then:
 
         <ol>
          <li><p>Set <var>response</var> to <var>storedResponse</var>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -1387,8 +1387,8 @@ Unless stated otherwise, it is unset.
   fetch will be made to update the entry in the HTTP cache. If the HTTP cache contains a matching
   <a>stale response</a>, a conditional network fetch will be returned to update the entry in
   the HTTP cache. Otherwise, a non-conditional network fetch will be returned to update the entry
-  in the HTTP cache.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]] [[!STALE-WHILE-REVALIDATE]]
+  in the HTTP cache. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  [[!STALE-WHILE-REVALIDATE]]
 
   <dt>"<code>no-store</code>"
   <dd>Fetch behaves as if there is no HTTP cache at all.
@@ -1881,15 +1881,15 @@ is a <a>filtered response</a> whose
  <li><p>Return <var>newResponse</var>.
 </ol>
 
-<p>A <dfn export id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose
-age is within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
+<p>A <dfn id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose age is
+within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
 
-<p>A <dfn export id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
+<p>A <dfn id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
 <a for=/>response</a> that is not a <a>fresh response</a>, and its current age is within
 the <a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
 
-<p>A <dfn export id=concept-stale-response>stale response</dfn> is
-a <a for=/>response</a> is a response that is not a <a>fresh response</a> or a
+<p>A <dfn export id=concept-stale-response>stale response</dfn> is a <a for=/>response</a> is a
+<a for=/>response</a> that is not a <a>fresh response</a> or a
 <a>stale-while-revalidate response</a>.
 
 
@@ -4221,28 +4221,39 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
       <!-- cache hit -->
       <ol>
        <li>
-        <p>If <a for=request>cache mode</a> is "<code>default</code>" and
-        <var>storedResponse</var> is a <a>stale-while-revalidate response</a>, then:
+        <p>If <a for=request>cache mode</a> is "<code>default</code>" and <var>storedResponse</var>
+        is a <a>stale-while-revalidate response</a>, then:
+
         <ol>
          <li><p>Set <var>response</var> to <var>storedResponse</var>.
+
          <li><p>Set <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
-         <li><p>Let <var>revalidateRequest</var> be a <a for=request>clone</a> of <var>request</var>.
+
+         <li><p>Let <var>revalidateRequest</var> be a <a for=request>clone</a> of
+         <var>request</var>.
+
          <li><p>Set <var>revalidateRequest</var>'s <a for=request>cache mode</a> set to
          "<code>no-cache</code>".
-         <li><p>Set <var>revalidateRequest</var>'s <a for=request>prevent no-cache cache-control header modification flag</a>.
+
+         <li><p>Set <var>revalidateRequest</var>'s
+         <a for=request>prevent no-cache cache-control header modification flag</a>.
+
          <li><p>Set <var>revalidateRequest</var>'s <a for=request>service-workers mode</a> set to
          "<code>none</code>".
+
          <li>
-          <p><a>In parallel</a> perform <a for=main>main fetch</a> using <var>revalidateRequest</var>.
+          <p><a>In parallel</a>, perform <a for=main>main fetch</a> using
+          <var>revalidateRequest</var>.
 
           <p class=note>Note the parallel fetch is only meant to update the state of the HTTP cache
           and the response will be unused until another cache access. The stale response will be used
           as the response to current request. The parallel fetch will be issued in the context of
           the browsing context so if it is destroyed the request will be aborted.
-
         </ol>
+
        <li>
         <p>Otherwise:
+
         <ol>
          <li><p>If <var>storedResponse</var> is a <a>stale response</a>, then set the
          <var>revalidatingFlag</var>.
@@ -4254,9 +4265,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
           <ol>
            <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
-           <a for="header list">contains</a> `<code>ETag</code>`, then <a for="header list">append</a>
-           `<code>If-None-Match</code>` with its value to <var>httpRequest</var>'s
-           <a for=request>header list</a>.
+           <a for="header list">contains</a> `<code>ETag</code>`, then
+           <a for="header list">append</a> `<code>If-None-Match</code>` with its value to
+           <var>httpRequest</var>'s <a for=request>header list</a>.
 
            <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
            <a for="header list">contains</a> `<code>Last-Modified</code>`, then
--- a/fetch.bs
+++ b/fetch.bs
@@ -4225,7 +4225,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
        <li>
         <p>If <a for=request>cache mode</a> is "<code>default</code>", <var>storedResponse</var>
         is a <a>stale-while-revalidate response</a>, and <var>httpRequest</var>'s
-        <a for=request>client</a> is not null, then:
+        <a for=request>client</a> is non-null, then:
 
         <ol>
          <li><p>Set <var>response</var> to <var>storedResponse</var>.


fetch.bs
```diff
@@ -4482,6 +4492,11 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
  <li><p>If <i>authentication-fetch flag</i> is set, then create an <a>authentication entry</a>
  for <var>request</var> and the given realm.
 
+ <li><p>If <i>revalidatingFlag</i> is set and <var>httpRequest</var>'s <a for=request>cache mode</a>
+ is "<code>async-stale-revalidated</code>" then asynchronously perform a
```
annevk: This needs a comma before then. "asynchronously" is not a thing, perhaps "in parallel"?

fetch.bs
```diff
@@ -4482,6 +4492,11 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+ <a>HTTP-network-or-cache fetch</a> using <var>request</var> with <a for=request>cache mode</a>
```
annevk: This will bypass the service worker. Is that okay? It seems problematic that you can create a network request like that.

kinu: SwR is a concept of HTTP cache, which basically sits below the service worker, so service workers being not able to observe revalidation requests sounds somewhat natural to me from implementation pov.  That's said... in this way service workers may keep holding stale responses in cache storage, and now that we're talking about Fetch layer's SwR handling, it might be worth discussing.

annevk:
cc @jakearchibald

fetch.bs
```diff
@@ -1542,6 +1549,11 @@ whose <a for=request>destination</a> is "<code>audio</code>", "<code>audioworkle
 "<code>script</code>", "<code>style</code>", "<code>track</code>", "<code>video</code>",
 "<code>xslt</code>", or the empty string.
 
+<p>A <dfn export>stale permitting subresource request</dfn> is a <a for=/>request</a>
+whose <a for=request>destination</a> is "<code>font</code>", "<code>image</code>",
+"<code>script</code>", "<code>style</code>", "<code>xslt</code>", or
+<a for=request>initiator</a> is "<code>prefetch</code>".
```
kinu: Allowing stale request for prefetch feels slightly odd to me because responses for prefetch won't be used downstream (per definition).  This one can just trigger a revalidation therefore the next request (the real one that wants to use the response) could use a fresh resource?

dtapuska: Yes I agree. Removed.

dtapuska:
This pull request now has tests that have been merged, I think the label can be removed.

dtapuska:
@annevk 

We don't do anything different for 401s or redirects. It is exactly the same as another sub-resource; standard processing occurs. This is the benefit of having the revalidation associated with the context that initiated the resource request.

The text indicates "in parallel perform" is that the naming? I indicated "Done" on the comment above. Is that the naming you are referring to?

Did Kinuko's response not suffice? Top level requests are sent to the service worker if associated. Can you explain why you think the revalidation should go back to the service worker? Fetches created from the service worker have a "" destination.

annevk:
@dtapuska if standard processing occurs, the current PR is wrong, right? It invokes an algorithm that doesn't handle redirects.

The naming comment is about "async" in the new enum value. Fetching is already asynchronous for all intents and purposes so this name is rather confusing.

As for the service worker, we don't have many fetches that can bypass it, so the problem would be that now there is a way and the service worker can do nothing about it. That breaks a design decision around service workers.

fetch.bs
```diff
@@ -4153,6 +4164,11 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    `<code>User-Agent</code>`/<a>default `<code>User-Agent</code>` value</a>
    to <var>httpRequest</var>'s <a for=request>header list</a>.
 
+   <li><p>If <var>httpRequest</var>'s <a for=request>allows stale content flag</a> is set and
+   <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">contains</a>
+   <code>Cache-Control: Must-Revalidate</code>, then unset <var>httpRequest</var>'s
+   <a for=request>allows stale content flag</a>.
```
annevk: I'm sorry, I had missed the approach changed substantially (at least I thought there was an enum involved originally). This doesn't quite work. You need to extract the header value and parse/compare it in some manner.

annevk:
I'm not sure, it's still not clear to me what the processing model is, and I'd like someone other than me to judge it's suddenly okay to bypass service workers when making network requests. That seems like a new primitive and rather isolated at that.

yoavweiss:
If we'd re-route the async revalidation requests through "HTTP fetch", that means they go again through SW, preflight and CORS checks before properly handling redirects.

Seems to me that making sure these requests don't go through SW *again* would be fine, so setting the SW mode to "none" probably makes sense.

Regarding preflights, if a preflight was triggered during the initial request, but didn't have `Access-Control-Max-Age` header, this will trigger a preflight request *again*. That doesn't seem ideal :/

Repeating the CORS check OTOH seems necessary, as the response's CORS headers may have changed between the stale cached response and the shiny new one.

Does that make sense? If so, seems like we'd need both SW mode as "none" as well as another flag that bypasses preflights in this particular case where we're repeating a request that was just triggered and didn't fail the preflight last time around.

/cc @jakearchibald and @yutakahirano for their opinions

annevk:
Do we have tests for all those scenarios? In particular the CORS-related ones seems good to extensively test in case of subtle security issues.

annevk:
The problem I see is that in principle SWs are supposed to handle all requests. And now we have some kind of request that gets made independent of that. So if the SW say wants to control all network access for an origin, that's no longer something it can do.

I guess in general I'm a bit skeptical of giving HTTP caches this capability as this is the kind of problem SWs are supposed to help solve.

yutakahirano:
Normal requests and CORS preflights live in different caches, so repating preflight sounds reasonable to me.

dtapuska:
I've pushed the change that loops back to the main fetch. 

Unfortunately I don't think you can test the preflight stuff because it is only applicable to unsafe resources right? And we are only allowing SWR on a subset of safe resources.

If the redirect mode is "follow" a service worker won't see the redirects so "SWs don't see all requests" nor do they see the requests with the revalidation headers.

yutakahirano:
If you want to limit the use of S-W-R, I think you need to reset the flag on request at "14: if input is not empty",  in [the request constructor](https://fetch.spec.whatwg.org/#dom-request).

fetch.bs
```diff
@@ -4482,6 +4498,13 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
  <li><p>If <i>authentication-fetch flag</i> is set, then create an <a>authentication entry</a>
  for <var>request</var> and the given realm.
 
+ <li><p>If <i>revalidatingFlag</i> and <var>request</var>'s <a for=request>allows stale content flag</a>
+ are set, then in parallel perform <a for=main>main fetch</a> using <var>request</var> with
```
yoavweiss: Why "main fetch" and not "HTTP fetch"?

dtapuska: There are some CORS checks in main and it is consistent with the other parts of the fetch spec.

dtapuska:
To summarize people's comments above:
- SW isn't notified of the revalidation request. I think this makes sense. The SW was already told about the initial request and can't initiate a SWR request via fetch if it delegates it.
- SWR is restricted based on destination. I can define it as for all non-prefetch subresources with the unsafe request flag not set instead. What are the thoughts about this?

fetch.bs
dtapuska: I've made it clear that we are looking for a directive in the header. Please let me know if this isn't sufficient. I've looked for other examples of parsing the cache-control header but didn't find any.

fetch.bs
```diff
@@ -1410,9 +1413,11 @@ Unless stated otherwise, it is unset.
  <dl>
   <dt>"<code>default</code>"
   <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network.
-  If there is a fresh response it will be used. If there is a stale response a conditional request
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  If there is a fresh response it will be used. If there is a stale response it will be returned immediately
+  if the request <a for=request>allows stale content flag</a> is set. A conditional request will be created
```
jakearchibald: I feel like I'm missing something, but isn't this going to cause most requests to return stale content? It looks like "allows stale content flag" will be frequently set, as long as it "is a request that is a  subresource request and that does not have the unsafe-request flag set. "

jakearchibald:
I'm struggling to get my head around the changes here, but I'm happy for stale-while-revalidate requests to bypass the service worker if the service worker allowed the initial request to go to the network (even though it ultimately came from the cache).

For instance, if I used a cache mode of "only-if-cached", there shouldn't be a stale-while-revalidate request to the network.

jakearchibald:
Ah, things make a lot more sense in the latest version.

Right now stale-while-revalidate seems like a mode of the request, whereas https://tools.ietf.org/html/rfc5861 suggests it's triggered by the entry in the cache. Is there a plan to add both to the fetch spec?

dtapuska:
So by default. Fetches that enter the fetch spec (not via Request.init(...))) ie. HTML spec and friends that are GET subresources may set the allows stale content flag. Which will cause an additional request that doesn't have the revalidation bit set on if the response received is stale.

If you are asking will this be eventually exposed such that service workers can explicitly request stale content it can be done in the future via an explicit API (say on the cache mode). But I don't believe it is needed at this time.

jakearchibald:
@dtapuska interesting. So individual APIs will have to opt into this by setting the mode to `"stale-while-revalidate"`?

mnot:
@dtapuska you seem to be conflating "the request allows stale content" and "the use of stale content will kick off a revalidating request asynchronously." The former is `max-stale` in [HTTP caching](https://httpwg.org/specs/rfc7234.html#cache-request-directive.max-stale); the latter is `stale-while-revalidate`.

I was concerned about similar things to @jakearchibald above in #852; HTTP already defines how responses can be used for SwR, but there's no hint of that here. 

@annevk this seems to be a recurring pattern; HTTP defines what happens on the wire, and Fetch defines a corresponding API for each feature. If that's going to be normal going forward, it'd probably be best to document the division of labour (as it were) so that people don't get confused.

Also, it'd be good to get alignment on whether or not browsers pay attention to request `Cache-Control` (see [tests](https://cache-tests.fyi/#cc-request)) and how that interacts with these Fetch directives.

dtapuska:
@mnot In looking at it again I can see how it is confusing. I've added definitions indicating "stale response in revalidate window" and "stale response". Hopefully that clarifies the conditions.

@jakearchibald I've removed the restriction on fetch as discussed offline.

fetch.bs
```diff
@@ -4295,31 +4326,40 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+          <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
+          <var>httpRequest</var>'s <a for=request>header list</a>.
+         </ol>
 
         <p class=note>See also the
         "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
         chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
 
-       <li><p>Otherwise, set <var>response</var> to <var>storedResponse</var> and set
+        <li><p>Otherwise, set <var>response</var> to <var>storedResponse</var> and set
```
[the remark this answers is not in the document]
jakearchibald: It might be because of the messed-up if statement, but it kinda looks like a stale response could be used even if the stale content mode is "disallow".

fetch.bs
```diff
@@ -1526,6 +1531,10 @@ Unless stated otherwise, it is zero.
 which is "<code>basic</code>", "<code>cors</code>", or "<code>opaque</code>".
 Unless stated otherwise, it is "<code>basic</code>".
 
+<p>A <a for=/>request</a> has an associated <dfn export for=request id=stale-content-mode>stale content mode</dfn>,
+which is "<code>allow</code>", "<code>disallow</code>" or "<code>default</code>". Unless stated
+otherwise, it is "<code>default</code>".
```
jakearchibald: I'm not 100% sure, but I'm not sure this is needed. Can't the current values of "cache mode" handle this?

fetch.bs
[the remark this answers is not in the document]
dtapuska: An <ol> is missing here. The last Otherwise is supposed to be inside the main Otherwise.

fetch.bs
```diff
@@ -4295,31 +4326,40 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
-
-         <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
-         <a for="header list">contains</a> `<code>Last-Modified</code>`, then
-         <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
-         <var>httpRequest</var>'s <a for=request>header list</a>.
+         <li><p>Set <var>response</var> to <var>storedResponse</var>.
+         <li><p>Set <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
+         <li><p>In parallel perform <a for=main>main fetch</a> using <var>request</var> with
+          <ul class=brief>
+          <li><p><i>stale content mode</i> set to "<code>disallow</code>" and
```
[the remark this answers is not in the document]
dtapuska: `Reload` doesn't work here because it won't cause a revalidation but a forced network load. Nor do we want to append any special Cache-Control headers.

I've added a note.

fetch.bs
```diff
@@ -1901,6 +1910,13 @@ is a <a>filtered response</a> whose
  <li><p>Return <var>newResponse</var>.
 </ol>
 
+<p>A <dfn export id=concept-stale-response>stale response</dfn> is a <a for=/>response</a> where
+the freshness lifetime calculation (excluding the stale-while-revalidate directive) has not
+exceeded the current age.
```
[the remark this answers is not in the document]
dtapuska: Yes you are correct

fetch.bs
```diff
@@ -1901,6 +1910,13 @@ is a <a>filtered response</a> whose
...
+
+<p>A <dfn export id=concept-stale-response>stale response inside revalidate window</dfn> is
+a <a>stale response</a> where the freshness lifetime calculation (including the stale-while-revalidate
```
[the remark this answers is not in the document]
dtapuska: Probably using explicit text here is better. Done.

fetch.bs
```diff
@@ -4163,6 +4179,21 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    `<code>If-Range</code>`, then set <var>httpRequest</var>'s
    <a for=request>cache mode</a> to "<code>no-store</code>".
 
+   <li><p>If <var>httpRequest</var>'s <a for=request>stale content mode</a> is
+   `<code>default</code>` then:
+
+    <ol>
+     <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>default</code>",
+     <var>httpRequest</var>'s <a for=request>method</a> is `<code>GET</code>`, and
```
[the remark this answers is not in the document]
dtapuska: No it isn't but perhaps the "stale content mode" name definitely gets confusing here. Not sure what to do.

fetch.bs
dtapuska: No they can't. There is no "must-revalidate" mode in the cache-mode. Reload causes it to skip the cache.

fetch.bs
```diff
@@ -1410,9 +1413,11 @@ Unless stated otherwise, it is unset.
  <dl>
   <dt>"<code>default</code>"
   <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network.
-  If there is a fresh response it will be used. If there is a stale response a conditional request
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  If there is a fresh response it will be used. If there is a <a>stale response inside revalidate window</a>
+  and the request has the <a for=request>stale content mode</a> set it will be returned
```
kinu: Looks like stale content mode is no longer a boolean but has a value, this text needs to be updated as well?

fetch.bs
kinu: What do we think about adding a new cache mode for that?  Currently it says the 'stale content mode' change is only applied for 'default', and other cache mode says something about cache behavior too, so it feels the mode can be one of the cache mode?

jakearchibald: If the only difference between "no-cache" and a SWR update request is the SWR update request doesn't add the `Cache-Control: max-age: 0` header, should we just add a flag to prevent that step?

fetch.bs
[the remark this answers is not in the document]
jakearchibald: We shouldn't call a fresh response a "stale response".

fetch.bs
dtapuska: This text is no longer present.

fetch.bs
```diff
@@ -1410,9 +1413,11 @@ Unless stated otherwise, it is unset.
...
+  immediately. If there is a <a>stale response</a> a conditional request will be created and the
+  response will update the HTTP cache. Otherwise, a normal network request will be issued.
```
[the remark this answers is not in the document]
dtapuska: Done

fetch.bs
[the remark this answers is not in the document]
dtapuska: Done

fetch.bs
dtapuska: I've added a "no-cache local only flag" and then I can use the no-cache mode. If you have a better name I can change it.

fetch.bs
```diff
@@ -1896,6 +1908,16 @@ is a <a>filtered response</a> whose
  <li><p>Return <var>newResponse</var>.
 </ol>
 
+<p>A <dfn export id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> where
```
jakearchibald: Remove "where"

fetch.bs
```diff
@@ -1404,10 +1407,16 @@ Unless stated otherwise, it is unset.
...
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network:
+  <ol>
+   <li>If the HTTP cache contains a matching <a>fresh response</a> it will be returned.
+   <li>If the HTTP cache contains a matching <a>stale-while-revalidate response</a> it will be returned,
+   and a conditional network fetch will be made to update the entry in the HTTP cache.
+   <li>If the HTTP cache contains a matching <a>stale response</a>, a conditional network will be created
+   and returned.
+   <li>Otherwise, a non-conditioanl network fetch will be returned.
```
jakearchibald: "conditional"

fetch.bs
```diff
@@ -1404,10 +1407,16 @@ Unless stated otherwise, it is unset.
 <div class="note no-backref">
  <dl>
   <dt>"<code>default</code>"
-  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network.
-  If there is a fresh response it will be used. If there is a stale response a conditional request
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network:
+  <ol>
+   <li>If the HTTP cache contains a matching <a>fresh response</a> it will be returned.
```
jakearchibald: I think @annevk likes these to be `<li><p>`

kinu:
Took another look, just wanted to say that this looks really good now :) (Thanks Jake for thorough comments)

jakearchibald:
@annevk us Chromefolk are happy with this now. Want to take another look?

fetch.bs
```diff
@@ -1404,10 +1407,16 @@ Unless stated otherwise, it is unset.
...
-  If there is a fresh response it will be used. If there is a stale response a conditional request
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network:
+  <ol>
+   <li><p>If the HTTP cache contains a matching <a>fresh response</a> it will be returned.
+   <li><p>If the HTTP cache contains a matching <a>stale-while-revalidate response</a> it will be returned,
+   and a conditional network fetch will be made to update the entry in the HTTP cache.
+   <li><p>If the HTTP cache contains a matching <a>stale response</a>, a conditional network will be created
+   and returned.
```
annevk: This drops the cache updating that the original text had. (Same for the next point.)

Also, perhaps this should be a long paragraph instead similar to the original? It's not really meant to be an algorithm. Just an explanation.

fetch.bs
```diff
@@ -1896,6 +1909,16 @@ is a <a>filtered response</a> whose
...
+<p>A <dfn export id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose
+age is within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
+
+<p>A <dfn export id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
+<a for=/>response</a> that is not a <a>fresh response</a>, and its current age is within
+the <a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
+
+<p>A <dfn export id=concept-stale-response>stale response</dfn> is
+a <a for=/>response</a> is a response that is not a <a>fresh response</a> or a
+<a>stale-while-revalidate response</a>.
```
annevk: Need an additional newline here since there were two before the `<h4>`.

fetch.bs
```diff
@@ -4151,7 +4174,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    `<code>If-Range</code>`, then set <var>httpRequest</var>'s
    <a for=request>cache mode</a> to "<code>no-store</code>".
 
-   <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>no-cache</code>" and
+   <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is "<code>no-cache</code>",
+   <a for=request>prevent no-cache cache-control header modification flag</a> is unset and
```
annevk: You need to lead this with _httpRequest_ again assuming that's where you grab the flag from.

annevk: Also need an Oxford comma.

fetch.bs
```diff
@@ -4283,32 +4307,50 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
       <!-- cache hit -->
       <ol>
-       <li><p>If <var>storedResponse</var> requires validation (i.e., it is not fresh), then set the
-       <var>revalidatingFlag</var>.
-
-       <li>
-        <p>If the <var>revalidatingFlag</var> is set and <var>httpRequest</var>'s
-        <a for=request>cache mode</a> is neither "<code>force-cache</code>" nor
-        "<code>only-if-cached</code>", then:
+       <li><p>If <a for=request>cache mode</a> is "<code>default</code>" and
```
annevk: The indenting is wrong here. If an `<li>` has multiple nested elements you need to put the `<p>` on a newline (and indent with an extra space).

fetch.bs
```diff
@@ -4283,32 +4307,50 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+        <var>storedResponse</var> is a <a>stale-while-revalidate response</a>, then:
+        <ol>
+         <li><p>Set <var>response</var> to <var>storedResponse</var>.
+         <li><p>Set <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
+         <li><p>In parallel perform <a for=main>main fetch</a> using <var>request</var> with
```
annevk: Same here.

annevk: Also use `<a>In parallel</a>`?

fetch.bs
```diff
@@ -4283,32 +4307,50 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+          <ul class=brief>
```
annevk: Indentation is wrong here.

fetch.bs
```diff
@@ -4283,32 +4307,50 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+          <li><p><a for=request>cache mode</a> set to "<code>no-cache</code>".
+          <li><p><a for=request>prevent no-cache cache-control header modification flag</a> set.
+          <li><p><a for=request>service-workers mode</a> set to "<code>none</code>".
+         </ul>
```
annevk: You're modifying the request here. It seems to me you'd need to copy it first?

fetch.bs
```diff
@@ -4283,32 +4307,50 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
-         `<code>If-None-Match</code>` with its value to <var>httpRequest</var>'s
-         <a for=request>header list</a>.
-
-         <li><p>If <var>storedResponse</var>'s <a for=response>header list</a>
-         <a for="header list">contains</a> `<code>Last-Modified</code>`, then
-         <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
-         <var>httpRequest</var>'s <a for=request>header list</a>.
+         <li><p>If <var>storedResponse</var> is a <a>stale response</a>, then set the
+          <var>revalidatingFlag</var>.
+          <p>If the <var>revalidatingFlag</var> is set and <a for=request>cache mode</a>
```
annevk: It seems this should be a new list item?

fetch.bs
```diff
@@ -4283,32 +4307,50 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+          is not one of "<code>force-cache</code>", "<code>only-if-cached</code>", then:
```
annevk: This sentence doesn't make sense? Now that there's no new enum value the original wording should be restored I think.

fetch.bs
annevk: This also dropped where you got "cache mode" from. Which request?

fetch.bs
dtapuska: done

fetch.bs
dtapuska: done

fetch.bs
dtapuska: done

fetch.bs
dtapuska: done

fetch.bs
dtapuska: done

fetch.bs
dtapuska: done

fetch.bs
dtapuska: done

dtapuska:
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1534698
Webkit: https://bugs.webkit.org/show_bug.cgi?id=195624
Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=348877

fetch.bs
```diff
@@ -4283,32 +4306,57 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
-         <a for="header list">append</a> `<code>If-Modified-Since</code>` with its value to
-         <var>httpRequest</var>'s <a for=request>header list</a>.
-        </ol>
+         <li><p>Set <var>response</var> to <var>storedResponse</var>.
+         <li><p>Set <var>response</var>'s <a for=response>cache state</a> to "<code>local</code>".
+         <li><p>Let <var>revalidateRequest</var> be a <a for=request>clone</a> of <var>request</var>.
+         <li><p>Set <var>revalidateRequest</var>'s <a for=request>cache mode</a> set to
+         "<code>no-cache</code>".
+         <li><p>Set <var>revalidateRequest</var>'s <a for=request>prevent no-cache cache-control header modification flag</a>
+         set.
```
annevk: You can drop "set" at the end here.

fetch.bs
```diff
@@ -4283,32 +4306,57 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+         <li><p>Set <var>revalidateRequest</var>'s <a for=request>service-workers mode</a> set to
+         "<code>none</code>".
+         <li><p><a>In parallel</a> perform <a for=main>main fetch</a> using
```
annevk: This `<li>` contains two children, so `<p>` needs to be on an indented newline.

fetch.bs
```diff
@@ -4283,32 +4306,57 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+         <p class=note>Note the parallel fetch is only meant to update the state of the HTTP cache
+         and the response will be unused until another cache access. The stale response will be used
+         as the response to current request. The parallel fetch will be issued in the context of
+         the browsing context so if it is destroyed the request will be aborted.
 
-        <p class=note>See also the
-        "<a href=https://tools.ietf.org/html/rfc7234#section-4.3.4>Sending a Validation Request</a>"
-        chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]].
+        </ol>
+       <li><p>Otherwise:
```
annevk: Indented newline.

fetch.bs
```diff
@@ -1404,10 +1407,14 @@ Unless stated otherwise, it is unset.
...
   <dt>"<code>default</code>"
-  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network.
-  If there is a fresh response it will be used. If there is a stale response a conditional request
-  will be created, and a normal request otherwise. It then updates the HTTP cache with the response.
-  [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
+  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network. If the HTTP cache
+  contains a matching <a>fresh response</a> it will be returned. If the HTTP cache contains a
+  matching <a>stale-while-revalidate response</a> it will be returned, and a conditional network
+  fetch will be made to update the entry in the HTTP cache. If the HTTP cache contains a matching
+  <a>stale response</a>, a conditional network will be created and returned to update the entry in
```
annevk: network fetch*

annevk: Also, "created and" doesn't seem needed given the other wording?

uasan:
Hello.

You are planning to implement support `stale-if-error`?
https://tools.ietf.org/html/rfc5861#section-4

This real cases, used in reverse proxy Nginx
https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_use_stale

Thanks.

dtapuska:
@uasan I don't have a plan for it at this time.

dtapuska:
@mayhemer  does opening https://bugzilla.mozilla.org/show_bug.cgi?id=1536511 mean that you have interest in proceeding with this PR. Ultimately we don't want to leave the PR in limbo forever.

mayhemer:
> 
> 
> @mayhemer does opening https://bugzilla.mozilla.org/show_bug.cgi?id=1536511 mean that you have interest in proceeding with this PR. Ultimately we don't want to leave the PR in limbo forever.

Yes, we intend to.  There doesn't seem to be any problem with it immediately visible, it's beneficial and easy to implement.

As I wrote at https://github.com/mozilla/standards-positions/issues/144#issuecomment-474475568 we should keep the HTTP-cache and fetch() parts separate.  We now have two bug (see that comment.)  The fetch() changes are still in development, in this thread, is that so?

dtapuska:
The fetch changes are done as proposed in Chrome but to merge the PR into the formal specification WHATWG requires two browser vendors to commit to it (otherwise it remains a PR indefinitely). And that is the question I'm asking.

mayhemer:
> 
> 
> The fetch changes are done as proposed in Chrome but to merge the PR into the formal specification WHATWG requires two browser vendors to commit to it (otherwise it remains a PR indefinitely). And that is the question I'm asking.

Aha, now it's clear to me.  Can you please add here or in our (mozilla) bug https://bugzilla.mozilla.org/show_bug.cgi?id=1534698 direct reference to the final form of the specification changes for fetch that Chrome is proposing?  

(In that bug's description, https://crbug.com/348877 is referring only the HTTP-cache part (mozilla mirror bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1536511).  The other link is for this thread.  And the last one just points at the mozilla positions discussion.)

Thanks.

coreyandersonbuilders:
Ignorance is bliss

mayhemer:
> 
> 
> The fetch changes are done as proposed in Chrome but to merge the PR into the formal specification WHATWG requires two browser vendors to commit to it (otherwise it remains a PR indefinitely). And that is the question I'm asking.

To properly answer: the fetch changes look good to me, I agree with them as they are stated in the proposed diff at the time of writing this comment.

There are two small things to add/note: 

1. as the background revalidation request's only constrain is to run at the original request's browsing context and be aborted with destruction of the context, we (Gecko) intend to lower its priority to a minimum to minimize its effect on the page load time during the critical loading phase.  It means, the revalidation request will hit the network likely no sooner than after the page load is finished, if a request for a stale-while-revalidate response is made early.

2. there is nothing said about possible coalescing of more than one such revalidation request.  the initial implementation for Gecko will not do any explicit coalescing, only implicit when two or more revalidation requests to the same URI happen to be started at the same time and race the common cache entry.  I intend to add telemetry probes on how this whole mechanism is going to work and if we find coalescing should but does not happen in significant(tm) amount of time, explicit coalescing may be implemented.

dtapuska:
1. We do this in Chrome as well, we fetch it with the very low priority as (same as beacons, pings, csp reports and lazy load)
2. Chrome's network stack does do coalescing of pending requests.

dtapuska:
I've rebased the patch on ToT. 
@annevk Now that we have confirmed support from Firefox and there is tests. Can we remove the "needs implementor interest" and "needs tests" and merge this PR?

annevk:
@dtapuska can you create a WPT PR that makes the tests no longer tentative? It'd also be nice to put the non-test resources into a `resources/` subdirectory. I'm also a little concerned about https://bugzilla.mozilla.org/show_bug.cgi?id=1540216.

I'll take some time to create and push editorial fixup here for you to review.

fetch.bs
```diff
@@ -1870,6 +1881,17 @@ is a <a>filtered response</a> whose
  <li><p>Return <var>newResponse</var>.
 </ol>
 
+<p>A <dfn id=concept-fresh-response>fresh response</dfn> is a <a for=/>response</a> whose age is
```
annevk: Should we link "age"?

fetch.bs
```diff
@@ -1870,6 +1881,17 @@ is a <a>filtered response</a> whose
...
+within its <a href=https://tools.ietf.org/html/rfc7234#section-4.2.1>freshness lifetime</a>.
+
+<p>A <dfn id=concept-stale-while-revalidate-response>stale-while-revalidate response</dfn> is a
+<a for=/>response</a> that is not a <a>fresh response</a>, and its current age is within
```
annevk: Is "current age" different from age?

fetch.bs
```diff
@@ -1870,6 +1881,17 @@ is a <a>filtered response</a> whose
...
+the <a href=https://tools.ietf.org/html/rfc5861#section-3>stale-while-revalidate lifetime</a>.
```
annevk: This reference doesn't define this term, which is a bit confusing.

fetch.bs
mayhemer: yes, according the rfc.  there is actually nothing like pure "age".  but there are apparent, cached, corrected received, corrected initial and current age.  current age is based on all of the others and is the one to compare with freshness lifetime.

fetch.bs
dtapuska: I've adjusted it to current age and lined it.

fetch.bs
dtapuska: I linked it to the definition of current age.

fetch.bs
dtapuska: Do you have a better suggestion? I linked it to the closest definition but I don't necessarily want to rewrite that spec.

dtapuska:
> @dtapuska can you create a WPT PR that makes the tests no longer tentative? It'd also be nice to put the non-test resources into a `resources/` subdirectory. I'm also a little concerned about https://bugzilla.mozilla.org/show_bug.cgi?id=1540216.
> 
> I'll take some time to create and push editorial fixup here for you to review.

PR for removing tentative and moving into resources is here:
https://github.com/web-platform-tests/wpt/pull/16337

mayhemer:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1543742

fetch.bs
```diff
@@ -4197,32 +4222,68 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+         "<code>none</code>".
+
+         <li>
+          <p><a>In parallel</a>, perform <a for=main>main fetch</a> using
+          <var>revalidateRequest</var>.
+
+          <p class=note>This fetch is only meant to update the state of the HTTP cache and the
+          response will be unused until another cache access. The stale response will be used as the
+          response to current request. This fetch is issued in the context of a client so if it goes
+          away the request will be terminated.
```
annevk: How is this enforced? Shouldn't we null check revalidateRequest's client before going here then?

fetch.bs
annevk: I'm not sure, it's not really my area of expertise. I guess we can leave it for now...

fetch.bs
dtapuska: I really don't know how to answer this. It is a note so it isn't part of the algorithm just calling out that the context of the revalidation occurs is in the browsing context's scope. Not some other random process.

jakearchibald: @dtapuska not all fetches have a browsing context. Your note says "This fetch is issued in the context of a client" but that guarantee isn't made in the prose. So, the question is, what do you want to happen to fetches that don't have a client? My gut feeling is that stale-while-revalidate should work there too.

@annevk can you give an example of a fetch without a client? I can never remember

annevk: OCSP, certain downloads, certain favicon fetches. (Not all of these are well-defined.)

dtapuska: And these are precisely things that Chrome's SWR won't work on because they are not fetched in the context of an agent. It will work for agents (workers, windows, service workers). 

@annevk would you feel more comfortable omitting this text entirely?

annevk: I'd be more comfortable if that was specified, as per my initial comment. 😊

jakearchibald: Although specifying it would solve the note issue, I'm worried that we're codifying an artefact of Chrome's implementation.

Does it make sense that SWR doesn't work in these cases? Thinking of how I'd explain it to developers, it feels a bit arbitrary.

dtapuska: @annevk I've added a not null condition to the if clause.

@jakearchibald That was largely a debate that was before for Chrome where SWR is implemented and having it at the top level (instead of the cache level) had these issues but provides the cancelibility that is desirable when a page terminates all outstanding requests terminate. Developers really shouldn't be relying on caching for functionality. 

For OCSP and downloads it doesn't make sense to do it for because they don't affect the initial display of the page which is what largely this feature targets. Do these requests without a client go through service workers as well?

jakearchibald: > Developers really shouldn't be relying on caching for functionality.

I guess we'd be telling developers "just because you use stale-while-revalidate doesn't mean the entry in the cache will ever be revalidated".

dtapuska: It would still be revalidated but it would just be done after expiry and revalidated synchronously (basically ignoring swr directive) on the next fetch of it. It would be nice to hear from Gecko what they did.  But I downloads nor OCSP likely hit the cache in Chrome anyways so it is probably an even more limited set of fetches that are unknown.

jakearchibald: I'm also interested in what Gecko did, but given that contextless fetches are rare, I'm happy that the cache isn't updated for these resources (now it's in the spec)

dtapuska: @mayhemer does the no client stuff look ok to you? ie. some fetches have no client so we avoid SWR in those cases.

mayhemer: Are you discussing fetch() behavior or HTTP request/response handling in general?

The initial implementation in Gecko is very simple, few unimplemented bits will be fixed as followups.  We respect SWR for any request (generally, not just for fetch() initiated requests), in any context, and even w/o a context.  There is no explicit coalescing of multiple revalidateRequests.  The revalidateRequest is not cancelled when the client (source page) is cancelled its load or otherwise goes away.

I have a followup [bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1542322) for the cancellation bit.  The exact functionality is to be written down yet.  Cancellation in a context makes sense.  No cancellation for context-less (usually internally initiated system) requests makes sense too.

So, to answer: if there is a context, cancel.  If there is no context, don't cancel.

I personally don't see a reason why we should not respect SWR for some selected requests.  The SWR spec comes from an HTTP RFC, not fetch() spec, and doesn't way anything about context.  SWR can be generally beneficial.

mayhemer: @annevk Is this PR change only affecting the fetch() spec?  If so, then the Chrome implementation seems (after discussion on IRC with @dtapuska) reasonable and Gecko will likely end up doing the same/close thing.  

Hence, specifying the non-null client condition to respect SWR for fetch() initiated requests would make sense.

It seems to me pointless to discuss downloads, OCSP and other types of browser initiated internal requests.  Those can potentially benefit from SWR, independently from fetch() and don't need cancellation.

dtapuska: @annevk Are we ok to proceed with merging this PR then?

annevk: We're not, because the fetch algorithm covers all requests, not just those from `fetch()`, right? You can't ever have a `fetch()`-initiated fetch with a null client (except perhaps from browser-privileged code, but that's not specified).

dtapuska: @annevk  Can you please advise what changes need to be made then?

annevk: I'm not sure as I don't understand the constraints well enough. But I think I do understand things to the point where the above exchange ends up being contradictory, as I tried to point out. Was it unclear that the fetch algorithm governs all fetches, not just those made by `fetch()`?

jakearchibald: @dtapuska it was my understanding that Chrome's implementation of SWR covers all fetches with a client (eg, an iframe's src, an `img`, a script's src etc etc). Is this true?

dtapuska: Yes Chrome's implementation covers all fetches with a client. Gecko's implementation does not have that requirement so it will be activated in more scenarios, albeit as @mayhemer points out some scenarios are limited because Gecko uses "no-store" policy for some of those fetches.

Chrome's implementation was done this way so outstanding SWR would be cancelled when a browsing context is destroyed. The Chrome networking team did not want SWR requests on going after a user has terminated the tab.

jakearchibald: > Gecko's implementation does not have that requirement

Don't requirements come from the spec? I think that's where this thread is getting confused.

annevk: I talked with Honza out-of-band and it sounds like Firefox will address the mismatches with the specification in an upcoming version.

annevk:
Thanks @dtapuska for seeing this through. If you have ideas for how we could make this more streamlined for you in the future those would be welcome. Perhaps synchronous communication at some point could have helped to set mutual expectations better?

uasan:
Hi.
We really lack the ability to handle events, receive a new response, to a background asynchronous browser request.
Are there any plans to add such an event so that we can handler to it?
Prototype handler interface:
```JS

addEventListener('cache-stale-updated', ({ data: url }) => myReRenderContentByUrl(url));

```

annevk:
@uasan no plans, but if you want this to be considered properly please file a new issue, ideally after reading https://whatwg.org/faq#adding-new-features. Thanks!

PR #884, opened 2019-03-23
Remove Client Hints

### LINKED ISSUE #867: Does Client Hints have multi vendor support?
The Fetch specification integrates support for Client Hints, most notably in step 7 of the [Fetching algorithm](https://fetch.spec.whatwg.org/#fetching).

Per [WHATWG Working Mode](https://whatwg.org/working-mode#additions), additions are supposed to have "support from at least two implementors", and "the support from implementers should be of the form 'we would like to implement this soon' and not just 'this seems like a reasonable idea'".

I know that Chrome/Blinks supports (and ships) Client Hints and Safari/WebKit does not ship Clients Hints or support it. Mozilla's [standards position on the topic](https://github.com/mozilla/standards-positions/issues/79) doesn't hard refuse, but also doesn't seem to indicate any kind of active support. Are there indications of multi-implementor support that I am missing?

I don't really want to be policing this, but since features shipping only in Safari have been removed from specs based on the multi-implementor support policy before, I think it should be applied consistently. There are also currently open issues for further extending this support, and I'm concerned they may give the impression that Client Hints is part of the interoperable multi-vendor Web Platform.

annevk:
The model described does not match implementations and there's only support from a single implementation at this point.

#726 works on reintroducing this feature with broader support and a better model.

Closes #867.

***
<a href="https://whatpr.org/fetch/884.html" title="Last updated on Mar 23, 2019, 9:26 AM UTC (4b9bb07)">Preview</a> | <a href="https://whatpr.org/fetch/884/265974f...4b9bb07.html" title="Last updated on Mar 23, 2019, 9:26 AM UTC (4b9bb07)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -53,14 +53,6 @@ url:https://tools.ietf.org/html/rfc7234#section-1.2.1;text:delta-seconds;type:df
     "WSP": {
         "aliasOf": "RFC6455"
     },
-    "CLIENT-HINTS": {
-        "authors": [
-            "Ilya Grigorik"
-        ],
-        "href": "https://tools.ietf.org/html/draft-ietf-httpbis-client-hints",
-        "publisher": "IETF",
-        "title": "HTTP Client Hints"
-    },
     "HTTPVERBSEC1": {
         "publisher": "US-CERT",
         "href": "https://www.kb.cert.org/vuls/id/867593",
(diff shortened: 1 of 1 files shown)


annevk:
@igrigorik @yoavweiss I created this as the existing text continues to raise confusion, for both Apple and Mozilla. Perhaps this will also make it easier to establish the newer model as there's no longer an old model to compare it with.

yoavweiss:
Thanks for removing those out-of-date definitions! Once merged, I'll rebase https://github.com/whatwg/fetch/pull/773

Issue #760, opened 2018-06-08
Extend Cross-Origin-Resource-Policy to take origin values


annevk:
In #687 there was a strong interest, notably by @arturjanc, to make `Cross-Origin-Resource-Policy` accept literal origins.

Things to decide:

* Multiple values or not.
* Byte-for-byte matching after any splitting if we do multiple values (yes please).

This also means we can no longer fail open. E.g., the tests in https://github.com/web-platform-tests/wpt/pull/11427 would have to be flipped so they expect rejection instead since we do not know that the unrecognized values are not origins.

cc @johnwilander @youennf

anforowicz:
What is the main problem that this issue attempts to solve?  Is this mostly about the problem described by @arturjanc in https://github.com/whatwg/fetch/issues/687#issuecomment-385048931:

> The problem with not allowing granular control in From-Origin is that it will be impossible for developers whose resources are loaded cross-site to use the header, even if they fully control the requesting domain.

cc @mikewest

anforowicz:
Also (for helping with prioritization), do you think supporting cross-site Cross-Origin-Resource-Policy (e.g. by makeing Cross-Origin-Resource-Policy accept literal origins) is more or less important than working on finalizing and implementing (Sec-Fetch-Site)[https://mikewest.github.io/sec-metadata/#sec-fetch-site-header]?

cc @csreis

PaperStrike:
From an individual website developer's view, this can effectively address the hotlinking issues. The region where my website's users are located demands expensive server bandwidth, and service providers like Vercel and Cloudflare, which are more cost-effective or even free, cannot offer stable services in that area. When some larger websites unauthorizedly use certain images/videos, the impact on my website is no less than DDoS attacks (pricewise). Nearly every local CDN provider offers anti-hotlinking services based on Referer. However, Referer is no longer accurate enough for long; `referrerpolicy` can easily bypass it, simply blocking empty referers also blocks many legitimate accesses. The `same-site` feature of CORP to a certain extent can address this issue, but often the CDN and the user visiting domain are not the same site, which is not flexible enough. If it were possible to specify the scope of the resources, the adoption rate of CORP would likely be much higher.

PR #893, opened 2019-04-12
Sketch out a CORP-only mode.

mikewest:
A straw-proposal to flesh out the discussion in https://github.com/whatwg/html/issues/4175#issuecomment-481496963.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2979,7 +2979,7 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-si
   the same response will similarly fall through to <b>allowed</b> below.
 
  <li>
-  <p>If <var>policy</var> is null, and <a>request</a>'s <a for=request>client</a>'s
+  <p>If <var>policy</var> is null, and <a for=/>request</a>'s <a for=request>client</a>'s
   <a for="environment settings object">responsible browsing context</a>'s <a>active document</a>'s
   <b>x-bikeshed-cross-origin-resource-policy-policy flag</b> is set, then set <var>policy</var> to
   "<code>same-origin</code>".
@@ -3014,7 +3014,7 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-si
       </ul>
       <p class=note>`<code>Cross-Origin-Resource-Policy: same-site</code>` does not consider a
       response delivered via a secure transport to match a non-secure requesting origin, even if
-      their hosts are otherwise <a>same-site</a>. Securely-transported responses will only match a
+      their hosts are otherwise <a>same site</a>. Securely-transported responses will only match a
       securely-transported initiator.
 
      <li><p>Otherwise, return <b>blocked</b>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -2940,35 +2940,33 @@ run these steps:</p>
 
 <h3 id=cross-origin-resource-policy-header>`<code>Cross-Origin-Resource-Policy</code>` header</h3>
 
-<p>The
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2940,67 +2940,99 @@ run these steps:</p>
...
+   <li><p><var>request</var>'s <a for=request>mode</a> is "<code>same-origin</code>",
+   "<code>cors</code>", or "<code>websocket</code>".
 
-  <p class="note no-backref">While redirects that carry  a
-  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header are checked, redirects
-  without such a header resulting in <var>response</var> do not contribute to this algorithm. I.e.,
-  <var>request</var>'s <a for=request>tainted origin flag</a> is not checked.
+   <li><p><var>request</var> is a <a>navigation request</a> whose <a for=request>reserved client</a>
+   is an <a for=/>environment</a> whose <a for=environment>target browsing context</a> is neither a
+   <a>nested browsing context</a> nor an <a>auxiliary browsing context</a>.
```
annevk: This should explain why the auxiliary navigating the non-auxiliary is not a problem.

fetch.bs
```diff
@@ -2940,67 +2940,99 @@ run these steps:</p>
...
-  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` headers will have the same effect.
-
- <li><p>If <var>policy</var> is `<code>same-origin</code>`, then return <b>blocked</b>.
+  ends up as <b>allowed</b> as it will not match any of the explicit options in the switch below.
+  Two or more `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` headers delivered in
+  the same response will similarly fall through to <b>allowed</b> below.
 
  <li>
-  <p>If the following are true
+  <p>If <var>policy</var> is null, and <a>request</a>'s <a for=request>client</a>'s
```
annevk: Is this what we want? I kinda feel like an erroneous policy should fail closed when this TBD boolean is set.

fetch.bs
mikewest: Well, hrm. I guess that actually is a problem without process isolation. `attacker.site` could open `attacker.site` in a new window, the latter could navigate the former to `victim.site`, and sadness would ensue. Perhaps this should instead ask whether we're navigating within a unit of related browsing contexts? Or whatever the new hotness is? "User agent cluster"?

fetch.bs
mikewest: I want to be able to support new values for this header in the future (for example, a list of origins a la https://github.com/whatwg/fetch/issues/760). I'm happy to tighten this up, but I'd like to figure out how to do so in a forward-compatible way.

fetch.bs
annevk: Without process isolation on the agent cluster level, right.

I think the navigation checks only make sense when the flag is set on a document. And that flag only makes sense if COOP is also set. (COOP guaranteeing process isolation on the browsing context group level.)

annevk: New hotness background:

* https://html.spec.whatwg.org/#integration-with-the-javascript-agent-cluster-formalism
* https://html.spec.whatwg.org/#groupings-of-browsing-contexts
* https://github.com/whatwg/html/issues/4361

fetch.bs
annevk: I guess the question then is what the migration strategy has to be. In equivalent cases we've failed for potentially new syntax and new servers will have to update (e.g., all the recentish and proposed CORS changes).

annevk:
9ff55e42934b360e374591cf1d0934f9ecb540f0 from PR #1030 addressed this.

PR #908, opened 2019-05-29
Let Origin header honor referrer policy for non CORS request

JuniorHsu:
Let Origin header honor referrer policy for non CORS request

Tests: https://github.com/web-platform-tests/wpt/pull/14260

***
<a href="https://whatpr.org/fetch/908.html" title="Last updated on Jun 26, 2019, 12:41 PM UTC (81c4f70)">Preview</a> | <a href="https://whatpr.org/fetch/908/49b92ee...81c4f70.html" title="Last updated on Jun 26, 2019, 12:41 PM UTC (81c4f70)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1548,6 +1548,30 @@ run these steps:
  <li><p>If <var>request</var>'s <a for=request>tainted origin flag</a> is set, then return
  `<code>null</code>`.
 
+ <li><p>If <var>request</var>'s <i>CORS flag</i> is unset, then switch on <var>request</var>'s
+ <a for=request>referrer policy</a>:
+
+    <dl class=switch>
+     <dt>"<code>no-referrer</code>"
+     <dd><p> return `<code>null</code>`.
+    </dl>
+
+    <dl class=switch>
+     <dt>"<code>no-referrer-when-downgrade</code>"
+     <dt>"<code>strict-origin</code>"
+     <dt>"<code>strict-origin-when-cross-origin</code>"
+     <dd><p>If <var>request</var>'s <a for=request>current URL</a>'s <var>scheme</var> is
+     "<code>http</code>" and <var>request</var>'s <a for=request>origin</a>'s
+     <var>scheme</var> is "<code>https</code>", then return `<code>null</code>`.
+    </dl>
+
+    <dl class=switch>
+     <dt>"<code>same-origin</code>"
+     <dd><p>If <var>request</var>'s <a for=request>origin</a> is not <a>same origin</a> with
+     <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, then return
+     `<code>null</code>`.
+    </dl>
+
  <li><p>Return <var>request</var>'s <a for=request>origin</a>,
  <a lt="ASCII serialization of an origin">serialized</a> and <a>isomorphic encoded</a>.
 </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4068,39 +4068,42 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <a for=request>referrer</a>, <a lt="url serializer">serialized</a> and <a>isomorphic encoded</a>,
    to <var>httpRequest</var>'s <a for=request>header list</a>.
 
-   <li><p>If the <i>CORS flag</i> is set, <var>httpRequest</var>'s <a for=request>method</a> is
-   neither `<code>GET</code>` nor `<code>HEAD</code>`, or <var>httpRequest</var>'s
-   <a for=request>mode</a> is "<code>websocket</code>", then:
+   <li><p>Run these steps:
    <ol>
-    <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
+    <li><p>Let <var>serializedOrigin</var> be the result of <a>serializing a request origin</a> with
     <var>httpRequest</var>.
-    <li><p>If <var>request</var>'s <i>CORS flag</i> is unset, then switch on
-    <var>httpRequest</var>'s <a for=request>referrer policy</a>:
+    <li><p>If the <i>CORS flag</i> is set, or <var>httpRequest</var>'s <a for=request>mode</a> is
+    "<code>websocket</code>", then <a for="header list">append</a>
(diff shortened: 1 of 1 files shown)


annevk:
I think we should only take this into account when setting the `Origin` header, otherwise it will also affect various CORS algorithms and that does not seem like the right thing to do.

Also, for some cross-origin navigation requests the `Origin` header was always `null` per the tests, right? That also needs to be reflected somehow.

JuniorHsu:
> I think we should only take this into account when setting the `Origin` header, otherwise it will also affect various CORS algorithms and that does not seem like the right thing to do.
Okay, it's fair. 
> 
> Also, for some cross-origin navigation requests the `Origin` header was always `null` per the tests, right? That also needs to be reflected somehow.
For form POST navigation, 
(a) the redirect case is covered with `tainted origin flag`
(b) other cases are all the same with non-CORS fetch

Was I missing something?

yutakahirano:
Sorry for naive questions:
 - What is the rationale?
 - Why is this only for non CORS requests?

JuniorHsu:
>     * What is the rationale?
> 
Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1504085#c7
At the beginning, we don't want to leak `Origin:` for HTTPS->HTTP cases.
And we find out `Referrer-Policy` is a good target to honor.

>     * Why is this only for non CORS requests?
Please see https://github.com/web-platform-tests/wpt/pull/15937#discussion_r270465836
Tag @annevk who might have some ideas.

annevk:
CORS is something both sides opt into so it seems reasonable that in that case we leak the origin. We previously decided that the Referrer Policy would not influence CORS for that reason. However, Origin is also used outside of CORS and we have not effectively dealt with it there.

yutakahirano:
Thank you.

fetch.bs
```diff
@@ -4070,9 +4070,38 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
    <li><p>If the <i>CORS flag</i> is set, <var>httpRequest</var>'s <a for=request>method</a> is
    neither `<code>GET</code>` nor `<code>HEAD</code>`, or <var>httpRequest</var>'s
-   <a for=request>mode</a> is "<code>websocket</code>", then <a for="header list">append</a>
-   `<code>Origin</code>`/the result of <a>serializing a request origin</a> with
-   <var>httpRequest</var>, to <var>httpRequest</var>'s <a for=request>header list</a>.
+   <a for=request>mode</a> is "<code>websocket</code>", then:
+   <ol>
+    <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
+    <var>httpRequest</var>.
+    <li><p>If <var>request</var>'s <i>CORS flag</i> is unset, then switch on
```
annevk: I suspect we also don't want to apply these steps when mode is "websocket" as that is effectively CORS too. And at that point I wonder if we can restructure this a bit. Perhaps something like:

1. Let serializedOrigin be the result of serializing a request origin with ...
2. If the CORS flag is set or httpRequest's mode is "websocket", then append Origin/serializedOrigin to httpRequest's header list.
3. Otherwise, if httpRequest's method is neither GET nor HEAD:
   1. The Referrer Policy sanitizing origin steps...

What do you think?

JuniorHsu:
I'm with you for the view of websocket. I addressed this with my new patch. 

> Thanks for the update, this mostly looks good to me now except for some minor things. Test coverage is good and I think other browsers are on board as well. I can take care of filing bugs on them once we land this.

Thanks for taking care of filing bug. FWIW, we also need a bug for navigation from `about:blank`
> 
> For the branching on referrer policy adding "Otherwise" and "Do nothing." as its steps might also be good, to be explicit about that scenario.

This issue is gone since I change the structure
> 
> And we should add a Note explaining why we adhere to Referrer Policy only sometimes. I can also take that on before I merge this unless you want to give it a go.

Thanks for taking care of the note.

fetch.bs
```diff
@@ -4068,11 +4068,47 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+
+    <dl class=switch>
+     <dt>"<code>no-referrer</code>"
+     <dd><p><a for="header list">Append</a> `<code>Origin</code>`/`<code>null</code>` to
+     <var>httpRequest</var>'s <a for=request>header list</a>.
+
+     <dt>"<code>no-referrer-when-downgrade</code>"
+     <dt>"<code>strict-origin</code>"
+     <dt>"<code>strict-origin-when-cross-origin</code>"
+     <dd><p>If <var>request</var>'s <a for=request>origin</a>'s <var>scheme</var> is
```
annevk: While reformatting I realized that this cannot work as there's no guarantee request's origin is an origin that has a scheme. So this also needs to account for the origin being an opaque origin.

Another problem here is that if this conditional isn't met, no header gets appended. That seems unlikely to be the intention?

One thing I was thinking that might make this easier if we split this out into its own algorithm, perhaps right below "Serializing a request origin". Something like "Appending a request Origin header".

As I know you're somewhat otherwise occupied let me know if I should help out here.

JuniorHsu: > While reformatting I realized that this cannot work as there's no guarantee request's origin is an origin that has a scheme. So this also needs to account for the origin being an opaque origin.

Done. Not sure if the fetch spec provide short-circuit behavior for (and/or)
> 
> Another problem here is that if this conditional isn't met, no header gets appended. That seems unlikely to be the intention?

It's horrible. Thanks for picking this.
> 
> One thing I was thinking that might make this easier if we split this out into its own algorithm, perhaps right below "Serializing a request origin". Something like "Appending a request Origin header".
> 

It's a bit complicated now, so IMO it's a good idea to isolate it.

@annevk Please take a look for the new patch. Thanks!

JuniorHsu:
@annevk Yes, I'm happy :)  Thanks for those tremendous help!

annevk:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1424076
* https://bugs.chromium.org/p/chromium/issues/detail?id=979142
* https://bugs.webkit.org/show_bug.cgi?id=199261

annevk:
And thank you for pushing this through @JuniorHsu!

PR #916, opened 2019-07-16
Set CORS preflight requests' `mode` to `cors`.

mikewest:
CORS preflight requests' `mode` is now exposed via the `Sec-Fetch-Mode` header, and it
makes sense to set it to `cors`, as it's part of the CORS protocol.

Fixes https://github.com/w3c/webappsec-fetch-metadata/issues/35.

***
<a href="https://whatpr.org/fetch/916.html" title="Last updated on Jul 16, 2019, 12:31 PM UTC (f74a395)">Preview</a> | <a href="https://whatpr.org/fetch/916/7e088f6...f74a395.html" title="Last updated on Jul 16, 2019, 12:31 PM UTC (f74a395)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4779,7 +4779,7 @@ run these steps:
   <a for=request>origin</a> is <var>request</var>'s <a for=request>origin</a>,
   <a for=request>referrer</a> is <var>request</var>'s <a for=request>referrer</a>,
   <a for=request>referrer policy</a> is <var>request</var>'s <a for=request>referrer policy</a>,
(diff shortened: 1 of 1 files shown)


mikewest:
Moved it down, thanks. I do aim to add tests to cover the preflight expectations in https://chromium-review.googlesource.com/c/chromium/src/+/1702311.

annevk:
I noticed one more thing that I can push a commit for. This is not byte sequence so it should use normal quotes.

PR #925, opened 2019-08-28
"Set" Structured Header values in header lists.

mikewest:
As discussed in w3c/webappsec-fetch-metadata#40, this patch takes a
small first step towards supporting Structure Header values in Fetch by
wrapping header lists' existing "set" algorithm with a version that
first serializes a Structured Header object into a string
representation.

***
<a href="https://whatpr.org/fetch/925.html" title="Last updated on Sep 13, 2019, 1:52 PM UTC (fb49277)">Preview</a> | <a href="https://whatpr.org/fetch/925/ce287b0...fb49277.html" title="Last updated on Sep 13, 2019, 1:52 PM UTC (fb49277)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -16,7 +16,6 @@ url:https://tools.ietf.org/html/rfc7230#section-3.2;text:field-content; type:dfn
 url:https://tools.ietf.org/html/rfc7230#section-3.2;text:field-value; type:dfn;spec:http
 url:https://tools.ietf.org/html/rfc7230#section-3.1.2;text:reason-phrase;type:dfn;spec:http
 url:https://tools.ietf.org/html/rfc7234#section-1.2.1;text:delta-seconds;type:dfn;spec:http-caching
-url:https://tools.ietf.org/html/draft-ietf-httpbis-header-structure#section-1;text:structured header;type:dfn;spec:header-structure
 url:https://tools.ietf.org/html/draft-ietf-httpbis-header-structure#section-4.1;text:serializing structured headers;type:dfn;spec:header-structure
 </pre>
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -564,6 +572,22 @@ A: 3
...
+<ol>
+ <li>Let <var>serialized value</var> be the result of executing the <a>serializing structured
+ headers</a> algorithm on <var>structured header</var>.
+ <li><a for="header list">Set</a> <var>name</var>/<var>serialized value</var> in <var>list</var>.
+</ol>
+
+<p class="note">Structured header values are defined as objects which HTTP can (eventually)
+serialize in interesting and efficient ways. [[!HEADER-STRUCTURE]] For the moment, Fetch only
+supports setting these objects in <a for=/>header lists</a> by first serializing them into a
+string representation.
```
annevk: I suppose we should nevertheless encourage folks to use this setter to receive benefits later.

fetch.bs
```diff
@@ -564,6 +572,22 @@ A: 3
  <var>name</var> and <a for=header>value</a> is <var>value</var> to <var>list</var>.
 </ol>
 
+<p>To <dfn export for="header list" id=concept-header-list-set-structured-header>set a structured
+header</dfn> <a for=header>name</a>/<a>structured header</a>
```
annevk: I filed https://github.com/httpwg/http-extensions/issues/908 as this reads rather awkward.

fetch.bs
```diff
@@ -81,6 +83,12 @@ url:https://tools.ietf.org/html/rfc7234#section-1.2.1;text:delta-seconds;type:df
     },
     "OCSP": {
         "aliasOf": "RFC2560"
+    },
+    "HEADER-STRUCTURE": {
+        "authors": ["Mark Nottingham","Paul-Henning Kamp"],
+        "href": "https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html",
```
annevk: I suppose we should use the tools.ietf.org version, unless we want to update all?

mikewest: Shifted to https://tools.ietf.org/html/draft-ietf-httpbis-header-structure.

fetch.bs
mikewest: I added another line about future efficiency gains. Any additional suggestions for the encouragement you'd like to provide here?

annevk:
Based on the discussion with @mnot it seems like we might want a different approach here whereby we do support structured headers underneath and pass those on to the HTTP layer directly.

mikewest:
> ... pass those on to the HTTP layer directly.

I assume that would require @mnot, et al. to add some hooks to HTTP. Should we land something like this in the meantime so that Fetch is the only spec that needs to change when those new hooks are created?

annevk:
In the email you were also copied on he claimed that the combination of HTTP and Structured Headers allows for this... But yeah, on reflection I think that preceding with this isn't harmful and a useful incremental step. I'll see about merging it when I'm back next Tuesday (please do ping).

mikewest:
>  I'll see about merging it when I'm back next Tuesday (please do ping).

I was out Tuesday. But it's Wednesday! Ping! ;)

annevk:
What do you think about my changes?

mikewest:
I think your changes are good. I would like for there to be a link for "structured header value", as I worry a bit that "name/structured header value name/structuredValue" will blur into each other and be a bit confused if that term isn't a clearly demarcated link, but presumably that will resolve itself as soon as https://github.com/httpwg/http-extensions/commit/da67208c034fc53dae221b254489664d97578d34 is published somewhere.

Also, I'd personally say "an RFC", not "a RFC" (but I'm American, and therefore have a weird way of saying "R". :) )

annevk:
Yeah, maybe we should teach the spelling rule about RFC, but since it's only a comment I'm not gonna bother. I agree about "structured header value", hence the XXX comment. I don't really see a way to do it otherwise other than linking the httpwg.org document which is also somewhat meh.

Also, due to recent changes this is now blocked on https://github.com/heycam/webidl/issues/795 as are all Fetch PRs. 😟

annevk:
I added a reference for "structured header value" based on the section where it'll end up in. Seems good enough.

Issue #904, opened 2019-05-09
Double-keyed HTTP cache


annevk:
The idea here is that the "browser's address bar origin" is an additional key for its HTTP cache, to prevent certain classes of attacks.

Safari ships a variant of this (uses registrable domain, not origin), but seems willing to adjust to origin. Other implementers are interested in shipping and are at various stages of experimentation.

This will require making all accesses of "the HTTP cache" more contextual, by accessing the HTTP cache of X whereby X is some defined origin. (Other ideas welcome, @mnot?)

I'm not sure where to store the defined origin. We could do a browsing context ancestor walk and that might be okay as I think all fetches always require a fully active document, but would be nice to have that confirmed.

(I'm also assuming that auxiliary browsing contexts are not special here and behave like other top-level browsing contexts for the purposes of this.)

cc @youennf @whatwg/security

youennf:
Safari is partitioning service workers so that a service worker of an iframe B in a top level page A will be using the partition of top level page A.
This makes the loading more consistent in general (whether intercepted or not, the iframe loads will use the same partition) and improves on privacy.

jkarlin:
Very interested. We'd like to mitigate security issues such as [x-site search](https://terjanq.github.io/Bug-Bounty/Google/cache-attack-06jd2d2mz2r0/index.html).

Chrome's experimenting with double keying by (top-frame origin, url) as well as triple-keying (top-frame origin, initiator origin, url). Triple keying protects caches of frames within a page from each other. The performance hit from double-keying seems reasonable at first glance, but it's important that we address x-origin prefetch in a multi-keyed cache world. We don't yet have data on triple-keying.

wanderview:
This issue should also consider the differences between partitioned http cache and partitioned origin storage.  I believe webkit's default partitioning include both http and origin storage.  I don't think the chrome experiment includes any origin storage partitioning, though.

mnot:
+1

It would be nice to try to factor out cache key computation to support not only this, but things like variants, etc.

annevk:
Mozilla is interested in partitioning other bits as well, but for this issue I'd like to focus solely on the HTTP cache. Some of the infrastructure we might be able to reuse for the other bits mentioned above, but I don't think there's any strong reason to couple them from the start.

johnwilander:
WebKit also partitions LocalStorage on eTLD+1 and used to partition cookies up until a couple of months ago (now the same cookies for third parties are just blocked instead).

[...]

PR #943, opened 2019-09-26
HTTP cache partitioning

shivanigithub:
Adds a section to define the HTTP cache associated with a request as per HTTP cache partitioning.
github issue: https://github.com/whatwg/fetch/issues/904
Explainer: https://github.com/shivanigithub/http-cache-partitioning/

***
<a href="https://whatpr.org/fetch/943.html" title="Last updated on Jun 9, 2020, 5:59 AM UTC (5e45c5b)">Preview</a> | <a href="https://whatpr.org/fetch/943/57305b8...5e45c5b.html" title="Last updated on Jun 9, 2020, 5:59 AM UTC (5e45c5b)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2118,6 +2118,50 @@ clearly stipulates that <a>connections</a> are keyed on
      WebSocket saner -->
 
 
+<h3 id=http-cache-partitions>HTTP cache partitions</h3>
+
+<p>To
+<dfn lt="determine the HTTP cache partition|determining the HTTP cache partition">determine the HTTP cache partition</dfn>,
+given <var>request</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>topLevelOrigin</var> be null.
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>reserved client</a> is non-null, then:
+
+  <ol>
+   <li><p>Set <var>topLevelOrigin</var> to <var>request</var>'s <a for=request>reserved client</a>'s
+   <a for="environment">top-level origin</a>.
+
+   <li>
+    <p>If <var>topLevelOrigin</var> is null, then set <var>topLevelOrigin</var> to
+    <var>request</var>'s <a for=request>reserved client</a>'s
+    <a for="environment">top-level creation URL</a>'s <a for=url>origin</a>.
+
+    <p class=note>This happens for top-level navigations only.
+  </ol>
+
+ <li><p>Otherwise, if <var>request</var>'s <a for=request>client</a> is non-null, set
+ <var>topLevelOrigin</var> to <var>request</var>'s <a for=request>client</a>'s
+ <a for="environment">top-level origin</a>.
+
+ <li><p>Otherwise, return null.
+
+ <li><p>Assert: <var>topLevelOrigin</var> is an <a for=/>origin</a>.
+
+ <li><p>Let <var>topLevelSite</var> be the result of <a lt="obtain a site">obtaining a site</a>,
+ given <var>topLevelOrigin</var>.
+
+ <li>
+  <p>Return the HTTP cache associated with <var>topLevelSite</var> and, possibly, a second key.
+  [[!HTTP-CACHING]]
+
+  <p class=XXX>The second key is intentionally a little vague as the finer points are still
+  evolving. See <a href=https://github.com/whatwg/fetch/issues/1035>issue #1035</a>.
+</ol>
+
+
 <h3 id=port-blocking>Port blocking</h3>
 
 <p>To determine whether fetching a <a for=/>request</a> <var>request</var>
@@ -4130,6 +4174,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
  <li><p>Let <var>storedResponse</var> be null.
 
+ <li><p>Let <var>httpCache</var> be null.
+
  <li><p>Let the <var>revalidatingFlag</var> be unset.
 
  <li>
@@ -4365,14 +4411,20 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     <var>httpRequest</var>'s
     <a for=request>credentials mode</a>.
 
+   <li><p>Set <var>httpCache</var> to the result of <a>determining the HTTP cache partition</a>,
+   given <var>httpRequest</var>.
+
+   <li><p>If <var>httpCache</var> is null, then set <var>httpRequest</var>'s
+   <a for=request>cache mode</a> to "<code>no-store</code>".
+
    <li>
     <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is neither "<code>no-store</code>"
     nor "<code>reload</code>", then:
 
     <ol>
      <li>
-      <p>Set <var>storedResponse</var> to the result of selecting a response from the HTTP cache,
-      possibly needing validation, as per the
+      <p>Set <var>storedResponse</var> to the result of selecting a response from the
+      <var>httpCache</var>, possibly needing validation, as per the
       "<a href=https://tools.ietf.org/html/rfc7234#section-4>Constructing Responses from Caches</a>"
       chapter of <cite>HTTP Caching</cite> [[!HTTP-CACHING]], if any.
 
@@ -4410,10 +4462,10 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
           <p><a>In parallel</a>, perform <a for=main>main fetch</a> using
           <var>revalidateRequest</var>.
 
-          <p class=note>This fetch is only meant to update the state of the HTTP cache and the
-          response will be unused until another cache access. The stale response will be used as the
-          response to current request. This fetch is issued in the context of a client so if it goes
-          away the request will be terminated.
+          <p class=note>This fetch is only meant to update the state of <var>httpCache</var>
+          and the response will be unused until another cache access. The stale response will be used
+          as the response to current request. This fetch is issued in the context of a client so if
+          it goes away the request will be terminated.
         </ol>
 
        <li>
@@ -4476,8 +4528,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li><p>If <var>httpRequest</var>'s <a for=request>method</a> is
    <a href=https://tools.ietf.org/html/rfc7231#safe.methods>unsafe</a> and
    <var>forwardResponse</var>'s <a for=response>status</a> is in the range <code>200</code> to
-   <code>399</code>, inclusive, invalidate appropriate stored responses in the HTTP cache, as per
-   the "<a href=https://tools.ietf.org/html/rfc7234#section-4.4>Invalidation</a>" chapter of
+   <code>399</code>, inclusive, invalidate appropriate stored responses in <var>httpCache</var>, as
+   per the "<a href=https://tools.ietf.org/html/rfc7234#section-4.4>Invalidation</a>" chapter of
    <cite>HTTP Caching</cite>, and set <var>storedResponse</var> to null. [[!HTTP-CACHING]]
 
    <li>
@@ -4504,8 +4556,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
      <li><p>Set <var>response</var> to <var>forwardResponse</var>.
 
      <li>
-      <p>Store <var>httpRequest</var> and <var>forwardResponse</var> in the HTTP cache, as per the
-      "<a href=https://tools.ietf.org/html/rfc7234#section-3>Storing Responses in Caches</a>"
+      <p>Store <var>httpRequest</var> and <var>forwardResponse</var> in <var>httpCache</var>, as per
+      the "<a href=https://tools.ietf.org/html/rfc7234#section-3>Storing Responses in Caches</a>"
       chapter of <cite>HTTP Caching</cite>. [[!HTTP-CACHING]]
 
       <p class=note>If <var>forwardResponse</var> is a <a>network error</a>, this effectively caches
@@ -4624,6 +4676,12 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
  <li><p>Let <var>credentials</var> be true if <i>credentials flag</i> is set, and false otherwise.
 
  <li><p>Let <var>response</var> be null.
+
+ <li><p>Let <var>httpCache</var> be the result of <a>determining the HTTP cache partition</a>, given
+ <var>httpRequest</var>.
+
+ <li><p>If <var>httpCache</var> is null, then set <var>request</var>'s <a for=request>cache mode</a>
+ to "<code>no-store</code>".
 
  <li>
   <p>Switch on <var>request</var>'s <a for=request>mode</a>:
@@ -4751,11 +4809,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <a for=body>total bytes</a> to that payload body length.
    <!-- XXX xref payload body -->
 
-   <li><p>If <var>response</var> is not a
-   <a>network error</a> and <var>request</var>'s
-   <a for=request>cache mode</a> is not "<code>no-store</code>",
-   update <var>response</var> in the HTTP cache for <var>request</var>.
-   <!-- XXX xref HTTP cache -->
+   <li><p>If <var>response</var> is not a <a>network error</a> and <var>request</var>'s
+   <a for=request>cache mode</a> is not "<code>no-store</code>", then update <var>response</var> in
+   <var>httpCache</var> for <var>request</var>.
 
    <li>
     <p>If <i>credentials flag</i> is set and the user agent is not configured to block cookies for
@@ -7268,6 +7324,7 @@ Sendil Kumar N,
 Shao-xuan Kang,
 Sharath Udupa,
 Shivakumar Jagalur Matt,
+Shivani Sharma,
 Sigbjørn Finne,
 Simon Pieters,
 Simon Sapin,


fetch.bs
```diff
@@ -2049,6 +2049,12 @@ clearly stipulates that <a>connections</a> are keyed on
 <!-- See https://github.com/whatwg/fetch/issues/114#issuecomment-143500095 for when we make
      WebSocket saner -->
 
+<h3 id=http-cache>HTTP Cache</h3>
+
+<p>A given request has an associated HTTP cache derived from the primary key(s). The primary key
+may be the top-frame origin of the document and also possibly the origin of the subframe
+associated with the request. When performing an <a>HTTP-network-or-cache fetch</a>, the associated
+HTTP cache is used to fetch the request URL.
```
annevk: I was thinking that we'd do this much more explicit. In particular, update HTML to define and propagate the top-level origin concept appropriately and then whenever we mention HTTP cache in this document instead invoke an algorithm that returns the appropriate cache. With this approach there's a lot of action-at-a-distance which is hard to maintain and often leads to all kinds of subtle issues.

shivanigithub: Sounds good. Created a pull request on the html spec for defining top level origin. PTAL, thanks!
https://github.com/whatwg/html/pull/4966

shivanigithub:
@domenic , @annevk: 
Uploaded the patch as per earlier discussions. This is now dependent on the [top-level origin](https://github.com/whatwg/html/pull/4966).
PTAL, thanks!

shivanigithub:
friendly ping.

fetch.bs
```diff
@@ -1417,42 +1417,44 @@ Unless stated otherwise, it is unset.
 "<code>no-cache</code>", "<code>force-cache</code>", or
 "<code>only-if-cached</code>". Unless stated otherwise, it is "<code>default</code>".
 
+<p>Let <var>HTTPCache</var> be the result of determining the <a>HTTP cache</a> partition for
```
domenic: We've traditionally cased variables in camelCase, so this would be `<var>httpCache</var>`.

fetch.bs
```diff
@@ -1417,42 +1417,44 @@ Unless stated otherwise, it is unset.
...
   <dt>"<code>default</code>"
-  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network. If the HTTP cache
-  contains a matching <a>fresh response</a> it will be returned. If the HTTP cache contains a
-  matching <a>stale-while-revalidate response</a> it will be returned, and a conditional network
-  fetch will be made to update the entry in the HTTP cache. If the HTTP cache contains a matching
-  <a>stale response</a>, a conditional network fetch will be returned to update the entry in
-  the HTTP cache. Otherwise, a non-conditional network fetch will be returned to update the entry
-  in the HTTP cache. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
-  [[!STALE-WHILE-REVALIDATE]]
+  <dd><a for=/>Fetch</a> will inspect the <var>HTTPCache</var> on the way to the network. If the
```
domenic: Since it's a variable reference, it should be `<var>httpCache</var>`, not `the <var>HTTPCache</var>`, here and throughout.

fetch.bs
```diff
@@ -2049,6 +2051,29 @@ clearly stipulates that <a>connections</a> are keyed on
 <!-- See https://github.com/whatwg/fetch/issues/114#issuecomment-143500095 for when we make
      WebSocket saner -->
 
+<h3 id=http-cache-partition>HTTP Cache partition</h3>
+
+<p>To determine the associated <dfn export id=http-cache>HTTP cache</dfn> partition, given
```
domenic: I think this should define HTTP cache partition, not HTTP cache. The `id` is also redundant. So just `<dfn export>HTTP cache partition</dfn>` would work.

fetch.bs
```diff
@@ -2049,6 +2051,29 @@ clearly stipulates that <a>connections</a> are keyed on
...
+<var>request</var>, run these steps:
+
+<ol>
+ <li>
+  <p>If <var>request</var>'s <a for=request>client</a> is null, terminate these steps.</li>
```
domenic: You can't terminate the steps without returning something. Maybe return null? Per https://github.com/whatwg/html/pull/4966#issuecomment-577316426 it sounds like in that case nothing will be cached, so, you'd probably need to update the processing model to account for when the return value of this algorithm is null.

fetch.bs
```diff
@@ -2049,6 +2051,29 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+ <li>
+  <p>Let <var>environmentSettings</var> be <var>request</var>'s <a for=request>client</a>.</li>
+
+ <li>
+  <p>Let <var>topLevelOrigin</var> be <var>environmentSettings</var>'s top-level origin and
```
domenic: Before we merge this you'll want to change this to `<a for="environment settings object">top-level origin</a>`. But, don't do that until https://github.com/whatwg/html/pull/4966 has merged and had a day to propagate into the linking database, as otherwise it will just cause spec compilation errors.

fetch.bs
```diff
@@ -2049,6 +2051,29 @@ clearly stipulates that <a>connections</a> are keyed on
...
+  <var>origin</var> be <var>environmentSettings</var>'s origin.</li>
```
domenic: This should be `<a for="environment settings object">origin</a>`.

fetch.bs
```diff
@@ -2049,6 +2051,29 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+ <li>
+  <p>Return the HTTP cache associated with <var>topLevelOrigin</var> and optionally,
+  <var>origin</var>.</li>
+</ol>
+
+<p class="note no-backref">User agents can use the registrable domains associated with
+<var>topLevelOrigin</var> and <var>origin</var> instead of the origins themselves.
```
domenic: Hmm, I don't like how many degrees of freedom this gives us. It seems like there are now six configurations: (top-level origin + origin, top-level origin) x (registrable domain instead of top-level origin, top-level origin as-is) x (registrable domain instead of origin, origin as-is). Do we expect browsers to implement all six possible points on this matrix? That seems like a pretty bad failure of the consensus process. Can we get agreement on just one, or maybe two?

fetch.bs
```diff
@@ -4692,7 +4719,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li><p>If <var>response</var> is not a
    <a>network error</a> and <var>request</var>'s
    <a for=request>cache mode</a> is not "<code>no-store</code>",
-   update <var>response</var> in the HTTP cache for <var>request</var>.
+   update <var>response</var> in the <a>HTTP cache</a> for <var>request</var>.
```
domenic: This should be `<var>httpCache</var>`, and not the general definition of a HTTP cache, right?

fetch.bs
shivanigithub: Do you think we could leave the note regarding registrable domain altogether and let the spec just say "Return the HTTP cache associated with the <var>topLevelOrigin</var>, and optionally <var>origin</var>". In other words, is it ok for the spec to not be specific enough to mention registrable domain detail?

/cc @annevk : Does Firefox plan to use origin or registrable domain?

domenic: So the proposal would be to reduce it to two possibilities? That's still not as good as one, but it'd be an improvement over six.

annevk: We are reconsidering "site" (aka "agent cluster key" https://html.spec.whatwg.org/#obtain-agent-cluster-key) based on Chrome's feedback. Additionally Chrome's triple-keying made us reconsider omni-keying (each parent up to the top-level adds a key), but it's still early days in proving it's feasible.

fetch.bs
shivanigithub: done

fetch.bs
shivanigithub: Since Chrome and Firefox are considering "site" and Safari already does that, changed it to use "agent cluster key". Although I could not get it to link to https://html.spec.whatwg.org/#obtain-agent-cluster-key. Would that require change in the way obtain-agent-cluster-key is defined in the html spec? Sorry, I am not very familiar with spec language.

fetch.bs
shivanigithub: done

fetch.bs
shivanigithub: Done.

fetch.bs
shivanigithub: Changed it to <dfn export>determining the HTTP cache partition</dfn>

fetch.bs
shivanigithub: done

fetch.bs
shivanigithub: done

shivanigithub:
Sorry for the slow movement here as I was OOO.  
Addressed feedback. PTAL, thanks!

fetch.bs
```diff
@@ -1417,42 +1417,45 @@ Unless stated otherwise, it is unset.
 "<code>no-cache</code>", "<code>force-cache</code>", or
 "<code>only-if-cached</code>". Unless stated otherwise, it is "<code>default</code>".
 
+<p>Let <var>httpCache</var> be the result of <a>determining the HTTP cache partition</a> for
```
domenic: This step doesn't seem to be located in the correct place. What uses this httpCache variable? It needs to be inside the fetch algorithm, not standing alone at top level. Then you can pass it another variable representing the current request as an argument, instead of passing it the general concept of a request.

fetch.bs
```diff
@@ -1417,42 +1417,45 @@ Unless stated otherwise, it is unset.
...
-  <dd><a for=/>Fetch</a> will inspect the HTTP cache on the way to the network. If the HTTP cache
-  contains a matching <a>fresh response</a> it will be returned. If the HTTP cache contains a
-  matching <a>stale-while-revalidate response</a> it will be returned, and a conditional network
-  fetch will be made to update the entry in the HTTP cache. If the HTTP cache contains a matching
-  <a>stale response</a>, a conditional network fetch will be returned to update the entry in
-  the HTTP cache. Otherwise, a non-conditional network fetch will be returned to update the entry
-  in the HTTP cache. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
-  [[!STALE-WHILE-REVALIDATE]]
+  <dd><a for=/>Fetch</a> will inspect the <var>httpCache</var> on the way to the network. If the
+  <var>httpCache</var> contains a matching <a>fresh response</a> it will be returned. If the
```
domenic: These non-normative explanations of different cache modes should not refer to a variable. Probably they should be reverted; they were reasonable before.

fetch.bs
```diff
@@ -2049,6 +2052,34 @@ clearly stipulates that <a>connections</a> are keyed on
 <!-- See https://github.com/whatwg/fetch/issues/114#issuecomment-143500095 for when we make
      WebSocket saner -->
 
+<h3 id=http-cache-partitions>HTTP cache partitions</h3>
+
+<p>For <dfn export>determining the HTTP cache partition</dfn>, given <var>request</var>,
+run these steps:
```
domenic: Instead of "for determining the HTTP cache partition", it should be "to determine the HTTP cache partition".

fetch.bs
```diff
@@ -2049,6 +2052,34 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+<ol>
+ <li>
+  <p>Let <var>environmentSettings</var> be <var>request</var>'s <a for=request>client</a>, if
+  present, otherwise return null.</p></li>
```
domenic: This is a bit confusing. I would just have a second step that says

1. If _environmentSettings_ is null, then return null.

fetch.bs
```diff
@@ -2049,6 +2052,34 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+ <li>
+  <p>Let <var>topLevelOrigin</var> be <var>environmentSettings</var>'s top-level origin and
+  <var>origin</var> be <a for="environment settings object">origin</a>.</p></li>
```
domenic:
```suggestion
  <var>origin</var> be its <a for="environment settings object">origin</a>.</p></li>
```

fetch.bs
```diff
@@ -2049,6 +2052,34 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+ <li>
+  <p>Let <var>topLevelOriginAgentClusterKey</var> be the result of <span
+   data-x="obtain-agent-cluster-key">obtaining an agent cluster key</span>, given
```
domenic: To cross link this, use `<a spec="HTML">obtaining an agent cluster key</a>`. Then we can make changes on the HTML side to allow you to remove the `spec="HTML"` (which is a way of overriding the fact that HTML currently doesn't export this term).

domenic: @annevk may have editorial feedback on this---e.g. in the past he's wanted to create a first-class "site" concept, instead of using the term "agent cluster key". But, we can worry about that after we get the processing model set up.

fetch.bs
```diff
@@ -3161,13 +3192,11 @@ which is unset unless otherwise specified.
...
-ignore the suspension request if the ongoing fetch is updating the response in the HTTP cache for
-the request.
+<dfn export for=fetch id=concept-fetch-resume>resumed</dfn>. The user agent should ignore the
+suspension request if the ongoing fetch is updating the response in the HTTP cache for the request.
 
-<p class="note no-backref">The user agent does not update the entry in the HTTP cache for a
-<a for=/>request</a> if request's cache mode is "no-store" or a
-`<code>Cache-Control: no-store</code>` header appears in the response.
+<p class="note no-backref">The user agent does not update the entry in the HTTP cache if request's
+cache mode is "no-store" or a `<code>Cache-Control: no-store</code>` header appears in the response.
```
domenic: Why make these two changes?

fetch.bs
domenic: Ah, I see, you do include it below. Yeah, then just remove all changes to this section. This section is defining concepts and doesn't need to get into these details.

fetch.bs
```diff
@@ -4303,14 +4332,19 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     <var>httpRequest</var>'s
     <a for=request>credentials mode</a>.
 
+   <li>
+    <p>Let <var>httpCache</var> be the result of <a>determining the HTTP cache partition</a> for
+    <var>httpRequest</var>. If <var>httpCache</var> is null, fetch behaves as if there is no HTTP cache
+    at all and the <a>cache mode</a> is considered "<code>no-cache</code>".
```
domenic: "fetch behaves as if there is no HTTP cache at all and the <a>cache mode</a> is considered "<code>no-cache</code>"." is not precise enough. You should add a second step which says something like

```html
If <var>httpCache</var> is null, then set <var>httpRequest</var>'s <a for=request>cache mode</a> to "<code>no-cache</code>".
```

fetch.bs
```diff
@@ -4303,14 +4332,19 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
    <li>
     <p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is neither "<code>no-store</code>"
     nor "<code>reload</code>", then:
 
     <ol>
      <li>
-      <p>Set <var>storedResponse</var> to the result of selecting a response from the HTTP cache,
-      possibly needing validation, as per the
+      <p>Set <var>storedResponse</var> to the result of selecting a response from the
+      <var>httpCache</var>, possibly needing validation, as per the
```
domenic: This step will be reached even if the mode is no-cache. Is that desired? Maybe no-cache is not correct, and it should be no-store instead?

Your goal is to prevent any steps that reference _httpCache_ from being reached if _httpCache_ is null, because that would cause a null pointer exception.

fetch.bs
```diff
@@ -4414,8 +4448,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li><p>If <var>httpRequest</var>'s <a for=request>method</a> is
    <a href=https://tools.ietf.org/html/rfc7231#safe.methods>unsafe</a> and
    <var>forwardResponse</var>'s <a for=response>status</a> is in the range <code>200</code> to
-   <code>399</code>, inclusive, invalidate appropriate stored responses in the HTTP cache, as per
-   the "<a href=https://tools.ietf.org/html/rfc7234#section-4.4>Invalidation</a>" chapter of
+   <code>399</code>, inclusive, invalidate appropriate stored responses in the <var>httpCache</var>,
```
domenic:
```suggestion
   <code>399</code>, inclusive, invalidate appropriate stored responses in <var>httpCache</var>,
```

fetch.bs
```diff
@@ -4442,8 +4476,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
      <li><p>Set <var>response</var> to <var>forwardResponse</var>.
 
      <li>
-      <p>Store <var>httpRequest</var> and <var>forwardResponse</var> in the HTTP cache, as per the
-      "<a href=https://tools.ietf.org/html/rfc7234#section-3>Storing Responses in Caches</a>"
+      <p>Store <var>httpRequest</var> and <var>forwardResponse</var> in the <var>httpCache</var>, as
```
domenic:
```suggestion
      <p>Store <var>httpRequest</var> and <var>forwardResponse</var> in <var>httpCache</var>, as
```

fetch.bs
```diff
@@ -4692,7 +4726,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <li><p>If <var>response</var> is not a
    <a>network error</a> and <var>request</var>'s
    <a for=request>cache mode</a> is not "<code>no-store</code>",
-   update <var>response</var> in the HTTP cache for <var>request</var>.
+   update <var>response</var> in the <var>httpCache</var> for <var>request</var>.
```
domenic:
```suggestion
   update <var>response</var> in <var>httpCache</var> for <var>request</var>.
```

fetch.bs
shivanigithub: Added triple keying as the default option here.

fetch.bs
shivanigithub: Removed

fetch.bs
shivanigithub: done

fetch.bs
shivanigithub: done

fetch.bs
shivanigithub: Made this change

fetch.bs
shivanigithub: done

fetch.bs
shivanigithub: Ah that's right, it should be no-store. Updated.

fetch.bs
shivanigithub: Using <a spec> gave an error in make deploy, so I've changed it to <a href>

fetch.bs
shivanigithub: done

shivanigithub:
Addressed feedback, made triple keying and agent cluster key the default based on discussion with Firefox. Also added httpCache to http-network-fetch section. PTAL, thanks!

fetch.bs
```diff
@@ -2049,6 +2049,36 @@ clearly stipulates that <a>connections</a> are keyed on
...
+ <li>
+  <p>If <var>environmentSettings</var> is null, then return null.</p></li>
+
+ <li>
+  <p>Let <var>topLevelOrigin</var> be <var>environmentSettings</var>'s top-level origin and
+  <var>origin</var> be its <a for="environment settings object">origin</a>.</p></li>
+
+ <li>
+  <p>Let <var>topLevelOriginAgentClusterKey</var> be the result of
+  <a href="https://html.spec.whatwg.org/#obtain-agent-cluster-key">obtaining an agent cluster key</a>,
```
domenic: These should also be updated to `<a>obtaining an agent cluster key</a>`, but will need HTML-side updates before that correctly auto-links. No action for now but something to keep in mind before the editors do the full merge.

shivanigithub:
/cc @jkarlin

Thanks Domenic!

annevk:
This looks really good to me, very nice work! We'll have to do some additional abstraction as we deploy this key across more features, but it's almost self-evident how that would be done.

The main thing I'm worried about here is the concern @domenic had as well, that it's not entirely clear where we will end up. https://github.com/shivanigithub/http-cache-partitioning/issues/2 and https://github.com/privacycg/proposals/issues/4 go into this.

I wonder if the initial take could be a bit more hand-wavy about the second key (the top-level one seems like a given at this point) and point to an outstanding issue that captures the debate.

shivanigithub:
Addressed the concern about triple keying and added a note pointing to the issue.

fetch.bs
domenic: Within 24 hours of this comment, this (and the blow one about obtaining an agent cluster key) should be fixable, since the HTML PR merged.

fetch.bs
annevk: I created https://github.com/whatwg/html/pull/5354 to discuss this.

annevk:
I addressed a bunch of nits and created an HTML PR draft for the remaining integration issue. I'm going to be out for two weeks and hopefully in that time there will be some more progress on the keying debates linked in https://github.com/whatwg/fetch/pull/943#issuecomment-595297650 as I'm not entirely comfortable merging this as-is. Perhaps we should only list the top-level origin's site as key and state that additional keys may be added or some such for now. (The HTTP cache is already optional so that doesn't seem too bad.)

@shivanigithub have you looked at the impact on web-platform-tests btw? Should we have some coverage for this? Are tests impacted?

shivanigithub:
We do have web platform tests for this. Here's the chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/1890952

shivanigithub:
Removed the definition of originAgentClusterKey and made the second key a bit more vague.

annevk:
Thanks @shivanigithub. I updated HTML to define "site" and "obtain a site" and we should be able to start using those tomorrow (takes a while before exported terms are indeed exported).

So I expect we can merge this then or the day after. If anyone needs more time this would be a good time to say so.

domenic:
Do we want web platform tests?

shivanigithub:
Web platform tests exist [here](https://github.com/web-platform-tests/wpt/blob/master/fetch/http-cache/split-cache.tentative.html)

domenic:
Thanks! That seems to only cover the popup case, right? Not iframes?

fetch.bs
```diff
@@ -4303,14 +4331,21 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     <var>httpRequest</var>'s
     <a for=request>credentials mode</a>.
 
+   <li><p>Set <var>httpCache</var> to the result of
+   <a lt="determine the HTTP cache partition">determining the HTTP cache partition</a> given
+   <var>httpRequest</var>.</p></li>
+
+   <li><p>If <var>httpCache</var> is null, then set <var>httpRequest</var>'s
+   <a for=request>cache mode</a> to "<code>no-store</code>".</p></li>
```
mnot: How common will this be? Note that currently it can have the effect of disabling upstream caches as well; see #722.

annevk: Hmm that's a good point. It should basically never be null, but I think this is buggy for user-initiated top-level navigations at the moment as they don't have a client. (But that might be buggy in Fetch overall...)

shivanigithub:
Uploaded another patch with a nit: renaming topLevelOriginAgentClusterKey to topLevelSite

fetch.bs
```diff
@@ -2050,6 +2050,32 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+ <li><p>If <var>environmentSettings</var> is null, then return null.
+
+ <li><p>Let <var>topLevelOrigin</var> be <var>environmentSettings</var>'s
+ <a for="environment settings object">top-level origin</a>.
+
+ <li><p>Let <var>topLevelSite</var> be the result of
+ <a lt="obtain a site">obtaining a site</a>, given <var>topLevelOrigin</var>.
+
+ <li><p>Return the HTTP cache associated with <var>topLevelSite</var> and possibly, a second key.
```
domenic:
```suggestion
 <li><p>Return the HTTP cache associated with <var>topLevelSite</var> and, possibly, a second key.
```

fetch.bs
shivanigithub: I put this null handling here to handle the case where "determine the HTTP cache partition" returns null, which it does if request's [client](https://fetch.spec.whatwg.org/#concept-request-client) is null. Not sure when that happens, though.

shivanigithub:
> Thanks! That seems to only cover the popup case, right? Not iframes?

Created https://bugs.chromium.org/p/chromium/issues/detail?id=1064765 to add a test for the iframes case.

fetch.bs
annevk: One example is a user-initiated top-level navigation. For those we should use the site of the request URL, but this would not do that.

I suspect the client is also wrong if tab A navigates tab B. We wouldn't want to use the key from tab A.

shivanigithub:
Checking in: is the spec merge blocked on something from my side? (Last comment is regarding adding an iframe wpt, is that a blocker here?)

annevk:
Both issues mentioned https://github.com/whatwg/fetch/pull/943#discussion_r399082089 seem pretty problematic to me:

> One example is a user-initiated top-level navigation. For those we should use the site of the request URL, but this would not do that.
>
> I suspect the client is also wrong if tab A navigates tab B. We wouldn't want to use the key from tab A.

Unfortunately I don't really have a complete handle on the appropriate fix.

For navigation requests we probably need to use the reserved client. However, that doesn't always have the top-level origin at the moment.

shivanigithub:
> Both issues mentioned [#943 (comment)](https://github.com/whatwg/fetch/pull/943#discussion_r399082089) seem pretty problematic to me:
> 
> > One example is a user-initiated top-level navigation. For those we should use the site of the request URL, but this would not do that.
> > I suspect the client is also wrong if tab A navigates tab B. We wouldn't want to use the key from tab A.
> 
> Unfortunately I don't really have a complete handle on the appropriate fix.
> 
> For navigation requests we probably need to use the reserved client. However, that doesn't always have the top-level origin at the moment.

Ah, I didn't realize that navigation requests settings object does not have the destination origin.

shivanigithub:
> > Both issues mentioned [#943 (comment)](https://github.com/whatwg/fetch/pull/943#discussion_r399082089) seem pretty problematic to me:
> > > One example is a user-initiated top-level navigation. For those we should use the site of the request URL, but this would not do that.
> > > I suspect the client is also wrong if tab A navigates tab B. We wouldn't want to use the key from tab A.
> > 
> > 
> > Unfortunately I don't really have a complete handle on the appropriate fix.
> > For navigation requests we probably need to use the reserved client. However, that doesn't always have the top-level origin at the moment.
> 
> Ah, I didn't realize that navigation requests settings object does not have the destination origin.

Uploaded a patch that uses target browsing context's settings object if the request is a navigation request. I am assuming that the target browsing context has been set up with the settings object at this point in fetch. Can you take a look if this solves the issue. Thanks!

fetch.bs
```diff
@@ -2050,6 +2050,34 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+<p>To
+<dfn lt="determine the HTTP cache partition|determining the HTTP cache partition">determine the HTTP cache partition</dfn>,
+given <var>request</var>, run these steps:
+
+<ol>
+ <li><p>If <var>request</var> is a <a>navigation request</a>, set <var>environmentSettings</var> to
+ <var>request</var>'s <a for="environment">target browsing context</a>'s <a>active document</a>'s
+ <a>environment settings object</a>. Otherwise let <var>environmentSettings</var> be
+ <var>request</var>'s <a for=request>client</a>.
```
annevk, line 2063: You need the request's reserved client before you can get to a target browsing context. Any reason not to use the creation URL's origin of the reserved client?

shivanigithub: Added the request's reserved client.
Regarding using the creation URL's origin, in case of iframe, that will not give the top-level origin, right and we want the top-level origin as the partitioning key for an iframe navigation request

shivanigithub:
@annevk : friendly ping.

fetch.bs
annevk: That's a good point, but I don't think this works either as the active document isn't the document we are navigating towards so it might have the wrong set of origins.

annevk: We probably need to put top-level origin on an environment.

shivanigithub: And that's because the active document is not yet set and will be set after this fetch, right? 

Then another option could be to only special case the main frame's navigation fetch request and not the iframe's, since iframe navigation request's partitioning key is the same as that of a subresource in the main frame.
And if we are only special casing the main frame navigation request then we could use the creation URL's origin, I guess. Does that make sense?

annevk: I'm afraid there's another corner case. If you have A which embeds B and popups C which embeds D, A or B can sometimes navigate D, but we should use the key from C. (Unfortunately it's not clearly defined when A/B can navigate D. It's a very longstanding issue against HTML.)

shivanigithub: In the scenario that you mention D is an iframe, and if we leave the iframe's navigation request case handled similarly as a subresource, then it should be ok?

annevk: Here the client would be A/B, not C. For a subresource of C the client would always be C. Frames are different.

shivanigithub: Got it. 
I was thinking the following for getting the top-level origin for iframe and top-level navigation requests:

- iframe navigation request: request's reserved client's target browsing context's top-level browsing context's active document's origin. As I understand it, In the case given above target browsing context's top-level browsing context will be C and not A/B, right?
- top-level navigation request: request's origin

annevk: That might work, yeah! It seems cleaner to make top-level origin available on reserved client though.

shivanigithub: Created https://github.com/whatwg/html/pull/5491 to define the target top-level origin in the reserved client.

shivanigithub:
Added https://github.com/whatwg/html/pull/5491 to define the target top-level origin in the reserved client and uploaded a patch here to use that.

annevk:
I squashed all existing commits and rebased on master. I also reworded the initial commit message to include some more information.

In a subsequent commit I will attempt to account for the new infrastructure from https://github.com/whatwg/html/pull/5491 as discussed at https://github.com/whatwg/html/issues/5558.

annevk:
This still fails to build because the HTML side hasn't landed yet.

@shivanigithub could you take a look?

fetch.bs
```diff
@@ -2118,6 +2118,55 @@ clearly stipulates that <a>connections</a> are keyed on
...
+   <a for="environment">top-level origin</a>.
+
+   <li>
+    <p>If <var>topLevelOrigin</var> is null, then set <var>topLevelOrigin</var> to
+    <var>request</var>'s <a for=request>reserved client</a>'s
+    <a for="environment">top-level creation URL</a>'s <a for=url>origin</a>.
+
+    <p class=note>This happens for top-level navigations.
+  </ol>
+
```
shivanigithub, line 2155: How does https://github.com/whatwg/fetch/pull/943#discussion_r410351876 work with this? As per the first condition it will be client's topLevelOrigin, but wouldn't that map to A/B instead of C?

annevk: Thanks. We should use the reserved client if non-null and only then use the regular client if it's still null. I'll fix that.

fetch.bs
```diff
@@ -2128,29 +2128,31 @@ given <var>request</var>, run these steps:
...
    <a for="environment">top-level origin</a>.
 
+   <li>
+    <p>If <var>topLevelOrigin</var> is null, then set <var>topLevelOrigin</var> to
+    <var>request</var>'s <a for=request>reserved client</a>'s
+    <a for="environment">top-level creation URL</a>'s <a for=url>origin</a>.
+
+    <p class=note>This happens for top-level navigations only.
+
    <li><p>Assert: <var>topLevelOrigin</var> is an <a for=/>origin</a>.
```
shivanigithub, line 2144: Since topLevelOrigin can be null here, is it ok to have this assert for null?

shivanigithub: Isn't that being checked in 2148

annevk: That only applies if the reserved client was null, but I think I can maybe improve this a bit by using otherwise more.

fetch.bs
```diff
@@ -2118,6 +2118,50 @@ clearly stipulates that <a>connections</a> are keyed on
...
+
+ <li><p>Let <var>topLevelSite</var> be the result of <a lt="obtain a site">obtaining a site</a>,
+ given <var>topLevelOrigin</var>.
+
+ <li>
+  <p>Return the HTTP cache associated with <var>topLevelSite</var> and, possibly, a second key.
+  [[!HTTP-CACHING]]
+
+  <p class=XXX>The second key is intentionally a little vague as the finer points are still
+  evolving. See <a href=https://github.com/whatwg/fetch/issues/904>issue #904</a>.
```
shivanigithub, line 2161: Would this be the best link for this or should we add either the privacycg storage partitioning issue link or the https://github.com/shivanigithub/http-cache-partitioning/issues/2?

annevk: I guess that issue and https://privacycg.github.io/storage-partitioning/ probably do a better job of capturing things. What do you think about porting that issue into this repository as a new issue, link that from here, and make this PR close 904?

shivanigithub: Do you mean the shivanigithub/http-cache-partitioning#2 issue to be ported to a new issue in https://privacycg.github.io/storage-partitioning/ ? I think this plan sounds good.

annevk: I meant here in the Fetch repository, but that seems fine too. I'll get to this on Monday.

annevk:
@shivanigithub @domenic was asking for additional test coverage above, do you know if that has been sorted since then? Is there a PR to remove `.tentative` from the tests? Any failures in Safari we should file a bug for? (As Chrome and Firefox haven't shipped yet, but are working on this, no bugs needed for them as far as I know.)

shivanigithub:
No updates on the web platform tests as of now.

fetch.bs
shivanigithub: Yeah fetch repository seems fine too.

shivanigithub:
https://bugs.chromium.org/p/chromium/issues/detail?id=1064765 is the bug for tracking the new test to be added. 
/cc @alexmturner will start working on that.
@annevk : the additional tests don't block this spec change, right?

annevk:
They normally do, we try to be pretty strict about tests as they often uncover issues that might warrant further specification changes. See https://whatwg.org/working-mode#changes. I guess if @domenic is okay with it I'm happy with going ahead here, but we should at least have a PR for removing `.tentative` from the existing tests and know browser results for them.

shivanigithub:
Iframe test added:
https://github.com/web-platform-tests/wpt/pull/24075

and .tentative removed:
https://github.com/web-platform-tests/wpt/pull/24070 

Thanks @alexmturner!

annevk:
That's very strange. @youennf thoughts?

@alexmturner thanks! As for the tests, from a quick skim they look good, but the naming could use a bit of work to instead talk about top-level site and also cross-site rather than cross-origin. One thing that might be worth doing is to cover a bunch more MIME types to ensure that any kind of specialized caches are also covered. @bakulf pointed me to https://searchfox.org/mozilla-central/source/toolkit/components/antitracking/test/browser/browser_staticPartition_cache.js#158-183 which unfortunately were not written as web-platform-tests. (I don't think that should block this PR though.)

fetch.bs
```diff
@@ -2118,6 +2118,50 @@ clearly stipulates that <a>connections</a> are keyed on
...
+   <li><p>Set <var>topLevelOrigin</var> to <var>request</var>'s <a for=request>reserved client</a>'s
+   <a for="environment">top-level origin</a>.
+
+   <li>
+    <p>If <var>topLevelOrigin</var> is null, then set <var>topLevelOrigin</var> to
+    <var>request</var>'s <a for=request>reserved client</a>'s
+    <a for="environment">top-level creation URL</a>'s <a for=url>origin</a>.
+
+    <p class=note>This happens for top-level navigations only.
+  </ol>
```
domenic, line 2143: Can you remind me why we're doing this here, instead of having HTML ensure top-level origin is always non-null by setting it to the origin of the top-level creation URL?

alexmturner:
Small update re Safari and the WPTs. I believe the WPT failure is likely related to Safari failing the first test in https://wpt.fyi/results/fetch/http-cache/freshness.html ("HTTP cache reuses a response with a future Expires"). Strangely, when I test locally (on wpt.live), Safari passes all freshness and split-cache tests. I'm not sure what the cause of the disparity is.

I'll update the naming to reflect the switch from origin to site and have a look at extending the tests for different MIME types soon.

fetch.bs
annevk: My thinking was that a null -> origin transition would be clearer than an origin -> origin transition. And it would also make it clearer here that we compute the origin directly from a URL rather than the response.

domenic: Hmm, that feels like it kind of defeats the point of environment/ESO being an abstraction that other specs don't have to worth about the details of, but can just use and let HTML handle the management. But I don't feel too strongly.

annevk: I think generally that's true and will remain true, but there's a couple places that will need to care about the distinction and networking is one of those. (Other places generally only get access to an environment settings object.)

shivanigithub:
The terminology in tests was updated via https://github.com/web-platform-tests/wpt/pull/24173

@annevk : Can the "needs tests" label be now removed? Is there anything else blocking the merge of this PR?

annevk:
Is there a PR for renaming the tests? That is, to drop .tentative?

shivanigithub:
Yes, it's here: https://github.com/web-platform-tests/wpt/pull/24070

annevk:
Thanks for making this happen @shivanigithub! This is a pretty big step for https://privacycg.github.io/storage-partitioning/ thanks to your efforts over the past ten months. We now have all the infrastructure in place for partitioning network state. 🎉

shivanigithub:
Thanks @annevk, @domenic for all your help and reviews on this!

PR #948, opened 2019-10-07
Revisit `destination` for `<iframe>`, `<frame>`, `<embed>`, and `<object>`.

mikewest:
As discussed in w3c/webappsec-fetch-metadata#45, splitting 'document' into a set of
destination values that developers can use to determine whether the request is for
a top-level document or a nested document will allow us to simplify Fetch Metadata
checks performed server side.

***
<a href="https://whatpr.org/fetch/948.html" title="Last updated on Jun 5, 2020, 8:23 AM UTC (939613d)">Preview</a> | <a href="https://whatpr.org/fetch/948/9ac2b5b...939613d.html" title="Last updated on Jun 5, 2020, 8:23 AM UTC (939613d)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1236,7 +1236,7 @@ not always relevant and might require different behavior.
    <td>CSP, NEL reports.
   <tr>
    <td>"<code>document</code>"
-   <td>HTML's navigate algorithm.
+   <td>HTML's navigate algorithm (top-level only).
   <tr>
    <td>"<code>frame</code>"
    <td><code>child-src</code>
@@ -3265,8 +3265,8 @@ the request.
      <li><p>Let <var>value</var> be `<code>*/*</code>`.
 
      <li>
-      <p>Otherwise, a user agent should set <var>value</var> to the first matching statement, if
-      any, switching on <var>request</var>'s <a for=request>destination</a>:
+      <p>A user agent should set <var>value</var> to the first matching statement, if any, switching
+      on <var>request</var>'s <a for=request>destination</a>:
       <!-- https://github.com/whatwg/fetch/issues/43#issuecomment-97909717 -->
 
       <dl class=switch>
--- a/fetch.bs
+++ b/fetch.bs
@@ -1185,6 +1185,8 @@ the empty string,
 "<code>document</code>",
 "<code>embed</code>",
 "<code>font</code>",
+"<code>frame</code>",
+"<code>iframe</code>",
 "<code>image</code>",
 "<code>manifest</code>",
 "<code>object</code>",
@@ -1228,7 +1230,7 @@ not always relevant and might require different behavior.
    <th>CSP directive
    <th>Features
   <tr>
-   <td rowspan=18>""
+   <td rowspan=19>""
    <td>"<code>report</code>"
    <td rowspan=2>&mdash;
    <td>CSP, NEL reports.
@@ -1236,9 +1238,13 @@ not always relevant and might require different behavior.
    <td>"<code>document</code>"
    <td>HTML's navigate algorithm.
   <tr>
-   <td>"<code>document</code>"
+   <td>"<code>frame</code>"
    <td><code>child-src</code>
-   <td>HTML's <code>&lt;iframe></code> and <code>&lt;frame></code>
+   <td>HTML's <code>&lt;frame></code>
+  <tr>
+   <td>"<code>iframe</code>"
+   <td><code>child-src</code>
+   <td>HTML's <code>&lt;iframe></code>
   <tr>
    <td>""
    <td><code>connect-src</code>
@@ -1604,19 +1610,15 @@ whose <a for=request>destination</a> is "<code>audio</code>", "<code>audioworkle
 "<code>script</code>", "<code>style</code>", "<code>track</code>", "<code>video</code>",
 "<code>xslt</code>", or the empty string.
 
-<p>A <dfn export>potential-navigation-or-subresource request</dfn> is a
-<a for=/>request</a> whose
-<a for=request>destination</a> is
-"<code>object</code>" or "<code>embed</code>".
-
 <p>A <dfn export>non-subresource request</dfn> is a <a for=/>request</a>
-whose <a for=request>destination</a> is "<code>document</code>",
-"<code>report</code>", "<code>serviceworker</code>", "<code>sharedworker</code>",
-or "<code>worker</code>".
+whose <a for=request>destination</a> is "<code>document</code>", "<code>embed</code>",
+"<code>frame</code>", "<code>iframe</code>", "<code>object</code>", "<code>report</code>",
+"<code>serviceworker</code>", "<code>sharedworker</code>", or "<code>worker</code>".
 
 <p>A <dfn export>navigation request</dfn> is a <a for=/>request</a> whose
 <a for=request>destination</a> is
-"<code>document</code>".
+"<code>document</code>", "<code>embed</code>", "<code>frame</code>", "<code>iframe</code>",
+or "<code>object</code>".
 
 <p class=note>See <a for=/>handle fetch</a> for usage of these terms.
 [[!SW]]
@@ -3262,16 +3264,17 @@ the request.
     <ol>
      <li><p>Let <var>value</var> be `<code>*/*</code>`.
 
-     <li><p>If <var>request</var> is a <a>navigation request</a>, a user agent should set
-     <var>value</var> to
-     `<code>text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8</code>`.
-
      <li>
       <p>Otherwise, a user agent should set <var>value</var> to the first matching statement, if
       any, switching on <var>request</var>'s <a for=request>destination</a>:
       <!-- https://github.com/whatwg/fetch/issues/43#issuecomment-97909717 -->
 
       <dl class=switch>
+       <dt>"<code>document</code>"
+       <dt>"<code>frame</code>"
+       <dt>"<code>iframe</code>"
+       <dd>`<code>text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8</code>`
+
        <dt>"<code>image</code>"
        <dd>`<code>image/png,image/svg+xml,image/*;q=0.8,*/*;q=0.5</code>`
 
@@ -5823,7 +5826,7 @@ dictionary RequestInit {
   any window; // can only be set to null
 };
 
-enum RequestDestination { "", "audio", "audioworklet", "document", "embed", "font", "image", "manifest", "object", "paintworklet", "report", "script", "sharedworker", "style",  "track", "video", "worker", "xslt" };
+enum RequestDestination { "", "audio", "audioworklet", "document", "embed", "font", "frame", "iframe", "image", "manifest", "object", "paintworklet", "report", "script", "sharedworker", "style",  "track", "video", "worker", "xslt" };
 enum RequestMode { "navigate", "same-origin", "no-cors", "cors" };
 enum RequestCredentials { "omit", "same-origin", "include" };
 enum RequestCache { "default", "no-store", "reload", "no-cache", "force-cache", "only-if-cached" };


mikewest:
I'll follow this up with an HTML patch in a moment.

fetch.bs
```diff
@@ -1567,13 +1573,13 @@ whose <a for=request>destination</a> is "<code>audio</code>", "<code>audioworkle
 "<code>object</code>" or "<code>embed</code>".
 
 <p>A <dfn export>non-subresource request</dfn> is a <a for=/>request</a>
-whose <a for=request>destination</a> is "<code>document</code>",
-"<code>report</code>", "<code>serviceworker</code>", "<code>sharedworker</code>",
-or "<code>worker</code>".
+whose <a for=request>destination</a> is "<code>document</code>", "<code>frame</code>",
+"<code>iframe</code>", "<code>report</code>", "<code>serviceworker</code>",
+"<code>sharedworker</code>", or "<code>worker</code>".
```
annevk: This isn't accurate anymore. Currently for embed/object the destination would be document if they have a browsing context that is navigated and they would end up classified here as a non-subresource request.

mikewest: Currently, `embed` and `object` are "potential-navigation-or-subresource" requests, and don't get categorized into either "subresource" or "non-subresource". I don't think this patch changes that.

Looking at "display a plugin", it looks like `<embed>` and `<object>` will always create a nested browsing context. Is that how we're categorizing "subresource" vs "non-subresource"? If so, then I agree that it seems like we should lump them into the "non-subresource" bucket.

It's not clear to me what the impact is there, though. Where is "non-subresource request" used?

annevk: My understanding is that `embed` (and `object`) are initially no-cors / embed, but change to navigate / document the moment they display a non-XXX resource (maybe XXX is only images?) and instantiate a nested browsing context.

https://w3c.github.io/ServiceWorker/ uses non-subresource.

mikewest: As far as Fetch is concerned, I think `<embed>` and `<object>` are both always `no-cors` / {`embed`,`object`}.

Tracing through from https://html.spec.whatwg.org/#the-embed-element-setup-steps:

1.  Step 3.4 fetches with `no-cors` / `embed`.
2.  Step 3.4.3 switches on the type of the response (in a way that I'm not sure matches implementations; both Firefox and Chrome render `text/html`, for instance) to either [display a plugin](https://html.spec.whatwg.org/#display-a-plugin) or navigate to the response.
3.  [Displaying a plugin](https://html.spec.whatwg.org/#display-a-plugin) drops any nested browsing context in step 1, and hand-waves at finding a plugin for the response type.
4.  Navigating to the response does not re-fetch with `navigation` / `document`; it navigates to the response that was given for the `no-cors`/`embed` fetch above.

`<object>` looks pretty similar, though more capacious with regard to the types it handles specially.

(It's probably worth reevaluating all of this complexity in the light of Flash's impending death. Perhaps we can make `<object>` and `<embed>` act like strangely-spelled `<iframe>`s at some point?)

> https://w3c.github.io/ServiceWorker/ uses non-subresource.

https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm uses "non-subresource request" only after punting entirely on "potential-navigation-or-subresource" requests, which means that the categorization would have no practical effect there.

annevk: If you navigate a browsing context whose owner is `embed`, how does it end up with `no-cors` / `embed`?

mikewest: Huh. I hadn't thought about that case. My initial thoughts are:

1. Don't do that.
2. That's surprising behavior!

Should we special-case this in the navigate algorithm by checking `<embed>` and `<object>` in https://github.com/whatwg/html/pull/4976 along with `<iframe>` and `<frame>`?

annevk: I think it makes sense that it becomes `navigate` as at that point it is navigation and cannot be used as a subresource anymore. It might be okay that it stays `embed`/`object` though if we go that way for `frame`/`iframe`.

mikewest:
> at that point it is navigation and cannot be used as a subresource anymore

I hesitate to accept this claim. :) Per spec, you're probably right. But short of auditing Chromium's plugin loading code, I'm not sure that it's practically the case in that engine. The code is a big mess, and it's not clear to me that it couldn't backslide into having an effect on the document in which it's embedded. I really can't wait until we can lock most of the weirdness behind a PPAPI flag for enterprise.

I'm willing to be convinced that `navigate` / `embed` is reasonable, but I think I'd like to err on the side of advertising the worst-case potential with `no-cors`.

annevk:
Wouldn't that end up affecting other parts of the system in weird ways? E.g., how a service worker is selected to handle the request? And is `no-cors` always worse than `navigate`?

mikewest:
> Wouldn't that end up affecting other parts of the system in weird ways? E.g., how a service worker is selected to handle the request?

The way things are specified today, I'm not sure any of our current behaviors around this space are correct. :)

What would you like the behavior to be for Service Workers? Could we treat `<embed>` navigations the same way we treat the original request from the `<embed>` element itself?

cc @jakearchibald who possibly has opinions.

> And is no-cors always worse than navigate?

My assumption is that `no-cors` can/will end up in an attacker's memory, whereas a `navigate` probably won't (assuming out-of-process frames). @arturjanc probably has an opinion from the server side of things...

annevk:
If A frames B using `embed`, we cannot use A's service worker for navigations that happen in B. Navigations have to be in scope of a service worker, which only a service worker from B can be. We could say they always bypass the service worker, maybe, but I doubt that's what's going on.

mikewest:
My reading of the service worker spec is that we skip the service worker for these requests. But I don't know if that's what vendors actually implement. Pinging @jakearchibald again who might have opinions (and data?).

@arturjanc apparently has the plague today, but when he recovers, he might be able to weigh in on the threat model around `no-cors` and `navigate`, and whether my assertions above match the mental model his team is working with.

(In the hopes of avoiding surprise, I currently think that this is the last open question we need to answer before shipping a first pass at `sec-fetch-dest`, which @ivanlish has started to work through in Chromium. Once we resolve this, I'd like to send an intent to ship for the feature so Artur, et al, can start rolling it out server-side.)

annevk:
I'm really curious whether we skip service workers once a browsing context is established there. I'd like to see that tested at least.

From Firefox's perspective solving `embed`/`object` is the main blocker too.

arturjanc:
> weigh in on the threat model around `no-cors` and `navigate`, and whether my assertions above match the mental model his team is working with.

They do match our mental model: we expect developers to focus on blocking cross-site `no-cors` loads while allowing navigations so they protect from XS-Leaks but still allow incoming links to work.

One counterexample could be sites that don't care about XS-Leaks but want to more tightly control navigations to defend from XSS and e.g. prevent cross-site navigations to anything but a tightly controlled set of application endpoints. But for this -- less common -- scenario developers already have `X-Frame-Options` to fall back on so the resource shouldn't render when it's embedded; we can also publish reference logic that will consider a cross-site `mode`=`no-cors` + `dest`=`embed`/`object` request as a potential navigation (in addition to `mode`=`navigate`). This complication in the less likely use case for Fetch Metadata seems like a good trade-off compared to requiring everyone who cares about XS-Leaks to implement special handling for loads via `embed`/`object`.

Overall, I like the model of having initial loads via `embed`/`object` be `no-cors`, but turn into `navigate` when the browsing context gets navigated.

mikewest:
> Overall, I like the model of having initial loads via `embed`/`object` be `no-cors`, but turn into `navigate` when the browsing context gets navigated.

@annevk I think you would be happy with this too, given your comments in https://github.com/whatwg/fetch/pull/948#discussion_r331992431. Is that still your preference?

I don't think this distinction will be easy to implement in Chromium, given the way `<embed>`/`<object>` are intertwined with navigation code, but I'm happy to assume that it's possible, spec and test it accordingly, and move on to more interesting debates. :)

---

For clarity, I think we're saying that the initial request from `<embed>` would send:

```
Sec-Fetch-Dest: embed
Sec-Fetch-Mode: no-cors
Sec-Fetch-Site: whatever
```

and subsequent requests from `<embed>` caused either by internal interaction or by poking at `window.frames` would send:

```
Sec-Fetch-Dest: embed
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: whatever
```

annevk:
The way the standard describes `<embed src>`/`<object data>` is roughly:

1. Let response be the result of fetching something.
2. If response is something that necessitates a browsing context, then:
   1. Create one.
   2. Navigate that browsing context to response.
3. Otherwise, display response without browsing context.

And then once there is a browsing context, you'd navigate that normally. It's surprising that Chrome has apparently a very different model from this. I haven't checked with folks familiar with Gecko's loading code recently, but I'm pretty sure the above is somewhat accurate for Gecko (now whether it's easy to get the right values might still be up for debate there of course).

fetch.bs
mikewest: So, given that the potential resolution to the underlying question in https://github.com/whatwg/fetch/pull/948#issuecomment-543560997 puts the `embed` and `object` destinations into two categories (depending on the mode), what would you like me to do with these definitions ("navigation request", "non-subresource request"?

Should we leave them as they are in this patch, or do you want to include/excude requests with a `mode` of `navigate` as appropriate?

mikewest:
> It's surprising that Chrome has apparently a very different model from this.

Chromium is nothing if not surprising. The challenge I'm thinking of is that we're currently keying the `Sec-Fetch-Dest` and `Sec-Fetch-Mode` header values off the framing element type (see https://cs.chromium.org/chromium/src/content/browser/frame_host/navigation_request.cc?rcl=5514046a3db301445a8e1c8aa23f638523e4f462&l=323 for example), and we run through the same codepath for the initial navigation from `<embed>` as we do for subsequent navigations. That is, Chromium treats the request in your step 1 as a navigation regardless of whether we're going to render it as a plugin or not. My recollection here is a bit fuzzy, but I think we end up synthesizing a [`PluginDocument`](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/html/plugin_document.h?q=plugindocument&dr=CSs&l=37) to hold the plugin, so when we extracted navigation from the renderer process, `<embed>` and `<object>` came right along with it.

jakearchibald:
I'm a little lost about what's trying to be achieved here, so I'm sorry if this is useless:

With service worker, we're happy for a service worker to intercept a request for `<script src="https://example.com">` instead deliver a resource from `evil.com`. The script runs in the context of the page, and the service worker owns the page, so it's only lying to itself.

A service worker on `evil.com` cannot intercept the request for `<iframe href="http://example.com">`, because, as far as I know, this would allow `evil.com` to run content from a different origin within `example.com`.

This means requests which form a client (which have their own CSP, and run against a particular origin based on their final request), will go through a service worker that matches the request scope, rather than the controller of the client that initiated the request.

This all works nicely because we can tell if a response is going to be used to create a client based on the destination. An `` doesn't, an `<iframe>` does, `<script>` doesn't, `new Worker` does etc etc.

We punted on `<object>`/`<embed>` because they're an exception. Here, the type of the *response* judges whether it creates a client or not. `image/jpeg` doesn't, but `text/html` does. This leads to the kind of mess you see in [step 4.9 of the object element](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-object-element) where it flip-flops around when `<object>` meets appcache.

If we can guarantee in advance whether a `<object>`/`<embed>` resource is going to create a client, we can feed it through service worker, otherwise it feels safest to avoid.

annevk:
@mikewest so `embed`/`object` navigate and if the result of the navigation is an image, the browsing context being navigated is tossed away and the image is rendered? And some care is taken to not exposing the browsing context being navigated to the embedding document?

mikewest:
@annevk:
> so embed/object navigate and if the result of the navigation is an image, the browsing context being navigated is tossed away and the image is rendered?

No. We do the same thing as we do for more general navigation to an image resource: synthesize an `ImageDocument` which wraps the image we just downloaded in a minimal HTML document for rendering. :)

> And some care is taken to not exposing the browsing context being navigated to the embedding document?

I'm not sure what you mean here? We use the origin of the resource as the browsing context's origin, and access controls flow naturally from that.

annevk:
Very interesting, but there's no initial about:blank it seems, so there's no exposed browsing context (e.g., via `self[0]`) until the `load` event fires, as far as I can tell. So it's still a bit different from how `iframe` works, but much less so than in Firefox.

I'm not sure what we want to do in that case.

cc @bzbarsky

mikewest:
> Very interesting, but there's no initial about:blank it seems, so there's no exposed browsing context (e.g., via self[0]) until the load event fires, as far as I can tell. So it's still a bit different from how iframe works, but much less so than in Firefox.

I think that's correct. If/when we actually get rid of plugins, it might be reasonable to make `embed` and `object` work more like strangely-spelled `iframe` elements? Given their usage in the wild, I think it's going to be hard to make substantial changes to their behavior.

Implementation details to the side, does the behavior in https://github.com/whatwg/fetch/pull/948#issuecomment-543560997 seem reasonable to you? If so, I'll figure out how to make it work in HTML and Chromium.

annevk:
If Firefox is okay with moving slowly towards making `embed`/`object` always create browsing contexts as well (even if it's only visible once the `load` event fires), I think the better solution is your initial take of always using mode `navigate`. I've pinged some folks internally and also bz above to see what they think.

mikewest:
I think my original proposal was `no-cors` for everything. `navigate` for everything seems reasonable once we're in a world where plugins aren't a problem. `no-cors` seems reasonable for requests whose response might end up in a plugin depending on type.

I don't think it makes a huge difference either way, as @arturjanc and team would just need to look at the destination in addition to the mode for navigation requests, which they already need to do for `iframe` and `frame`. That said, less well-informed developers might be inclined to treat `navigate` as totally fine.

Better documentation and example code might help mitigate that risk? I can live with either, I think.

bzbarsky:
> it might be reasonable to make embed and object work more like strangely-spelled iframe elements?

They have quite different layout behavior in the image case, no?  Including in Chrome.

> If Firefox is okay with moving slowly towards making embed/object always create browsing contexts as well (even if it's only visible once the load event fires)

What web-observables would this actually affect?

Ignoring for the moment plug-ins (let's pretend we killed them already), there are basically two cases for embed/object in the spec and in Gecko's implementation right now: document and image.  Which case we're in is decided based on the MIME type of the last load triggered via the "src"/"data" attribute respectively.

If you set the relevant attr to a URL that returns a non-image MIME type that the browser is able to render, you get a browsing context (once we get the headers from the server) and general iframe-like layout behavior.  The resulting browsing context can be navigated via its `Location`, including to image URLs (which will render like they would in an iframe).  But setting the relevant attr again will (always? I'd have to check the code) get rid of the browsing context and then run the selection logic again once the response is received.

If you set the relevant attr to a URL that returns an image MIME type, we don't create a browsing context at all.  We just have the image data (like `` does), and create a layout object identical to the one for ``, which knows how to do the right intrinsic sizing, etc.  This _could_ perhaps be done on top of a browsing context but would require some work to make the layout behavior correct.  This is work that Chrome is apparently already doing, or it's doing _something_ at least.  If I do `<object data="100x100px-image" style="height: 200px"> in Chrome, I get a 200x2000 image shown, which is totally not how iframes work.  Also, note that the renderings of an iframe pointing to an image are NOT interoperable across Chrome and Firefox, nor are they standardized, unlike the case of `<object>` pointing to an image.

mikewest:
>They have quite different layout behavior in the image case, no? Including in Chrome. 

That's true, my claim earlier in the thread was incorrect: on testing again, I realize that Chrome has exactly the difference in behavior you're describing for `<embed src="[image URL]">` on the one hand, and direct navigating to an image from within a browsing context associated with `<embed src="[HTML URL]">`. I was testing the latter case, but not the former.

Still, both cases run through the same navigation architecture in Chrome, and we make a decision about whether to commit into a new browsing context vs. render the result.

I suspect that the cases in which the image-like display nature you're pointing to is something we could change if we wanted to simplify the model. Based on the same (sketchy because Canary) data source as above, <0.0001% of page views contain either `<embed>` or `<object>` used to render an image.

bzbarsky:
> Still, both cases run through the same navigation architecture in Chrome, and we make a decision about whether to commit into a new browsing context vs. render the result.

I'm not sure that's the whole story, by the way.  `<object data="image-src">` creates a web-visible `self[0]` (and hence browsing context?) in Chrome if the src is same-origin but not if it's cross-origin, with both having the "image" rendering...  I can't speak to whether both have a browsing context under the hood.

mikewest:
> > Still, both cases run through the same navigation architecture in Chrome, and we make a decision about whether to commit into a new browsing context vs. render the result.
>
> I'm not sure that's the whole story, by the way.

All I was saying here is that the outgoing request runs through `NavigationRequest` in Chromium. I haven't dug into that code enough to determine where we make the decision that it really isn't a navigation, and ought to have different Blink-side behavior, because I agree with you that it is, in fact different. :)

This is getting a little bit into the weeds, though. I'd like to get us back to the high-level question of how the `mode` and `destination` of `<embed>` and `<object>` _ought_ to work, assuming that we can change how they _do_ work today. Given the presence of plugins today, the proposal in https://github.com/whatwg/fetch/pull/948#issuecomment-543560997 still seems reasonable to me (the initial request (and subsequent requests that modify the `<embed>` directly sends `no-cors`/`embed` as they might cause plugin creation, whereas requests that navigate a browsing context created in the `<embed>` would send `navigate`/`embed`).

If we can do away with plugins, and stop special-casing images, then `navigate`/`embed` might be appropriate for everything?

annevk:
Long term we should be able to get rid of plugins, I think. I wanted to double check what ends up happening in Chrome, but I cannot find a plugin example that works `<embed src=flash>` triggers a download, as you would expect from a navigation.

mikewest:
> I wanted to double check what ends up happening in Chrome, but I cannot find a plugin example that works

Try `<embed src="https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf" type="application/pdf">`. That sends `no-cors`/`embed` in Chrome 79, as we currently render PDF via a PPAPI plugin.

annevk:
As far as I can tell that creates a (navigated) browsing context in both Chrome and Firefox (i.e., `frames[0]` is not undefined). In Firefox the resulting `WindowProxy` is always cross-origin, in Chrome a bunch of implementation details are exposed (such as `embed` having a thing called `internalid`) when I put that code on the same origin (via DevTools, but no real reason to distrust). That puts me even more firmly in the navigate camp.

mikewest:
Weird.

The first Flash game I found uses `<embed src="https://img-hws.y8.com/cloud/y8-flash-game/contents/item_versions/flash_games/50996/original/super_drift_2.swf?1521598154" class="item-direct-container resizable" width="850" height="478" id="gamefileEmbed" name="flash-content" align="middle" wmode="window" allowfullscreen="false" allowfullscreeninteractive="false" fullscreenaspectratio="" quality="high" play="true" loop="true" menu="" allowscriptaccess="always" haspriority="true" type="application/x-shockwave-flash" pluginspage="https://www.adobe.com/go/getflashplayer" style="display: block;">"`. Jamming that into https://example.com/ via devtools results in an undefined `frames[0]` (in Chrome 78). I suspect PDF is doing something unexpected. :/

annevk:
Unexpected, but better? I can confirm that Flash (apparently the flash file I tried to use doesn't have the correct MIME type anymore) has the behavior you describe, but if everything else behaves as navigation I'm not sure that should override, given that it's on its way out.

mikewest:
> on its way out

In a year. Probably. :) It seems to me that while it's a thing that's possible, marking the outgoing request as `no-cors` is safer than marking it as `navigate`. After 2020, when we're sure that requests are going to spin up browsing contexts because we've finally gotten rid of Flash, it seems pretty reasonable to revisit this mechanism?

Honestly, I don't have strong opinions about this either way, and I'm pretty sure that @arturjanc has the same take; Google servers will know to look at `sec-fetch-dest` and evaluate the risk accordingly. We can add a note in the spec to that effect and hope other folks pay attention as well. I think it would be helpful to developers to distinguish the request that might or might not create a browsing context from requests that will definitely land in a browsing context, but I also think that it's not a distinction I'm going to jump up and down and insist upon. :)

I think we've played out the arguments here pretty well: what would you like to see happen, @annevk? I'll defer to you for a decision here. `navigate` everywhere for `<embed>` and `<object>`?

annevk:
I would say navigate since flipping later seems problematic from a compatibility perspective. If we also make `X-Frame-Options` always work here there might be less of a security risk?

bzbarsky:
Fwiw, I don't have strong opinions on what the destination should be, as long as "navigate" does not in fact promise that we end up with a browsing context in the image case and we can sort that case out separately.

annevk:
I cannot reproduce Chrome doing something with images that does not match `<iframe>` by the way, experimenting a bit in Live DOM Viewer. That is, setting a height changes the container, but not the image. E.g., `<embed src=image style=height:600px>`. I also tried navigating a document to an image using `target` and `<object name>` and that also did not result in resizing (perhaps as expected), and the same vice versa.

bzbarsky:
That's because Chrome is buggy in spec-violating ways and has different behaviors depending on the value of the `type` attribute.  Try adding `type="image/gif"`.

annevk:
To address your statement above, I think I'm okay with saying that `navigate` does not imply images will have to create a (visible) browsing context. It would be interesting to investigate to what extent sites set `type` to image MIME types. If they don't, it seems we could simplify this given the "broken" behavior in Chrome and Safari.

fetch.bs
mikewest: Coming back to this: since we've decided that `mode` for `embed` and `object` is going to be `navigate`, do you still want changes in this section, or are you comfortable with the current text of this PR?

mikewest: Ok. Is that something you want me to do in _this_ patch, which introduces `iframe` and `frame`? Or would you be ok with me splitting it out into a separate patch (which I think I'd like to do, because some things about changing `<embed>`'s mode to `navigate` aren't clear to me, and will likely require changes in both Fetch and HTML).

If you're alright with deferring it to a future patch, is there anything else we need to change about this one?

annevk: I'm a little worried it's gonna linger and never get fixed, despite being somewhat important from an interoperability perspective.

And note that before it'd go from `no-cors`/`embed` to `navigate`/`document` and that is being changed too.

And given that it also changes for `iframe` et al, there needs to be an associated HTML PR already for this change.

mikewest: > I'm a little worried it's gonna linger and never get fixed, despite being somewhat important from an interoperability perspective.
>
> And note that before it'd go from no-cors/embed to navigate/document and that is being changed too.

I agree that we should fix it in the spec and lock it in with tests. I'm not sure _this_ patch is the right place to do it, given how simple it is today, and how not-simple I think the `<embed>` change is going to be more generally. :)

If you'd prefer one big patch that addresses `<embed>`/`<object>`/`<frame>`/`<iframe>` all together, I'll do that! But I think I'd prefer to just land this and the associated HTML PR, and worry about `<embed>` and `<object>` in a subsequent set of PRs.

> And given that it also changes for iframe et al, there needs to be an associated HTML PR already for this change.

https://github.com/whatwg/html/pull/4976 is that change.

annevk: I might be missing something, but apart from the tests I don't think it's as big of a change on this side? The HTML side would require a lot of work, but not sure all of that is blocking given the many issues embed/object have already.

mikewest: Step 5 of main fetch (https://fetch.spec.whatwg.org/#main-fetch) has pretty different behavior between "navigate" and "no-cors". I think it'll turn out to be incorrect for the initial request from `<embed>`, but I haven't through through what needs to change (or whether I'm even correct that it does need to change).

mikewest: (There's also a minor MIME type change in step 1.3.2 of Fetching (https://fetch.spec.whatwg.org/#fetching), but that's likely not a big deal to change.)

annevk: Ah, what `Accept` should say.

arturjanc:
I think we all want to make a decision here and I don't want to derail a productive discussion, but IMHO there are 3 problems with `mode`=`navigate` for `embed`/`object`:

1. It seems counter-intuitive: in my mental model an `<embed>` or `<object>` work similarly to `` or any other element that loads a non-document subresource. The fact that browsers create a browsing context for some MIME types seems like an implementation detail; I doubt developers are aware of this or rely on it -- I've never seen a browsing context created via `embed` be navigated like an iframe.
2. It is currently unsafe (with respect to Fetch Metadata) because it allows loading images via `<embed>`, and Firefox ignores `X-Frame-Options`/`frame-ancestors` in this case. If a site prevents `cross-site` loading of images but allows `cross-site` navigations, and sets `X-Frame-Options` to protect from embedding, an attacker will still be able to load images cross-site via `<embed>` and e.g. exflitrate them via Spectre.
3. It will complicate writing Fetch Metadata policies that protect from framing because developers will need to consider framed loads via `object` and `embed`, in addition to `frame` and `iframe`. (But this would arguably be the same with @mikewest's `no-cors-on-initial-load-and-then-navigate` model.)

Basically, I think that `navigate` here prioritizes what makes sense to a browser over what makes sense to web developers, and `no-cors` would be a better fit (for a developer, though not necessarily for a browser) and be safer. That said, if we can guarantee that `<embed>`/`<object>` always works exactly the same as an iframe, e.g. respects `X-Frame-Options` and doesn't give the embedder any new ways to interact with the loaded resource such as triggering events that a load in an iframe wouldn't trigger, then I think I can live with `navigate`.

annevk:
Honoring `X-Frame-Options` (and friend) consistently for `embed` and `object` seems good. Could Chrome start doing that for plugins? I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1595491 to consider this on Mozilla's side.

(As for your point 1, the problem is that it doesn't match your mental model, right? SVG will execute script. It won't when fetched with ``. I wouldn't call that a detail.)

mikewest:
> Honoring `X-Frame-Options` (and friend) consistently for embed and object seems good. Could Chrome start doing that for plugins? I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1595491 to consider this on Mozilla's side.

(Lack of) specs to the side, it appears that there's already reasonable interop when creating a browsing context under `<embed>`: compare `data:text/html,<embed src="https://avatars3.githubusercontent.com/u/665379?s=88&v=4" type="image/jpeg">`, which loads @arturjanc's pixelated head as an image, to `data:text/html,<embed src="https://paypal.com/">`, which doesn't load `https://paypal.com/` because of its `frame-ancestors` assertions, and `data:text/html,<embed src="https://avatars3.githubusercontent.com/u/665379?s=88&v=4">`, which doesn't load @arturjanc's pixelated head because of XFO.

We can try do the same for plugin content. I wouldn't be surprised if we found breakage for PDF, but I'm happy to add metrics.

> If a site prevents cross-site loading of images but allows cross-site navigations, and sets X-Frame-Options to protect from embedding, an attacker will still be able to load images cross-site via `<embed>` and e.g. exflitrate them via Spectre.

It seems like the way we're recommending that folks deal with this would be to examine `Sec-Fetch-Dest`, and make a decision based on `img` vs `embed` vs `iframe`?  I mean, it is an additional check, and that does create some complexity, but it's not actually _complicated_. It seems to me that sites are either going to accept the risk associated with a given resource being loaded via `<embed>` or not, and that there's not actually a decision we expect anyone to make about the qualities of that load.

I'd also suggest that `Cross-Origin-Resource-Policy` (and the nascent COEP mechanism) should more or less resolve this concern by requiring explicit permission from a site to embed its bits and pieces iff those pieces would be embedded in a powerful-enough context. Does that mitigate some of your concerns? Or do you see it as tangential?

---

Bottom line, I think there's a reasonable philosophical argument for `no-cors/embed` followed by `navigate/embed` if we create a context and then navigate it. My mental model matches Artur's on this point. That said, I have a hard time getting worked up about it practically. I don't have the impression that sending `navigate/embed` for both actually changes the risk analysis that we'd ask developers to perform, and it seems straightforward to modify the policy recommendations put forth on sites like https://secmetadata.appspot.com/ to match.

bzbarsky:
> which doesn't load @arturjanc's pixelated head because of XFO.

Er..  Both pixelated heads load in Firefox, because they are both treated as the image case (unlike Chrome, which has different behavior depending on the `type` attribute, as you note).

(Note that the Firefox behavior is the correct one per spec at least for `<object>`: the type attribute is advisory and should not affect how the content is handled if it is handled at all.)

mikewest:
> Er.. Both pixelated heads load in Firefox, because they are both treated as the image case (unlike Chrome, which has different behavior depending on the type attribute, as you note).

Indeed. To keep this focused on the alignment in XFO behavior instead of Chrome's non-spec-compliant interpretation of `type`, please pretend that last item read `data:text/html,<embed src="https://vox.com/">` instead.

arturjanc:
I chatted about this with Mike and the tl;dr is that `navigate` for `object`/`embed` is something developers can deal with, under the following assumptions:
1. Browsers will treat browsing contexts created by `<object>` and `<embed>` equivalently to iframes for security purposes, i.e. apply `X-Frame-Options` and `frame-ancestors`, put the browsing context in a separate process (for browsers with out-of-process frames), and not allow any ways to access information about the loaded resource that isn't allowed in an iframe (e.g. don't reveal `width` and `height` as we would for ``).
2. Developers may need to add an extra condition to their policies to prevent their resources from being used as an `<embed>` (because it will appear as a navigation instead of a resource load). This will make policies a little more complicated (which we can handle by sharing reference implementations) and will require developers to consult the value of `Sec-Fetch-Dest` -- so it will be important for browsers to have complete implementations that send the header.

Does this sound reasonable?

annevk:
1. This is already not true for SVG, right? We treat it identical to `<iframe>` in most respects, but also resize the element based on the intrinsic dimensions.

Sounds okay otherwise and I'll file a bug on Firefox to at least honor those headers.

annevk:
I think I might have misunderstood some of @mikewest's comments above. What I'd like to see is consistently honoring XFO/frame-ancestors, including for cases such as  `data:text/html,<embed src="https://avatars3.githubusercontent.com/u/665379?s=88&v=4">` or `data:text/html,<embed src="https://avatars3.githubusercontent.com/u/665379?s=88&v=4" type="image/jpeg">`. Is that on the table?

mikewest:
That's on the table. I worry that we'll end up breaking PDFs in Chrome that send headers like that, as I recall some bugs in the past around CSP enforcement, but I'm willing to add metrics to see what the breakage looks like, and it seems like a reasonable goal.

The image case shouldn't be a problem at all, given its (apparently) very low usage.

arturjanc:
> 1. This is already not true for SVG, right? We treat it identical to `<iframe>` in most respects, but also resize the element based on the intrinsic dimensions.

This kinda sucks :/ If we go with `navigate` it will mean that, for example, if an application hosts SVG files with thumbnails of user documents and blocks `no-cors` loads, it will still be vulnerable to various XS-Leaks (e.g. an attacker loading an image cross-site to determine if a user has access to a given resource and infer whether the load succeeded from the dimensions of the `<object>`).

Basically, developers would now also need to restrict requests with mode=`navigate` and always check `Sec-Fetch-Dest` in their logic and/or to fall back on `X-Frame-Options` / `frame-ancestors` for images, which is not common practice. It seems suboptimal to make every developer writing Fetch Metadata logic learn about these quirks or have a potentially bypassable security configuration; it's a wart that large sites can live with, but which will probably come back to haunt us one way or another.

mikewest:
I may be cynical about this, Artur, but my expectation is not that every developer writing Fetch Metadata logic will (or should) "learn about these quirks", but that instead they will "copy and paste from the documentation that clever folks like @arturjanc write about how to use the raw data exposed by this feature" (a la https://csp.withgoogle.com/docs/strict-csp.html).

I also don't think that this is really a lot to learn. Especially if we write an implementation considerations section with the rules we might recommend for folks to play around with and put it into the spec so that we can point to it with a degree of clarity that I, at least, would find compelling.

Can we do that instead of circling around this question for another month? :)

arturjanc:
Yes, we can go with `navigate` here, I just wanted to make it abundantly clear that this is a footgun and has a cost that will be borne by all developers who use Fetch Metadata.

annevk:
That attack is possible regardless of whether the initial fetch is `no-cors` or `navigate`. You can also navigate the `object` element from a same-origin document to the target SVG, same problem.

Or is your argument (now, I think it was not before) that navigations of `object` elements should also be `no-cors`?

arturjanc:
Yes -- before you mentioned the fact that `object` will resize based on image intrinsic dimensions I didn't realize that this is a web-observable difference between `iframe` and `object` and thought it should be safe to have `navigate` for a subsequent `object` navigation. Knowing this, I think at least from a security perspective the hybrid model isn't really useful and it would ideally be `no-cors` always.

That said, I think @mikewest is reasonably annoyed at me for unproductively griping here :) Practically, we can do all the things Mike mentioned above to reduce the likelihood of `navigate` for `object` being a problem. It has drawbacks, but it's *an* outcome, it's somewhat principled, and as long as browsers are consistent in what they do here, we'll be able to deal with it. So, uh, let's ship it?

mikewest:
> before you mentioned the fact that object will resize based on image intrinsic dimensions I didn't realize that this is a web-observable difference between iframe and object and thought it should be safe to have navigate for a subsequent object navigation.

I'm fairly certain that `<embed>`/`<object>` only does the weird sizing trick on the initial load. If it creates a browsing context, navigating that browsing context to an image does not resize the element in either Chrome or Firefox.

> That said, I think @mikewest is reasonably annoyed at me for unproductively griping here :) 

I'm not annoyed at your griping. I'm mostly just annoyed that a conversation that I feel like we could have hammered out in an hour if we were all in the same room has taken over a month over GitHub comments. :)

arturjanc:
> I'm fairly certain that `<embed>`/`<object>` only does the weird sizing trick on the initial load. If it creates a browsing context, navigating that browsing context to an image does not resize the element in either Chrome or Firefox.

This is what I had thought as well, but I checked before replying last night :) Sadly: https://arturjanc.com/object-resize.html

mikewest:
> This is what I had thought as well, but I checked before replying last night :) Sadly: https://arturjanc.com/object-resize.html

Confirmed this in Firefox and Chrome via `data:text/html,<embed src="https://arturjanc.com/object-resize-frame.html">`. Weird.

As a counterexample, loading `data:text/html,<embed src="https://example.com/">` in Firefox and Chrome and then putting `window.frames[0].location = "https://mikewest.org/static/20111231/images/mikewest.jpg";` into the console does not cause a resize.

That behavior surprises me. Why do we rerun the SVG bits of https://html.spec.whatwg.org/#the-embed-element-setup-steps upon navigation?

mikewest:
Ok. I've added `embed` and `object`'s recategorization as navigation requests to this PR, and dropped "potential-navigation-or-subresource requests" (see also https://github.com/w3c/ServiceWorker/pull/1486). I'll follow up with the HTML patch's update in a minute.

mikewest:
Ping on this. It's not clear to me what else you'd like me to do with the patch; I think we landed on a reasonable approach, but I'd appreciate you taking another look, @annevk. :)

fetch.bs
```diff
@@ -1561,19 +1567,15 @@ whose <a for=request>destination</a> is "<code>audio</code>", "<code>audioworkle
...
-or "<code>worker</code>".
+whose <a for=request>destination</a> is "<code>document</code>", "<code>embed</code>",
+"<code>frame</code>", "<code>iframe</code>", "<code>object</code>", "<code>report</code>",
+"<code>serviceworker</code>", "<code>sharedworker</code>", or "<code>worker</code>".
 
 <p>A <dfn export>navigation request</dfn> is a <a for=/>request</a> whose
 <a for=request>destination</a> is
-"<code>document</code>".
+"<code>document</code>", "<code>embed</code>", "<code>frame</code>", "<code>iframe</code>",
+or "<code>object</code>".
```
annevk: Does this work for the uses in https://w3c.github.io/ServiceWorker/? Also, do these always have a reserved client and replaces client id? I don't think so right? All of that is rather messy at the moment I suppose and embed/object are somewhat poorly understood, so maybe waggling in the right direction is good enough.

cc @jakearchibald @jungkees

mikewest:
> Also, do we have exhaustive test coverage and are existing tests for destination updated?

I think we have pretty good coverage in https://wpt.fyi/results/fetch/metadata?label=experimental&label=master&aligned. Tests like https://wpt.fyi/results/fetch/metadata/iframe.tentative.https.sub.html?label=experimental&label=master&aligned verify that `dest` is `iframe` and so on.

> It would help if you could write the commit message that goes with this change which summarizes all the interconnected changes and tests and links those.

Will do.

fetch.bs
mikewest: 1. Didn't we decide that `object` and `embed` requests bypassed Service Workers?

2. Assuming it's not a no-op from that perspective, it looks like "navigation request" is used three times in `ServiceWorker`. Perhaps @jakearchibald and @jungkees could weigh in on the latter two, as I'm not sure I understand what they're trying to do:

    * 12.3 of https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm (which is a no-op, as it will never match "the navigation triggering it was initiated with a shift+reload or equivalent")
    * 12.7 of https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm 
    * 19.9 of https://w3c.github.io/ServiceWorker/#terminate-service-worker-algorithm

annevk: 1. Sounds agreeable to me, but would be good to have tests for that, especially where embed/object is being navigated.

jungkees: 1. Yes `object` and `embed` return early in [Handle Fetch](https://w3c.github.io/ServiceWorker/#handle-fetch) step 11.
1. I think all the three cases are OK: 1) is force-reload as @mikewest said 2) is checking if the request is for main resource for navigation preload 3) is to fill up fetchEvent.replacesClientId for navigation requests. Fetch requests went through [process a navigate fetch](https://html.spec.whatwg.org/#process-a-navigate-fetch) should have a replaces client id at this point.

mikewest:
> > It would help if you could write the commit message that goes with this change which summarizes all the interconnected changes and tests and links those.
>
> Will do.

I've rebased and updated the commit message. Does it more or less match what you were looking for?

annevk:
This looks good to me, thanks! Is there a PR to make the tests non-tentative?

annevk:
And is there a test that a service worker doesn't intercept a navigation in embed/object?

mikewest:
> This looks good to me, thanks! Is there a PR to make the tests non-tentative?

I evidently forgot to rebase/land https://chromium-review.googlesource.com/c/chromium/src/+/1995162. I'll do that separately.

> And is there a test that a service worker doesn't intercept a navigation in embed/object?

https://wpt.fyi/results/service-workers/service-worker/embed-and-object-are-not-intercepted.https.html?label=experimental&label=master&aligned looks like it. I don't think it covers the navigation case; I'll see about extending it.

domenic:
Did those test changes ever get finished? It'd be nice to merge this and the corresponding HTML Standard PR (https://github.com/whatwg/html/pull/4976).

wanderview:
Apparently this landed in chromium with WPT tests in:

https://chromium-review.googlesource.com/c/chromium/src/+/1886863

But this spec pr has not merged yet.  What is the status of the spec?  This is confusing some web developers.  Specifically, there are people gating on `destination==='document'` for their service worker navigation preload logic.

annevk:
@wanderview do they cover the scenario discussed in https://github.com/whatwg/fetch/pull/948#discussion_r377651947?

wanderview:
I don't think it covers object/embed and service workers.  I don't see any service worker tests at first glance in that CL.

Anyway, it would be really great to get this resolved since apparently its shipping in chrome 81, but developers are looking at the spec and getting confused.

annevk:
Agreed!

wanderview:
I'm reviewing tests from Mike in this CL:

https://chromium-review.googlesource.com/c/chromium/src/+/2209456

So hopefully will unblock soon.

mikewest:
Your nit LGTM, @annevk. Thank you.

annevk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1643611

annevk:
Thanks @mikewest and @wanderview for getting this done, I think all is merged now and all relevant bugs have been filed and linked.

Issue #601, opened 2017-09-11
Header to opt out of opaque redirect


jakearchibald:
Something like `Access-Control-Allow-Visible-Redirect: *`

This would make a redirect responses visible. If the request was cross-origin, it would still have to pass existing CORS checks, and would be filtered accordingly.

annevk:
Rough duplicate of #75.

I don't think we should call it `Access-Control` as it's not really about CORS with redirects as we hide them same-origin as well (for cross-origin redirect resources you'd still need CORS of course).

Can you explain the use case and do you know to what extent Chrome is interested in implementing? And other browsers?

jakearchibald:
The suggestion came from an internal team who wanted to act conditionally on the type of redirect they received. However, they may not have the server control they need to add this header anyway.

I was just adding it for completeness. Should be considered low priority unless we get demand.

rajkjag:
In our specific case, the navigation preload request, is being DOS blocked and redirected to a CAPTCHA page. However the preload response resolves to a response with  type "opaqueredirect" and no access to the redirect url (sent back in the Location: header).The server which performs the DOS block is not under our control and happens upstream.

While we can work around this specific case, it might be easier if ServiceWorker code on the client, has access to the redirect URL for error handling and recovery.

annevk:
@rajkjag that would violate https://fetch.spec.whatwg.org/#atomic-http-redirect-handling.

rajkjag:
The proposal on the table was to have server specify via a response header the list of headers to expose.

annevk:
Okay, I got confused since you said the server wasn't under your control and it would be easier if service workers had access to the redirect URL.

rajkjag:
I took a closer look at our setup; While the server that generates the captcha URL is a different server, it is being operated in a slave mode and we seem to have control on the final response and headers. Apologies on the confusion.
1. Primary server gets requests.
2. Contact DOS slave server, gets back indication on whether to block or redirect.
3. Returns a redirect response with the redirect URL.

If we cannot implement a header and have the client respect it and expose the redirect URL, then our fallback option would be to embed the redirect URL inside the body of the redirect response (If and when the server determines it is safe to do so).

ibnesayeed:
Being able to expose status code, location, and various other headers will be great when redirects are handled manually in a fetch. The name of this redirect mode is confusing and misleading. However, if we can someone opt-out of the `opaqueredirect` then the name can be justified.

[...]

PR #955, opened 2019-10-29
Add TAO check

npm1:
This adds a Timing Allow Origin (TAO) check, trying to be as close to a CORS check as possible. The check is added very close to where the CORS check is, in the HTTP fetch algorithm. It is added so that service worker responses are also checked. When the check fails, it is stored in the request (and once it fails, future redirects cannot cause it to pass). This check is similar to https://w3c.github.io/resource-timing/#dfn-timing-allow-check but the tainted bool is replaced with the tainted origin flag to align with CORS.

***
<a href="https://whatpr.org/fetch/955.html" title="Last updated on Dec 3, 2019, 6:27 PM UTC (648f46e)">Preview</a> | <a href="https://whatpr.org/fetch/955/ae6435f...648f46e.html" title="Last updated on Dec 3, 2019, 6:27 PM UTC (648f46e)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1549,7 +1549,7 @@ Unless stated otherwise, it is unset.
 Unless stated otherwise, it is unset.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=timing-allow-failed-flag>timing allow failed flag</dfn>. Unless stated
+<dfn export for=request id=timing-allow-failed>timing allow failed flag</dfn>. Unless stated
 otherwise, it is unset.
 
 <p class="note no-backref">A <a for=/>request</a>'s <a for=request>tainted origin flag</a>,
@@ -1810,8 +1810,8 @@ being provided to an API that didn't make a range request. See the flag's usage 
 description of the attack.
 
 <p>A <a for=/>response</a> has an associated
-<dfn for=response id=concept-response-timing-allow-passed-flag>timing allow passed flag</dfn>, which
-is initially unset.
+<dfn for=response id=concept-response-timing-allow-passed>timing allow passed flag</dfn>, which is
+initially unset.
 
 <p class=note>This is used so that the caller to a fetch can determine if sensitive timing data is
 allowed on the resource fetched by looking at the flag of the response returned. Because the flag on
@@ -4979,7 +4979,7 @@ run these steps:
 <p>A <dfn>cache entry</dfn> consists of:
 
 <ul class=brief>
- <li><dfn id=concept-cache-origin for="cache entry">serialized origin</dfn> (a
+ <li><dfn id=concept-cache-origin for="cache entry">byte-serialized origin</dfn> (a
  <a for=/>byte sequence</a>)
  <li><dfn id=concept-cache-url for="cache entry">URL</dfn> (a <a for=/>URL</a>)
  <li><dfn id=concept-cache-max-age for="cache entry">max-age</dfn> (a number of seconds)
@@ -5002,7 +5002,7 @@ be removed before that moment arrives.
   <p>Let <var>entry</var> be a <a>cache entry</a>, initialized as follows:
 
   <dl>
-   <dt><a for="cache entry">serialized origin</a>
+   <dt><a for="cache entry">byte-serialized origin</a>
    <dd><p>The result of <a>byte-serializing a request origin</a> with <var>request</var>
 
    <dt><a for="cache entry">URL</a>
@@ -5027,15 +5027,15 @@ be removed before that moment arrives.
 
 <p>To <dfn id=concept-cache-clear>clear cache entries</dfn>, given a <var>request</var>,
 <a for=list>remove</a> any <a>cache entries</a> in the user agent's <a>CORS-preflight cache</a>
-whose <a for="cache entry">serialized origin</a> is the result of
+whose <a for="cache entry">byte-serialized origin</a> is the result of
 <a>byte-serializing a request origin</a> with <var>request</var> and whose
 <a for="cache entry">URL</a> is <var>request</var>'s <a for=request>current URL</a>.
 
 <p>There is a <dfn id=concept-cache-match>cache entry match</dfn> for a <a>cache entry</a>
 <var>entry</var> with <var>request</var> if <var>entry</var>'s
-<a for="cache entry">serialized origin</a> is the result of <a>byte-serializing a request origin</a>
-with <var>request</var>, <var>entry</var>'s <a for="cache entry">URL</a> is <var>request</var>'s
-<a for=request>current URL</a>, and one of
+<a for="cache entry">byte-serialized origin</a> is the result of
+<a>byte-serializing a request origin</a> with <var>request</var>, <var>entry</var>'s
+<a for="cache entry">URL</a> is <var>request</var>'s <a for=request>current URL</a>, and one of
 
 <ul class=brief>
  <li><var>entry</var>'s <a for="cache entry">credentials</a> is true
--- a/fetch.bs
+++ b/fetch.bs
@@ -1590,11 +1590,15 @@ run these steps:
 
 <ol>
  <li><p>If <var>request</var>'s <a for=request>tainted origin flag</a> is set, then return
- `<code>null</code>`.
+ "<code>null</code>".
 
  <li><p>Return <var>request</var>'s <a for=request>origin</a>,
- <a lt="ASCII serialization of an origin">serialized</a> and <a>isomorphic encoded</a>.
+ <a lt="ASCII serialization of an origin">serialized</a>.
 </ol>
+
+<p><dfn>Byte-serializing a request origin</dfn>, given a <a for=/>request</a> <var>request</var>,
+is to return the result of <a>serializing a request origin</a> with <var>request</var>,
+<a>isomorphic encoded</a>.
 
 <hr>
 
@@ -2430,8 +2434,8 @@ origin                           = <a for=url>scheme</a> "://" <a for=url>host</
 given a <a for=/>request</a> <var>request</var> with an optional <i>CORS flag</i>, run these steps:
 
 <ol>
- <li><p>Let <var>serializedOrigin</var> be the result of <a>serializing a request origin</a> with
- <var>request</var>.
+ <li><p>Let <var>serializedOrigin</var> be the result of <a>byte-serializing a request origin</a>
+ with <var>request</var>.
 
  <li><p>If the <i>CORS flag</i> is set or <var>request</var>'s <a for=request>mode</a> is
  "<code>websocket</code>", then <a for="header list">append</a>
@@ -3542,7 +3546,8 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
  <!-- If you are ever tempted to move this around, carefully consider responses from about URLs,
       blob URLs, service workers, HTTP cache, HTTP network, etc. -->
 
- <li><p>If <var>request</var>'s <a for=request>timing allow failed flag</a> is unset, then set <var>internalResponse</var>'s <a for=response>timing allow passed flag</a>.
+ <li><p>If <var>request</var>'s <a for=request>timing allow failed flag</a> is unset, then set
+ <var>internalResponse</var>'s <a for=response>timing allow passed flag</a>.
 
  <li><p><a href=https://w3c.github.io/webappsec-csp/#set-response-csp-list>Set <var>internalResponse</var>'s CSP list</a>.
  [[!CSP]]
@@ -4998,7 +5003,7 @@ be removed before that moment arrives.
 
   <dl>
    <dt><a for="cache entry">serialized origin</a>
-   <dd><p>The result of <a>serializing a request origin</a> with <var>request</var>
+   <dd><p>The result of <a>byte-serializing a request origin</a> with <var>request</var>
 
    <dt><a for="cache entry">URL</a>
    <dd><p><var>request</var>'s <a for=request>current URL</a>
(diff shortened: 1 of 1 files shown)


npm1:
@annevk @yoavweiss @yutakahirano what do you think? Can't add reviewers so just tagging you in comment.

fetch.bs
```diff
@@ -5075,6 +5083,34 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+<p>To perform a <dfn id=concept-tao-check>TAO check</dfn> for a <var>request</var> and
+<var>response</var>, run these steps:
+
+<ol>
+ <li><p>If <var>response</var>'s <a for=request>timing allow failed flag</a> is set, then return
+ failure.
+
+ <li><p>If <var>request</var>'s <a for=request>tainted origin flag</a> is unset and
+ <var>response</var>'s <a for=response>location URL</a>'s <a for=url>origin</a> is
+ <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then return success.
```
annevk: Instead of this, I think we should skip this check when request's response tainting is "basic". Otherwise, when you have A -> B -> A, we won't do a TAO check for the final A, right? That would be problematic.

fetch.bs
```diff
@@ -3896,6 +3901,9 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
     or <a for=/>responses</a> from a service worker for that matter, it is applied here.
   </ol>
 
+ <li><p>If the <a>TAO check</a> for <var>request</var> and <var>response</var> returns failure, then
+  set <var>request</var>'s <a for=request>timing allow failed flag</a>.
```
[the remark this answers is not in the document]
npm1: Ok, I've moved it. I guess it's just not clear to me how the service worker handles the checks.

fetch.bs
```diff
@@ -3894,6 +3908,10 @@ optional <i>CORS flag</i> and <i>CORS-preflight flag</i>, run these steps:
     <p class="note no-backref">As the <a>CORS check</a> is not to be applied to
     <a for=/>responses</a> whose <a for=response>status</a> is <code>304</code> or <code>407</code>,
     or <a for=/>responses</a> from a service worker for that matter, it is applied here.
+
+   <li><p>If the <a>TAO check</a> for <var>request</var> and <var>response</var> returns failure,
+   then set <var>request</var>'s <a for=request>timing allow failed flag</a> and set
+   <var>response</var>'s <a for=response>timing allow failed flag</a>.
```
annevk: I think we should set this at the same time we set URLs on a response, in main fetch either before or after

> If internalResponse’s URL list is empty, then set it to a clone of request’s URL list.

fetch.bs
annevk: It's possible, if the header says `*` (which arguably is also problematic for credentialed requests, but so be it) or `null` (which is what the `Origin` request header will contain).

fetch.bs
npm1: Done

fetch.bs
npm1: Hmm ok, I'll ask around to see if someone can explain to me the tainted origin flag within CORS. I'd like to better understand what it solves so that I can see how it could also solve problems with TAO. In the meantime, I'm ok including the 'serializing a request origin' since hopefully this would just be an edge case in the wild, thus not breaking a lot of performance monitoring.

npm1: @yutakahirano explained reasoning for the tainted origin flag in CORS. While it's hard to see how to apply it to TAO, I'm not opposed to aligning here.

annevk: Okay, so this still needs changes, right?

fetch.bs
```diff
@@ -1800,6 +1805,15 @@ initially unset.
 being provided to an API that didn't make a range request. See the flag's usage for a detailed
 description of the attack.
 
+<p>A <a for=/>response</a> has an associated
+<dfn for=response id=concept-response-timing-allow-failed-flag>timing allow failed flag</dfn>, which
```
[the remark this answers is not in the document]
npm1: It makes sense for the consumer but I saw that all existing flags are initially unset. Request's flag must start in 'success' so that once it flips to failure it never goes back. In order for it to start in success and unset, it must be the failure flag.

And I think it would be odd to have 'timing allow failed flag' in requests and 'timing allow succeeded flag' in responses. There's already 'trailer failed flag' in response, so overall for consistency it seems preferable to have the 'failed' flags here, what do you think?

fetch.bs
npm1: I've added the check so that the same-origin check is only used when response tainting is "basic". I'm also using the 'serializing a request origin' to obtain the origin, so that the tainted origin flag affects the outcome. What's missing?

annevk: Well, you're checking location URL and that really only works if the response is a redirect, which isn't checked in this step and I also don't think matters. If from origin A you fetched a redirect from origin A it does not matter whether that redirect goes to A or B. You don't need a TAO check for it, right? Or are we requiring TAO for all redirect responses, even same-origin? (That might make some sense too, given how we don't normally expose redirects.)

fetch.bs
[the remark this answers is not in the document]
annevk: I'm somewhat okay with it, but I still think a positive flag would be better for the response side, starting out unset. That would also mean the default is safe. (The trailer stuff should maybe change a bit, but also might go away as nobody implements it.)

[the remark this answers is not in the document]
npm1: Ok, your call! I've changed response's flag to be positive.

fetch.bs
npm1: Ah ok, I think I see what you mean. If I understand correctly, up to this point it suffices to check the response being basic.

annevk:
@yoavweiss can you review this as well? And in particular confirm that same-origin redirects do not require TAO headers even if they're not otherwise exposed for security reasons (well, with some caveats, it's mostly not exposed when there's multiple).

yoavweiss:
> @yoavweiss can you review this as well? 
I'll review shortly

> And in particular confirm that same-origin redirects do not require TAO headers even if they're not otherwise exposed for security reasons (well, with some caveats, it's mostly not exposed when there's multiple).

Currently we don't expose specific timing of such redirects, although we are planning to at some point. Can you elaborate on the security reasons those are not exposed in CORS?

yoavweiss:
> @yoavweiss can you review this as well? 
I'll review shortly

> And in particular confirm that same-origin redirects do not require TAO headers even if they're not otherwise exposed for security reasons (well, with some caveats, it's mostly not exposed when there's multiple).

Currently we don't expose specific timing of such redirects, although we are planning to at some point. Can you elaborate on the security reasons those are not exposed in CORS?

annevk:
CORS doesn't have anything to do with same-origin redirects. The reason we don't expose them is because they can contain secrets: https://fetch.spec.whatwg.org/#atomic-http-redirect-handling.

yoavweiss:
> CORS doesn't have anything to do with same-origin redirects. The reason we don't expose them is because they can contain secrets: https://fetch.spec.whatwg.org/#atomic-http-redirect-handling.

The nature of the attack is not 100% clear to me in a same-origin context without any cross-origin redirects. Is the concern that it'd enable an origin to guess its own http-only cookies?

fetch.bs
```diff
@@ -1800,6 +1805,15 @@ initially unset.
 being provided to an API that didn't make a range request. See the flag's usage for a detailed
 description of the attack.
 
+<p>A <a for=/>response</a> has an associated
+<dfn for=response id=concept-response-timing-allow-failed-flag>timing allow check flag</dfn>, which
```
yoavweiss: The id and the name don't match. It's "failed" in one and "check" in the other

fetch.bs
```diff
@@ -1800,6 +1805,15 @@ initially unset.
...
+is initially unset.
+
+<p class=note>This is used so that the caller to a fetch can determine if sensitive timing data is
```
yoavweiss: Interesting! Did we have requests to expose TAO availability to JS? Or is it just a "let's expose it because we can and it's cheap"?

fetch.bs
yoavweiss: Would be better to make the name more explicit. IIUC, this flag is set when TAO check passed successfully. So maybe "timing allow passed flag"?

yoavweiss:
Other than the same-origin redirects discussion, LGTM % nits.

fetch.bs
yutakahirano: IIUC this is not exposed to JS - it's a concept field.

yoavweiss: oh, sorry. Misunderstood the note (jetlag...)

arturjanc:
Re: timing same-origin redirects, this is already possible so I don't think there's any danger if TAO isn't required to expose timing information in this case.

For example, you can time the redirect via Fetch's `redirect: 'manual'` ([example](https://arturjanc.com/time-same-origin-redirect.html)) or by CSP's `SecurityPolicyViolationEvent` if you issue a cross-origin redirect that violates your policy.

Even if this wasn't possible, I don't think the risk of timing is comparable to the risk of directly exposing the `Location` value. So, basically, exposing timing information for all same-origin requests doesn't seem particularly scary to me (or, scary at all).

annevk:
@yoavweiss it was hidden for any secrets the `Location` header might contain (can contain login tokens for (cross-origin) resources). And I guess any header (or body) at this point.

@arturjanc so it wasn't clear to me if the timing data is broken down and currently you cannot detect a second redirect to my knowledge (although maybe some other API that isn't layered appropriately broke that, forgot). None of this seems particularly scary, indeed, but I'd like to be rather clear when we reveal more networking data than previously as sometimes the combination of a number of those sidechannels leads to issues.

yoavweiss:
> currently you cannot detect a second redirect to my knowledge

That is correct. Currently you only get timing data of the very-first redirect's start and the very-lst redirect's end.

> None of this seems particularly scary, indeed, but I'd like to be rather clear when we reveal more networking data than previously as sometimes the combination of a number of those sidechannels leads to issues.

Might be worthwhile to add a note to Resource Timing indicating that same-origin redirects may contain sensitive information and we want to be careful when exposing more details about them. We already have requests to expose more information about redirects, but we could probably gate that behind TAO or some other opt-in (even for same origin) once we actually plan to do that.

annevk:
It would be good to block any new features on a better layering. Perhaps #601 can be used for such functionality.

yoavweiss:
> It would be good to block any new features on a better layering. Perhaps #601 can be used for such functionality.

That's indeed the mode that Resource Timing has been operating under. New functionality is blocked until integrating with Fetch happens.

fetch.bs
npm1: Done

fetch.bs
```diff
@@ -5075,6 +5094,32 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+ <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>basic</code>", then return success.
+
+ <li><p>Let <var>values</var> be the result of
+ <a for="header list">getting, decoding, and splitting</a>
+ `<code>Timing-Allow-Origin</code></a>` from <var>response</var>'s <a for=response>header list</a>.
+
+ <li><p>If <var>values</var> contains `<code>*</code>`, then return success.
+
+ <li><p>Let <var>serializedOrigin</var> be the result of <a>serializing a request origin</a> with
+ <var>request</var>.
```
annevk: Is there a test for this scenario and browser bugs given that it sounded liked this was not yet how browsers behaved?

How are we going to integrate this with the resource timing specification?

npm1: > Is there a test for this scenario and browser bugs given that it sounded liked this was not yet how browsers behaved?

I'll modify https://github.com/web-platform-tests/wpt/blob/master/resource-timing/crossorigin-sandwich-TAO.sub.html because it should no longer pass the TAO check with this change. And will also add a crossorigin sandwich which uses '*' in the header to ensure that it still passes. I imagine browser bugs should be filed once the spec change has landed?
 
> How are we going to integrate this with the resource timing specification?

We'll replace the timing allow check algorithm with a query into the response's flag.

annevk: They can be filed now (having a brief heads up window is nice), and the test change can be created as well. Ideally we also have a PR ready for Resource Timing. That way we can land it all together and know that it integrates well.

npm1:
Ok, I'll work on adding tests. In the meantime, as an update we noticed that there's an unresolved problem with how to deal with iframes https://github.com/whatwg/fetch/issues/968

fetch.bs
npm1: https://github.com/web-platform-tests/wpt/blob/master/resource-timing/crossorigin-sandwich-TAO.sub.html
https://bugs.chromium.org/p/chromium/issues/detail?id=1022816
https://bugzilla.mozilla.org/show_bug.cgi?id=1597808
https://bugs.webkit.org/show_bug.cgi?id=204375

npm1: Also https://github.com/w3c/resource-timing/pull/218 (yes, still hand-wavy because what is a resource, but need to fix the use of 'resource' throughout in order to solve that problem).

npm1:
Ping @annevk do we need anything else before merging?

annevk:
I pushed a commit to address some remaining issues I spotted. As I mentioned on IRC to @yoavweiss eventually we should also capture the timings here, but this is a good incremental step.

Could you please review my commit?

fetch.bs
```diff
@@ -4998,7 +5003,7 @@ be removed before that moment arrives.
 
   <dl>
    <dt><a for="cache entry">serialized origin</a>
```
npm1: This should be named byte-serialized origin now.

fetch.bs
```diff
@@ -5059,7 +5083,7 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
  <li><p>If <var>request</var>'s <a for=request>credentials mode</a> is not "<code>include</code>"
  and <var>origin</var> is `<code>*</code>`, then return success.
```
npm1:
```suggestion
  and <var>origin</var> is "<code>*</code>", then return success.
```

annevk: In this case it's bytes, right? The value isn't decoded here.

npm1: Ah yes, you're right.

npm1:
Looks good now, thanks

annevk:
Implementation bugs and tests are covered by https://github.com/whatwg/fetch/pull/955#discussion_r348196456.

PR #959, opened 2019-11-07
Use ReadableStreamPipeTo in Request constructor

### LINKED ISSUE #463: Specify body transfer in Request constructor more precisely

yutakahirano:
We have used an imprecise sentence for creating a ReadableStream "from which
one can read the exactly same data as one could read" from the original
ReadableStream, because transform streams and piping were not defined. Now they
are available, so let's use them.

Fixes #463.

***
<a href="https://whatpr.org/fetch/959.html" title="Last updated on Dec 13, 2019, 9:03 AM UTC (81c5b23)">Preview</a> | <a href="https://whatpr.org/fetch/959/7db8ac5...81c5b23.html" title="Last updated on Dec 13, 2019, 9:03 AM UTC (81c5b23)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6224,17 +6224,19 @@ constructor must run these steps:
   <p>If <var>inputBody</var> is <var>body</var> and <var>inputBody</var> is non-null, then:
 
   <ol>
-   <li>
-    <p>Let <var>rs</var> bs a {{ReadableStream}} object from which one can read the exactly
-    same data as one could read from <var>inputBody</var>'s <a for=body>stream</a>.
+   <li><p>Let <var>ws</var> and <var>rs</var> be the <a>writable side</a> and <a>readable side</a>
+   of an <a>identity transform stream</a>, respectively.</li>
 
-    <p class=XXX>This will be specified more precisely once
-    <a href=https://streams.spec.whatwg.org/#ts-model>transform stream</a> and
-    <a href=https://streams.spec.whatwg.org/#pipe-chains>piping</a> are precisely defined.
-    See <a href=https://github.com/whatwg/fetch/issues/463>issue #463</a>.
+   <li>
+    <p>Let <var>promise</var> be the result of calling
+    <a abstract-op>ReadableStreamPipeTo</a>(<var>inputBody</var>, <var>ws</var></var>, false,
+    false, false, undefined).
 
     <p class="note no-backref">This makes <var>inputBody</var>'s <a for=body>stream</a>
     <a for=ReadableStream>locked</a> and <a for=ReadableStream>disturbed</a> immediately.
+   </li>
+
+   <li><p>Set <var>promise</var>.\[[PromiseIsHandled]]</var> to true.
 
    <li><p>Set <var>body</var> to a new <a for=/>body</a> whose <a for=body>stream</a> is
    <var>rs</var>, whose <a for=body>source</a> is <var>inputBody</var>'s <a for=body>source</a>, and


yutakahirano:
@annevk @ricea can you PTAL?

fetch.bs
```diff
@@ -6224,17 +6224,19 @@ constructor must run these steps:
...
 
   <ol>
-   <li>
-    <p>Let <var>rs</var> bs a {{ReadableStream}} object from which one can read the exactly
-    same data as one could read from <var>inputBody</var>'s <a for=body>stream</a>.
+   <li><p>Let <var>ws</var> and <var>rs</var> be the
+   <a href="https://streams.spec.whatwg.org/#writable-side">writable side</a> and
+   <a href="https://streams.spec.whatwg.org/#readable-side">readable side</a> of an
+   <a href="https://streams.spec.whatwg.org/#identity-transform-stream">identity transform
+   stream</a> respectively.</li>
```
ricea: I think it would be better to have a comma before "respectively", but not 100% confident of this.

fetch.bs
```diff
@@ -6224,17 +6224,19 @@ constructor must run these steps:
...
 
-    <p class=XXX>This will be specified more precisely once
-    <a href=https://streams.spec.whatwg.org/#ts-model>transform stream</a> and
-    <a href=https://streams.spec.whatwg.org/#pipe-chains>piping</a> are precisely defined.
-    See <a href=https://github.com/whatwg/fetch/issues/463>issue #463</a>.
+   <li>
+    <p>Call <a abstract-op>ReadableStreamPipeTo</a>(<var>inputBody</var>,
```
ricea: I think we need to set [[PromiseIsHandled]] on the promise returned from ReadableStreamPipeTo, similar to [how pipeThrough does it](https://streams.spec.whatwg.org/#rs-pipe-through). This will stop uncaught promise rejections being visible when a stream is errored.

fetch.bs
```diff
@@ -6228,16 +6228,19 @@ constructor must run these steps:
...
-    <var>ws</var></var>, false, false, false, undefined).
+    <p>Let <var>promsie</var> be the result of calling
+    <a abstract-op>ReadableStreamPipeTo</a>(<var>inputBody</var>, <var>ws</var></var>, false,
+    false, false, undefined).
 
     <p class="note no-backref">This makes <var>inputBody</var>'s <a for=body>stream</a>
     <a for=ReadableStream>locked</a> and <a for=ReadableStream>disturbed</a> immediately.
    </li>
 
+   <li><p>set <var>promise</var>.\[[PromiseIsHandled]]</var> to true.
```
ricea: Capital S for "set".

fetch.bs
```diff
@@ -6228,16 +6228,19 @@ constructor must run these steps:
    <a href="https://streams.spec.whatwg.org/#writable-side">writable side</a> and
    <a href="https://streams.spec.whatwg.org/#readable-side">readable side</a> of an
    <a href="https://streams.spec.whatwg.org/#identity-transform-stream">identity transform
-   stream</a> respectively.</li>
+   stream</a>, respectively.</li>
 
    <li>
-    <p>Call <a abstract-op>ReadableStreamPipeTo</a>(<var>inputBody</var>,
-    <var>ws</var></var>, false, false, false, undefined).
+    <p>Let <var>promsie</var> be the result of calling
```
ricea: s/promsie/promise/

fetch.bs
yutakahirano: Fixed.

fetch.bs
yutakahirano: Fixed.

fetch.bs
yutakahirano: Fixed.

fetch.bs
yutakahirano: Fixed.

fetch.bs
```diff
@@ -6224,17 +6224,22 @@ constructor must run these steps:
...
+   <li>
+    <p>Let <var>promise</var> be the result of calling
+    <a abstract-op>ReadableStreamPipeTo</a>(<var>inputBody</var>, <var>ws</var></var>, false,
+    false, false, undefined).
 
     <p class="note no-backref">This makes <var>inputBody</var>'s <a for=body>stream</a>
     <a for=ReadableStream>locked</a> and <a for=ReadableStream>disturbed</a> immediately.
+   </li>
+
+   <li><p>Set <var>promise</var>.\[[PromiseIsHandled]]</var> to true.
```
annevk: Is there a way to avoid this? Manipulating internal state of promises seems problematic and especially doing so in a specification that isn't likely to be notified if ECMAScript is refactored in some fashion.

yutakahirano: I'm not aware of one other than calling `then` (or `catch`) explicitly. @domenic, do you have any recommendation?

domenic: In my opinion this is an exceptional internal slot that it is OK for other standards to manipulate. (As an alternative to uselessly calling PerformPromiseThen() with a no-op catch handler.)

fetch.bs
```diff
@@ -6224,17 +6224,22 @@ constructor must run these steps:
   <p>If <var>inputBody</var> is <var>body</var> and <var>inputBody</var> is non-null, then:
 
   <ol>
-   <li>
-    <p>Let <var>rs</var> bs a {{ReadableStream}} object from which one can read the exactly
-    same data as one could read from <var>inputBody</var>'s <a for=body>stream</a>.
+   <li><p>Let <var>ws</var> and <var>rs</var> be the
+   <a href="https://streams.spec.whatwg.org/#writable-side">writable side</a> and
+   <a href="https://streams.spec.whatwg.org/#readable-side">readable side</a> of an
+   <a href="https://streams.spec.whatwg.org/#identity-transform-stream">identity transform
```
domenic: These should be linked with just `<a>writable side</a>` or maybe `<a for="transform stream">writable side</a>`. If that is not possible now then they should be exported from streams so that it becomes possible.

fetch.bs
annevk: What do you think of moving this into IDL somehow with other specifications having "**Handle** _promise_." or some such?

domenic: That seems like a reasonable future cleanup, yeah. As-is right now we're still in the world where streams operates on ES promises directly.

annevk:
Remaining:

* Domenic's cross-reference issue.
* Follow-up issues against IDL and specifications manipulating Promise internals.
* Tests? Not sure if this is observable or not.

fetch.bs

yutakahirano:
> Domenic's cross-reference issue.

Done.

> Follow-up issues against IDL and specifications manipulating Promise internals.

Filed webidl/issues/829.

> Tests? Not sure if this is observable or not.

I think this is needless. This is a kind of cleanup change.

PR #960, opened 2019-11-08
Remove the CORS flag

annevk:
As far as I can tell we use "response tainting" for the same thing.

This also fixes an issue whereby we checked origin tainting rather than response tainting in main fetch which results in a minor bug when it comes to opaquing A -> B -> A chains.

***
<a href="https://whatpr.org/fetch/960.html" title="Last updated on Nov 11, 2019, 8:50 AM UTC (70ffc79)">Preview</a> | <a href="https://whatpr.org/fetch/960/7db8ac5...70ffc79.html" title="Last updated on Nov 11, 2019, 8:50 AM UTC (70ffc79)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2387,13 +2387,12 @@ if the result of calling
 request <a for=/>header</a> indicates where a
 <a for=/>fetch</a> originates from.
 
-<p class="note no-backref">The `<a http-header><code>Origin</code></a>` header is a version
-of the `<code>Referer</code>` [sic] header that does not reveal a
-<a for=url>path</a>. It is used for
-all <a lt="HTTP fetch">HTTP fetches</a> whose <i>CORS flag</i> is
-set as well as those where <a for=/>request</a>'s
-<a for=request>method</a> is neither `<code>GET</code>` nor `<code>HEAD</code>`. Due to
-compatibility constraints it is not included in all
+<p class="note no-backref">The `<a http-header><code>Origin</code></a>` header is a version of the
+`<code>Referer</code>` [sic] header that does not reveal a <a for=url>path</a>. It is used for all
+<a lt="HTTP fetch">HTTP fetches</a> whose <a for=/>request</a>'s
+<a for=request>response tainting</a> is "<code>cors</code>" as well as those where
+<a for=/>request</a>'s <a for=request>method</a> is neither `<code>GET</code>` nor
+`<code>HEAD</code>`. Due to compatibility constraints it is not included in all
 <a lt=fetch for=/>fetches</a>.
 <!-- Ian Hickson told me Adam Barth researched that -->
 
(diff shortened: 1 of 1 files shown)


annevk:
@JuniorHsu interested in reviewing as well?

fetch.bs
```diff
@@ -3375,8 +3374,8 @@ with a <i>CORS flag</i> and <i>recursive flag</i>, run these steps:
 
   <dl class=switch>
    <dt><var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> is
-   <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, <var>request</var>'s
-   <a for=request>tainted origin flag</a> is unset, and the <i>CORS flag</i> is unset
+   <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, and <var>request</var>'s
+   <a for=request>response tainting</a> is "<code>basic</code>"
```
domfarolino: It's likely I'm missing something, but I don't exactly see how this condition changes, could you please explain:

> This also fixes an issue whereby we checked origin tainting rather than response tainting in main fetch which results in a minor bug when it comes to opaquing A -> B -> A chains.

From what I can tell, assuming the current URL's origin is same-origin with request's origin, then:
 - Whenever _tainted origin flag_ is unset, _response tainting_ is always "basic" here, and
 - Whenever _tainted origin flag_ is set, _response tainting_ is not "basic"

annevk: I think you're right. A -> B -> A does set the tainted origin flag. I guess that means this whole change is editorial. I still think it's for the better though.

domfarolino: I agree, I think favoring response tainting is clearer. Thanks.

yutakahirano:
The change looks good, but I have a few questions for the future direction.

At this moment we have two options for "bookkeeping".

1. As a parameter for fetch algorithms (e.g., CORS preflight flag)
1. As a field of the concept request (e.g., request's tainted origin flag)

Do we have a general preference to one over the other? What do you think about separating such bookkeeping fields from the concept request? To be honest I'm not very comfortable with these fields as I believe they should be part of loader (or fetcher), not the request.

fetch.bs
```diff
@@ -2387,13 +2387,12 @@ if the result of calling
...
-of the `<code>Referer</code>` [sic] header that does not reveal a
-<a for=url>path</a>. It is used for
-all <a lt="HTTP fetch">HTTP fetches</a> whose <i>CORS flag</i> is
-set as well as those where <a for=/>request</a>'s
-<a for=request>method</a> is neither `<code>GET</code>` nor `<code>HEAD</code>`. Due to
-compatibility constraints it is not included in all
+<p class="note no-backref">The `<a http-header><code>Origin</code></a>` header is a version of the
+`<code>Referer</code>` [sic] header that does not reveal a <a for=url>path</a>. It is used for all
+<a lt="HTTP fetch">HTTP fetches</a> whose <a for=/>request</a>'s
+<a for=request>response tainting</a> is "<code>cors</code>" as well as those where
```
sideshowbarker:
```suggestion
<a for=request>response tainting</a> is "<code>cors</code>", as well as those where
```

annevk:
@yutakahirano I think putting this state on some other "object" makes a lot of sense. I suspect this is still a reasonable incremental step towards that though. If you have a design in mind, could you describe that in a new issue? And if you think solving that blocks this PR, please say so.

yutakahirano:
> @yutakahirano I think putting this state on some other "object" makes a lot of sense. I suspect this is still a reasonable incremental step towards that though. If you have a design in mind, could you describe that in a new issue? And if you think solving that blocks this PR, please say so.

Thank you. I don't want to block this PR, and I agree that this is a good change.

PR #965, opened 2019-11-11
Editorial: use origin-based "same site" definition

domenic:
Follows https://github.com/whatwg/html/pull/5076.

/cc @mikewest 

I'm unsure whether this should be "same site" or "schemelessly same site". I guess for parity with the previous it should be "schemelessly same site"? But that seems like a pretty bad mismatch with the web-dev-visible header name... what's going on?

***
<a href="https://whatpr.org/fetch/965.html" title="Last updated on Nov 19, 2019, 5:01 PM UTC (1d6002f)">Preview</a> | <a href="https://whatpr.org/fetch/965/65138f3...1d6002f.html" title="Last updated on Nov 19, 2019, 5:01 PM UTC (1d6002f)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3107,9 +3107,22 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" ; case-sensit
 
  <li><p>If <var>policy</var> is `<code>same-origin</code>`, then return <b>blocked</b>.
 
- <li><p>If <var>request</var>'s <a for=request>origin</a> is <a>same site</a> with
- <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, then return
- <b>allowed</b>.
+ <li>
+  <p>If the following are true
+
+  <ul class=brief>
+   <li><var>request</var>'s <a for=request>origin</a> is <a>same site</a> with
+   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>
+   <li><var>request</var>'s <a for=request>origin</a>'s <a for=url>scheme</a> is
+   "<code>https</code>" or <var>response</var>'s <a for=response>HTTPS state</a> is
+   "<code>none</code>"
+  </ul>
+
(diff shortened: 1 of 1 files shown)


mikewest:
I think the goal will be to change our collective behavior such that we take the scheme into account for mechanisms that work through today's "same site" concept. So yes, today it's a mismatch. Tomorrow, it ideally won't be.

annevk:
The second bullet point which you did not touch is important here and follows from a suggestion by @youennf in https://github.com/whatwg/fetch/issues/687#issuecomment-395551229, also supported by @arturjanc and @johnwilander at the time.

Soon those mixed content images will no longer be requested over HTTP though so I think we can indeed simplify this, but that means two things for this PR:

1. The second bullet point needs to go and perhaps that will let us remove the list altogether.
2. It's no longer editorial.

domenic:
OK, I think I changed this as people were suggesting. However I'm a bit unsure whether or how we should tie this up with the updates for https://github.com/whatwg/url/issues/448. Any guidance would be appreciated.

To be clear, I was mostly hoping to make editorial changes with this effort, and would rather not sign up for e.g. updating web platform tests for CORP, if we're making normative changes to its behavior.

annevk:
In that case you should replace same site with schemelessly same site in the existing text and we can file a follow-up to consider making this stricter (which would then end up with the simplified text).

domenic:
Ok, and keep the second bullet point? In that case I think it's just a revert of my second commit, right?

annevk:
Well, and s/same site/schemelessly same site/ as that was what same site used to do.

domenic:
OK cool, this should be ready to merge when the Bikeshed link database updates. Commit message:

```
Editorial: use origin-based "schemelessly same site"

Follows https://github.com/whatwg/html/pull/5076. See
https://github.com/whatwg/fetch/issues/969 for a potential normative
follow-up to change this to use "same site"
```

Issue #944, opened 2019-10-01
Drop developer-controlled Authorization header on cross-origin redirects


annevk:
User-agent-controlled credentials are only included for matching requests, but developer-controlled credentials will be copied from request to request.

There's a proposal to scope a developer-controlled `Authorization` header to the origin of the initial request. (If you use other headers to carry credentials you are out of luck.)

This might be reasonably compatible as `Authorization` is a header that requires a preflight (and does not allow wildcards) and redirects for preflights were not followed until recently.

What's needed to move this forward:

 - [ ] Implementers need to be interested.
 - [ ] Tests need to be written to ensure it's dropped at the appropriate time (and other headers are not).
 - [ ] The specification needs to be updated to account for this. Perhaps by reusing https://fetch.spec.whatwg.org/#cors-non-wildcard-request-header-name somehow.

cc @whatwg/security

ptoomey3:
> (If you use other headers to carry credentials you are out of luck.)

I'm sure there is a good reason for this, so apologies if this has been discussed to death; is there a thread that discusses the motivation of sending custom-set headers by default on redirects to cross-origin domains? The `Authorization` header seems an obvious candidate for not being sent cross-origin. But, what if folks are doing stuff like `MyApp-API-Token: .....`? Intentional redirect to third-parties are not super rare, and it seems surprising/a possible foot-gun to relay all custom headers to another origin without the caller opting into that behavior.

annevk:
@ptoomey3 it's similar to the body being sent there if you use 307/308 (or 301/302 with a method that is not POST). In general the server receiving all the secrets could relay them if they want.

The threat model here seems to be that the server is cooperating but has an XSS of sorts around its `Location` header handling. I suppose another solution here could be to introduce a new redirect mode, whereby you can only redirect within the origins provided by the request URL and yourself, or some such.

arturjanc:
I think the threat model is a little different here: it's the application interpolating data into URLs of requests to which the `Authorization` header is attached.

For example, let's say my application takes the `user` parameter from the query or fragment, and then make a request to `/api/{user}/num_frombles`. An attacker will be able to set the `user` to `../../redirect?url=https://evil.example#` and make the victim's browser leak their credentials to the attacker's origin (assuming there's a redirector somewhere in the application, which is common.)

[...]

Issue #609, opened 2017-09-28
Should Content-Type header be stripped when redirecting from POST to GET?


ricea:
The reporter of Chrome issue https://bugs.chromium.org/p/chromium/issues/detail?id=765898 ("XHR preserves Content-Type on cross-origin 303 redirect") argues that when a 303 redirect causes a request to be redirected from POST to GET and the request body stripped, the Content-Type header should also be removed.

Chrome's current behaviour differs depending on the type of fetch that is being performed.

Before settling on a single behaviour in Chrome it would be nice to be definitive about what the standard behaviour is.

My understanding from reading the above Chrome issue is that the standard currently implies that the header is not removed, but removing it has better compatibility with servers.

annevk:
I guess we should do this for 301/302 as well when we change the method to GET there? If someone is willing to write tests I'm happy to file all the browser bugs (if any) and change the specification.

It seems a little weird, but since we modify the request anyway we might as well do this too.

Should we remove other `Content-*` headers?

annevk:
cc @mnot

mnot:
Yeah, for 303 it's truely a different request, so it doesn't make sense to talk about the old request body.

For 301 and 302 where you change the method, I *think* the same argument applies; it doesn't make sense to talk about a request body that isn't there.

annevk:
@mnot thanks, but we already drop the body. The question is what should happen with `Content-Type` and potentially other `Content-*` request headers.

mnot:
Yep; `Content-*` headers talk about the body. 

There are cases when there *are* headers that reference a missing body, but I don't think this is one of them; they're mostly around responses that refer to things in caches or the state of the resource after a change.

annevk:
So should we drop all headers that start with `Content-` or only a set of known header names? E.g., if a developer set `Content-Pony`, would it be okay to drop that for these redirects?

annevk:
@mnot ping.

wisniewskit:
I just closed XHR issue 192 (https://github.com/whatwg/xhr/issues/192) of this, but was wondering if the inconsistency I found there is related to this. Browsers are currently passing the "301 POST with string and explicit Content-Type safelisted" XHR WPT in /xhr/send-redirect-to-cors.htm, but I don't believe they should if we're meant to strip these content- headers (since it's a 301 redirect that expects the body to be nulled, so it seems it should not be expecting the content-type to remain).

mnot:
Sorry, missed the notification.

[...]

PR #977, opened 2019-12-03
strip request-body-header for redireting from POST to GET

JuniorHsu:
***
<a href="https://whatpr.org/fetch/977.html" title="Last updated on Dec 9, 2019, 6:56 PM UTC (23ecbb3)">Preview</a> | <a href="https://whatpr.org/fetch/977/493c021...23ecbb3.html" title="Last updated on Dec 9, 2019, 6:56 PM UTC (23ecbb3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -890,8 +890,8 @@ from APIs using <a for=/>fetch</a> that allow control over
  <li>`<code>Set-Cookie2</code>`
 </ul>
 
-<p>A <dfn export>request-body-header name</dfn> is a <a for=/>header</a>
-<a for=header>name</a> that is a <a>byte-case-insensitive</a> match for one of:
+<p>A <dfn export>request-body-header name</dfn> is a <a for=/>header</a> <a for=header>name</a> that
+is a <a>byte-case-insensitive</a> match for one of:
 
 <ul class=brief>
  <li>`<code>Content-Disposition</code>`
@@ -4022,16 +4022,13 @@ optional <i>CORS-preflight flag</i>, run these steps:
 
   <p>then:
 
-  <ol class=brief>
+  <ol>
    <li><p>Set <var>request</var>'s <a for=request>method</a> to `<code>GET</code>` and
    <var>request</var>'s <a for=request>body</a> to null.
+
    <li><p><a for="list">For each</a> <var>headerName</var> of <a>request-body-header name</a>:
-
-   <ol class=brief>
-    <li><p><a for="header list">Delete</a> <var>headerName</var> from <var>request</var>'s
-    <var>headers</var>'s <a for=Headers>header list</a>.
-   </ol>
-
+   <a for="header list">delete</a> <var>headerName</var> from <var>request</var>'s
+   <a for=request>header list</a>.
   </ol>
 
  <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -898,7 +898,6 @@ is a <a>byte-case-insensitive</a> match for one of:
  <li>`<code>Content-Language</code>`
  <li>`<code>Content-Location</code>`
  <li>`<code>Content-Type</code>`
- <li>`<code>Content-Length</code>`
 </ul>
 
 <hr>
--- a/fetch.bs
+++ b/fetch.bs
@@ -890,6 +890,19 @@ from APIs using <a for=/>fetch</a> that allow control over
  <li>`<code>Set-Cookie2</code>`
 </ul>
 
+<p>A <dfn export>request-body-header name</dfn> is a <a for=/>header</a>
+<a for=header>name</a> that is a <a>byte-case-insensitive</a> match for one of:
+
+<ul class=brief>
+ <li>`<code>Content-Disposition</code>`
+ <li>`<code>Content-Encoding</code>`
+ <li>`<code>Content-Language</code>`
+ <li>`<code>Content-Location</code>`
+ <li>`<code>Content-Range</code>`
+ <li>`<code>Content-Type</code>`
+ <li>`<code>ETag</code>`
+</ul>
+
 <hr>
 
 <p>To <dfn export lt="extract header values|extracting header values">extract header values</dfn>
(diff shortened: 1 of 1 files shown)


JuniorHsu:
This addresses #609. I'll work on test and gecko implementation. 
@annevk does the patch make sense?

fetch.bs
```diff
@@ -4007,8 +4020,20 @@ optional <i>CORS-preflight flag</i>, run these steps:
...
   </ul>
 
-  <p>then set <var>request</var>'s <a for=request>method</a> to `<code>GET</code>` and
-  <var>request</var>'s <a for=request>body</a> to null.
+  <p>then:
+
+  <ol class=brief>
+   <li><p> Set <var>request</var>'s <a for=request>method</a> to `<code>GET</code>` and
+   <var>request</var>'s <a for=request>body</a> to null.
+   <li><p> <p><a for="list">For each</a> <var>headerName</var> of
```
annevk: There's a markup error here.

fetch.bs
```diff
@@ -4007,8 +4020,20 @@ optional <i>CORS-preflight flag</i>, run these steps:
    <var>request</var>'s <a for=request>method</a> is not `<code>HEAD</code>`
   </ul>
 
-  <p>then set <var>request</var>'s <a for=request>method</a> to `<code>GET</code>` and
-  <var>request</var>'s <a for=request>body</a> to null.
+  <p>then:
+
+  <ol class=brief>
+   <li><p> Set <var>request</var>'s <a for=request>method</a> to `<code>GET</code>` and
```
annevk: No need for the space before "Set".

annevk:
#944 is a bit related to this, if you wanted to pick up more issues in this space.

ricea:
lgtm. It would be nice if there were tests, but I guess whichever browser implements this first will write them. CC: @yutakahirano who was also active on https://bugs.chromium.org/p/chromium/issues/detail?id=765898.

annevk:
@davidben are you on board with stripping more headers than `Content-Type` and `Content-Length`?

davidben:
I like how in 2019, we *still* don't know what HTTP redirects are. :-)

This is a list of request headers to remove, right? This list is strange. `Content-Disposition`, `Content-Range`, and `ETag` are defined as response headers.

Adding @MattMenke2 and @mnot who may have thoughts about this. One hopes the risk of changes here is tiny, but we should incur that risk at most once.

MattMenke2:
There's also Transfer-Encoding (No idea if Transfer-Encoding is generally used in place of Content-Encoding for uploads, but if we're encouraging using these sorts of entity-body headers symmetrically for request and response bodies, seems like it should be here).

Does seem like we should create some more formal grouping of entity-body headers that can apply both to uploads and downloads, if we want to have some sort of support for consumers setting them in web-platform uploads.

davidben:
Can `Transfer-Encoding` be set by web content? I would hope not since it's a property of how the browser arranges bytes on the wire, rather than the request itself. In that case, hopefully it gets applied below redirect handling anyway.

MattMenke2:
The same is true of Content-Encoding, no?  I mean, they're nominally different, but in practice, not so much.  Regardless, there's discussion of a streaming upload API, which would use chunked uploads.  No idea if Chrome's pre-existing chunked upload support uses.

annevk:
`Transfer-Encoding` is blocked, `Content-Encoding` is not, see https://fetch.spec.whatwg.org/#forbidden-header-name.

MattMenke2:
But we still have rules for how forbidden headers are mutated as we follow redirects.  Also, I don't think any browser should be letting web content set "Content-Encoding: chunked" directly.

MattMenke2:
Though I suppose chunked is generally advertised as a Transfer-Encoding, it just seems not a great idea.

davidben:
I don't think `Content-Encoding: chunked` means anything, so it shouldn't be any different from `Content-Encoding: garbage`. While `Transfer-Encoding` nominally can also do compression, I don't think any browser supports it.

The two are effectively orthogonal. `Content-Encoding` is whether there is compression applied to the body and should be supplied by whoever supplied the body. `Transfer-Encoding` is largely a symptom of HTTP/1.1's poor layering. It should be supplied by the low-level HTTP/1.1 (and not HTTP/2) encoding logic, well below Fetch and the web platform. Alas, HTTP headers are a leaky abstraction, so we're stuck needling to forbid `Transfer-Encoding` at the Fetch level, but we should layer things so we otherwise don't need to think about it.

MattMenke2:
I do think that consumers have just as much control over Transfer-Encoding: chunked (Assuming we expose a streaming upload API) as they have over Content-Length.  Neither one can they set directly, but they can set them indirectly (By their selection of upload body or their choice of upload API).

fetch.bs
```diff
@@ -894,13 +894,11 @@ from APIs using <a for=/>fetch</a> that allow control over
...
 
 <ul class=brief>
- <li>`<code>Content-Disposition</code>`
  <li>`<code>Content-Encoding</code>`
  <li>`<code>Content-Language</code>`
  <li>`<code>Content-Location</code>`
- <li>`<code>Content-Range</code>`
  <li>`<code>Content-Type</code>`
- <li>`<code>ETag</code>`
+ <li>`<code>Content-Length</code>`
```
annevk: `Content-Length` is a forbidden header name so doesn't need to be listed. It's always computed just before the request, so nulling body does effectively cause it to be omitted.

JuniorHsu: It meets my experiment. Thanks.

JuniorHsu:
We're wondering if we need to remove request-body-header (Content-Type, Content-Language... ) from request headers for GET->GET redirection if the client adds them, e.g,  `fetch(url, {headers: {"Content-Encoding": "Identity"}}` for a 301/302/303 response.

We suspect that leaving GET->GET alone is probably the quickest way to consensus, and I'd like to include GET->GET in the web-platform tests.

annevk:
@JuniorHsu if we want to leave GET -> GET alone we do have to add a check for 303 in the specification: that method is not GET or HEAD. As the current wording does affect GET -> GET for 303s.

JuniorHsu:
> @JuniorHsu if we want to leave GET -> GET alone we do have to add a check for 303 in the specification: that method is not GET or HEAD. As the current wording does affect GET -> GET for 303s.

Thanks for finding this. It's good to be consistent

JuniorHsu:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1530230

annevk:
More bugs:
* https://bugs.webkit.org/show_bug.cgi?id=205119
* https://bugs.chromium.org/p/chromium/issues/detail?id=1033056

Issue #930, opened 2019-08-30
Support Structured Headers in the Headers API


annevk:
In particular:

* An algorithm that takes a header name _name_ and a structured header type _type_ and returns the corresponding header's value in some kind of useful object-representation for _type_ (or an error).
* An algorithm that takes a header name _name_ and a structured header value _value_ and replaces all corresponding headers with a new structured header (or errors if _value_ cannot be coerced).

annevk:
This topic came up in the HTTP Workshop and it made me think about the requirements some more. The complicated piece here are the getters from an ergonomics perspective (as for the setter you can rely on the type of the value). There's three separate types and for the return values there's parameters. Returning the value without parameters is useful and often what you want. Taken together this might suggest something like:

```webidl
typedef (boolean or long long or double or ByteString or Uint8Array) SFBareItem; // TODO: distinguish Token and String

interface Headers {
  SFBareItem getItem(ByteString name);
  SFItem getFullItem(ByteString name);
  …
};
```
(There's a question of course as to how much header processing happens in client-side JS to begin with. Perhaps this can be a library first.)

jasnell:
@annevk ... I'd like to see if we can revive this a bit. Having structured header (https://datatracker.ietf.org/doc/html/rfc8941) support in `Headers` would be useful. I do think we might be able to keep it simple by not messing around the with the `set`/`append`/`get` methods and instead focus on introducing new statics on `Headers` that only deal with the serialization/deserialization.

```js
h = new Headers();
h.set('abc', Headers.fromItem('a'));
h.set('xyz', Headers.fromList([1,2,3]);
h.set('foo', Headers.fromDictionary(...);

Headers.toDictionary(h.get('foo'));
// etc
```

The idea here would be to introduce six new static methods on `Headers` whose purpose to only to serialize/deserialize the structured header syntax:

* `Headers.fromItem(...)`
* `Headers.fromList(...)`
* `Headers.fromDict(...)`
* `Headers.toItem(...)`
* `Headers.toList(...)`
* `Headers.toDict(...)`

PR #983, opened 2019-12-13
"Get" a structured header from a header list.

mikewest:
As discussed in #930 and mikewest/corpp#2, this patch adds a "get"
method on header lists that allows consistent extraction of
structured headers.

***
<a href="https://whatpr.org/fetch/983.html" title="Last updated on Dec 21, 2019, 8:54 AM UTC (9125901)">Preview</a> | <a href="https://whatpr.org/fetch/983/b93383c...9125901.html" title="Last updated on Dec 21, 2019, 8:54 AM UTC (9125901)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -381,6 +381,50 @@ for consistency.
 <p class="note no-backref">A <a for=/>header list</a> is essentially a
 specialized multimap. An ordered list of key-value pairs with potentially duplicate keys.
 
+<p>To
+<dfn export for="header list" id=concept-header-list-get-structured-header>get a structured header</dfn>
+given a <var>name</var> and a <var>type</var> from a <a for=/>header list</a> <var>list</var>, run
+these steps:
+
+<ol>
+ <li><p>Assert: <var>type</var> is one of "<code>dictionary</code>", "<code>list</code>", or
+ "<code>item</code>".
+
+ <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var> from
+ <var>list</var>.
+
+ <li><p>If <var>value</var> is null, then return null.
+
+ <li><p>Let <var>result</var> be the result of executing the <a>parsing structured headers</a>
+ algorithm with <var ignore>input_string</var> set to <var>value</var>, and
+ <var ignore>header_type</var> set to <var>type</var>.
+
+ <li><p>If parsing failed, then return failure.
+
+ <li><p>Return <var>result</var>.
+</ol>
+
+<p>To
+<dfn export for="header list" id=concept-header-list-set-structured-header>set a structured header</dfn>
+<a for=header>name</a>/<a>structured header value</a> <var>name</var>/<var>structuredValue</var>
+pair in a <a for=/>header list</a> <var>list</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>serializedValue</var> be the result of executing the
+ <a>serializing structured headers</a> algorithm on <var>structuredValue</var>.
+
+ <li><p><a for="header list">Set</a> <var>name</var>/<var>serializedValue</var> in <var>list</var>.
+</ol>
+
+<p class=note><a>Structured header values</a> are defined as objects which HTTP can (eventually)
+serialize in interesting and efficient ways. For the moment, Fetch only supports <a for=/>header</a>
+<a for=header>values</a> as <a for=/>byte sequences</a>, which means that these objects can be set
+in <a for=/>header lists</a> only via serialization, and they can be obtained from
+<a for=/>header lists</a> only by parsing. In the future the fact that they are objects might be
+preserved end-to-end. [[!HEADER-STRUCTURE]]
+
+<hr>
+
 <p>A <a for=/>header list</a> <var>list</var>
 <dfn export for="header list" lt="contains|does not contain">contains</dfn> a <a for=header>name</a>
 <var>name</var> if <var>list</var> <a for=list>contains</a> a <a for=/>header</a> whose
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -584,10 +585,32 @@ pair in a <a for=/>header list</a> <var>list</var>, run these steps:
...
-objects in <a for=/>header lists</a> by serializing them. In the future the fact that they are
-objects might be preserved end-to-end. [[!HEADER-STRUCTURE]]
+<p>To
+<dfn export for="header list" id=concept-header-list-get-structured-header>get a structured header</dfn>
+given a <var>name</var> and a <var>type</var> from a <a for=/>header list</a> <var>list</var>, run
+these steps:
+
+<ol>
+ <li><p>If <var>type</var> is not one of "<code>dictionary</code>", "<code>list</code>", or
+ "<code>item</code>", return failure.
```
annevk: This should probably be an assert, right? At least until we have an API and at that point we could use an enum on the IDL boundary to enforce it.

fetch.bs
```diff
@@ -584,10 +585,32 @@ pair in a <a for=/>header list</a> <var>list</var>, run these steps:
...
+ <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var> from
+ <var>list</var>.
+ <li><p>If <var>value</var> is null, return null.
```
```suggestion
 <li><p>If <var>value</var> is null, then return null.
```

fetch.bs
```diff
@@ -584,10 +585,32 @@ pair in a <a for=/>header list</a> <var>list</var>, run these steps:
...
+ algorithm with <var ignore>input_string</var> set to <var>value</var>, and
+ <var ignore>header_type</var> set to <var>type</var>.
+ <li><p>If parsing failed, return failure.
+ <li><p>Return <var>result</var>.
+</ol>
+
+<p class=note><a>Structured header values</a> are defined as objects which HTTP can (eventually)
+serialize in interesting and efficient ways. For the moment, Fetch only supports <a for=/>header</a>
+<a for=header>values</a> as <a for=/>byte sequences</a>, which means that these objects can be set
+in <a for=/>header lists</a> only via serialization, and they can be obtained from <a for=/>header
```
annevk: Please keep elements like `<a>` on a single line.

fetch.bs
```diff
@@ -584,10 +585,32 @@ pair in a <a for=/>header list</a> <var>list</var>, run these steps:
...
+<ol>
+ <li><p>If <var>type</var> is not one of "<code>dictionary</code>", "<code>list</code>", or
+ "<code>item</code>", return failure.
+ <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var> from
+ <var>list</var>.
+ <li><p>If <var>value</var> is null, return null.
+ <li><p>Let <var>result</var> be the result of executing the <a>parsing structured headers</a>
+ algorithm with <var ignore>input_string</var> set to <var>value</var>, and
+ <var ignore>header_type</var> set to <var>type</var>.
+ <li><p>If parsing failed, return failure.
```
```suggestion
 <li><p>If parsing failed, then return failure.
```

fetch.bs
```diff
@@ -584,10 +585,32 @@ pair in a <a for=/>header list</a> <var>list</var>, run these steps:
...
+</ol>
+
+<p class=note><a>Structured header values</a> are defined as objects which HTTP can (eventually)
+serialize in interesting and efficient ways. For the moment, Fetch only supports <a for=/>header</a>
+<a for=header>values</a> as <a for=/>byte sequences</a>, which means that these objects can be set
+in <a for=/>header lists</a> only via serialization, and they can be obtained from <a for=/>header
+lists</a> only by parsing. In the future the fact that they are objects might be preserved
+end-to-end. [[!HEADER-STRUCTURE]]
+
+</div>
```
annevk: It seems this and the newline before it were added accidentally? Or is this fixing an existing mistake?

zcorpan:
> Thanks, so the one question I have is whether we want to further map a structured header value to an Infra value?

I think that seems nice. But how would we differentiate between token and string?

annevk:
We'd have a struct consisting of type, value, and parameters, I suspect. (We'd need something similar for the JavaScript API.) But I'm okay leaving that for later and refactoring things as appropriate.

(An alternative that might be nice is that we define implicit mapping, similar to how we deal with various IDL concepts.)

domenic:
I'm really confused why we'd introduce this wrapper. All it seems to add over using the SH spec directly is an assert and a redundant "if failure, return failure".

In https://wicg.github.io/origin-policy/#parse-an-origin-policy-header I used the SH spec directly and I see no compelling reason to change.

I am -1 on this PR.

annevk:
Well, you could reduce the caller algorithm slightly because of this as you no longer have to grab the header value separately. Fetch contains a lot of algorithms like that which make the ergonomics better and generally makes specifications easier to read.

Also, if at some point a header list is able to contain header values that do not need to be parsed because they are already of the correct type, we'd have one place to change to account for that rather than a dozen.

mikewest:
Like @annevk, I see this patch as creating a single point we can change in the future if/when we have cleaner support for structured headers as objects in themselves that are persisted end-to-end. I didn't hang type information in this patch, but I think it's probably worth doing in a subsequent PR.

fetch.bs
mikewest: Done.

fetch.bs
mikewest: Accidental, thanks!

fetch.bs
mikewest: Changed to an assert.

annevk:
Pushed a hopefully final nit to this branch. Happy to merge it now but will wait a bit to see if there's final comments.

mikewest:
Still LGTM, FWIW. Happy holidays! :)

annevk:
Happy holidays to you all too!

PR #993, opened 2020-01-10
Integrate with Fetch Metadata.

### LINKED ISSUE #885: Consider supporting Fetch Metadata.
https://mikewest.github.io/sec-metadata/ is starting to look pretty reasonable. I plan to put up some PRs to sketch out what the integration would look like, based on the thoughts in https://mikewest.github.io/sec-metadata/#fetch-integration. I think we'd need ~3 changes, but who knows? :)

1.  Add a boolean `user activation` flag to `request` objects, defaulting to `false`. We'd set it in HTML's navigation algorithm, based on whether the navigation was triggered by user activation.

2.  Call [set the Fetch Metadata headers for a request](https://mikewest.github.io/sec-metadata/#abstract-opdef-set-the-fetch-metadata-headers-for-a-request). Probably from "main fetch"?

3.  Define "`nested-navigate`", as per #755.

WDYT?

mikewest:
This patch integrates Fetch Metadata processing into Fetch's "main
fetch" algorithm, and defines a "user activation flag" on requests that
will be populated during HTML's "process a navigate fetch" algorithm.

Closes #885.

- [X] At least two implementers are interested (and none opposed):
    * Chrome is shipping this as of 76.
    * Mozilla has labeled the feature as "[worth prototyping](https://mozilla.github.io/standards-positions/#fetch-metadata)".
    * https://wpt.fyi/results/fetch/metadata?label=experimental&label=master&aligned
    * @mikewest is writing a test for the specific bit around Service Worker's access to the headers in https://github.com/web-platform-tests/wpt/pull/27857.
    * Chrome: Shipped.
    * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1695911
    * Safari: https://bugs.webkit.org/show_bug.cgi?id=204744

***
<a href="https://whatpr.org/fetch/993.html" title="Last updated on Mar 4, 2021, 8:14 AM UTC (b36a125)">Preview</a> | <a href="https://whatpr.org/fetch/993/24a8670...b36a125.html" title="Last updated on Mar 4, 2021, 8:14 AM UTC (b36a125)">Diff</a>


mikewest:
As discussed on public-webappsec@. WDYT, @annevk?

fetch.bs
```diff
@@ -3335,6 +3340,9 @@ optionally with a <i>recursive flag</i>, run these steps:
    <li><p><a href=https://w3c.github.io/webappsec-upgrade-insecure-requests/#upgrade-request>Upgrade <var>request</var> to a potentially secure URL, if appropriate</a>.
    [[!UPGRADE]]
 
+   <li><p><a href=https://w3c.github.io/webappsec-fetch-metadata/#abstract-opdef-set-the-fetch-metadata-headers-for-a-request>Set the Fetch metadata headers for <var>request</var></a>.
```
annevk: This should be an exported term. You can file a bug against tabatkins/bikeshed to get Fetch Metadata indexed. If all goes well we can land this PR Monday.

mikewest: Hrm. It is exported from Fetch Metadata, and that spec shows up in Bikeshed's biblio entries, but not in its anchor index. Weird!

Filed https://github.com/tabatkins/bikeshed/issues/1584

annevk: Biblio is for references at the end of the document, Shepherd is for inline cross-references and does not pick up new specifications to crawl automatically.

mikewest:
Coming back to this as a prereq to https://github.com/whatwg/html/pull/5203.

WDYT?

fetch.bs
```diff
@@ -3335,6 +3340,9 @@ optionally with a <i>recursive flag</i>, run these steps:
    <li><p><a href=https://w3c.github.io/webappsec-upgrade-insecure-requests/#upgrade-request>Upgrade <var>request</var> to a potentially secure URL, if appropriate</a>.
    [[!UPGRADE]]
 
+   <li><p><a abstract-op lt="set the Fetch metadata headers for a request">Set the Fetch metadata headers for <var>request</var></a>.
```
annevk: Which is this an abstract-op? I thought those were only for JavaScript.

mikewest: Bikeshed defines `abstract-op` as "[English-language algorithms](https://tabatkins.github.io/bikeshed/#dfn-types)"; that's generally why I've been using it to identify algorithms. What type would you use?

mikewest:
> As Yoav once discovered, this does not work. As https://fetch.spec.whatwg.org/#cors-unsafe-request-header-names is currently invoked, setting Sec- headers before the network layer forces a preflight. We'd have to add an exception there (or its caller).

Would you like me to put these headers specifically into https://fetch.spec.whatwg.org/#cors-safelisted-request-header? Or would you like us to work out a `Sec-*` carveout generally?

fetch.bs
annevk: Interesting, I tend to use dfn for those and only use abstract-op for the IAmACamelCaseAlgorithm types from ECMAScript. I guess I'm not going to block on this then, but copying @domenic and @tabatkins for thoughts.

annevk:
I think we should carve it out generically, right?

Now actually copying @yoavweiss so he can share any thoughts as to why carving out `Sec-` generically might be bad.

mikewest:
Would you like me to carve it out generically in this CL, or in a separate patch? Probably the latter, right?

annevk:
This also does not address https://github.com/w3c/webappsec-fetch-metadata/issues/29 as far as I can tell. We need a solution for that.

Also, https://github.com/w3c/webappsec-fetch-metadata/issues/38 changed right as this PR would expose them to service workers?

Could you maybe do another triage round so I know what upstream issues might end up impacting Fetch still?

fetch.bs
tabatkins: I think it's appropriate to use abstract-op here, fwiw.

tabatkins: (When I finally add some cool functionality around registering algorithms with arguments and types and whatnot, I expect that I'll hook abstract-op as the type for that.)

mikewest:
I merged the last few months of changes in https://github.com/whatwg/fetch/pull/993/commits/f7800972676ffd16648ed401603ce3a9696a6fbe, and moved the integration out of "main fetch" and into "HTTP-network-or-cache" fetch in https://github.com/whatwg/fetch/pull/993/commits/58d9d1870d9adc2bd348033c3bf897fe38729ae2. I believe that addresses the following concerns:

*   We do not expose the headers to the `Request.headers` accessor in Service workers, as they're added after SW have a shot at handling the request. That should allow us to resolve https://github.com/w3c/webappsec-fetch-metadata/issues/38.
*   The HTTPS->HTTP->HTTPS redirect behavior described in the spec (and in WPT) should fall naturally out of this change for the same reason (we recalculate on each hop), which should allow us to resolve https://github.com/w3c/webappsec-fetch-metadata/issues/29.
*    We can punt the question of a more generic `Sec-*` carveout for CORS preflights, as this change ensures that headers are added after the point at which we would have send a preflight (in HTTP fetch).

I've skimmed very quickly through other bugs, but nothing other than https://github.com/whatwg/html/pull/5203 jumps out at me as needing to be resolved before landing this integration. I'll try to label things accordingly if I can find some time.

WDYT?

fetch.bs
```diff
@@ -1661,6 +1662,10 @@ Unless stated otherwise, it is unset.
 
 <p class="note no-backref">This flag is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-user-activation-flag>user-activation flag</dfn>. Unless
+stated otherwise, it is unset.
```
annevk, line 1667: Let's change this into a boolean, similar to keepalive. I'd like to avoid adding flags and slowly convert existing flags to booleans. (I'm happy to help with this for the Fetch and HTML PRs if needed, but would prefer if someone else did it for Fetch Metadata.)

I think we should also add a note that this is for exclusive usage by HTML's navigate algorithm. At least for now?

fetch.bs
```diff
@@ -4203,6 +4208,9 @@ steps. They return a <a for=/>response</a>.
 
    <li><p><a>Append a request `<code>Origin</code>` header</a> for <var>httpRequest</var>.
 
+   <li><p><a abstract-op lt="set the Fetch metadata headers for a request">Set the Fetch metadata
+   headers for <var>request</var></a>. [[!FETCH-METADATA]]
```
annevk, line 4212: _httpRequest_, right?

annevk: And ideally we would use "append" here as we do elsewhere since there should be nothing to override.

annevk (commented):
Thanks for getting around to updating this. The plan looks good to me, I only have nits.

annevk:
I've also added the pull request template to OP as this is a normative change. Mainly so we can find tests and bugs later. I think this has support from Chrome/Firefox.

Maybe @youennf can weigh in for Safari?

fetch.bs
mikewest: Updated in https://github.com/whatwg/fetch/pull/993/commits/c0050629590f01932fcaf4931699705c54029daa. I'll take care of the HTML PR and Fetch Metadata once you're happy with this patch.

fetch.bs
mikewest: Used the right variable, and renamed the algorithm. I'll update Fetch Metadata accordingly once you're happy with the rest.

annevk (approved):
Yeah this looks good, thanks.

fetch.bs
mikewest: HTML is addressed in the updates to https://github.com/whatwg/html/pull/5203.

Fetch Metadata is renamed as well: https://w3c.github.io/webappsec-fetch-metadata/#abstract-opdef-append-the-fetch-metadata-headers-for-a-request.

mikewest:
Updated dependent PRs and specs, and filled out the template above.

annevk:
Thanks @mikewest!

annevk:
Unfortunately, I just realized I forgot to note https://github.com/whatwg/html/pull/5203 in the commit message. Oh well.

PR #1011, opened 2020-03-26
Update query parameters example

MikiDi:
The current [example for using query parameters](https://fetch.spec.whatwg.org/#fetch-blob-example) **writes** to the [`searchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) property of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), which is now [**readonly**](https://www.fxsitecompat.dev/en-CA/docs/2015/urlutils-searchparams-is-now-readonly/). This patch updates the example to make use of [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams) to set query parameters instead.

***
<a href="https://whatpr.org/fetch/1011.html" title="Last updated on Mar 26, 2020, 7:47 PM UTC (22e1a7a)">Preview</a> | <a href="https://whatpr.org/fetch/1011/00239c1...22e1a7a.html" title="Last updated on Mar 26, 2020, 7:47 PM UTC (22e1a7a)">Diff</a>


domenic:
This isn't quite right. readonly just means you cannot do `url.searchParams = foo`. The `append()` method still works.

annevk:
Yeah, closing this therefore. I would accept a PR to modernize the code in the examples somewhat though, e.g., sprinkle some `const` dust.

MikiDi:
:man_facepalming:

PR #1013, opened 2020-03-30
Make Cors requests not take Referrer-Policy into account

rwlbuis:
The test fetch/origin/assorted.window.js indicates that CORS (mode
is Cors) requests should not take Referrer-Policy into account, however
basing this on the tainting value seems to invalidate that, as same-origin
Cors fetches set tainting to Basic, not Cors, and would take
Referrer-Policy into account, so instead of tainting base Step 2
on mode.

***
<a href="https://whatpr.org/fetch/1013.html" title="Last updated on Mar 31, 2020, 12:53 PM UTC (208d6c6)">Preview</a> | <a href="https://whatpr.org/fetch/1013/00239c1...208d6c6.html" title="Last updated on Mar 31, 2020, 12:53 PM UTC (208d6c6)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2436,7 +2436,7 @@ request <a for=/>header</a> indicates where a
 <p class="note no-backref">The `<a http-header><code>Origin</code></a>` header is a version of the
 `<code>Referer</code>` [sic] header that does not reveal a <a for=url>path</a>. It is used for all
 <a lt="HTTP fetch">HTTP fetches</a> whose <a for=/>request</a>'s
-<a for=request>response tainting</a> is "<code>cors</code>", as well as those where
+<a for=request>mode</a> is "<code>cors</code>", as well as those where
 <a for=/>request</a>'s <a for=request>method</a> is neither `<code>GET</code>` nor
 `<code>HEAD</code>`. Due to compatibility constraints it is not included in all
 <a lt=fetch for=/>fetches</a>.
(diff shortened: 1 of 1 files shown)


annevk:
We also need to update the note above which also talks about using response tainting.

Looking at cc80ec58d24668413b7a3c7160d9b4d83ace7b20 I wonder if we need to take both into account, as otherwise same origin requests will include the `Origin` header, which at least per the specification it has never done.

annevk:
cc @JuniorHsu

rwlbuis:
> We also need to update the note above which also talks about using response tainting.

Good point, I missed it, fixed.

> Looking at [cc80ec5](https://github.com/whatwg/fetch/commit/cc80ec58d24668413b7a3c7160d9b4d83ace7b20) I wonder if we need to take both into account, as otherwise same origin requests will include the `Origin` header, which at least per the specification it has never done.

To be clear, I think we want to either fix the specification or the test, I have no big preference so just let me know what direction we will go and I can make the patch if needed.

annevk:
Sorry, I don't have it all paged in. Looking more closely at `fetch/origin/assorted.window.js` and results in browsers it seems that they mostly _do_ include the `Origin` header for same-origin "cors" requests, despite the specification suggesting the opposite. Unfortunately it doesn't test non-POST methods as extensively. It would be good to have it confirmed that a same-origin "cors" request with method GET also includes the `Origin` header at all times. If that's the case I think we should move forward with this PR as-is.

rwlbuis:
> Sorry, I don't have it all paged in. Looking more closely at `fetch/origin/assorted.window.js` and results in browsers it seems that they mostly _do_ include the `Origin` header for same-origin "cors" requests, despite the specification suggesting the opposite. Unfortunately it doesn't test non-POST methods as extensively. It would be good to have it confirmed that a same-origin "cors" request with method GET also includes the `Origin` header at all times. If that's the case I think we should move forward with this PR as-is.

I created this to check above, let's see what happens:
https://github.com/web-platform-tests/wpt/pull/22567

yutakahirano:
Assuming the expectations of https://github.com/web-platform-tests/wpt/pull/22567 are correct, the first condition should be:

If `mode` is "websocket" or `response tainting` is "cors" or (`mode` is "cors" and `method` is neither "GET" nor "HEAD"):

annevk:
The way I read the expectation is that we could close this and make no change. Since "cors" same-origin would go into step 3 of https://fetch.spec.whatwg.org/#append-a-request-origin-header which means "cors" same-origin GET does not an `Origin` header "cors" same-origin POST is influenced by referrer policy.

yutakahirano:
Ah I see thank you.

rwlbuis:
I am closing this now the test change has been merged.

PR #1016, opened 2020-04-01
Add `cross-origin` to allowed CORP header values.

mikewest:
This patch adds `cross-site` to the list of allowed CORP header values. It's a no-op
at this point, but I've started asking folks to set this header in preparation for COEP
rollout, and one has pointed me to the spec, noting that the value I'm suggesting they
add isn't technically legal.

***
<a href="https://whatpr.org/fetch/1016.html" title="Last updated on Apr 1, 2020, 2:32 PM UTC (1665249)">Preview</a> | <a href="https://whatpr.org/fetch/1016/00239c1...1665249.html" title="Last updated on Apr 1, 2020, 2:32 PM UTC (1665249)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3122,7 +3122,7 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
 <p>Its <a for=header>value</a> <a>ABNF</a>:
 
 <pre><code class=lang-abnf>
-Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" ; case-sensitive
+Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-site" ; case-sensitive
 </code></pre>
 
 <p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <var>request</var> and
--- a/fetch.bs
+++ b/fetch.bs
@@ -3122,7 +3122,7 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
 <p>Its <a for=header>value</a> <a>ABNF</a>:
 
 <pre><code class=lang-abnf>
-Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-site" ; case-sensitive
+Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-origin" ; case-sensitive
 </code></pre>
 
 <p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <var>request</var> and


mikewest:
WDYT about this very complicated patch, @annevk?

mikewest:
Aaaand I just realized that I apparently typo'd this as `cross-origin` in the COEP spec. Ugh. I am dumb, and my dumbness apparently bled into our implementation in Chromium. That's unfortunate. Did Mozilla spell this the same way?

/cc @yutakahirano

mikewest:
Ugh. https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpChannel.cpp#7679

`cross-site` is better than `cross-origin`, but maybe we should support both for some period of time because I'm dumb?

annevk:
I don't think `cross-site` is a good idea: https://github.com/whatwg/html/issues/5388.

mikewest:
Hrm. I see `same-origin`, `same-site`, and `cross-site` as an expanding set of buckets, each encompassing the previous. In that version of the tale, `cross-site` means `cross-site-and-also-same-site-because-if-you-can-load-me-i-can-obviously-load-me`. I recognize the ambiguity, but `cross-origin` doesn't seem significantly better (given that some cross-origin resources are, in fact, same-site.

Meh. Given that we have implementations with this wording in both Chromium and Firefox, I guess I can live with `cross-origin`, and I'll update this patch accordingly. Spelling is hard.

annevk:
My reasoning was that this is a cross-origin policy and therefore:

* same-origin: no other origins
* same-site: same-site origins
* cross-origin: all origins, whereas with cross-site you might think it would exclude same-site-but-not-same-origin origins

At some point we might want to spell this out if this continues to come up. And yeah, all tests and implementations use `cross-origin` at the moment (and I PR'd corpp for this at some point in the past).

annevk:
It seems your membership of the Google GitHub organization is not public. You'll need to change that. It might also take a day or so as we need something in HTML to change for Fetch to build again.

mikewest:
> And yeah, all tests and implementations use `cross-origin` at the moment

Yup. Like I said, I can live with it, and I don't have different ideas for non-ambiguous spellings.

> It seems your membership of the Google GitHub organization is not public. You'll need to change that.

Fixed.

> It might also take a day or so as we need something in HTML to change for Fetch to build again.

No rush. Thank you!

yutakahirano:
FTR: https://github.com/WICG/cross-origin-embedder-policy/pull/1

mikewest:
Yup. I LGTM'd it and then apparently forgot about it. As above, me == dumb.

Issue #985, opened 2019-12-18
Have CORP check rely on request's response tainting rather than request's mode


yutakahirano:
Brought up at https://chromium-review.googlesource.com/c/chromium/src/+/1971810, related with https://github.com/w3c/ServiceWorker/issues/1490.

We would like to run [the CORP check](https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check) in CacheStorage for COEP (https://github.com/w3c/ServiceWorker/issues/1490), but currently the CORP check relies on the request mode which is not directly accessible there. [request's response tainting](https://fetch.spec.whatwg.org/#concept-request-response-tainting) is stored as [response type](https://fetch.spec.whatwg.org/#concept-response-type) and we can use it. What do you think about changing the first item as follows?

If _request_'s response tainting is not "opaque", then return __allowed__.

In the CacheStorage spec, we will restore a _request_ for the CORP check as follows:

 - Let _request_ be a new request.
 - Set _request_'s URL to the given URL.
 - Set _request_'s origin to the current realm's origin
 - Set _request_'s response tainting to 
   - "basic" if _response_'s type is "basic", "default" or "opaqueredirect"
   - "opaque" if _response_'s type is "opaque"
   - "cors" if _response_'s type is "cors"

@ArthurSonzogni @wanderview @annevk

annevk:
Given the check the only caller does today is

> If httpRequest’s response tainting is not "cors"

per https://fetch.spec.whatwg.org/#ref-for-cross-origin-resource-policy-check perhaps that initial step

> If request’s mode is not "no-cors", then return allowed.

is even redundant?

I think we should investigate how this should change together with the change to enforce this for a service worker response.

yutakahirano:
Yes, that is a possible approach. Do you think it's better?

Regarding the service worker response, [it](https://mikewest.github.io/corpp/#integration-sw) is now part of https://github.com/mikewest/corpp.

annevk:
I think I would prefer putting all the checks in Fetch, if possible (Fetch gets the response from the service worker and can do it, after all), rather than putting security checks in Service Workers for this.

yutakahirano:
I prefer running a CORP check in cache.match(https://github.com/w3c/ServiceWorker/issues/1490), which means calling it from the SW spec.

annevk:
Fair, I guess that would be the exception, but I don't think we need it for `respondWith()`.

wanderview:
[...]

PR #1030, opened 2020-06-01
Integrate CORP and COEP

yutakahirano:
This is part of the introduction of COEP
(https://github.com/whatwg/html/pull/5454). The CORP check now takes
COEP into account. Also, responses coming from service workers
are checked.

***
<a href="https://whatpr.org/fetch/1030.html" title="Last updated on Jun 26, 2020, 7:14 AM UTC (ea40eaf)">Preview</a> | <a href="https://whatpr.org/fetch/1030/b228c33...ea40eaf.html" title="Last updated on Jun 26, 2020, 7:14 AM UTC (ea40eaf)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1751,14 +1751,17 @@ source of security bugs. Please seek security review for features that deal with
 
 <hr>
 
-<p>To <dfn export>serialize a request URL for reporting</dfn>, given a <a for=/>request</a>
-<var>request</var>, run these steps:
+<p>To <dfn export>serialize a response URL for reporting</dfn>, given a <a for=/>response</a>
+<var>response</var>, run these steps:
 
 <ol>
- <li>
-  <p>Let <var>url</var> be a copy of <var>request</var>'s <a for=request>URL</a>.
+ <li><p>Assert: <var>response</var>'s <a for=response>URL list</a> <a for=list>is not empty</a>.
 
-  <p class="note">This is not <var>request</var>'s <a for=request>current URL</a> in order to avoid
+ <li>
+  <p>Let <var>url</var> be a copy of <var>response</var>'s <a for=response>URL list</a>'s first
+  element.
+
+  <p class="note">This is not <var>response</var>'s <a for=response>URL</a> in order to avoid
   leaking information about redirect targets (see
   <a href="https://w3c.github.io/webappsec-csp/#security-violation-reports">similar considerations for CSP reporting</a>
   too). [[CSP]]
@@ -1771,26 +1774,6 @@ source of security bugs. Please seek security review for features that deal with
  <i>exclude fragment flag</i> set.
 </ol>
 
-
-<p>To <dfn export for=request id=concept-request-serialize-url-for-reporting>serialize a request URL
-for reporting</dfn>, given a <a for=/>request</a> <var>request</var>, run these steps:
-
-<ol>
- <li>
-  <p>Let <var>url</var> be a copy of <var>request</var>'s <a for=request>URL</a>.
-
-  <p class="note">This is not <var>request</var>'s <a for=request>current URL</a> in order to avoid
-  leaking information about redirect targets (see
-  <a href="https://w3c.github.io/webappsec-csp/#security-violation-reports">similar considerations
-  for CSP reporting</a> too).
-
- <li><p><a>Set the username</a> given <var>url</var> and the empty string.
-
- <li><p><a>Set the password</a> given <var>url</var> and the empty string.
-
- <li><p>Return <var>url</var> <a lt="url serializer">serialized</a>with the
- <i>exclude fragment flag</i> set.
-</ol>
 
 <h4 id=responses>Responses</h4>
 
@@ -3167,23 +3150,54 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
 Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-origin" ; case-sensitive
 </code></pre>
 
-<p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <var>request</var> and
-<var>response</var>, run these steps:</p>
+<p>To perform a <dfn export>cross-origin resource policy check</dfn>, given an <a for=url>origin</a>
+<var>origin</var>, an <a for=/>environment settings object</a> <var>settingsObject</var>, a
+<a for=/>response</a> <var>response</var>, and an optional boolean <var>forNavigation</var>, run
+these steps:
 
 <ol>
- <li><p>If <var>request</var>'s <a for=request>mode</a> is not "<code>no-cors</code>", then return
- <b>allowed</b>.
+ <li><p>Set <var>forNavigation</var> to false if it is not given.
+
+ <li><p>Let <var>embedderPolicy</var> be <var>settingsObject</var>'s embedder policy.
 
  <li>
-  <p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
-  <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, then return
-  <b>allowed</b>.
+  <p>If the <a>cross-origin resource policy internal check</a> with <var>origin</var>,
+  "<code>unsafe-none</code>", <var>response</var>, and <var>forNavigation</var> returns
+  <b>blocked</b>, then return <b>blocked</b>.
 
-  <p class="note no-backref">While redirects that carry  a
-  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header are checked, redirects
-  without such a header resulting in <var>response</var> do not affect the outcome as the default is
-  <b>allowed</b>.
-  <!-- This changes with COEP's cross-origin value. -->
+  <p class="note">This step is needed because we don't want to report violations not related to
+  Cross-Origin Embedder Policy below.
+
+ <li><p>If the <a>cross-origin resource policy internal check</a> with <var>origin</var>,
+ <var>embedderPolicy</var>'s report only value, <var>response</var>, and <var>forNavigation</var>
+ returns <b>blocked</b>, then <a>queue a cross-origin embedder policy CORP violation report</a>
+ with <var>response</var>, <var>settingsObject</var>, and true.
+
+ <li><p>If the <a>cross-origin resource policy internal check</a> with <var>origin</var>,
+ <var>embedderPolicy</var>'s value, <var>response</var>, and <var>forNavigation</var> returns
+ <b>allowed</b>, then return <b>allowed</b>.
+
+ <li><p><a>Queue a cross-origin embedder policy CORP violation report</a> with <var>response</var>,
+ <var>settingsObject</var>, and false.
+
+ <li><p>Return <b>blocked</b>.
+</ol>
+
+<p class="note no-backref">Only HTML's navigate algorithm uses this check with
+<var>forNavigation</var> set to true, and it's always for nested navigations. Otherwise,
+<var>response</var> is either the <a for=internal>internal response</a> of an
+<a>opaque filtered response</a> or a <a for=/>response</a> which will be the
+<a for=internal>internal response</a> of an
+<a>opaque filtered response</a>. [[HTML]]
+
+<p>To perform a <dfn>cross-origin resource policy internal check</dfn>, given an
+<a for=url>origin</a> <var>origin</var>, an embedder policy value <var>embedderPolicyValue</var>,
+a <a for=/>response</a> <var>response</var>, and a boolean <var>forNavigation</var>, run these
+steps:
+
+<ol>
+ <li><p>If <var>forNavigation</var> is true and <var>embedderPolicyValue</var> is
+ "<code>unsafe-none</code>", then return <b>allowed</b>.
 
  <li>
   <p>Let <var>policy</var> be the result of <a for="header list">getting</a>
@@ -3191,33 +3205,90 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
   <a for=response>header list</a>.
 
   <p class=note>This means that `<code>Cross-Origin-Resource-Policy: same-site, same-origin</code>`
-  ends up as <b>allowed</b> below as it will never match anything. Two or more
+  ends up as <b>allowed</b> below as it will never match anything, as long as
+  <var>embedderPolicyValue</var> is "<code>unsafe-none</code>". Two or more
   `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` headers will have the same effect.
 
- <li><p>If <var>policy</var> is `<code>same-origin</code>`, then return <b>blocked</b>.
+ <li><p>If <var>policy</var> is neither `<code>same-origin</code>`, `<code>same-site</code>`, nor
+`<code>cross-origin</code>`, then set <var>policy</var> to null.
+
+ <li><p>If <var>policy</var> is null and <var>embedderPolicyValue</var> is
+ "<code>require-corp</code>", then set <var>policy</var> to `<code>same-origin</code>`.
 
  <li>
-  <p>If the following are true
+  <p>Switch on <var>policy</var>:
 
-  <ul class=brief>
-   <li><var>request</var>'s <a for=request>origin</a> is <a>schemelessly same site</a> with
-   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>
-   <li><var>request</var>'s <a for=request>origin</a>'s <a for=url>scheme</a> is
-   "<code>https</code>" or <var>response</var>'s <a for=response>HTTPS state</a> is
-   "<code>none</code>"
-  </ul>
+  <dl class=switch>
+   <dt>null
+   <dt>`<code>cross-origin</code>`
+   <dd><p>Return <b>allowed</b>.
 
-  <p>then return <b>allowed</b>.
+   <dt>`<code>same-origin</code>`
+   <dd>
+    <p>If <var>origin</var> is <a>same origin</a> with <var>response</var>'s <a for=response>URL</a>'s
+    <a for=url>origin</a>, then return <b>allowed</b>.
 
-  <p class=note>This prevents HTTPS responses with
-  `<code>Cross-Origin-Resource-Policy: same-site</code>` from being accessed without secure
-  transport.
+    <p>Otherwise, return <b>blocked</b>.
 
- <li><p>If <var>policy</var> is `<code>same-site</code>`, then return <b>blocked</b>.
+   <dt>`<code>same-site</code>`
+   <dd>
+    <p>If the following are true
 
- <li><p>Return <b>allowed</b>.
+    <ul class=brief>
+     <li><p><var>origin</var> is <a>schemelessly same site</a> with <var>response</var>'s
+     <a for=response>URL</a>'s <a for=url>origin</a>
+
+     <li><p><var>origin</var>'s <a for=url>scheme</a> is "<code>https</code>" or
+     <var>response</var>'s <a for=response>HTTPS state</a> is "<code>none</code>"
+    </ul>
+
+    <p>then return <b>allowed</b>.
+
+    <p>Otherwise, return <b>blocked</b>.
+
+    <p class=note>`<code>Cross-Origin-Resource-Policy: same-site</code>` does not consider a
+    response delivered via a secure transport to match a non-secure requesting origin,
+    even if their hosts are otherwise same site. Securely-transported responses will only
+    match a securely-transported initiator.
+  </dl>
 </ol>
 
+<p>To <dfn>queue a cross-origin embedder policy CORP violation report</dfn>, given a
+<a for=/>response</a> <var>response</var>, an <a for=/>environment settings object</a>
+<var>settingsObject</var>, and a boolean <var>reportOnly</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>endpoint</var> be <var>settingsObject</var>'s embedder policy's
+ report only reporting endpoint if the <var>reportOnly</var> is true and
+ <var>settingsObject</var>'s embedder policy's reporting endpoint otherwise.
+
+ <li><p>Let <var>serialized url</var> be the result of
+ <a lt="serialize a response URL for reporting">serializing a response URL for reporting</a> with
+ <var>response</var>.
+
+ <li>
+  <p>Let <var>body</var> be a new object containing the following properties:
+
+  <table>
+   <thead>
+    <th>key
+    <th>value
+   </thead>
+   <tbody>
+    <tr>
+     <td>"<code>type</code>"
+     <td>"<code>corp</code>"
+    </tr>
+    <tr>
+     <td>"<code>blocked-url</code>"
+     <td><var>serialized url</var>
+    </tr>
+   </tbody>
+  </table>
+
+ <li><p><a for="reporting">Queue</a> <var>body</var> as the "<code>coep</code>"report type for
+ <var>endpoint</var> on <var>settingsObject</var>. [[!REPORTING]]
+</ol>
 
 
 <h2 id=fetching>Fetching</h2>
@@ -3609,9 +3680,8 @@ optionally with a <i>recursive flag</i>, run these steps:
   <p>If <var>internalResponse</var>'s <a for=response>URL list</a> <a for=list>is empty</a>, then
   set it to a <a for=list>clone</a> of <var>request</var>'s <a for=request>URL list</a>.
 
-  <p class=note>A <a for=/>response</a>'s <a for=response>URL list</a> will typically be empty at
-  this point, unless it came from a service worker, in which case it will only be empty if it was
-  created through <a lt="Response()" constructor><code>new Response()</code></a>.
+  <p class=note>A <a for=/>response</a>'s <a for=response>URL list</a> can be empty when it's for
+  "<code>about</code>" URLs for example.
  <!-- If you are ever tempted to move this around, carefully consider responses from about URLs,
       blob URLs, service workers, HTTP cache, HTTP network, etc. -->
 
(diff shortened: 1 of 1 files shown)


yutakahirano:
@domenic: PTAL

I cannot make COEP related links, which I __guess__ blocked by whatwg/html#5454. Please let me know if I'm wrong.

Note:
 - The reporting logic in the CORP check in https://wicg.github.io/cross-origin-embedder-policy/ is wrong, so I fixed the logic in this PR.
 - The CORP check between the original client and the SW (described in https://wicg.github.io/cross-origin-embedder-policy/#integration-sw) is included, because IIRC @annevk wanted to avoid checking CORP in places other than the fetch spec as much as possible.

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+ "<code>coep</code>" for <var>reporting endpoint</var> on <var>request</var>'s
+ <a for=request>client</a>. [[!REPORTING]]
 </ol>
 
 
+<p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <a for=/>request</a>
+<var>request</var> and a <a for=/>response</a> <var>response</var>, run these steps:</p>
+<ol>
+ <li><p>Let <var>embedder policy</var> be <var>request</var>'s <a for=request>client</a>'s
+ embedder policy.
```
domenic, line 3262: This will need to cross-link

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+
+ <li>
+  <p>If the <a>cross-origin resource policy internal check</a> with "<code>unsafe-none</code>",
+  <var>request</var> and <var>response</var> returns <b>blocked</b>, then return <b>blocked</b>.
+
+  <p class="note no-backref">This is to queue only COEP related violation reports.
```
domenic, line 3268:
```suggestion
  <p class="note">This is to queue only COEP-related violation reports.
```

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+
+ <li><p>If the <a>cross-origin resource policy internal check</a> with <var>embedder policy</var>'s
+ report only value, <var>request</var> and <var>response</var> returns <b>blocked</b>, then
```
domenic, line 3271: "report only value" should cross-link. I guess we should export that from HTML too.

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+ <a>queue a cross-origin embedder policy CORP violation report</a> with
+ <var>request</var> and <var>embedder policy</var>'s report only reporting endpoint.
```
domenic, line 3273: Same for "report only reporting endpoint".

Hmm, could we make this algorithm that an "embedder policy" and a mode or something, which is "normal" or "report only"? So it doesn't have to look into the internals of the "embedder policy" struct as much? I dunno, maybe it's fine to have these algorithms so tightly coupled. But it is strange to me that they are split across different specs if so.

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
-<var>response</var>, run these steps:</p>
+<p>To perform a <dfn>cross-origin resource policy internal check</dfn>, given a string
+<var>embedder policy value</var>, a <a for=/>request</a> <var>request</var> and
+a <a for=/>response</a> <var>response</var>, run these steps:</p>
 
 <ol>
- <li><p>If <var>request</var>'s <a for=request>mode</a> is not "<code>no-cors</code>", then return
- <b>allowed</b>.
+ <li><p>Assert: <var>request</var>'s <a for=request>mode</a> is "<code>navigate</code>" or
+ "<code>no-cors</code>".
```
domenic, line 3152: This is very different than https://wicg.github.io/cross-origin-embedder-policy/#integration-fetch. I guess that is intentional?

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
-  <b>allowed</b>.
+  <p>If <var>request</var>'s <a for=request>mode</a> is "<code>navigate</code>":
 
-  <p class="note no-backref">While redirects that carry  a
-  `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` header are checked, redirects
-  without such a header resulting in <var>response</var> do not affect the outcome as the default is
-  <b>allowed</b>.
-  <!-- This changes with COEP's cross-origin value. -->
+  <p class="note no-backref">Only HTML's navigate algorithm uses this check for the
+  "<code>navigate</code>" <a for=request>mode</a>, for nested navigations. [[!HTML]]
```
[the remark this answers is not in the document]
yutakahirano: Thank you. I don't know the meaning of "!" here, can you explain it?

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+    <p>Otherwise, return <b>blocked</b>.
 
- <li><p>If <var>policy</var> is `<code>same-site</code>`, then return <b>blocked</b>.
+   <dt>"<code>same-site</code>"
+   <dd>
+    <p>If the following are true
 
- <li><p>Return <b>allowed</b>.
+    <ul class=brief>
+     <li><p><var>request</var>'s <a for=request>origin</a> is <a>schemelessly same site</a> with
```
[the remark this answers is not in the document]
yutakahirano: I think I didn't change anything here. The current https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check uses "schemelessly same site".

fetch.bs
yutakahirano: Yes, please see #985. The responsibility to handle requests with modes other than "no-cors" has not been clear up until now. This change clarifies that, stating that it is caller's responsibility.

fetch.bs
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+  <p>Switch on <var>policy</var>:
 
-  <ul class=brief>
-   <li><var>request</var>'s <a for=request>origin</a> is <a>schemelessly same site</a> with
-   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>
-   <li><var>request</var>'s <a for=request>origin</a>'s <a for=url>scheme</a> is
-   "<code>https</code>" or <var>response</var>'s <a for=response>HTTPS state</a> is
-   "<code>none</code>"
-  </ul>
+  <dl class=switch>
```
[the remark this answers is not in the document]
yutakahirano: Good catch, thanks. I added a statement sanitizing _policy_ above.

fetch.bs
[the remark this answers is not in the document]
domenic: See https://tabatkins.github.io/bikeshed/#biblio

fetch.bs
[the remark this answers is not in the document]
domenic: I meant compared to https://wicg.github.io/cross-origin-embedder-policy/#abstract-opdef-cross-origin-resource-policy-internal-check. I think the change from Fetch -> COEP spec is intentional, but I am unsure if it is tested...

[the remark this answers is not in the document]
yutakahirano: Ah I see. @mikewest, did you change that intentionally?

fetch.bs
```diff
@@ -3871,6 +3980,10 @@ optional <i>CORS-preflight flag</i>, run these steps:
 
        <li><var>request</var>'s <a for=request>redirect mode</a> is not "<code>follow</code>" and
        <var>response</var>'s <a for=response>URL list</a> has more than one item.
+
+       <li><p><var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>" and
+       the <a>cross-origin resource policy check</a> with <var>request</var> and <var>response</var>
+       returns <b>blocked</b>.
```
annevk, line 3986: Instead of calling it twice both here and in the HTTP layer, could we call it in a step before "If actualResponse’s status is a redirect status, then:"? Then we would only have to check the response once I think, whether it came from a service worker or the network.

It seems that would miss the 401/407 scenarios, but I wonder if those should not also invoke HTTP fetch (maybe the latter shouldn't though) as they end up bypassing the service worker as currently defined. Hmm.

Either way we should document what we're doing and why, potentially just in a source comment for now.

fetch.bs
```diff
@@ -3123,58 +3143,147 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
 Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-origin" ; case-sensitive
 </code></pre>
 
-<p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <var>request</var> and
-<var>response</var>, run these steps:</p>
+<p>To perform a <dfn>cross-origin resource policy internal check</dfn>, given an
```
annevk, line 3146: I don't think this algorithm works now we also use it for responses from a service worker as it primarily uses the request as authority, whereas the request has almost zero authority in such situations.

This is similar to the problem with the Cache API check and needs a similar solution.

yutakahirano:
Hi Anne, I changed the CORP check, and it now now uses response's URL list instead of request's URL list. Can you take a look again?

fetch.bs
```diff
@@ -3145,58 +3148,145 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
 Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-origin" ; case-sensitive
 </code></pre>
 
-<p>To perform a <dfn>cross-origin resource policy check</dfn>, given a <var>request</var> and
-<var>response</var>, run these steps:</p>
+<p>To perform a <dfn>cross-origin resource policy internal check</dfn>, given a
+<a for=/>response</a> <var>response</var>, an <a for=url>origin</a>, <var>origin</var>, an
+embedder policy value <var>embedder policy value</var> and a boolean
+<var>for navigation</var>, run these steps:
```
annevk, line 3154: Since we're checking if origin, using policy X, can read a response I'd reorder the arguments a bit.

fetch.bs
```diff
@@ -3145,58 +3148,145 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
   <a for=response>header list</a>.
 
   <p class=note>This means that `<code>Cross-Origin-Resource-Policy: same-site, same-origin</code>`
-  ends up as <b>allowed</b> below as it will never match anything. Two or more
+  ends up as <b>allowed</b> below as it will never match anything, as long as
+  <var>embedder policy value</var> is "<code>unsafe-none</code>". Two or more
   `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` headers will have the same effect.
 
- <li><p>If <var>policy</var> is `<code>same-origin</code>`, then return <b>blocked</b>.
+ <li><p>If <var>policy</var> is neither "<code>same-origin</code>", "<code>same-site</code>" nor
```
annevk, line 3170: Oxford comma before nor.

fetch.bs
annevk, line 3154: Please camelCase the variable names.

fetch.bs
annevk, line 3170: You need to use backticks here as the original text did since the result of getting a header value is a byte sequence, not a string.

fetch.bs

fetch.bs

fetch.bs

fetch.bs

fetch.bs
```diff
@@ -3145,58 +3148,145 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
 
+<p>To perform a <dfn export>cross-origin resource policy check</dfn>, given an <a for=url>origin</a>
+origin, an <a for=/>environment settings object</a> <var>settingsObject</var>, a
+<a for=/>response</a> <var>response</var>, and an optional <i>navigation flag</i>, run these steps:
+
+<p class="note no-backref">Only HTML's navigate algorithm uses this check with the
+<i>navigation flag</i> set, and it's always for nested navigations. Otherwise, <var>response</var>
+is either the <a for=internal>internal response</a> of an <a>opaque filtered response</a> or
+a <a for=/>response</a> which will be the <a for=internal>internal response</a> of an
+<a>opaque filtered response</a>. [[HTML]]
```
annevk, line 3261: If we make HTML not invoke this when the embedder policy is "unsafe-none", wouldn't that be easier?

yutakahirano: Yeah I agree that that wouldn't work with the report-only value.

yutakahirano:
> Not needed for this PR, but what do you think about moving the URL list copying and CORP check into the scheme fetch algorithm? We might not even need as much URL list copying for the other schemes as redirecting to them is forbidden.

I think having the CORP check (only) in the scheme fetch is not enough because of redirects.

yutakahirano:
Anne, are you fine with this change except for missing links defined in the HTML spec? If so I'll fix the HTML and Service Worker PRs accordingly.

fetch.bs
```diff
@@ -3145,58 +3148,145 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+    <p class=note>`<code>Cross-Origin-Resource-Policy: same-site</code>` does not consider a
+    response delivered via a secure transport to match a non-secure requesting origin,
+    even if their hosts are otherwise same site. Securely-transported responses will only
+    match a securely-transported initiator.
+  </dl>
+</ol>
+
+<p>To <dfn>queue a cross-origin embedder policy CORP violation report</dfn>, given a
+<a for=/>response</a> <var>response</var>, an <a for=/>environment settings object</a>
+<var>settingsObject</var>, and a <i>report only flag</i>, run these steps:
```
annevk, line 3216:
```suggestion
<var>settingsObject</var>, and a boolean <var>reportOnly</var>, run these steps:
```

fetch.bs
```diff
@@ -3145,58 +3148,145 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+
+ <li><p><a href="https://w3c.github.io/reporting/#queue-report">Queue</a> <var>body</var> as
+ "<code>coep</code>" for <var>endpoint</var> on <var>settingsObject</var>.
+ [[!REPORTING]]
 </ol>
 
 
+<p>To perform a <dfn export>cross-origin resource policy check</dfn>, given an <a for=url>origin</a>
+origin, an <a for=/>environment settings object</a> <var>settingsObject</var>, a
+<a for=/>response</a> <var>response</var>, and an optional <i>navigation flag</i>, run these steps:
```
annevk, line 3255:
```suggestion
<a for=/>response</a> <var>response</var>, and an optional boolean <var>forNavigation</var>, run these steps:
```

fetch.bs
annevk: Can we move this note to be after the algorithm? It reads a little weird being before the list of steps.

fetch.bs
```diff
@@ -1749,14 +1749,17 @@ source of security bugs. Please seek security review for features that deal with
 
 <hr>
 
-<p>To <dfn export>serialize a request URL for reporting</dfn>, given a <a for=/>request</a>
-<var>request</var>, run these steps:
+<p>To <dfn export>serialize a response URL for reporting</dfn>, given a <a for=/>response</a>
+<var>response</var>, run these steps:
 
 <ol>
+ <li><p>Assert: <var>response</var>'s <a for=response>URL list</a> is not empty.
```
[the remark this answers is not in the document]

fetch.bs
```diff
@@ -3145,58 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+ <li><p>Set <var>forNavigation</var> to false if it is not given.
+
+ <li><p>Let <var>embedderPolicy</var> be <var>settingsObject</var>'s embedder policy.
+
+ <li>
+  <p>If the <a>cross-origin resource policy internal check</a> with <var>origin</var>,
+  "<code>unsafe-none</code>", <var>response</var>, and <var>forNavigation</var> returns
+  <b>blocked</b>, then return <b>blocked</b>.
+
+  <p class="note">This is to queue only COEP-related violation reports.
```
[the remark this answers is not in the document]

fetch.bs
```diff
@@ -3145,58 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+
+ <li><p>Return <b>blocked</b>.
+</ol>
+
+<p class="note no-backref">Only HTML's navigate algorithm uses this check with
+<var>forNavigation</var> set to true, and it's always for nested navigations. Otherwise,
+<var>response</var> is either the <a for=internal>internal response</a> of an
+<a>opaque filtered response</a> or a <a for=/>response</a> which will be the
+<a for=internal>internal response</a> of an
+<a>opaque filtered response</a>. [[HTML]]
```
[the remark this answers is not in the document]
yutakahirano: Done.

yutakahirano:
Updated: Other than addressing comments I started using the "coep" report type (without link, as it is defined in whatwg/html#5454) to address whatwg/html#5634.

```make deploy``` still shows an error when I use ```<a for="reporting">queue</a>```. Should I wait further or is there anything I should do?

File: fetch.bs [hunk 19]
```diff
@@ -3143,58 +3143,144 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+     <td>"<code>corp</code>"
+    </tr>
+    <tr>
+     <td>"<code>blocked-url</code>"
+     <td><var>serialized url</var>
+    </tr>
+   </tbody>
+  </table>
+
+ <li><p><a href="https://w3c.github.io/reporting/#queue-report">Queue</a> <var>body</var> as
```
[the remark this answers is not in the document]
annevk: Filed https://github.com/tabatkins/bikeshed/issues/1706 as it appears it's not indexed.

annevk:
@yutakahirano I filed an issue for that. Could you please resolve the comments here that have been resolved? @domenic do you want to take another look?

fetch.bs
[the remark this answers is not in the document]
annevk: So it is indexed in Shepherd, but https://github.com/w3c/reporting/commit/f31f654852f7db3dcae9c01570d0cf6e533a763c did not build. I now filed https://github.com/w3c/reporting/issues/213.

fetch.bs
```diff
@@ -3145,57 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+  </dl>
 </ol>
 
+<p>To <dfn>queue a cross-origin embedder policy CORP violation report</dfn>, given a
+<a for=/>response</a> <var>response</var>, an <a for=/>environment settings object</a>
+<var>settingsObject</var>, and a boolean <var>reportOnly</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>endpoint</var> be <var>settingsObject</var>'s embedder policy's
+ report only reporting endpoint if the <var>reportOnly</var> is true and
```
domenic, line 3220: These should cross-link to HTML, which won't work until the HTML PR is merged.

fetch.bs
```diff
@@ -3145,57 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+    </tr>
+    <tr>
+     <td>"<code>blocked-url</code>"
+     <td><var>serialized url</var>
+    </tr>
+   </tbody>
+  </table>
+
+ <li><p><a href="https://w3c.github.io/reporting/#queue-report">Queue</a> <var>body</var> as
+ the "<code>coep</code>"report type for <var>endpoint</var> on
```
domenic, line 3248:
```suggestion
 the "<code>coep</code>" report type for <var>endpoint</var> on
```

fetch.bs
```diff
@@ -3145,57 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+ <var>settingsObject</var>.
+ [[!REPORTING]]
+</ol>
+
+<p>To perform a <dfn export>cross-origin resource policy check</dfn>, given an <a for=url>origin</a>
```
domenic, line 3253: This algorithm being separate from the internal version is a bit confusing. I might make it the first thing in this section, instead?

fetch.bs
```diff
@@ -3145,57 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+<var>origin</var>, an <a for=/>environment settings object</a> <var>settingsObject</var>, a
+<a for=/>response</a> <var>response</var>, and an optional boolean <var>forNavigation</var>, run
+these steps:
+
+<ol>
+ <li><p>Set <var>forNavigation</var> to false if it is not given.
+
+ <li><p>Let <var>embedderPolicy</var> be <var>settingsObject</var>'s embedder policy.
```
domenic, line 3261: TODO cross-link when HTML is merged

fetch.bs
```diff
@@ -3145,57 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+
+ <li>
+  <p>If the <a>cross-origin resource policy internal check</a> with <var>origin</var>,
+  "<code>unsafe-none</code>", <var>response</var>, and <var>forNavigation</var> returns
+  <b>blocked</b>, then return <b>blocked</b>.
+
+  <p class="note">This is to queue only Cross-Origin Embedder Policy violation reports.
```
domenic, line 3268: I don't understand this note

fetch.bs
```diff
@@ -3145,57 +3148,146 @@ response <a for=/>header</a> can be used to require checking a <a for=/>request<
...
+ <li><p><a>Queue a cross-origin embedder policy CORP violation report</a> with <var>response</var>,
+ <var>settingsObject</var>, and false.
+
+ <li><p>Return <b>blocked</b>.
+</ol>
+
+<p class="note no-backref">Only HTML's navigate algorithm uses this check with
+<var>forNavigation</var> set to true, and it's always for nested navigations. Otherwise,
+<var>response</var> is either the <a for=internal>internal response</a> of an
+<a>opaque filtered response</a> or a <a for=/>response</a> which will be the
```
domenic, line 3288:
```suggestion
<a>opaque filtered response</a>, or a <a for=/>response</a> which will be the
```

fetch.bs
```diff
@@ -3586,9 +3678,8 @@ optionally with a <i>recursive flag</i>, run these steps:
   <p>If <var>internalResponse</var>'s <a for=response>URL list</a> <a for=list>is empty</a>, then
   set it to a <a for=list>clone</a> of <var>request</var>'s <a for=request>URL list</a>.
 
-  <p class=note>A <a for=/>response</a>'s <a for=response>URL list</a> will typically be empty at
-  this point, unless it came from a service worker, in which case it will only be empty if it was
-  created through <a lt="Response()" constructor><code>new Response()</code></a>.
+  <p class=note>A <a for=/>response</a>'s <a for=response>URL list</a> can be empty when it's for
+  "<code>about</code>" URLs for example.
```
domenic, line 3682: This change seems a bit out of place? And the phrasing is a bit weird. If we keep it then I'd say

> A response's URL list can be empty (for example, when the response represents an `about:` URL).

fetch.bs

fetch.bs
yutakahirano: Updated. Is it now clear?

fetch.bs

yutakahirano:
> Several TODOs around HTML cross-linking, but I'm not sure what's the latest on the sequencing of the PRs...

This PR heavily uses concepts defined in whatwg/html#5454. whatwg/html#5454 depends on two things in this PR.

 1. serialize a response URL for reporting.
 2. the modified parameters for the CORP check

It is possible to split this change to break the circular dependency, but I think it's just OK to land the HTML PR first because it seems OK to have a missing link there in terms of building.

domenic:
GitHub says this has conflicts with master, so is probably a good idea to rebase or merge it to make landing it easier.

fetch.bs
jugglinmike: The latest revision doesn't use @domenic's suggested phrasing.

fetch.bs
```diff
@@ -4513,13 +4614,12 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     </ol>
   </ol>
 
+ <li><p>Set <var>response</var>'s <a for=response>URL list</a> to a <a for=list>clone</a> of
+ <var>httpRequest</var>'s <a for=request>URL list</a>.
```
jugglinmike, line 4618: @yutakahirano I'm having trouble understanding the implications of this change. Would you mind explaining this a little?

fetch.bs
yutakahirano: Sorry, it seems I dropped the edit during merging. Fixed.

fetch.bs
yutakahirano: This PR changes a parameter of the CORP check. Currently it takes a request, but with this change it will take a response. This specific change is to set the correct URL list to the response so that the CORP check keeps working correctly.

annevk:
This looks good to me (and Domenic I think and presumably Yutaka), it has implementer support, and it has tests. Both Chrome and Firefox have implementation bugs, including for reporting. Should we file a general one on Safari for COOP and COEP?

Commit message seems reasonable (mostly deferring to HTML for information). We should probably mention that this changes "serialize a request URL for reporting".

If I understand it correctly the plan would be to merge HTML first, then wait a day, address the remaining integration issues as a result of Shepherd having new terms, and then merge this?

domenic:
> If I understand it correctly the plan...

Yep!

fetch.bs
```diff
@@ -4057,15 +4058,16 @@ optional <i>CORS-preflight flag</i>, run these steps:
   </ol>
 
  <li>
-  <p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>"
-  and the <a>cross-origin resource policy check</a> with <var>request</var>'s
-  <a for=request>origin</a>, <var>request</var>'s <a for=request>client</a>, and
-  <var>actualResponse</var> returns <b>blocked</b>, then return a <a>network error</a>.
+  <p>If either <var>request</var>'s <a for=request>response tainting</a> or <var>response</var>'s
```
yutakahirano, line 4061: I changed here because the previous logic didn't work well for responses coming from the service worker.

annevk:
Tentative commit message:
```
Integrate CORP and COEP

This adds support for cross-origin embedding policy, which primarily is a way of enforcing the Cross-Origin-Resource-Policy header to be set on responses.

See also this HTML change which links tests and the various issues and standards efforts involved here: https://github.com/whatwg/html/pull/5454.

The earlier added "serialize a request URL for reporting" has been replaced with "serialize a response URL for reporting" as centering things around responses was found more logical.
```
Any final feedback?

domenic:
LGTM!

fetch.bs
jugglinmike: Got it. Thank you!

PR #1036, opened 2020-06-11
Add style rules

yutakahirano:
Add style rules for new lines and names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,10 +108,9 @@ etc.
 ```html
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-redirect-mode>redirect mode</dfn>,...
-
-...
+```
+```html
 <p>Let <var>redirectMode</var> be <var>request</var>'s <a for=request>redirect mode</a>.
-...
 ```
 
 ## Merge policy
--- a/README.md
+++ b/README.md
@@ -80,29 +80,30 @@ is.
 End tags must be omitted when possible and attribute values must not be quoted (if necessary, use
 double quotes).
 
-Place one new line between paragraphs (including list elements). Place two or more new lines before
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,40 @@ is.
 End tags must be omitted when possible and attribute values must not be quoted (if necessary, use
 double quotes).
 
+Place one new line between paragraphs (including list elements). Place two or more new lines before
```
annevk, line 83: newline* throughout.

README.md
```diff
@@ -80,6 +80,40 @@ is.
...
+headings (e.g., `<h3>`). This doesn't apply when a nested heading follows the parent heading.
```
annevk, line 84: Headings depend on their heading level, but I've probably not always been consistent. `<h2>` gets three newlines, the remainder two. As the size of the document increases it might make sense to adjust this to give even more spacing for the larger sections.

README.md
```diff
@@ -80,6 +80,40 @@ is.
...
+
+<h3>Place two or more new lines above.</h3>
+
+<h4>Placing one new line is OK here.</h4>
+
+
+<h4>Place two or more new lines above.</h4>
+```
+
+Use camelCase for variable names. On the other hand, it is OK to use spaced names for defined
```
annevk, line 104: I would strike "it is OK to" as it's in fact the style to use spaced names there and it would probably not be okay to use camel case there.

README.md
yutakahirano: Done.

README.md
yutakahirano: Done.

README.md
yutakahirano: Done.

PR #1037, opened 2020-06-12
Meta: update repository files

foolip:
See https://github.com/whatwg/spec-factory for details.


annevk:
Here too it seems Travis CI is confused and GitHub Actions didn't do anything.

foolip:
Same comment as in https://github.com/whatwg/dom/pull/872#issuecomment-643359061 applies.

foolip:
Success: https://fetch.spec.whatwg.org/

Issue #950, opened 2019-10-09
Refreshing of stale response assumed to always be successful


gterzian:
Step 7.4 of https://fetch.spec.whatwg.org/#http-network-or-cache-fetch says to:
- If the revalidatingFlag is set and forwardResponse’s status is 304, then: 
    - Update storedResponse’s header list using forwardResponse’s header list, as per the       "Freshening Stored Responses upon Validation" chapter of HTTP Caching
    - Set response to storedResponse.

However, [Freshening Stored Responses upon Validation](https://tools.ietf.org/html/rfc7234#section-4.3.4) is subject to successfully selecting a stored response based on strong/weak/no validators found in the headers of the 304 response. 

Some cases contain a `MUST NOT` directive preventing the cache from using the 304 from updating stored responses, for example:

> If the new response contains a strong validator (see Section 2.1
      of [RFC7232]), then that strong validator identifies the selected
      representation for update.  All of the stored responses with the
      same strong validator are selected.  If none of the stored
      responses contain the same strong validator, then the cache MUST
      NOT use the new response to update any stored responses.

----

There are a few potential issues that can be identified:

1. Fetch assumes that the 304 ` forwardResponse` will update the previously set `storedResponse`, while in fact the cache might have to select a different stored resource for update, based on the validators found in the 304 response.
2. Step 4.2 always does a `Set response to storedResponse. `, while in fact the cache might not have been able to refresh `storedResponse`, or any other cached resource. 
3. If the cache fails to refresh any stored response based on the 304, it seems that fetch should do a "normal" network fetch of the resource, and such hook is not found in the algorithm. Instead step 5 says to `Set response to forwardResponse. `(if response is null), so if an implementation were to not update response as part of Step 4, due to a failure to select a resource for update, it would seem that the 304 would then be set to be the response. 

It would seem that a solution could be:

a. making Step 4 conditional upon selecting a resource for update based on the the 304 response, and setting `storedResponse` to the option result from that operation, 
b. and if unsuccesful, the algorithm should make a new network request that is not a validation request, and set `forwardResponse` to the result of HTTP-network fetch.

Please let me know what you think.

mnot:
SGTM

gterzian:
Ok I will follow-up with a PR

mnot:
I probably should have mentioned; for figuring out what to align to, you'll want to refer to [this](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#freshening.responses), not 7234.

[...]

PR #1041, opened 2020-06-20
Align spec more closely with cache freshening

gterzian:
FIX https://github.com/whatwg/fetch/issues/950

   * …
   * …
   * …
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: …
   * Firefox: …
   * Safari: …


gterzian:
@mnot I'm wondering if the test added in https://github.com/servo/servo/pull/26999 would be testing for the correct behavior?

gterzian:
@annevk I think this is probably worth some implementation bugs, just to check that this behavior is indeed what others are doing?

I'm not sure about "implementer interest", I think this simply aligns fetch with the http-caching spec, and I wouldn't be surprised if this already matches what implementations are doing.

annevk:
1. It seems this allows for reducing the effective scope of _storedResponse_. We should probably make the text reflect that.
2. And yeah, I think we should have some tests here to be sure. I also would not expect implementations to do a lot of this properly. E.g., combining of headers is one of the areas with known bugs (and perhaps some lack of clarity in the protocol as to which headers can be updated and such; see https://github.com/httpwg/http-core/issues/165).

PR #1044, opened 2020-06-26
Adapt to changes in Streams

### LINKED ISSUE #780: Fetch claims to define ReadableStream, but Streams is the actual location
Currently Fetch has a definition for ReadableStream, despite Streams having the actual definition. I suspect this predates use of Bikeshed, as it immediately has a biblio reference to Streams, and the "have a local definition but biblio-link to the real dfn's spec" was a common practice in The Bygone Days.

Can we remove this definition, or at least mark it as "local only" by giving it `lt="" local-lt="ReadableStream"`?

### LINKED ISSUE #976: ReadableStream and in-parallel/event-loop integration question
I have question regarding the use of streams in fetch, in the specific case of [`transmit the body`](https://fetch.spec.whatwg.org/#concept-request-transmit-body) of a request. 

So I think in all cases, when "transmit the body" is called, we are on "in-parallel" steps of the spec, hence I think we shouldn't be directly interacting with javascript objects, as described in https://html.spec.whatwg.org/multipage/#event-loop-for-spec-authors:in-parallel

So I'm wondering how, when transmitting the body of a request from parallel steps, the spec then says to:

1. [get a reader for the stream](https://fetch.spec.whatwg.org/#concept-get-reader)
2. Start [reading chunks](https://fetch.spec.whatwg.org/#concept-read-chunk-from-readablestream) from it until it is done or errored.

Wouldn't this require interacting with the corresponding javascript stream, which could consists of an arbitrary number of other streams piped/transformed into it?

For example, let's say the JS wants to upload some data to a server, which would be the end-result of a pipeline of streams. When "transmitting the body" of the request, we would effectively have to pull chunks from that pipeline and therefore run the corresponding javascript on a event-loop, correct? 

Would it perhaps be more correct to queue a task on the event-loop which would start reading chunks from the stream, and for each chunk queue "in-parallel" steps back on fetch that would transmit those chunks over the network?

----

Similarly, when [`http-network-fetch`](https://fetch.spec.whatwg.org/#http-network-fetch) [enqueues](https://fetch.spec.whatwg.org/#ref-for-concept-enqueue-readablestream%E2%91%A0) bytes to the response stream, shouldn't this also happen via a task queued on the event-loop(since otherwise the queuing is done immediately from parallel steps)?

Since enqueuing calls into the `strategySizeAlgorithm` of the controller, while that initial strategy is set by the user-agent [here](https://fetch.spec.whatwg.org/#ref-for-concept-construct-readablestream%E2%91%A0), if the response body passed to the JS via one of the queued fetch tasks has been piped by the JS to other streams, I guess their strategy size will be called into as well at that point(meaning we're interacting with JS objects from parallel steps)?

domenic:
Fixes #780. Fixes #976. Follows https://github.com/whatwg/streams/pull/1045.

(At least, I think this fixes #976. That's a long issue with lots of text, but this significantly reduces the amount of "in parallel" that happens in upload streaming, so it seems likely to address the problem.)

This should land after https://github.com/whatwg/streams/pull/1045 and needs some updates once it does, so tagging "do not merge yet". But it's ready for review.

The "looping" algorithms could potentially use better names; bikeshed help welcome.

***
<a href="https://whatpr.org/fetch/1044.html" title="Last updated on Jul 7, 2020, 4:34 PM UTC (b2777e3)">Preview</a> | <a href="https://whatpr.org/fetch/1044/b228c33...b2777e3.html" title="Last updated on Jul 7, 2020, 4:34 PM UTC (b2777e3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -9,6 +9,7 @@ Markup Shorthands: css off
 Translate IDs: typedefdef-bodyinit bodyinit,dictdef-requestinit requestinit,typedefdef-requestinfo requestinfo,enumdef-requestdestination requestdestination,enumdef-requestmode requestmode,enumdef-requestcredentials requestcredentials,enumdef-requestcache requestcache,enumdef-requestredirect requestredirect,dictdef-responseinit responseinit,enumdef-responsetype responsetype
 </pre>
 
+<!-- TODO remove whatpr.org streams links before merging -->
 <pre class=anchors>
 url:https://tools.ietf.org/html/rfc7230#section-3.1.1;text:method;type:dfn;spec:http
 url:https://tools.ietf.org/html/rfc7230#section-3.2;text:field-name;type:dfn;spec:http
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1673,53 +1682,67 @@ is to return the result of <a>serializing a request origin</a> with <var>request
...
-       <li>
-        <p>Transmit <var>bs</var>. Whenever one or more bytes are transmitted, increase
-        <var>body</var>'s <a for=body>transmitted bytes</a> by the number of transmitted bytes and
-        <a>queue a fetch task</a> on <var>request</var> to <a>process request body</a>
-        for <var>request</var>.
+       <p class="note no-backref">This step blocks until <var>bs</var> is fully transmitted.
 
-        <p class="note no-backref">This step blocks until <var>bs</var> is fully transmitted.
+      <li><p>If the ongoing fetch is not <a for=fetch>terminated</a>, then perform the
+      <a>transmit-body loop</a> given <var>request</var>, <var>body</var>, and <var>reader</var>.
```
ricea, line 1719: This seems like cheating. Don't we have to post a task to get back out of "in parallel"?

domenic: Yes, oops, I meant to do that.

domenic:
The corresponding streams PR has been merged.

domenic:
This is ready to merge, @annevk.

PR #1056, opened 2020-07-14
Return null if parsing a structured field value fails

### LINKED ISSUE #1053: Structured Field Values (formerly Structured Headers) changes
Reportedly SFV is the final name of https://tools.ietf.org/html/draft-ietf-httpbis-header-structure, though the RFC is not yet published.

This would mean we have to change

* https://fetch.spec.whatwg.org/#concept-header-list-get-structured-header
* https://fetch.spec.whatwg.org/#concept-header-list-set-structured-header

though maybe keep using the old IDs for some stability.

The other change @domenic and I have been contemplating is that we do not distinguish between no header present and header fails to parse for the entirety of the web platform. This would mean changing step 5 of the first algorithm to return null instead. I don't believe this affects any existing callers as they all already do so, but please let me know if you have any concerns.

Standards impacted by these naming changes are HTML, https://w3c.github.io/webappsec-fetch-metadata/, https://w3c.github.io/webappsec-feature-policy/, https://w3c.github.io/reporting/, ...? From a quick check they all treat null and failure as equivalent already.

cc @whatwg/http @clelland @mikewest @domenic

annevk:
All usage of structured fields in the web platform should not distinguish between the header not being there and failing to parse per the structured field parsing algorithm.

This also aligns Fetch slightly with terminology changes in HTTP that more clearly separate headers and trailers.

A future change might rename other concepts to also align with that (e.g., header list to field list).

Fixes #1053.

***
<a href="https://whatpr.org/fetch/1056.html" title="Last updated on Jul 15, 2020, 4:41 AM UTC (b15a670)">Preview</a> | <a href="https://whatpr.org/fetch/1056/801dc8b...b15a670.html" title="Last updated on Jul 15, 2020, 4:41 AM UTC (b15a670)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -75,9 +75,7 @@ url:https://tools.ietf.org/html/draft-ietf-httpbis-header-structure#section-4.2;
         "title": "HTTP proxy default configurations allow arbitrary TCP connections."
     },
     "EXPECT-CT": {
-        "authors": [
-            "Emily Stark"
-        ],
+        "authors": ["Emily Stark"],
         "href": "https://tools.ietf.org/html/draft-ietf-httpbis-expect-ct-02",
         "publisher": "IETF",
         "title": "Expect-CT Extension for HTTP"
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -397,28 +395,28 @@ these steps:
...
 
- <li><p>Let <var>result</var> be the result of executing the <a>parsing structured headers</a>
- algorithm with <var ignore>input_string</var> set to <var>value</var>, and
- <var ignore>header_type</var> set to <var>type</var>.
+ <li><p>Let <var>result</var> be the result of <a>parsing structured fields</a> with
+ <var ignore>input_string</var> set to <var>value</var> and <var ignore>header_type</var> set to
+ <var>type</var>.
 
- <li><p>If parsing failed, then return failure.
+ <li><p>If parsing failed, then return null.
```
domenic, line 402: I think it's worth an explicit note how the Fetch wrapper does not distinguish between absent and unparseable structured fields, and that this is intentional in order to encourage uniform processing across the web platform.

Issue #917, opened 2019-07-16
Double-keyed connection pools


sleevi:
Related to #904, the idea here is that user agents make connections conditionally based upon not just the destination origin, but on the initiating/'top level' origin.

For UAs that implement #904 for privacy-preserving reasons, failed to also implement the same for connection pools allows for the easy re-establishment of identifying information.

Consider a UA that sets [**credentials**](https://fetch.spec.whatwg.org/#concept-connection) to false for third-party requests, for privacy reasons, and allows **credentials** to be defined based on the [**credentials mode**](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) otherwise.

* When the UA visits https://a.example, and https://a.example attempts to load https://b.example/tracker, the parameters for the connection to `b.example` are `{ "https://b.example", false}`.
* When the UA visits https://b.example as a navigation, the parameters for the connection to `b.example` are `{"https://b.example", true}`.
* However, `b.example` can use the local storage it has (e.g. `LocalStorage`, cookies, etc) and can initiate a request with the fetch() **credentials mode** of `omit`, for `https://b.example/setUserId?id={some identifier}`. This connection will be initiated as `{"https://b.example", false}`.

If the **connection** is reused between the fetch in the context of `a.example` and `b.example`, then the information from `b.example` may be linkable on the server to the request in `a.example`, and may even be used to set responses or storage in the context of `a.example` (e.g. a response that `a.example` then uses to set an `a.example` cookie for `b.example`'s ID)

Broadly speaking, partitioning the HTTP cache, as described in #904, can not be reasoned to have any privacy-enhancing properties in the absence of also partitioning the **connection pool** on the same key. Any collapsing of key or state between the **fetch** and the **connection**, from something finer grained to something coarser, ensures that the finer-grained separation cannot be reasoned to actually work.

annevk:
I think it makes sense to do this for anti-tracking purposes and Firefox will likely do this. This should be relatively straightforward to add once #904 is fixed.

Note that the partitioning of the HTTP cache has another motivation, namely thwarting XSLeaks-style of attacks (except for those through popups, sites will need to use COOP to thwart those). Those seem slightly less applicable here, though perhaps timing attacks on connection setup are feasible? (There is also a connection exhaustion attack, due to there being a global limit.)

sleevi:
Correct. We are aware that you can use resource exhaustion - including at the OS level - to mount XSLeaks-style attacks. This does not mitigate such attacks.

[...]

PR #1063, opened 2020-07-28
Partition HTTP Connections

MattMenke2:
Extend the HTTP cache partitioning logic to partition connections as well. This doesn't cover everything that will need to be partitioned, since much of it isn't even capturing in the fetch spec yet (browser-managed DNS caches, WebSocket connections over H2/H3 proxies, H2/H3 sessions, TLS/H3 session resumption cache, cache of metadata about H3-capable servers, HTTP auth cache, etc).

This addresses parts of issue #917.

   * Chrome
   * FireFox
   * ...
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/963480
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1572544
   * Safari: …

***
<a href="https://whatpr.org/fetch/1063.html" title="Last updated on Aug 6, 2020, 2:51 PM UTC (8cbe436)">Preview</a> | <a href="https://whatpr.org/fetch/1063/4ad496d...8cbe436.html" title="Last updated on Aug 6, 2020, 2:51 PM UTC (8cbe436)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4837,7 +4837,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
    <dt>Otherwise
    <dd><p>Let <var>connection</var> be the result of
    <a lt="obtain a connection" for=connection>obtaining a connection</a>, <var>networkPartitionKey</var>,
-   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, and <var>credentials</var>.
+   given <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, and
+   <var>credentials</var>.
   </dl>
 
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -2066,20 +2066,19 @@ for each associated <a for="fetch group">fetch record</a> whose
 
 <h3 id=connections>Connections</h3>
 
-<p>A user agent has an associated
-<dfn export id=concept-connection-pool for=connection>connection pool</dfn>. A
-<a for=connection>connection pool</a> consists of zero or more
-<dfn lt=connection export id=concept-connection>connections</dfn>. Each
-<a>connection</a> is identified by a <b>key</b> (a <a>network partition key</a>), an
-<b>origin</b> (an <a for=/>origin</a>), and <b>credentials</b> (a boolean).
+<p>A user agent has an associated <dfn export id=concept-connection-pool>connection pool</dfn>. A
+<a>connection pool</a> consists of zero or more
+<dfn lt=connection export id=concept-connection>connections</dfn>. Each <a>connection</a> is
+identified by a <b>key</b> (a <a>network partition key</a>), an <b>origin</b> (an
+<a for=/>origin</a>), and <b>credentials</b> (a boolean).
 
-<p>To <dfn export id=concept-connection-obtain for=connection>obtain a connection</dfn>, given
-<var>key</var>, <var>origin</var>, and <var>credentials</var>, run these steps:
+<p>To <dfn export id=concept-connection-obtain>obtain a connection</dfn>, given a <var>key</var>,
+<var>origin</var>, and <var>credentials</var>, run these steps:
 
 <ol>
- <li><p>If <a for=connection>connection pool</a> contains a <a>connection</a> whose <b>key</b> is
- <var>key</var>, <b>origin</b> is <var>origin</var>, and <b>credentials</b> is <var>credentials</var>,
- then return that <a>connection</a>.
+ <li><p>If the user agent's <a>connection pool</a> contains a <a>connection</a> whose <b>key</b> is
+ <var>key</var>, <b>origin</b> is <var>origin</var>, and <b>credentials</b> is
+ <var>credentials</var>, then return that <a>connection</a>.
 
  <li><p>Let <var>connection</var> be null.
 
@@ -2107,8 +2106,8 @@ for each associated <a for="fetch group">fetch record</a> whose
    <li><p>Return failure.
   </ol>
 
- <li><p>Add <var>connection</var> to the <a for=connection>connection pool</a> with <b>key</b>
- being <var>key</var>, <b>origin</b> being <var>origin</var>, and <b>credentials</b> being
+ <li><p>Add <var>connection</var> to the user agent's <a>connection pool</a> with <b>key</b> being
+ <var>key</var>, <b>origin</b> being <var>origin</var>, and <b>credentials</b> being
  <var>credentials</var>.
 
  <li><p>Return <var>connection</var>.
@@ -2130,8 +2129,8 @@ clearly stipulates that <a>connections</a> are keyed on
 an <a>implementation-defined</a> value.
 
 <p>To
-<dfn lt="determine the network partition key|determining the network partition key">
-determine the network partition key</dfn>, given <var>request</var>, run these steps:
+<dfn lt="determine the network partition key|determining the network partition key">determine the network partition key</dfn>,
+given <var>request</var>, run these steps:
 
 <ol>
  <li><p>Let <var>topLevelOrigin</var> be null.
@@ -2162,13 +2161,13 @@ determine the network partition key</dfn>, given <var>request</var>, run these s
  <li><p>Let <var>topLevelSite</var> be the result of <a lt="obtain a site">obtaining a site</a>,
  given <var>topLevelOrigin</var>.
 
- <li><p>Let <var>secondKey</var> be null or an <a>implementation-defined</a> value.
+ <li>
+  <p>Let <var>secondKey</var> be null or an <a>implementation-defined</a> value.
 
   <p class=XXX>The second key is intentionally a little vague as the finer points are still
   evolving. See <a href=https://github.com/whatwg/fetch/issues/1035>issue #1035</a>.
 
  <li><p>Return (<var>topLevelSite</var>, <var>secondKey</var>).
-
 </ol>
 
 
@@ -2179,12 +2178,12 @@ determine the network partition key</dfn>, given <var>request</var>, run these s
 given <var>request</var>, run these steps:
 
 <ol>
- <li><p>Let <var>key</var> be the result of <a>determining the network partition key</a>.
+ <li><p>Let <var>key</var> be the result of <a>determining the network partition key</a> given
+ <var>request</var>.
 
  <li><p>If <var>key</var> is null, then return null.
 
- <li><p>Return the unique HTTP cache associated with the <var>key</var>.
- [[!HTTP-CACHING]]
+ <li><p>Return the unique HTTP cache associated with <var>key</var>. [[!HTTP-CACHING]]
 </ol>
 
 
@@ -4823,7 +4822,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
  <li><p>If <var>httpCache</var> is null, then set <var>request</var>'s <a for=request>cache mode</a>
  to "<code>no-store</code>".
 
- <li><p>Let <var>networkPartitionKey</var> be the result of <a>determining the network partition key</a>.
+ <li><p>Let <var>networkPartitionKey</var> be the result of
+ <a>determining the network partition key</a> given <var>request</var>.
 
  <li>
   <p>Switch on <var>request</var>'s <a for=request>mode</a>:
@@ -4836,9 +4836,9 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
    <dt>Otherwise
    <dd><p>Let <var>connection</var> be the result of
-   <a lt="obtain a connection" for=connection>obtaining a connection</a>, given
-   <var>networkPartitionKey</var>, <var>request</var>'s <a for=request>current URL</a>'s
-   <a for=url>origin</a>, and <var>credentials</var>.
+   <a lt="obtain a connection">obtaining a connection</a>, given <var>networkPartitionKey</var>,
+   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, and
+   <var>credentials</var>.
   </dl>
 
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2071,15 +2071,16 @@ for each associated <a for="fetch group">fetch record</a> whose
 <a for=connection>connection pool</a> consists of zero or more
 <dfn lt=connection export id=concept-connection>connections</dfn>. Each
 <a>connection</a> is identified by an <b>origin</b> (an
-<a for=/>origin</a>) and <b>credentials</b> (a boolean).
+<a for=/>origin</a>), <b>credentials</b> (a boolean), and
+<b>httpStatePartition</b> (a <a for=/>site</a> list or null).
```
MattMenke2, line 2075: I have no idea how to verify that this "for" link actually works.

MattMenke2:
@annevk @shivanigithub @domenic @mnot thoughts? Not sure if this is the best approach.

In terms of testing, it looks like BasicHttpServer doesn't exposed any sort of connection ID, but it does expose port. So it seems like the best we could do without touching BasicHttpServer itself is to make a request handler that exposes the client-side port each request is using, and make sure the same ports are used to service requests made by two top-level browsing contexts using different origins.  Haven't looked into the exact mechanics of how to share observed port data, but could just use link decoration with a list of previously observed ports, if there's nothing more elegant available.

MattMenke2:
I have no idea how to address those "make deploy" errors ("file:/home/runner/work/fetch/fetch/fetch.spec.whatwg.org/index.html":1958.5-1958.36: error: Element “aside” not allowed as child of element “dl” in this context. (Suppressing further errors from this subtree.))

I don't have access to index.html, so have no idea what's on line 1958. Running make deploy locally, it complains about GITHUB_REPOSITORY not existing, and setting it manually doesn't seem to address that. Looks like that's just for lines htat don't run, regardless, and the error appears to be on the step that's gated on the GITHUB_REPOSITORY step. Also, while running make deploy locally claims that it's generating index.html (".spec.whatwg.org/index.html"), I still don't have any idea how to access that file, to see what's at line 1958.

fetch.bs
```diff
@@ -4531,12 +4536,15 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     <var>httpRequest</var>'s
     <a for=request>credentials mode</a>.
 
-   <li><p>Set <var>httpCache</var> to the result of <a>determining the HTTP cache partition</a>,
+   <li><p>Set <var>httpStatePartition</var> to the result of <a>determining the HTTP state partition</a>,
    given <var>httpRequest</var>.
 
    <li><p>If <var>httpCache</var> is null, then set <var>httpRequest</var>'s
```
shivanigithub, line 4542: Since the httpCache is replaced by httpStatePartition above, <var>httpCache</var> here will not be a valid cache?

shivanigithub:
Regarding the deploy errors, I don't see an <aside> element in your changes which makes me wonder if the error exists without your changes as well?

shivanigithub:
Regarding the deploy errors, I don't see an \<aside\> element in your changes which makes me wonder if the error exists without your changes as well?

domenic:
Sorry about the deploy errors. That is a problem with the generated code via the build tool, and will be fixed by https://github.com/tabatkins/bikeshed/pull/1727.

fetch.bs
shivanigithub, line 2075: Does this mean that an httpStatePartition is a list of sites? May be add "key" to the name to be more clear that this is not the actual partition but the key "httpStatePartitionKey"?

fetch.bs
```diff
@@ -4796,7 +4804,7 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
  <li><p>Let <var>response</var> be null.
 
- <li><p>Let <var>httpCache</var> be the result of <a>determining the HTTP cache partition</a>, given
+ <li><p>Let <var>httpCache</var> be the result of <a>determining the HTTP state partition</a>, given
```
shivanigithub, line 4807: Let <var>httpCache</var> be the HTTP cache associated with the result of <a>determining the HTTP cache partition</a>

fetch.bs
MattMenke2: Thanks! Meant to put the line below this above it, but even that doesn't really handle the null case. Fixed.

fetch.bs
MattMenke2: Thanks! Meant to put the line below this above it, but even that doesn't really handle the null case. Fixed.

fetch.bs
MattMenke2: Done, everywhere (also "HTTP state partition" -> "HTTP state partition key")

fetch.bs
```diff
@@ -2123,11 +2125,11 @@ clearly stipulates that <a>connections</a> are keyed on
      WebSocket saner -->
 
 
-<h3 id=http-cache-partitions>HTTP cache partitions</h3>
+<h3 id=http-state-partition>HTTP state partition key</h3>
```
annevk, line 2128: Should we make this network or perhaps fetch partition key? Network might be better as we'll want to use it for things that do not go through Fetch as well, such as RTCDataChannel, right?

fetch.bs
```diff
@@ -4531,9 +4536,14 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
     <var>httpRequest</var>'s
     <a for=request>credentials mode</a>.
 
-   <li><p>Set <var>httpCache</var> to the result of <a>determining the HTTP cache partition</a>,
+   <li><p>Set <var>httpStatePartitionKey</var> to the result of <a>determining the HTTP state partition key</a>,
    given <var>httpRequest</var>.
 
+   <li><p>Set <var>httpCache</var> to null.
```
annevk, line 4542: How can it be non-null at this point?

fetch.bs
```diff
@@ -4796,8 +4806,8 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
 
  <li><p>Let <var>response</var> be null.
 
- <li><p>Let <var>httpCache</var> be the result of <a>determining the HTTP cache partition</a>, given
- <var>httpRequest</var>.
+ <li><p>Let <var>httpCache</var> be the result of <a>determining the HTTP state partition key</a>,
+ given <var>httpRequest</var>.
```
annevk, line 4810: This returns a key, not a cache. I think that this illustrates that we still want an easy way to obtain a partitioned cache though and not inline that in various places.

fetch.bs
MattMenke2: I assume we need to declare variables exist before conditionally setting them to something.  If this were psuedo-Javascript, it's undefined at this point, not null. If this is psuedo-C, it's a compile failure.

Anyhow, I've reverted this and added back a shortened HTTP cache partition section, per your other comment.

fetch.bs
MattMenke2: This would be just as accurate if this were null-checking checking the network partition key, and I just renamed httpCache here to networkPartitionKey - I just forgot to rename the variable.  That having been said, I've brought back the cache section, though it's pretty tiny now, and reverted this to directly reference the cache. That does seem less regression prone.

fetch.bs
annevk: There's a much earlier "Let _httpCache_ be null." step, no?

annevk:
@MattMenke2 you'll need to make your membership of the googlers organization public. Hopefully once https://github.com/tabatkins/bikeshed/pull/1742 is fixed the build issues will go away too.

fetch.bs
```diff
@@ -2123,11 +2125,11 @@ clearly stipulates that <a>connections</a> are keyed on
      WebSocket saner -->
 
 
-<h3 id=http-cache-partitions>HTTP cache partitions</h3>
+<h3 id=http-state-partition>Network partition key</h3>
```
annevk, line 2129: Let's add a definition here of a network partition key as a tuple consisting of a site and null or an implementation-defined value.

fetch.bs
```diff
@@ -2158,12 +2160,31 @@ given <var>request</var>, run these steps:
  <li><p>Let <var>topLevelSite</var> be the result of <a lt="obtain a site">obtaining a site</a>,
  given <var>topLevelOrigin</var>.
 
- <li>
-  <p>Return the HTTP cache associated with <var>topLevelSite</var> and, possibly, a second key.
-  [[!HTTP-CACHING]]
+ <li><p>Let <var>secondKey</var> be the second key, or null.
```
annevk, line 2163: Let's rephrase this to

> `be null or an <a>implementation-defined</a> value.`

to make it even more explicit.

fetch.bs
```diff
@@ -2158,12 +2160,31 @@ given <var>request</var>, run these steps:
...
 
   <p class=XXX>The second key is intentionally a little vague as the finer points are still
   evolving. See <a href=https://github.com/whatwg/fetch/issues/1035>issue #1035</a>.
+
+ <li><p>If <var>secondKey</var> is not null, return (<var>topLevelSite</var>, <var>secondKey</var>).
```
annevk, line 2168: Let's remove the conditional and the then unneeded step below. Implementations that do not have a second key can use a simpler data structure.

fetch.bs
```diff
@@ -2071,15 +2071,16 @@ for each associated <a for="fetch group">fetch record</a> whose
 <a for=connection>connection pool</a> consists of zero or more
 <dfn lt=connection export id=concept-connection>connections</dfn>. Each
 <a>connection</a> is identified by an <b>origin</b> (an
-<a for=/>origin</a>) and <b>credentials</b> (a boolean).
+<a for=/>origin</a>), <b>credentials</b> (a boolean), and
+<b>networkPartitionKey</b> (a <a for=/>site</a> list or null).
```
annevk, line 2075: I suggest we call this key and lead with it. So key, origin, credentials. And per below its type can be network partition key.

fetch.bs
```diff
@@ -2123,11 +2125,11 @@ clearly stipulates that <a>connections</a> are keyed on
      WebSocket saner -->
 
 
-<h3 id=http-cache-partitions>HTTP cache partitions</h3>
+<h3 id=http-state-partition>Network partition key</h3>
```
annevk, line 2128: The ID here needs updating and the title should be plural per convention.

fetch.bs
```diff
@@ -2158,12 +2160,31 @@ given <var>request</var>, run these steps:
...
+
+
+<h3 id=http-cache-partitions>HTTP cache partitions</h3>
+
+<p>To
+<dfn lt="determine the HTTP cache partition|determining the HTTP cache partition">determine the HTTP cache partition</dfn>,
+given <var>request</var>, run these steps:
+
+<ol>
+ <li><p>Set <var>networkPartitionKey</var> to the result of <a>determining the network partition key</a>.
```
annevk, line 2182: Use "Let" when initializing. See https://infra.spec.whatwg.org/#variables.

fetch.bs
```diff
@@ -2158,12 +2160,31 @@ given <var>request</var>, run these steps:
...
+
+ <li><p>If <var>networkPartitionKey</var> is non-null, return null.
```
annevk, line 2184: then return null.

But also, the conditional seems wrong?

fetch.bs
```diff
@@ -2158,12 +2160,31 @@ given <var>request</var>, run these steps:
...
+
+ <li><p>Otherwise, Return the unique HTTP cache associated with the <var>networkPartitionKey</var>.
```
annevk, line 2186: No need for otherwise.

fetch.bs
MattMenke2: Done. Renamed it in HTTP cache partitions section as well, but stuck with networkPartitionKey in the logic to find a connection, since that's a longer method.

fetch.bs
MattMenke2: Done.

fetch.bs
MattMenke2: Done.

fetch.bs
MattMenke2: You're correct - completely missed it, given how far up it is.

MattMenke2:
> @MattMenke2 you'll need to make your membership of the googlers organization public. Hopefully once [tabatkins/bikeshed#1742](https://github.com/tabatkins/bikeshed/pull/1742) is fixed the build issues will go away too.

My membership of the googlers organization isn't listed. Not sure how to resolve that.

MattMenke2:
> > 
> 
> My membership of the googlers organization isn't listed. Not sure how to resolve that.

Erm...To be clear, googlers doesn't appear in my organization list.

MattMenke2:
> > > 
> > 
> > 
> > My membership of the googlers organization isn't listed. Not sure how to resolve that.
> 
> Erm...To be clear, googlers doesn't appear in my organization list.

This should be taken care of (Thanks Domenic!).   It is rather unfortunate that this required me to hand over my personal phone number to github (I don't have a work phone, since I don't really need one).  :(

shivanigithub:
Looks good to me as well.

MattMenke2:
@annevk: Your changes all look good to me.

annevk:
Thanks all! (To be clear, I'm not blocking on tests as we don't have good infrastructure to test connections and I don't really see that changing anytime soon.)

MattMenke2:
> Thanks all! (To be clear, I'm not blocking on tests as we don't have good infrastructure to test connections and I don't really see that changing anytime soon.)

Thanks again, Anne! I do have some putative tests that I hope to send out this week or next. They're sufficiently weird that it's possible I'll face pushback on landing them, though. They track which tabs send requests to which server ports, with the server code assuming ports aren't reused, since the Python server class only exposes port information, not connection information, and doesn't expose when connections are closed.  I'll plan to go through review github instead of Chrome's magic mirror, to allow a broader set of folks to offer feedback or push back on the changes.

annevk:
Oh, that sounds great. I'll be out for week but happy to look at that mid-August.

PR #1069, opened 2020-08-06
Add examples for collect an HTTP quoted string

### LINKED ISSUE #1068: "get, decode, and split" seems incorrect
It's wrong for X-Content-Type-Options, but it seems wrong for other callers too. E.g., we don't want to support `Content-Type: "text/html"`.

What needs to happen is that this algorithm preserves the quotes in the output. It needs to be aware of them to ensure they are matched up with another quote, but it shouldn't end up stripping them.

annevk:
And remove a confusing note.

Closes #1068.

***
<a href="https://whatpr.org/fetch/1069.html" title="Last updated on Aug 6, 2020, 2:42 PM UTC (3e0b5da)">Preview</a> | <a href="https://whatpr.org/fetch/1069/44a94a6...3e0b5da.html" title="Last updated on Aug 6, 2020, 2:42 PM UTC (3e0b5da)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -323,7 +323,6 @@ and optionally an <var>extract-value flag</var>, run these steps:
    <td>"<code>"Hello" World</code>"
    <td>7
    <td>"<code>"Hello"</code>"
-   <td>"<code>"Hello"</code>"
    <td>"<code>Hello</code>"
   <tr>
    <td>"<code>"Hello \\ World\""</code>"
--- a/fetch.bs
+++ b/fetch.bs
@@ -307,9 +307,31 @@ and optionally an <var>extract-value flag</var>, run these steps:
  inclusive, within <var>input</var>.
 </ol>
 
-<p class="note no-backref">The <var>extract-value flag</var> argument makes this algorithm suitable
-for <a for="header list">getting, decoding, and splitting</a> and <a>parse a MIME type</a>, as well
-as other header value parsers that might need this.
+<div class=example id=example-http-quoted-string>
+ <table>
+  <tr>
+   <th>Input
+   <th>Final position (always starts at 0)
+   <th>Output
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -307,9 +307,30 @@ and optionally an <var>extract-value flag</var>, run these steps:
...
 </ol>
 
-<p class="note no-backref">The <var>extract-value flag</var> argument makes this algorithm suitable
-for <a for="header list">getting, decoding, and splitting</a> and <a>parse a MIME type</a>, as well
-as other header value parsers that might need this.
+<div class=example id=example-http-quoted-string>
+ <table>
+  <tr>
+   <th>Input
+   <th>Final position (always starts at 0)
```
domenic, line 314:
```suggestion
   <th>Final position (start position is always 0)
```

domenic: Although it might be better to just delete this column... it's not an output of the algorithm, and it's not clear to me what it's helping to illustrate.

annevk:
Should I put final position last?

fetch.bs
annevk: It's an "outparam", no?

domenic: Oh, right, that is important... hmm. Then I'd phrase it as something like "New position variable value", maybe?

Although I'm not sure readers are really going to manually count. It might be good to add a parenthetical value, e.g. 2 (end of string), 7 ( )? Or maybe a `<mark>` around the pointed-to-character? Meh, nothing works great there...

annevk: https://whatpr.org/fetch/1069.html#example-http-quoted-string looks reasonable to me. What do you think?

domenic: Yeah, that's pretty solid; ship it!

PR #1076, opened 2020-08-18
Set request header's guard if no-cors

### LINKED ISSUE #1074: The guard of headers created from no-cors request
In https://fetch.spec.whatwg.org/#dom-request, we set request's headers guard to "request" if _init_ is empty, even when created from a no-cors request. This allows web developers to modify headers that are not CORS-safelisted.

This seems like a spec bug.

See also: #560

@youennf @annevk @jakearchibald

jakearchibald:
Fixes #1074.

I think this is all it needs, because otherwise we want the header guard to be "request".

***
<a href="https://whatpr.org/fetch/1076.html" title="Last updated on Aug 18, 2020, 2:10 PM UTC (6fdbcde)">Preview</a> | <a href="https://whatpr.org/fetch/1076/1b1239a...6fdbcde.html" title="Last updated on Aug 18, 2020, 2:10 PM UTC (6fdbcde)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6385,6 +6385,23 @@ constructor steps are:
  <a for=Headers>guard</a> is "<code>request</code>".
 
  <li>
+  <p>If <a>this</a>'s <a for=Request>request</a>'s <a for=request>mode</a> is
+  "<code>no-cors</code>", then:
+
+  <ol>
+   <li><p>If <a>this</a>'s <a for=Request>request</a>'s <a for=request>method</a> is not a
+   <a>CORS-safelisted method</a>, then <a>throw</a> a {{TypeError}}.
+   <!-- This will throw in cases where APIs have created no-cors requests with a non-safe method and
+   the developer has called new Request(fetchEvent.request) or fetch(fetchEvent.request) within a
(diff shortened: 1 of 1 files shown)


annevk:
Do we still need 32.4.2 at this point? Perhaps we could run 32.4 in its entirety before 32?

jakearchibald:
Good point about the guard setting, but I'm less sure about "If this’s request’s method is not a CORS-safelisted method, then throw a TypeError." - are there cases where we have no-cors requests that use methods that we want to pass through, but don't want developers to be able to create themselves?

annevk:
Kind of seems like a bug if you can end up in that state as it would bypass the same-origin policy. I don't mind going this way for now though, but we should leave a source comment at least.

jakearchibald:
I've added a note about it.

annevk:
Sorry I explained myself badly. Unless we know that it actually happens I think it should be a comment in the source code for editors of this document. I'm pretty certain that you cannot get no-cors with a method other than GET/POST.

jakearchibald:
Hopefully I've understood this time 😄

annevk:
As this fixes a regression no tests or implementation bugs are needed.

PR #1079, opened 2020-08-20
Augment COEP violation report

yutakahirano:
- "blocked-url" is renamed to "blockedURL" (whatwg/html#5818).
 - Added "disposition" (whatwg/html#5391).
 - Added "destination" (whatwg/html#5391).

The CORP check now takes "destination" as a parameter.

See also: https://github.com/whatwg/html/pull/5848

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3232,9 +3232,9 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
 </code></pre>
 
 <p>To perform a <dfn export>cross-origin resource policy check</dfn>, given an <a for=url>origin</a>
-<var>origin</var>, an <a for=/>environment settings object</a> <var>settingsObject</var>, a
-<a for=/>response</a> <var>response</var>, and an optional boolean <var>forNavigation</var>, run
-these steps:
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -3350,10 +3351,13 @@ these steps:
  <var>settingsObject</var>'s <a for="environment settings object">embedder policy</a>'s
  <a for="embedder policy">reporting endpoint</a> otherwise.
 
- <li><p>Let <var>serialized url</var> be the result of
+ <li><p>Let <var>serializedURL</var> be the result of
  <a lt="serialize a response URL for reporting">serializing a response URL for reporting</a> with
  <var>response</var>.
 
+ <li><p>Let <var>disposition</var> be "<code>reporting</code>" if <var>reportOnly</var> is true,
+ and "<code>enforce</code>" otherwise.
```
annevk, line 3359:
```suggestion
 <li><p>Let <var>disposition</var> be "<code>reporting</code>" if <var>reportOnly</var> is true;
 otherwise "<code>enforce</code>".
```

annevk (commented):
This looks okay to me, modulo nit. Thanks for fixing it so quickly!

fetch.bs
annevk: FWIW, the background here is https://github.com/whatwg/infra/pull/250.

PR #1083, opened 2020-08-25
Allow a request body to be a byte sequence/string

### LINKED ISSUE #1073: Allow strings or byte sequences as bodies?
This is several issues in one...

HTML always invokes Fetch with byte sequence request bodies. The two invocations are:

- https://html.spec.whatwg.org/#form-submission-algorithm:concept-request-body
- https://html.spec.whatwg.org/#hyperlink-auditing:concept-request-body

I think we should allow this, instead of requiring HTML or other callers to create new `ReadableStream` instances. Fetch could, very early in the Fetch algorithm, [construct a fixed ReadableStream instance](https://fetch.spec.whatwg.org/#concept-construct-fixed-readablestream).

HTML also has a few places where it wants to supply a string as a response body in a synthetic response:

- https://html.spec.whatwg.org/#the-iframe-element:concept-response-body
- https://html.spec.whatwg.org/#navigating-across-documents:concept-response-body

These are interesting because I know that Blink, at least, has a direct path for javascript: URLs and iframe srcdoc to set the resulting document's value as a string. Bytes never get involved, and the charset defined for the document is ignored. (I don't know if the HTML spec defines the charset of the produced documents very well for those cases...)

I'm not sure what we should do here. Maybe HTML's navigate algorithm should grow a new type of "resource", the string resource, since this seems very specific to that case? Or HTML could UTF-8 encode and created a fixed stream itself. Maybe with a helper defined in Fetch, like "created a fixed string body"?

I can't think of an easy way for Fetch to handle this in general, because we probably want to confine string-bodies to these navigation special cases; certainly any cases that might pass through service workers need to be stream-bodies.

Finally, HTML also treats script response bodies as byte sequences:

- https://html.spec.whatwg.org/#fetching-scripts:concept-response-body and 3 other very similar places.

But this is already covered in https://github.com/whatwg/fetch/issues/661, so we can ignore it for this issue.

annevk:
This simplifies things for several callers.

Fixes #1073.

***
<a href="https://whatpr.org/fetch/1083.html" title="Last updated on Aug 26, 2020, 11:30 AM UTC (2e9e273)">Preview</a> | <a href="https://whatpr.org/fetch/1083/ad9a84a...2e9e273.html" title="Last updated on Aug 26, 2020, 11:30 AM UTC (2e9e273)">Diff</a>


annevk:
I think we could get rid of https://fetch.spec.whatwg.org/#concept-empty-readablestream too if we wanted to as it only has one caller and that might as well create a promise resolved with an empty byte sequence.

fetch.bs
```diff
@@ -1090,11 +1090,12 @@ outlawing <a>forbidden methods</a> and
...
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-body>body</dfn> (null or a
-<a for=/>body</a>). Unless stated otherwise it is null.
+<dfn export for=request id=concept-request-body>body</dfn> (null, a <a for=/>byte sequence</a>, a
+<a for=/>scalar value string</a>, or a <a for=/>body</a>). Unless stated otherwise it is null.
 
-<p class="note no-backref">This can be updated during redirects to null as described in
-<a>HTTP fetch</a>.
+<p class="note no-backref">A <a for=/>byte sequence</a> and a <a for=/>scalar value string</a> will
```
domenic, line 1096:
```suggestion
<p class="note no-backref">A <a for=/>byte sequence</a> or a <a for=/>scalar value string</a> will
```

fetch.bs
```diff
@@ -5759,7 +5769,7 @@ typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;</pre>
 
    <dt>{{ReadableStream}}
    <dd>
-    <p>If the <var>keepalive flag</var> is set, then <a>throw</a> a {{TypeError}}.
+    <p>If <a><var>keepalive</var></a> is true, then <a>throw</a> a {{TypeError}}.
```
domenic, line 5772:
```suggestion
    <p>If <var>keepalive</var> is true, then <a>throw</a> a {{TypeError}}.
```

Only call sites link; inside the algorithm we just treat them as variables.

annevk:
I was thinking that since the note links "safely extract" it's clear it'll be UTF-8 and we don't really do anything else for strings these days (and shouldn't).

domenic:
That's reasonable. However, I'd still err on the side of removing the string possibility until we have more than one concrete caller that would use it. It appears the only user would be reporting, which would not get much simpler, because it would just switch from "serialize JSON to bytes" to... I guess manually using %JSON.stringify%?

annevk:
Fair, I'll drop that for now. @yutakahirano opinions on removing the empty ReadableStream primitive?

fetch.bs
```diff
@@ -2489,8 +2477,16 @@ object <var>stream</var>, run these steps:
 </ol>
 
 <p>An <dfn export for=ReadableStream id=concept-empty-readablestream>empty</dfn> {{ReadableStream}}
-object is the result of <a lt="construct a fixed ReadableStream object">constructing</a> a fixed
-{{ReadableStream}} object with an empty list.
+object, run these steps:
```
annevk, line 2480: This needs rewording if we are to keep the "empty" construct.

yutakahirano:
I think we need to fix https://fetch.spec.whatwg.org/#dom-body-body too.

annevk:
@yutakahirano hey! What fix does that need? Do you think we need to preserve "empty"? Any other thoughts on this PR?

yutakahirano:
... or the body definition in https://whatpr.org/fetch/1083.html#request-class.

https://fetch.spec.whatwg.org/#concept-body-body expects the body to be a [body](https://fetch.spec.whatwg.org/#concept-body) or null, but with this change (concept) request's body is null or a [body](https://fetch.spec.whatwg.org/#concept-body) or a byte sequence.

annevk:
Isn't the byte sequence turned into a body well before the `Boxy` mixin gets hold of it? What about the other questions?

yutakahirano:
I'm not sure it is.

https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm creates a Request object from a concept request directly at step 21.3.2.

>
> Let requestObject be a new Request object associated with request and a new associated Headers object whose guard is "immutable".

I'm fine with removing https://fetch.spec.whatwg.org/#concept-empty-readablestream if it's unnecessary.

yutakahirano:
Ah I was wrong, that is covered by https://whatpr.org/fetch/1083/ad9a84a...ba5c0f3.html#fetching. Sorry for that.

fetch.bs
```diff
@@ -5710,17 +5716,13 @@ typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;</pre>
     <p>If <var>object</var>'s {{Blob/type}}
     attribute is not the empty byte sequence, set <var>Content-Type</var> to its value.
 
-    <p>Set <var>source</var> to <var>object</var>.
+   <dt><a for=/>byte sequence</a>
+   <dd><p><a for=ReadableStream>Enqueue bytes</a> given <var>object</var> and <var>stream</var>.
```
yutakahirano, line 5720: We need to close the stream below.

fetch.bs
```diff
@@ -5710,17 +5716,13 @@ typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;</pre>
...
-    <p><a for=ReadableStream>Enqueue</a> a <code>Uint8Array</code> object
-    wrapping an <code>ArrayBuffer</code> containing a copy of the bytes held by <var>object</var>
-    to <var>stream</var> and <a abstract-op>close</a>
-    <var>stream</var>. If that threw an exception,
-    <a abstract-op>error</a> <var>stream</var> with that exception.
-
-    <p>Set <var>source</var> to <var>object</var>.
+   <dd><p><a for=ReadableStream>Enqueue bytes</a> given a
+   <a lt="get a copy of the buffer source">copy of the bytes</a> held by <var>object</var> and
+   <var>stream</var>.
```
yutakahirano, line 5725: ditto

yutakahirano (changes requested):

yutakahirano (approved):
LGTM

PR #1085, opened 2020-08-27
Use new Streams algorithms

domenic:
Follows https://github.com/whatwg/streams/pull/1073.

***
<a href="https://whatpr.org/fetch/1085.html" title="Last updated on Sep 15, 2020, 9:33 AM UTC (b5d273c)">Preview</a> | <a href="https://whatpr.org/fetch/1085/cd38eca...b5d273c.html" title="Last updated on Sep 15, 2020, 9:33 AM UTC (b5d273c)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -92,6 +92,17 @@ url:https://tools.ietf.org/html/draft-ietf-httpbis-header-structure#section-4.2;
 }
 </pre>
 
+<pre class=link-defaults>
+spec:streams; type:dfn;
+    text:closed
+    text:errored
+    text:need more data
+    text:readable
+    text:disturbed
+    text:locked
+    text:cancel
+</pre>
+
 
 
 <h2 id=goals class="no-num short">Goals</h2>
--- a/fetch.bs
+++ b/fetch.bs
@@ -1024,7 +1024,7 @@ user-agent-defined <a for=header>value</a> for the
 <a for=/>body</a> <var>body</var>, run these steps:
 
 <ol>
- <li><p>Let «<var>out1</var>, <var>out2</var>» be the result of <a for=ReadableStream>teeing</a>
+ <li><p>Let « <var>out1</var>, <var>out2</var> » be the result of <a for=ReadableStream>teeing</a>
  <var>body</var>'s <a for=body>stream</a>.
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1024,8 +1024,8 @@ user-agent-defined <a for=header>value</a> for the
 <a for=/>body</a> <var>body</var>, run these steps:
 
 <ol>
- <li><p>Let «<var>out1</var>, <var>out2</var>» be the result of
- <a lt=tee for=ReadableStream>teeing</a> <var>body</var>'s <a for=body>stream</a>.
+ <li><p>Let «<var>out1</var>, <var>out2</var>» be the result of <a for=ReadableStream>teeing</a>
```
annevk, line 1027: Nit: should there be a space after « and before »?

fetch.bs
```diff
@@ -4990,20 +4801,20 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
          <li><p>If <var>bytes</var> is failure, then <a lt=terminated for=fetch>terminate</a> the
          ongoing fetch.
 
-         <li><p><a for=ReadableStream>Enqueue bytes</a> given <var>bytes</var> and
-         <var>stream</var>.
+         <li><p><a for=ReadableStream>Enqueue</a> a {{Uint8Array}} wrapping an {{ArrayBuffer}}
```
annevk, line 4804: Uint8Array object*
ArrayBuffer object*

And the same below. It's a little unfortunate this duplicates the creation of these objects again, but that's probably fine for now. At some point we probably want an operation that you give a byte sequence and a realm and it returns this.

annevk (commented):
Lovely! Many years in the making and we're finally almost there.

@yutakahirano would be great to get your take on this as well.

yutakahirano (approved):
Thank you! LGTM.

fetch.bs
domenic: I don't think adding the redundant "object" is a good idea. Ultimately it's your spec so the editorial conventions are up to you, but I wouldn't want to see that carry over to HTML, for example.

I agree that eventually we want something better here in Web IDL.

annevk: It doesn't seem entirely redundant to be specific about instance vs class.

PR #1106, opened 2020-11-02
Document how new protocols relate to port blocking

### LINKED ISSUE #956: Document bad port blocking rationale
In particular, it would be good to note that this does not (have to) apply to protocols that use TLS and ALPN.

annevk:
Closes #956.

cc @martinthomson

***
<a href="https://whatpr.org/fetch/1106.html" title="Last updated on Nov 3, 2020, 8:11 AM UTC (6502e1e)">Preview</a> | <a href="https://whatpr.org/fetch/1106/515e6bb...6502e1e.html" title="Last updated on Nov 3, 2020, 8:11 AM UTC (6502e1e)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2190,6 +2190,9 @@ given <var>request</var>, run these steps:
 
 <h3 id=port-blocking>Port blocking</h3>
 
+<p class=note>New protocols can avoid the need for blocking ports by negotiating the protocol
+through TLS. The protocol cannot be spoofed through HTTP requests in that case.
+
 <p>To determine whether fetching a <a for=/>request</a> <var>request</var>
 <dfn export lt="block bad port">should be blocked due to a bad port</dfn>,
 run these steps:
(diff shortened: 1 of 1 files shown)


ricea:
Drive-by lgtm.

ricea:
It may be more subtle than the comment implies. For example, SMTP/TLS could still be spoofed by treating it as an HTTPS server.

fetch.bs
```diff
@@ -2190,6 +2190,9 @@ given <var>request</var>, run these steps:
 
 <h3 id=port-blocking>Port blocking</h3>
 
+<p class=note>New protocols can avoid the need for blocking ports by negotiating the protocol
+through TLS. The protocol cannot be spoofed through HTTP requests in that case.
```
martinthomson, line 2194: I might cite RFC 7301 here.

martinthomson (approved):
I don't know what level of detail you are looking for.  This is certainly on the light side, but this is a good thing to say.

annevk:
@ricea thanks, in that case SMTP isn't negotiated though, is it? @martinthomson thanks, done.

ricea (approved):
Yes, protocols that require ALPN are safe (AFAIK). lgtm.

Issue #982, opened 2019-12-12
421 and upload streams


annevk:
Per #497 browsers implement 421. What does a 421 mean for an upload stream? Would we reveal the 421 to the caller after streaming completed?

Would clients that support early responses cancel the stream in addition?

cc @whatwg/http @yutakahirano

yoichio:
Since ReadableStream is not replayable, 421 handling should be referred to the web author.

I guess supporting early responses cancel is hard as same as bidirectional support given Chromium starts response handling after whole request body sending.
cc @MattMenke2

MattMenke2:
I'm not expert on the mechanics of sockets, but if we write and the socket is half-closed by the other end, without reading from it, my recollection is we tend to get connection resets, as opposed connection closed notifications.  This behavior is observed across all platforms.

Once Windows notifies a consumer of a connection reset, you can no longer read from the socket, even if there was a 421 response pending.  So just trying to read from a socket after uploading the request body fails, which would be much simpler than supporting full duplex sockets, won't give us a 421 response on Windows.  On other platforms, we do actually read from the body on connection resets of upload attempts, and return 4xx/5xx responses, if we get them.

Note that this only applies to H1, I'm not sure what H2/QUIC do in this case, or if they even have an analog of half-closed sockets.

yoichio:
Ping @annevk ?

annevk:
I don't really have a stake in the outcome, but when we pass it on to the web developer (and when we retry) needs to be clear.

yoichio:
If fetch upload failed, promise reject is called. If it succeeded and server returns 421, the web author gets 421.

annevk:
I meant in the specification. 😊

yoichio:
I don't understand yet what is special and unclear about 421 specification and the upload stream.
https://tools.ietf.org/html/rfc7540#section-9.1.2

annevk:
Well, the request cannot be retried and therefore the 421 is forwarded to the application, whereas in other cases retrying is allowed. (I haven't tested what browsers do with it in general though.)

yoichio:
I confirmed that both Chrome and Firefox retried the request once upon the 421 response for simple string fetch except
only Chrome prompted ```POST <url> 421 (Misdirected Request)``` error in devtool.

As [fetch() fails on redirect resolution](https://github.com/whatwg/fetch/issues/538#issuecomment-617563962), how about 
[...]

PR #1141, opened 2021-01-20
Add HTTP fetch step on 421.

### LINKED ISSUE #497: 421 Status Code
Should Fetch specify how to handle the 421 status code?

http://httpwg.org/specs/rfc7540.html#MisdirectedRequest

This is sort of like a redirect -- a client is supposed to automatically start a new connection and reissue the request, because the chosen connection isn't authoritative for it.

cc @mcmanus

yoichio:
  - Chrome
  - Firefox
-  [ ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   - https://github.com/web-platform-tests/wpt/pull/27322
Chrome: As-is for non-streaming request. https://crbug.com/1171011 for streaming request.
Firefox: As-is for non-streaming request. N/A as streaming request body has not been implemented
Safari: TBD for non-streaming request.  N/A as streaming request body has not been implemented

Fixes #497, #982

***
<a href="https://whatpr.org/fetch/1141.html" title="Last updated on Mar 5, 2021, 10:19 AM UTC (5efd769)">Preview</a> | <a href="https://whatpr.org/fetch/1141/51a664d...5efd769.html" title="Last updated on Mar 5, 2021, 10:19 AM UTC (5efd769)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2145,15 +2145,16 @@ identified by a <b>key</b> (a <a>network partition key</a>), an <b>origin</b> (a
 <a for=/>origin</a>), and <b>credentials</b> (a boolean).
 
 <p>To <dfn export id=concept-connection-obtain>obtain a connection</dfn>, given a <var>key</var>,
-<var>origin</var>, <var>credentials</var>, an optional boolean
-<dfn export for="obtain a connection"><var>http3Only</var></dfn> (default: false), and an optional
-boolean <dfn export for="obtain a connection"><var>dedicated</var></dfn> (default: false), run these
+<var>origin</var>, <var>credentials</var>, an optional boolean <var>bypassConnectionPool</var>
+(default false), an optional boolean
+<dfn export for="obtain a connection"><var>http3Only</var></dfn> (default false), and an optional
+boolean <dfn export for="obtain a connection"><var>dedicated</var></dfn> (default false), run these
 steps:
 <!-- http3Only and dedicated have been added for WebTransport -->
 
 <ol>
  <li>
-  <p>If <var>dedicated</var> is false, then:
+  <p>If <var>bypassConnectionPool</var> is false or if <var>dedicated</var> is false, then:
 
   <ol>
    <li><p>Let <var>connections</var> be a set of <a>connections</a> in the user agent's
@@ -4159,8 +4160,9 @@ steps. They return a <a for=/>response</a>.
 <h3 id=http-network-or-cache-fetch>HTTP-network-or-cache fetch</h3>
 
 <p>To <dfn id=concept-http-network-or-cache-fetch>HTTP-network-or-cache fetch</dfn>, given a
-<a for=/>fetch params</a> <var>fetchParams</var> and an optional boolean
-<var>isAuthenticationFetch</var> (default false), run these steps:
+<a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
+<var>isAuthenticationFetch</var> (default false), and an optional boolean
+<var>isNewConnectionFetch</var> (default false), run these steps:
 
 <p class=note>Some implementations might support caching of partial content, as per
 <cite>HTTP Range Requests</cite>. However, this is not widely supported by browser caches.
@@ -4525,7 +4527,7 @@ steps. They return a <a for=/>response</a>.
    "<code>only-if-cached</code>", then return a <a>network error</a>.
 
    <li><p>Let <var>forwardResponse</var> be the result of running <a>HTTP-network fetch</a> given
-   <var>httpFetchParams</var> and <var>includeCredentials</var>.
+   <var>httpFetchParams</var>, <var>includeCredentials</var>, and <var>isNewConnectionFetch</var>.
 
    <li><p>If <var>httpRequest</var>'s <a for=request>method</a> is
    <a href=https://tools.ietf.org/html/rfc7231#safe.methods>unsafe</a> and
@@ -4658,6 +4660,35 @@ steps. They return a <a for=/>response</a>.
    <var>fetchParams</var>.
   </ol>
 
+ <li>
+  <p>If the following are true
+
+  <ul class=brief>
+   <li><p><var>response</var>'s <a for=response>status</a> is 421
+   <li><p><var>isNewConnectionFetch</var> is false
+   <li><p><var>request</var>'s <a for=request>body</a> is null, or <var>request</var>'s
+   <a for=request>body</a> is non-null and <var>request</var>'s <a for=request>body</a>'s
+   <a for=body>source</a> is non-null
+  </ul>
+
+  <p>then:
+
+  <ol>
+   <li>
+    <p>If the ongoing fetch is <a for=fetch>terminated</a>, then:
+
+    <ol>
+     <li><p>Let <var>aborted</var> be the termination's aborted flag.
+
+     <li><p>If <var>aborted</var> is set, then return an <a>aborted network error</a>.
+
+     <li><p>Return a <a>network error</a>.
+    </ol>
+
+   <li><p>Set <var>response</var> to the result of running <a>HTTP-network-or-cache fetch</a> given
+   <var>fetchParams</var>, <var>isAuthenticationFetch</var>, and true.
+  </ol>
+
  <li><p>If <var>isAuthenticationFetch</var> is true, then create an <a>authentication entry</a> for
  <var>request</var> and the given realm.
 
@@ -4670,8 +4701,8 @@ steps. They return a <a for=/>response</a>.
 <h3 id=http-network-fetch>HTTP-network fetch</h3>
 
 <p>To <dfn id=concept-http-network-fetch>HTTP-network fetch</dfn>, given a <a for=/>fetch params</a>
-<var>fetchParams</var> and an optional boolean <var>includeCredentials</var> (default false), run
-these steps:
+<var>fetchParams</var>, an optional boolean <var>includeCredentials</var> (default false), and an
+optional boolean <var>bypassConnectionPool</var> (default false), run these steps:
 
 <ol>
  <li><p>Let <var>request</var> be <var>fetchParams</var>'s <a for="fetch params">request</a>.
@@ -4699,8 +4730,8 @@ these steps:
    <dt>Otherwise
    <dd><p>Let <var>connection</var> be the result of
    <a lt="obtain a connection">obtaining a connection</a>, given <var>networkPartitionKey</var>,
-   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, and
-   <var>includeCredentials</var>.
+   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>,
+   <var>includeCredentials</var>, and <var>bypassConnectionPool</var>.
   </dl>
 
  <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -2145,16 +2145,15 @@ identified by a <b>key</b> (a <a>network partition key</a>), an <b>origin</b> (a
 <a for=/>origin</a>), and <b>credentials</b> (a boolean).
 
 <p>To <dfn export id=concept-connection-obtain>obtain a connection</dfn>, given a <var>key</var>,
-<var>origin</var>, <var>credentials</var>, an optional boolean <var>bypassConnectionPool</var>
-(default false), an optional boolean
-<dfn export for="obtain a connection"><var>http3Only</var></dfn> (default false), and an optional
-boolean <dfn export for="obtain a connection"><var>dedicated</var></dfn> (default false), run these
-steps:
+<var>origin</var>, <var>credentials</var>, an optional boolean <var>forceNew</var> (default false),
+an optional boolean <dfn export for="obtain a connection"><var>http3Only</var></dfn> (default
+false), and an optional boolean <dfn export for="obtain a connection"><var>dedicated</var></dfn>
+(default false), run these steps:
 <!-- http3Only and dedicated have been added for WebTransport -->
 
 <ol>
  <li>
-  <p>If <var>bypassConnectionPool</var> is false or if <var>dedicated</var> is false, then:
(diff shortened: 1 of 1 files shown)


annevk:
Do we also want to do this if the service worker returns a 421 or only if the network returns one? We should probably have test coverage either way.

Also, I think we should cover what should happen for a non-streaming request.

yoichio:
> Do we also want to do this if the service worker returns a 421 or only if the network returns one? 

Both Chrome and Firefox retry only if the network returns 421. 

> Also, I think we should cover what should happen for a non-streaming request.

Chrome does this retry during "7.3. Set response to the result of making an HTTP request over connection using request" in 4.6. HTTP-network fetch.
How might we write it down ?

> We should probably have test coverage either way.

Sure.

> Could you also restore the pull request template?

Done.

annevk:
In https://fetch.spec.whatwg.org/#http-network-or-cache-fetch we have checks for 401 and 407. I think we should handle 421 there as well, in a similar manner. This means that if a service worker creates a 421 the API will see it. Is there any kind of limit on the number of retries?

yoichio:
Though fetch event is captured by a service worker or not, HTTP-network fetch retries only once.

yutakahirano:
The PR description is copied from elsewhere. Please update it.

yoichio:
> The PR description is copied from elsewhere. Please update it.

What kind of update is needed?

yoichio:
Updated #http-network-or-cache-fetch after 407 checking. PTAL.
I'm not sure how we can limit the retry only once.

yutakahirano:
https://crbug.com/1167013 is a different bug.

"N/A as streaming request body has not been implemented" is incorrect for this change because "we should cover what should happen for a non-streaming request."

yoichio:
Updated the title and the description:
- Add wpt pull-request
- Update  Implementation bugs

yutakahirano:
I think we should have a dedicated issue for Chromium (for streaming request).

yoichio:
Done (https://crbug.com/1171011: Fetch streaming upload should fail on a 421 (Misdirected Request) response).

yutakahirano:
The current logic leads to endless retry. Using another connection, which is required by the [RFC](https://tools.ietf.org/html/rfc7540#section-9.1.2), is also missing.

I'm not sure how in detail it should be specified, but at least some notes are needed. @annevk, thoughts?

annevk:
I think it's okay for the initial version to have something like 

> `User agents should have an <a>implementation-defined</a> recursion limit.`

as a separate paragraph below the recursive invocation. It's not immediately clear to me if this limit can be exploited in some way, but I guess we'll learn that over time...

annevk:
This will also fix #497 and we should note that in the final commit message.

fetch.bs
```diff
@@ -4593,6 +4593,31 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+    </ol>
+
+   <li>
+    <p>If <var>actualResponse</var>'s <a for=response>status</a> is 421,
+     <var>request</var>'s <a for=request>body</a> is non-null, and
+     <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is
+      null, then return a <a>network error</a>.
+
+   <li><p>Set <var>response</var> to the result of performing an <a>HTTP-network-or-cache fetch</a>
+   using <var>request</var>.
```
mnot, line 4618: The p
```suggestion
   using <var>request</var> on a new <a lt=connection>connection</a>.
```

The point of 421 is that the new request will be on a separate connection.

fetch.bs
```diff
@@ -4593,6 +4593,31 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+  <ol>
+   <li>
+    <p>If the ongoing fetch is <a for=fetch>terminated</a>, then:
+
+    <ol>
+     <li><p>Let <var>aborted</var> be the termination's aborted flag.
+
+     <li><p>If <var>aborted</var> is set, then return an <a>aborted network error</a>.
+
+     <li><p>Return a <a>network error</a>.
```
mnot, line 4608: I haven't followed the whole discussion, but is there a reason that 421 isn't just forwarded through? Surfacing it as a network error is opaque, and do will be difficult to debug / handle -- is there an underlying security issue here?

yoichio: I just followed the above 401 and 407 case handlings.
@annevk ,@yutakahirano, any thoughts?

yoichio:
Added recursion limit and updated the PR description including #497 .

fetch.bs
annevk: This could happen during the fetch initiated below if the initial fetch (that resulted in a 421) was explicitly aborted somehow. It seems safer to error in that case so aborting doesn't result in new information, but I'm not sure it matters a whole lot. I think I'd prefer keeping this as-is and revisiting if there are deployment issues. Note that implementations can still show the relevant information in DevTools and such.

fetch.bs
```diff
@@ -4593,6 +4593,33 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+   <li>
+    <p>If <var>actualResponse</var>'s <a for=response>status</a> is 421,
+     <var>request</var>'s <a for=request>body</a> is non-null, and
+     <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is
+      null, then return a <a>network error</a>.
+
+   <li><p>Set <var>response</var> to the result of performing an <a>HTTP-network-or-cache fetch</a>
+   using <var>request</var> on a new <a lt=connection>connection</a>.
+
+    <p class=XXX>User agents should have an <a>implementation-defined</a> recursion limit.
```
annevk, line 4620: This should be a normal paragraph, not an open issue.

annevk (commented):
Indentation could use some cleanup, but otherwise this looks good to me.

fetch.bs
yoichio: Fixed.

fetch.bs
mnot: OK. I was assuming that the common case here would be that the fetch aborts for some reason, and the initiating code wants to automatically re-fetch based upon request details that it has buffered. I suppose it can do so based upon the network error, but 421 conveys a little bit of additional information -- that the request definitely hasn't changed any server-side state.

fetch.bs
annevk: Oh I see, that seems important to call out indeed and might require more changes than this.

In particular this would need to invoke https://fetch.spec.whatwg.org/#concept-connection-obtain but add some argument to bypass the cache lookup I suppose.

yoichio: How about do HTTP-network fetch instead of HTTP-network-or-cache fetch to skip cache ?
Also since HTTP-network fetch includes obtaining a connection, "on a new connection" declaration is really needed here ?

annevk: I'm not talking about the HTTP cache, I'm talking about the connection pool "cache". This needs to ensure you get a new connection if I understand it correctly, not reuse an existing one. (It's probably fine to invoke HTTP-network fetch directly, but you'd still need to signal that it needs a fresh connection, perhaps with a new argument in a few places.)

yoichio: I see.
Added a description about the new connection.

annevk: I think we should address this in the way I suggested. Whereby obtain a connection has a new argument that forces the lookup to be bypassed so a new connection is returned. Trying to clarify this in the caller's prose doesn't make it clear that this is a feature an "obtain a connection" implementation needs to provide.

yoichio: O.K. Updated "obtain a connection" with new argument.

fetch.bs
```diff
@@ -4593,6 +4594,35 @@ Range Requests</cite>. [[HTTP-RANGE]] However, this is not widely supported by b
...
+     <var>request</var>'s <a for=request>body</a> is non-null, and
+     <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is
+      null, then return a <a>network error</a>.
+
+   <li>
+     <p>Set <var>response</var> to the result of performing an <a>HTTP-network-or-cache fetch</a>
+      using <var>request</var> on a new <a>connection</a>, which
+      is <a lt="obtain a connection">obtained</a> using true
+      <var>bypassConnectionPool</var>.
+      User agents should have an <a>implementation-defined</a> recursion limit.
```
annevk, line 4623: Similar to the above this needs to be a new argument to HTTP-network-or-cache fetch instead, right? So that it does the right thing when a connection is established.

fetch.bs
```diff
@@ -2073,10 +2073,11 @@ identified by a <b>key</b> (a <a>network partition key</a>), an <b>origin</b> (a
 <a for=/>origin</a>), and <b>credentials</b> (a boolean).
 
 <p>To <dfn export id=concept-connection-obtain>obtain a connection</dfn>, given a <var>key</var>,
-<var>origin</var>, and <var>credentials</var>, run these steps:
+<var>origin</var>, and <var>credentials</var>, with an optional boolean
+<var>bypassConnectionPool</var> (default false), run these steps:
 
 <ol>
- <li><p>If the user agent's <a>connection pool</a> contains a <a>connection</a> whose <b>key</b> is
+ <li><p>If <var>bypassConnectionPool</var> is false, the user agent's <a>connection pool</a> contains a <a>connection</a> whose <b>key</b> is
```
annevk, line 2080: is false and the user agent's*

fetch.bs
yoichio: Do you mean adding `bypassConnectionPool` argument to HTTP-network-or-cache fetch and HTTP-network fetch ?

annevk: Right, yeah. Let me know if you want help rebasing this btw.

fetch.bs
yoichio: Done.

fetch.bs
yoichio: Done.

fetch.bs
```diff
@@ -4575,6 +4579,37 @@ steps. They return a <a for=/>response</a>.
...
+     <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is
+      null, then return a <a>network error</a>.
+
+   <li>
+     <p>Set <var>bypassConnectionPool</var> to true.
+
+   <li>
+     <p>Set <var>response</var> to the result of running
+      <a>HTTP-network-or-cache fetch</a> given <var>fetchParams</var> and
+      <var>bypassConnectionPool</var>.
```
annevk, line 4609: This also needs to pass isAuthenticationFetch. Not sure what value it should be.

annevk (commented):
This looks good to me, except for the one missing parameter. I have some editorial nits I can push in a fixup commit.

fetch.bs
yoichio: Done. Just reused. BTW, no one call HTTP-network-or-cache fetch with isAuthenticationFetch parameter...

annevk: https://fetch.spec.whatwg.org/#ref-for-concept-http-network-or-cache-fetch%E2%91%A0 is the caller.

annevk: I pushed some formatting nits to this PR. I'd appreciate your review. I also still don't understand something about the tests and left a comment there. Apart from that I think this all turned out great.

yoichio:
I misunderstood column was in 80 characters instead of 100.
Thanks. LGTM.

annevk:
@yoichio I updated this per the discussion in https://github.com/httpwg/http-core/issues/789. In particular:

1. If you get a 421 while you do a stream upload, we just pass the 421 on to the caller.
2. If you get a 421 a second time, we just pass the 421 on the caller (and don't allow for another retry).

Are you willing to update the tests if that seems reasonable? I can probably also help out.

yoichio:
LGTM. I'll update the test.

annevk:
I rebased this again. So "obtain a connection" now has bypassConnectionPool and dedicated. bypassConnectionPool bypasses the connection pool on lookup, but still stores the result in the connection pool, and dedicated bypasses it completely. Should we organize this differently?

@martinthomson who had naming suggestions on the WebTransport PR might have insights here as well.

annevk:
One thought I had was to rename this parameter to _forceNew_ to make it clearer it only forces a new connection, nothing else.

yoichio:
_forceNew_ sounds good to me.

fetch.bs
```diff
@@ -2145,15 +2145,16 @@ identified by a <b>key</b> (a <a>network partition key</a>), an <b>origin</b> (a
...
+(default false), an optional boolean
+<dfn export for="obtain a connection"><var>http3Only</var></dfn> (default false), and an optional
+boolean <dfn export for="obtain a connection"><var>dedicated</var></dfn> (default false), run these
 steps:
 <!-- http3Only and dedicated have been added for WebTransport -->
 
 <ol>
  <li>
-  <p>If <var>dedicated</var> is false, then:
+  <p>If <var>bypassConnectionPool</var> is false or if <var>dedicated</var> is false, then:
```
yutakahirano, line 2157: We don't need the second "if".

fetch.bs
```diff
@@ -4658,6 +4660,35 @@ steps. They return a <a for=/>response</a>.
    <var>fetchParams</var>.
   </ol>
 
+ <li>
+  <p>If the following are true
+
+  <ul class=brief>
+   <li><p><var>response</var>'s <a for=response>status</a> is 421
```
yutakahirano, line 4667: Please place a newline below.

Please see https://github.com/whatwg/fetch#formatting for the style.
Ditto below.

fetch.bs
```diff
@@ -4658,6 +4660,35 @@ steps. They return a <a for=/>response</a>.
    <var>fetchParams</var>.
   </ol>
 
+ <li>
+  <p>If the following are true
```
yutakahirano, line 4664: I think

```
If all of the following are true
```

is better.

yutakahirano (approved):

annevk:
Thanks @yoichio and @yutakahirano. I'll merge this now and I'll file a bug against Safari so they can consider supporting 421 responses.

PR #1144, opened 2021-01-22
Tee request body on ServiceWorker interception

### LINKED ISSUE #572: Body should not be transmitted if service worker intercepts request

yutakahirano:
   * Chrome
   * Firefox
   * https://github.com/web-platform-tests/wpt/pull/27325
   * Chrome: https://crbug.com/1167013
   * Firefox: N/A as streaming request body has not been implemented
   * Safari: N/A as streaming request body has not been implemented

Fixes w3c/ServiceWorker#1560, fixes #572.

***
<a href="https://whatpr.org/fetch/1144.html" title="Last updated on Feb 17, 2021, 12:27 PM UTC (c13ef6f)">Preview</a> | <a href="https://whatpr.org/fetch/1144/3c90676...c13ef6f.html" title="Last updated on Feb 17, 2021, 12:27 PM UTC (c13ef6f)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3813,7 +3813,8 @@ these steps:
     <p>If <var>response</var> is not null, then:
 
     <ol>
-     <li><p><a>Transmit request body</a> given <var>fetchParams</var>.
+     <li>If <var>request</var>'s <a for=request>body</a> is not null, then
+     <a for=ReadableStream>cancel</a> <var>request</var>'s <a for=request>body</a> with undefined.
 
      <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
      <a>filtered response</a>, and to <var>response</var>'s
--- a/fetch.bs
+++ b/fetch.bs
@@ -3805,7 +3805,6 @@ these steps:
    <li><p>Let <var>requestForServiceWorker</var> be a <a for=request>clone</a> of
    <var>request</var>.
 
-
    <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
    <var>requestForServiceWorker</var>. [[!HTML]] [[!SW]]
 
@@ -3813,7 +3812,7 @@ these steps:
     <p>If <var>response</var> is not null, then:
 
     <ol>
-     <li>If <var>request</var>'s <a for=request>body</a> is not null, then
+     <li>If <var>request</var>'s <a for=request>body</a> is non-null, then
      <a for=ReadableStream>cancel</a> <var>request</var>'s <a for=request>body</a> with undefined.
 
      <li><p>Set <var>actualResponse</var> to <var>response</var>, if <var>response</var> is not a
--- a/fetch.bs
+++ b/fetch.bs
@@ -3806,27 +3806,26 @@ these steps:
    <a for=/>body</a>.
 
    <li>
-    <p>Let <var>copiedRequest</var> be a copy of <var>request</var>.
+    <p>Let <var>savedBody</var> be a copy of <var>request</var>'s <a for=request>body</a>.
 
     <p class="note no-backref">This step doesn't copy the contents of <a for=/>body</a>. If
-    non-null, <a lt=body for=request>bodies</a> of <var>request</var> and
-    <var>copiedRequest</var> contains references to the same <a for=body>stream</a>.
+    non-null, <var>request</var>'s <a for=request>body</a> and <var>savedBody</var>
+    contain references pointing to the same <a for=body>stream</a>.
 
    <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
-   <var>copiedRequest</var>. [[!HTML]] [[!SW]]
+   <var>request</var>. [[!HTML]] [[!SW]]
 
    <li>
-    <p>If <var>response</var> is null, <var>request</var>'s <a for=request>body</a> is
-    non-null, and its <a for=body>stream</a> is <a for=ReadableStream>disturbed</a> or
+    <p>If <var>response</var> is null, <var>savedBody</var> is non-null, and its
+    <a for=body>stream</a> is <a for=ReadableStream>disturbed</a> or
     <a for=ReadableStream>locked</a>, then:
 
     <ol>
-     <li><p>If <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is null,
-     then return a <a>network error</a>.
+     <li><p>If <var>savedBody</var>'s <a for=body>source</a> is null, then return a
+     <a>network error</a>.
 
      <li><p>Otherwise, set <var>request</var>'s <a for=request>body</a> to the first return value of
-     <a for=BodyInit>safely extracting</a> <var>request</var>'s <a for=request>body</a>'s
-     <a for=body>source</a>.
+     <a for=BodyInit>safely extracting</a> <var>savedBody</var>'s <a for=body>source</a>.
     </ol>
 
    <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -3802,11 +3802,18 @@ these steps:
   <p>If <var>request</var>'s <a>service-workers mode</a> is "<code>all</code>", then:
 
   <ol>
-   <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
-   <var>request</var>. [[!HTML]] [[!SW]]
-
    <li><p>Assert: <var>request</var> 's <a for=request>body</a> is either null or a
    <a for=/>body</a>.
+
+   <li>
+    <p>Let <var>copiedRequest</var> be a copy of <var>request</var>.
+
+    <p class="note no-backref">This step doesn't copy the contents of <a for=request>body</a>. If it
+    is non-null, <a lt=body for=request>bodies</a> of <var>request</var> and
+    <var>copiedRequest</var> reference to the same <a for=body>stream</a>.
+
+   <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
+   <var>copiedRequest</var>. [[!HTML]] [[!SW]]
 
    <li>
     <p>If <var>response</var> is null, <var>request</var>'s <a for=request>body</a> is
--- a/fetch.bs
+++ b/fetch.bs
@@ -3809,7 +3809,7 @@ these steps:
    <a for=/>body</a>.
 
    <li><p>If <var>response</var> is null, <var>request</var>'s <a for=request>body</a> is
-   non-null, its <a for=body>source</var> is null, and its <a for=body>stream</a> is
+   non-null, its <a for=body>source</a> is null, and its <a for=body>stream</a> is
    <a for=ReadableStream>disturbed</a> or <a for=ReadableStream>locked</a>, then return a
    <a>network error</a>.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -3802,31 +3802,12 @@ these steps:
   <p>If <var>request</var>'s <a>service-workers mode</a> is "<code>all</code>", then:
 
   <ol>
-   <li><p>Assert: <var>request</var> 's <a for=request>body</a> is either null or a
-   <a for=/>body</a>.
+   <li><p>Let <var>requestForServiceWorker</var> be a <a for=request>clone</a> of
+   <var>request</var>.
 
-   <li>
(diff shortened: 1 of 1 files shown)


yutakahirano:
@annevk @yoichio PTAL

annevk:
So what's the idea for non-streaming bodies that end up consumed by the service worker? It seems you decided to not clone the request for those scenarios?

yutakahirano:
Sorry you're right. The latest PR includes an explicit copy step.

annevk (commented):
This looks good I think, but unlike https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch we do not copy request here. Is request's body the only thing the service worker can affect? Is the headers list immutable? (I'm not saying this needs to change, but I'd like to ensure we have the right setup here.)

I'd also appreciate review from @wanderview and @jakearchibald.

yutakahirano:
> This looks good I think, but unlike https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch we do not copy request here. Is request's body the only thing the service worker can affect? Is the headers list immutable? (I'm not saying this needs to change, but I'd like to ensure we have the right setup here.)

The latest PR creates a copy of a request.

annevk:
Could you explain why you make a copy?

yutakahirano:
To make it clear that request mutation in service worker doesn't affect the network network.

annevk:
Yeah I was wondering if that was possible, but looking at https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm `Headers` should be immutable so I don't think we need it.

yutakahirano:
Reverted the change partially.

fetch.bs
```diff
@@ -3834,9 +3834,32 @@ optional <i>CORS-preflight flag</i>, run these steps:
   <p>If <var>request</var>'s <a>service-workers mode</a> is "<code>all</code>", then:
 
   <ol>
+   <li><p>Assert: <var>request</var> 's <a for=request>body</a> is either null or a
+   <a for=/>body</a>.
+
+   <li>
+    <p>Let <var>savedBody</var> be a copy of <var>request</var>'s <a for=request>body</a>.
```
jakearchibald, line 3841: What's the benefit of copying the body at this point?

jakearchibald (approved):
LGTM, but it isn't clear to me why the body is copied

fetch.bs
jakearchibald: Makes sense! Might be worth putting that in a note?

annevk: I'm not sure I understand. If you clone a request it uses https://fetch.spec.whatwg.org/#concept-body-clone to clone its body, right? This updates body's stream to a non-null value but its source remains the same. Couldn't we always extract from source again?

yutakahirano: The point I'm trying to make is that it overwrites the request's body's stream and that means request will not be disturbed nor locked. I'd like to fail the request in such a case too (if other conditions are met). This corresponds to "clone-and-ignore" tests in https://github.com/web-platform-tests/wpt/pull/27325.

annevk: So when you do a `fetch()` with a stream and then the service worker clones the `Request` and doesn't do anything else, the `fetch()` should be rejected? Why?

yutakahirano: The clone() logic runs in the service worker, so in order to make the request work we need to make the service worker alive until the end-of-stream. That is not desirable, given the developer chose the network fallback option.

annevk: And the service worker cannot obtain a lock and transfer it?

For the non-service worker case this is happening in https://fetch.spec.whatwg.org/#concept-request-transmit-body which seems to be invoked way too late as it should happen before `fetch()` has returned... This might also be a problem for responses?

yutakahirano: Sorry, I don't understand the last comment. Can you rephrase it?

annevk:
```js
const req = new Request("about:blank", { method: "POST", body: rs });
fetch(req);
req.text();
```
What causes the last line to throw? As far as I can tell `fetch()` doesn't obtain the reader for _rs_ until way too late.

yutakahirano: https://fetch.spec.whatwg.org/#fetch-method creates a new Request. That means the 39th step of https://fetch.spec.whatwg.org/#dom-request creates a proxy, which makes `req.body` locked and disturbed.

annevk: Thanks, I suspect we need to do something similar for passthrough stuff (though probably transfer instead). If the service worker doesn't touch a request with a stream, we do need to transfer it somehow as otherwise the service worker could later try to read from it. I don't think the service worker cloning it but not doing anything else should trigger any weird side effects.

yutakahirano: Thanks. Does that mean making [handle fetch](https://w3c.github.io/ServiceWorker/#handle-fetch) return a response __or a body__? That is similar to what we are planning to do in our implementation.

annevk: Yeah, that might be the way to do it. It's a little bit confusing as the body will be of the request so we it'll need some careful notes. It's also only needed when fetch handed the service worker a stream, right? Because only in that case fetch no longer has a body.

Although I suppose if you `clone()` and then read from the non-cloned request we might still want to fail with a network error even if fetch has a body whose source is a string or some such? Bit unclear to me what's best there.

yutakahirano: My gut feeling is that `req.clone()` should be treated as same as `req.text()`. Hence calling `req.clone()` and returning null in the service worker handler should mean network fallback when the body is null or the body has a non-null source, and that should result in a network error when the body is non-null and has a null source.

annevk: I disagree. It should be fine to clone a request for monitoring purposes or some such. Cloning should not result in failure all by itself.

yutakahirano: Hmm, what do you think about my previous comment?

> The clone() logic runs in the service worker, so in order to make the request work we need to make the service worker alive until the end-of-stream. That is not desirable, given the developer chose the network fallback option.

annevk: I don't understand it. Once fallback is chosen we obtain a lock on the stream (the proxy stuff) as per https://github.com/w3c/ServiceWorker/pull/1563 and the service worker returns null and the stream. Why does the service worker need to remain alive?

yutakahirano: Let's assume `req` has a body made from a ReadableStream, and it's intercepted in the service worker.

```
self.addEventListener('fetch', (e) => {
  const req = e.request;
  const clonedBody = req.clone();

  ...

  return null;
});
```

The above handler calls `clone()` and returns null. If we support this use case,

 1. [Tee](https://streams.spec.whatwg.org/#rs-tee) runs in the service worker realm,
 2. One of the streams made by [Tee](https://streams.spec.whatwg.org/#rs-tee) is transferred to the network process.

Hence shutting down the service worker shuts down the teeing logic, which breaks the transferred stream in the network process.

annevk: I'd love some input from @ricea @MattiasBuelens @domenic @jakearchibald. I see the problem and how this might be hard to optimize for, but it also seems very nonintuitive to web developers that merely cloning a request results in a network error. I personally don't think that's acceptable.

yutakahirano: (For people invited to the discussion, we are talking only about requests with a body made from a ReadableStream. Other requests such as a request with a blob body are out of scope.)

ricea: I agree that this behaviour is unintuitive, but given that `clone()` uses `Tee()` as its underlying implementation, I can't see another reasonable solution.

Keeping the service worker alive would be bad.

The tee algorithm is running on the service worker thread, so we can't use the new stream. The tee algorithm is tied to the JavaScript realm in which it was created, due to using promises. Even if the stream is transferred away, the algorithm still has to run on the original thread.

The original stream is read from when we call `Tee()` so there's no way to put it back the way it was before the call to `clone()`.

annevk: It seems that an implementation could use a different implementation though that would avoid keeping the service worker alive. The document that created the request stream (originalStream) is supplying the contents. The service worker gets that stream (requestStream), clones it, keeps one for itself (clonedStream) and hands the other back to the document that created the stream in the first place (requestStream -> originalStreamProxy). It seems an implementation could structure things such that at this point originalStreamProxy gets its contents directly from originalStream and originalStreamProxy gets them from originalStream as well.

I'm not exactly sure where the promises come in here, but it seems that for things that happen in the service worker there is only one realm to chose from and those in the document cannot use the realm from the service worker either.

jakearchibald: Just to check I'm understanding the above correctly:

1. The _originalRequest_ is sent to the service worker, and `event.request` represents that _originalRequest_.
2. `event.request.clone()` creates two new bodies by teeing, one for `event.request`, and one for the new cloned request.
3. The body associated with _originalRequest_ is now locked, due to the teeing process.

If the above is wrong, it's probably best to ignore the rest of this 😄 

---

We could clone the request (including teeing the body) before sending it to the service worker, giving us _originalRequest_ and _originalRequestForServiceWorker_.

1. _originalRequestForServiceWorker_ is sent to the service worker as `event.request`.
2. If `event.respondWith` is called, then _originalRequest_'s body can be closed. Otherwise, _originalRequestForServiceWorker_'s body can be closed (unless it's already locked).

I think this would allow cloning without increasing memory usage much. It would also mean you can call `event.request.text()` without calling `event.respondWith`, without creating a network error. I don't know if that's a feature or a bug.

MattiasBuelens: @jakearchibald I think that would work. 👍

I do agree that we should specify how passing a `Request` object between the original realm and the service worker's realm should work. It seems natural to make `Request` transferrable, or at least make sure that we [transfer the body's stream](https://streams.spec.whatwg.org/#rs-transfer).

annevk: I think if _originalRequest_ holds a stream, calling `event.request.text()` during the fetch event while not calling `event.repondWith()` should result in a network error still, even if we can make it work due to the additional teeing. It seems nicer to keep enforcing the "read once" rule as much as possible, even though we messed it up for non-stream backed requests in service workers.

yutakahirano: If we tee the request body before passing it to the service worker (it is hard to implement TBH - I think we will not be able to implement it in the first release), I would like to allow reading the body in the network fallback case. I'm uncomfortable with having magic in the clone operation, and that is more important to me than enforcing the "read once" rule. @annevk, what do you think?

yoichio: Why do yo prefer global "read once" rule to a stream body?

annevk: To give web developers an API that's as consistent as possible and also give implementations freedom to experiment with various memory optimization techniques.

yutakahirano: OK, now it clones the original request. I'm going to w3c/ServiceWorker#1563.

annevk:
@yutakahirano I don't see the closing of these streams as @jakearchibald suggested. Don't we need to do that? I guess it's not strictly speaking observable, but it seems somewhat bad to have a leak.

I like how this ended up being a lot simpler than some of the earlier approaches.

yutakahirano:
I added a cancel step at in https://github.com/w3c/ServiceWorker/pull/1563.

annevk:
Shouldn't we close _request_ here as well if we got a response back? We might also want to explain in a note what the setup is here.

yutakahirano:
> Shouldn't we close request here as well if we got a response back?

That is #572, I think (and yes, I think we should cancel it instead of transmitting).

annevk:
Yeah, I'm thinking that we should fix #572 as you suggested after all. In that only the network layer attempts to transmit chunks.

yutakahirano:
Done.

annevk (approved):
I think this is good now, I'd love if @ricea or @jakearchibald could do another round as well. Note that this goes for the tests as well: https://github.com/web-platform-tests/wpt/pull/27325. Thanks for all your work on this @yutakahirano!

annevk:
@yutakahirano can you write a tentative commit message that mentions the tests and closes the relevant issues?

jakearchibald (approved):
LGTM!

jakearchibald:
@yutakahirano thanks for all the work on this. I wouldn't have spotted the `request.clone()` issue.

yutakahirano:
Thank you!

PR #1148, opened 2021-01-29
Add port blocking for tftp, netbios-ns, snmp, rtsp, h323gatestat, h32…

mozfreddyb:
…3hostcall, pptp, sane-port

   * [blink intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/4Btz5xQ-gXc/m/iPDxYSEgAgAJ)
   * [Firefox](https://www.mozilla.org/en-US/security/advisories/mfsa2021-03/#CVE-2021-23961)
   * https://github.com/web-platform-tests/wpt/pull/27990
   * Chrome: See discussion.
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1677940
   * Safari: See discussion.

***
<a href="https://whatpr.org/fetch/1148.html" title="Last updated on Mar 10, 2021, 2:49 PM UTC (6601ba8)">Preview</a> | <a href="https://whatpr.org/fetch/1148/d070ea2...6601ba8.html" title="Last updated on Mar 10, 2021, 2:49 PM UTC (6601ba8)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2238,6 +2238,7 @@ run these steps:
  <tr><td>42<td>name
  <tr><td>43<td>nicname
  <tr><td>53<td>domain
+ <tr><td>69<td>tftp
  <tr><td>77<td>—
  <tr><td>79<td>finger
  <tr><td>87<td>—
@@ -2255,8 +2256,10 @@ run these steps:
  <tr><td>119<td>nntp
  <tr><td>123<td>ntp
  <tr><td>135<td>epmap
+ <tr><td>137<td>netbios-ns
  <tr><td>139<td>netbios-ssn
  <tr><td>143<td>imap
+ <tr><td>161<td>snmp
  <tr><td>179<td>bgp
  <tr><td>389<td>ldap
  <tr><td>427<td>svrloc
@@ -2271,6 +2274,7 @@ run these steps:
  <tr><td>532<td>netnews
  <tr><td>540<td>uucp
  <tr><td>548<td>afp
+ <tr><td>554<td>rtsp
  <tr><td>556<td>remotefs
  <tr><td>563<td>nntps
  <tr><td>587<td>submission
@@ -2278,12 +2282,16 @@ run these steps:
  <tr><td>636<td>ldaps
  <tr><td>993<td>imaps
  <tr><td>995<td>pop3s
+ <tr><td>1719<td>h323gatestat
+ <tr><td>1720<td>h323hostcall
+ <tr><td>1723<td>pptp
  <tr><td>2049<td>nfs
  <tr><td>3659<td>apple-sasl
  <tr><td>4045<td>npp
  <tr><td>5060<td>sip
  <tr><td>5061<td>sips
  <tr><td>6000<td>x11
+ <tr><td>6566<td>sane-port
  <tr><td>6665<td>ircu
  <tr><td>6666<td>ircu
  <tr><td>6667<td>ircu


annevk:
So Chrome did not block 554, does Firefox? @youennf does Safari?

mozfreddyb:
Thanks, I must have missed that. 
Firefox also blocks port 10080, which is omitted from this PR (but can be included if Safari will block it too).

youennf:
> So Chrome did not block 554, does Firefox? @youennf does Safari?

Safari is blocking it right now. @ricea, could you share the rationale for Chrome to not block 554?

> Firefox also blocks port 10080, which is omitted from this PR (but can be included if Safari will block it too).

We are not blocking it right now but can envision it. When did Firefox start blocking it? Is Chrome blocking it?

mozfreddyb:
Blocking for port 554 has landed in Firefox 84, which we released mid December.
Port 10080 has been blocked since Firefox 85, which we landed this Thursday (and we're currently getting feedback that port 10080 might be too popular. It might be too early to say though)

ricea:
I couldn't find a good reason to retain the block on 554 before, but just now I found https://www.juniper.net/documentation/en_US/junos/topics/topic-map/security-rtsp-alg.html which is worrying.

annevk:
So what will it be regarding 554 and 10080? Should we file a follow-up on those or do we want to wait a bit to see if we can come to agreement?

ricea:
I strongly think that 10080 should not be on the restricted list as we have significant evidence of use for HTTP servers, and the protocol is UDP-only so it's not exploitable via HTTP 1 or 2. My assumption is that QUIC provides sufficient guard against untrusted content that it can't be used to exploit UDP NAT algorithms.

I removed 554 from the restricted list for Chrome, but I no longer have confidence that it is safe. If it gets added here I will go through the normal intent process to see if we can block it without too much damage to users.

I support addressing these two ports in a follow-up.

youennf:
10080 might be best handled as a follow-up.

> I removed 554 from the restricted list for Chrome, but I no longer have confidence that it is safe. If it gets added here I will go through the normal intent process to see if we can block it without too much damage to users.

@ricea, would you prefer adding 554 to the list after Chrome intent process gets approved?
If so, it makes sense to also treat it as a follow-up.

ricea:
> @ricea, would you prefer adding 554 to the list after Chrome intent process gets approved?

Yes, but it looks like we are going to be inconsistent with other browsers either way, so it probably doesn't make much difference.

mozfreddyb:
(We'll currently consider port 554 and 10080 as blocked (it's possible we will re-evaluate if we see significant breakage))

@ricea you said the protocol on port 10080 is UDP only. I didn't see that on e.g., https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt which lists the amanda protocl as UDP and TCP. But it seems the typical linux kernel helper in netfilter/nf_conntrack_amanda.c is only working for UDP. I might be open to reconsider port 10080...

ricea:
I was incorrect about the protocol being UDP-only, sorry.

From my research, it appears that modern Amanda mostly uses TCP on port 10080 but historically Amanda used UDP. I haven't been able to find out whether the TCP Amanda protocol also embeds port numbers in the message. The only ALG implementation for Amanda I've been able to find is the Linux one, so it may be okay to consider TCP safe.

I've started gathering statistics on use of these ports and hope to have some preliminary results next week. I will share them when I have something.

mozfreddyb:
Thank you @ricea! Eaglery waiting for your stats. :)

ricea:
My statistics recording code still hasn't made it into the dev channel, so the numbers I have are only from canary. As a result, the absolute numbers are small and the population is not representative. Still, here's some preliminary results:

Port	| Ratio
------|-------
554|1%
1719|3%
1720|1%
1723|1%
6566|1%
10080|92%

Ports 69, 137, and 161 still haven't seen a single connection.

mozfreddyb:
So, we agree 10080 is the most popular port out of these listed. I suppose you also have numbers how much % of all requests fall into the whole group?

ricea:
I have sent a Intent to Ship to block port 554 in Chrome: https://groups.google.com/a/chromium.org/g/blink-dev/c/kyVo08TtOp8/m/nu4B94LcCAAJ.

Here's about two weeks worth of data from the Chrome dev channel:

Port                | Ratio
---------------|-----------
69                  | 0%
161                 | 0%
554                 | 5%
1719                | 1%
1720                | 0%
1723                | 0%
6566                | 0%
10080               | 93%

mozfreddyb:
Git is hard. Let me try and recover this pull request... :disappointed:

mozfreddyb:
It seems to me this PR has the necessary cross-browser agreement then?

youennf:
Safari does not block 10080 at the moment either.
@mozfreddyb, how is Firefox attempt to block 10080?

mozfreddyb:
We've blocked port 10080 since https://bugzilla.mozilla.org/show_bug.cgi?id=1677940, which landed in late November.
It went to release in [Firefox 85](https://www.mozilla.org/en-US/security/advisories/mfsa2021-03/#CVE-2021-23961) on January 26th. We saw some complaints and bugs filed, but suggested [this user-specific workaround](https://bugzilla.mozilla.org/show_bug.cgi?id=1689107#c6) if absolutely necessary.

I can undo my latest commit, if @youennf is also hesitant to block port 10080. OK for me, if we want to move forward without and then reconsider at a later stage.

youennf:
I would be tempted to land this without port 10080 since there is consensus for the other ports.
And either open a new issue specifically for 10080 or keep this one opened to continue making progress on 10080.

ricea (approved):

annevk:
From the discussion it seems we don't need implementation bugs here. Do we have test coverage?

annevk (approved):

PR #1149, opened 2021-01-29
Remove location URL as field on responses

### LINKED ISSUE #631: Clarify response's url list state when parsing a 3xx response's Location header value

### LINKED ISSUE #958: Editorial: refactor response's location URL so it's always present
Members that are conditionally present are hard to work with. We should just make it take several 3 string enum values or a URL and convert null and failure to 2 of those string enum values as we get a result from the header parser (the third value being the default).

### LINKED ISSUE #1146: Script-created redirects responses have no location URL
Edit: I misread the Fetch spec. See https://github.com/whatwg/fetch/issues/1146#issuecomment-769618653 and https://github.com/whatwg/fetch/issues/1146#issuecomment-769624814. Though instead we've identified an editorial change that is hopefully clearer and less complex to implement.

---

(CC @jakearchibald @wanderview who I talked to about this recently.)

Fetch responses have a location URL, computed from the Location header, when they're constructed from the network:
https://fetch.spec.whatwg.org/#ref-for-concept-response-location-url%E2%91%A0

And then the HTTP-redirect fetch algorithm never looks at the Location header and instead follows the location URL value. I also don't see any reference to setting the location URL in the Service Worker spec.
https://fetch.spec.whatwg.org/#concept-http-redirect-fetch

From what I can tell, the main platform-visible consequence of this that, if a Service Worker responds to a navigation FetchEvent of / with `ev.respondWith(fetch("/foo/", {redirect: "manual"})` and the server responds with `Location: bar.html`, the redirect resolves to /foo/bar.html rather than /bar.html.

Looks like Chrome and Firefox don't implement this (resulting in /bar.html) while Safari does. The spec's behavior seems preferable, especially given atomic HTTP redirect handling, so I'm looking at fixing that. In doing so, two fun cases came up:

* [Response.redirect](https://fetch.spec.whatwg.org/#dom-response-redirect) sets the Location header, but not the location URL. That means, if I understand the spec correctly, `respondWith(Response.redirect(...))` does not actually work. We could fix that by adding an extra step to the algorithm, but...
* It's also possible to construct a redirect with `new Response("", {status: 302: headers: {location: "bar.html"}})`. Since you can manipulate the headers after construction, there isn't a very clear place to stick in the location URL computation.

I can see two options here:

1. Add an extra step in `Response.redirect`, but leave `new Response(...302...)` broken.
2. Update HTTP-redirect fetch (or FetchEvent.respondWith) to say, if the response doesn't have a location URL, you rerun the steps to construct it from the status code and Location header and use that instead.

Interestingly, `Response.redirect` works in Chrome/Firefox/Safari, but `new Response(...302...)` only works in Chrome/Firefox. Safari seems to fail the fetch when you `respondWith(new Response(...302...))`. [Edit: it only fails for relative Location headers. Absolute Location headers are treated as redirects.]

(1) does mostly match one browser. I say mostly because I don't see anything in the spec that supports failing the fetch. It seems the spec would want you to actually commit the response, which is really weird. But, relative to Chrome/Firefox, it's a breaking change and I'm a little nervous about compatibility. It also seems generally confusing for `new Response(...302...)` to silently fail. (2) would avoid this.

I think (2) is thus preferable, even though repeating the Location resolution in two places is a little odd.

annevk:
It does not need to be stored on a response and therefore resulted in confusion.

Also clarify that synthetic responses need to have an absolute URL in the Location header field value (Response.redirect() does this automatically).

Corresponding HTML PR: TODO.

Tests: TODO.

Closes #631, closes #633, closes #958, and closes #1146. (Some of these can be closed due to #1030 making response's URL no longer null for network responses.)

   * Chrome
   * Firefox
   * https://chromium-review.googlesource.com/c/chromium/src/+/2665871
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1170379
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1690286
   * Safari: N/A

***
<a href="https://whatpr.org/fetch/1149.html" title="Last updated on Feb 2, 2021, 3:27 PM UTC (e63e41a)">Preview</a> | <a href="https://whatpr.org/fetch/1149/98f23db...e63e41a.html" title="Last updated on Feb 2, 2021, 3:27 PM UTC (e63e41a)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1879,16 +1879,6 @@ allowed on the resource fetched by looking at the flag of the response returned.
 the response of a redirect has to be set if it was set for previous responses in the redirect chain,
 this is also tracked internally using the request's <a for=request>timing allow failed flag</a>.
 
-<p>A <a for=/>response</a> can have an associated
-<dfn export for=response id=concept-response-location-url>location URL</dfn> (null, failure, or a
-<a for=/>URL</a>). Unless specified otherwise, <a for=/>response</a> has no
-<a for=response>location URL</a>.
-
-<p class="note no-backref">This concept is used for redirect handling in Fetch and in HTML's
-navigate algorithm. It ensures `<code>Location</code>` has
-<a lt="extracting header values">its value extracted</a> consistently and only once.
-[[!HTML]]
-
 <hr>
 
 <p>A <a for=/>response</a> whose
@@ -2005,6 +1995,36 @@ is a <a>filtered response</a> whose
 
 <p>A <dfn export id=concept-stale-response>stale response</dfn> is a <a for=/>response</a> that is
 not a <a>fresh response</a> or a <a>stale-while-revalidate response</a>.
+
+<hr>
+
+<p>The <dfn export for=response id=concept-response-location-url>location URL</dfn> algorithm of
+given a <a for=/>response</a> <var>response</var> is the following steps. They return null, failure,
+or a <a for=/>URL</a>.
+
+<ol>
+ <li><p>If <var>response</var>'s <a for=response>status</a> is not a <a>redirect status</a>, then
+ return null.
+
+ <li><p>Let <var>location</var> be the result of <a>extracting header list values</a> given
+ `<code>Location</code>` and <var>response</var>'s <a for=response>header list</a>.
+ <!-- https://github.com/whatwg/fetch/issues/814#issuecomment-431366126 -->
+
+ <li>
+  <p>If <var>location</var> is a <a for=header>value</a>, then set <var>location</var> to the result
+  of <a lt="url parser">parsing</a> <var>location</var> with <var>response</var>'s
+  <a for=response>URL</a>.
+
+  <p class=note>If <var>response</var> was constructed through the {{Response}} constructor,
+  <var>response</var>'s <a for=response>URL</a> will be null, meaning that <var>location</var> will
+  only parse successfully if it is an <a>absolute-URL-with-fragment string</a>.
+
+ <li><p>Return <var>location</var>.
+</ol>
+
+<p class=note>The <a for=response>location URL</a> algorithm is exclusively used for redirect
+handling in this standard and in <cite>HTML</cite>'s navigate algorithm which handles redirects
+manually. [[!HTML]]
 
 
 <h4 id=miscellaneous>Miscellaneous</h4>
@@ -3953,16 +3973,6 @@ optional <i>CORS-preflight flag</i>, run these steps:
 
     <p class=note>303 is excluded as certain communities ascribe special status to it.
 
-   <li><p>Let <var>location</var> be the result of <a>extracting header list values</a> given
-   `<code>Location</code>` and <var>actualResponse</var>'s <a for=response>header list</a>.
-
-   <li><p>If <var>location</var> is a <a for=header>value</a>, then set <var>location</var> to the
-   result of <a lt="URL parser">parsing</a> <var>location</var> with <var>actualResponse</var>'s
-   <a for=response>URL</a>.
-
-   <li><p>Set <var>actualResponse</var>'s
-   <a for=response>location URL</a> to <var>location</var>.
-
    <li>
     <p>Switch on <var>request</var>'s
     <a for=request>redirect mode</a>:
@@ -4003,18 +4013,14 @@ optional <i>CORS-preflight flag</i>, run these steps:
  <a>filtered response</a>, and <var>response</var>'s
  <a for=internal>internal response</a> otherwise.
 
- <li><p>If <var>actualResponse</var>'s <a for=response>location URL</a>
- is null, then return <var>response</var>.
+ <li><p>Let <var>locationURL</var> be <var>actualResponse</var>'s <a for=response>location URL</a>.
 
- <li><p>If <var>actualResponse</var>'s <a for=response>location URL</a>
- is failure, then return a <a>network error</a>.
- <!-- only Gecko does this; and even that is currently more complicated -->
+ <li><p>If <var>locationURL</var> is null, then return <var>response</var>.
 
- <li><p>If <var>actualResponse</var>'s
- <a for=response>location URL</a>'s
- <a for=url>scheme</a> is <em>not</em> an
- <a>HTTP(S) scheme</a>, then return a
- <a>network error</a>.
+ <li><p>If <var>locationURL</var> is failure, then return a <a>network error</a>.
+
+ <li><p>If <var>locationURL</var>'s <a for=url>scheme</a> is not an <a>HTTP(S) scheme</a>, then
+ return a <a>network error</a>.
 
  <li><p>If <var>request</var>'s <a for=request>redirect count</a> is
  twenty, return a <a>network error</a>.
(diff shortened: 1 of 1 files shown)


annevk:
cc @davidben @wanderview

fetch.bs
```diff
@@ -2006,6 +1996,34 @@ is a <a>filtered response</a> whose
...
+ <li><p>Let <var>location</var> be the result of <a>extracting header list values</a> given
+ `<code>Location</code>` and <var>response</var>'s <a for=response>header list</a>.
+ <!-- https://github.com/whatwg/fetch/issues/814#issuecomment-431366126 -->
+
+ <li>
+  <p>If <var>location</var> is a <a for=header>value</a>, then set <var>location</var> to the result
+  of <a lt="url parser">parsing</a> <var>location</var> with <var>response</var>'s
+  <a for=response>URL</a>.
+
+  <p class=note>If <var>response</var> was constructed in a service worker <var>response</var>'s
```
davidben, line 2017: Nit: comma after "service worker"?

fetch.bs
```diff
@@ -2006,6 +1996,34 @@ is a <a>filtered response</a> whose
...
+<ol>
+ <li><p>If <var>response</var>'s <a for=response>status</a> is not a <a>redirect status</a>, then
+ return null.
+
+ <li><p>Let <var>location</var> be the result of <a>extracting header list values</a> given
+ `<code>Location</code>` and <var>response</var>'s <a for=response>header list</a>.
+ <!-- https://github.com/whatwg/fetch/issues/814#issuecomment-431366126 -->
+
+ <li>
+  <p>If <var>location</var> is a <a for=header>value</a>, then set <var>location</var> to the result
```
domenic, line 2013: "extracting header list values" returns a list, not a value, if I am reading this right?

fetch.bs
```diff
@@ -2006,6 +1996,34 @@ is a <a>filtered response</a> whose
...
+  of <a lt="url parser">parsing</a> <var>location</var> with <var>response</var>'s
+  <a for=response>URL</a>.
```
domenic, line 2015: The URL parser takes an optional base URL, not a URL-or-null. So it'd be most correct to state this as something like "parsing _location_, with _response_'s URL if _response_'s URL is not null."

fetch.bs
```diff
@@ -2006,6 +1996,34 @@ is a <a>filtered response</a> whose
...
+
+  <p class=note>If <var>response</var> was constructed in a service worker <var>response</var>'s
+  <a for=response>URL</a> will be null, meaning that <var>location</var> will only parse
+  successfully if it is an <a>absolute-URL-with-fragment string</a>.
+
+ <li><p>Return <var>location</var>.
+</ol>
+
+<p class="note no-backref"><a>Extract a location URL</a> is used for redirect handling in Fetch and
+in HTML's navigate algorithm as it handles redirects manually. [[!HTML]]
```
domenic, line 2025: `<cite>` around names of standards?

domenic (commented):
IMO this should either be "extract a location URL" and take a header list (like "extract a MIME type"), or it should be "response's location URL" (like "response's/request's MIME type"). I prefer the latter.

davidben (commented):
Thanks for doing this!

fetch.bs
annevk: Yeah, but I'm not fixing this here.

fetch.bs
annevk: I plan to update the URL Standard to make the default more clear now that we have Infra language for it. I guess I might as well try to land that before this.

annevk:
A request never has a location URL so that seems kinda redundant.

fetch.bs
annevk: https://github.com/whatwg/url/pull/575

domenic:
I'm just suggesting changing the phrasing from "To **extract a location URL** given a response _response_" to "The **location URL** of a response _response_ is the value given by the following algorithm:". This lets the call sites be nicer, more like a "getter" instead of a "function call".

annevk:
@davidben if you have a changeset up for Chromium could you link it here? I would like to use that to satisfy the test criteria. I also take it https://github.com/web-platform-tests/wpt/pull/10449 can be closed in favor of that?

domenic (approved):

davidben:
> @davidben if you have a changeset up for Chromium could you link it here? I would like to use that to satisfy the test criteria. I also take it web-platform-tests/wpt#10449 can be closed in favor of that?

Here you are. Sorry about the delay. I tried to get a test working for relative script-constructed redirects. I... only sort of succeeded. It relies on network errors triggering iframe onload which seems to only be true in Chrome. (And, even then, It's Complicated.) Is there's a better trick for that sort of thing?
https://chromium-review.googlesource.com/c/chromium/src/+/2665871
https://chromium-review.googlesource.com/c/chromium/src/+/2665871/1/third_party/blink/web_tests/external/wpt/service-workers/service-worker/navigation-redirect-resolution.https.html

annevk:
It seems you could do a `fetch()` and assert it throws a TypeError, with a service worker intercepting it and returning a synthetic "faulty" redirect response (i.e., where `Location` is not an absolute URL, but everything else is in order)? (`iframe` not firing a `load` event for network errors is a known interop issue. Chrome's behavior there is being standardized upon.)

davidben:
Ah, yeah that'll work. Similar to how I previously misread it, I thought only navigations followed SW-constructed redirects, but that's not quite true since HTTP fetch step 3 still continues on to step 6, and step 3.2.3 only forbids opaqueredirect on non-manual requests. Synthetic ones are still fine. I'll switch it to that.

Issue #1061, opened 2020-07-27
A new Request(input, init) constructor step sounds weird.


yoichio:
https://fetch.spec.whatwg.org/#dom-request
A step 38th states:
"38. If inputBody is body and input is disturbed or locked, then throw a TypeError."

However If input is string, asking if it is disturbed or locked doesn't make sense because [disturbed](https://fetch.spec.whatwg.org/#concept-body-disturbed) and locked are defined only on a Body mixin.
It should be 
"38. If inputBody is body and a Request object, and input is disturbed or locked, then throw a TypeError." ?

annevk:
Well, _inputBody_ is only set to a non-null value if that is the case, no?

yoichio:
Do you mean the case is "If inputBody is body" ? No, it can be both inputBody is null and body is null.

yutakahirano:
Neither _body_ nor _inputBody_ can be a Request object. They are null or a [body](https://fetch.spec.whatwg.org/#concept-body). So 

```
inputBody is body and a Request object
```

 always returns false.

annevk:
But disturbed/locked also does a non-null check so that part is covered, no? (We could make step 38 step 39.0 I suppose and assert that _input_'s body is non-null, but would that make it better?)

yoichio:
@yutakahirano 
I missed. It should be
"If inputBody is body and input is a Request object, ..."

@annevk 
Disturbed/locked checks Body mixin's body is non-null but doesn't check Body's non-null.

annevk:
@yoichio I don't follow. A `Request` object's body is Body mixin's body. Body itself is a mixin that Request includes, it cannot be null.

yoichio:
> Body itself is a mixin that Request includes, it cannot be null.

Yes.

I feel discussion is going wrong. Back to the original question:
if _input_ is string and _init_["body"] doesn't exist, the 38th step is asking whether the _input_ is disturbed or locked.
That's strange.

annevk:
I see, so yeah, moving step 38 to 39.0 would be the right approach here as we need to know that _body_/_inputBody_ are non-null.

annevk:
I pushed a fix for this to #1152.

Issue #954, opened 2019-10-29
It is hard read this section because of 4 "body"s and reassigning each other.


yoichio:
https://fetch.spec.whatwg.org/commit-snapshots/ae6435fdfd08f10833bbe65bd0ba559192baa330/#request-class

> Let inputBody be input’s request’s body if input is a Request object, and null otherwise\.
> 
> If either init\["body"\] exists and is non\-null or inputBody is non\-null, and request’s method is \`GET\` or \`HEAD\`, then throw a TypeError\.
> 
> Let body be inputBody\.
> 
> If init\["body"\] exists and is non\-null, then:​
> 
> Let Content\-Type be null\.
> 
> If init\["keepalive"\] exists and is true, then set body and Content\-Type to the result of extracting init\["body"\], with the keepalive flag set\.
> 
> Otherwise, set body and Content\-Type to the result of extracting init\["body"\]\.
> 
> If Content\-Type is non\-null and r’s headers’s header list does not contain \`Content\-Type\`, then append \`Content\-Type\`/Content\-Type to r’s headers\.
> 
> If body is non\-null and body’s source is null, then:​
> 
> If r’s request’s mode is neither "same\-origin" nor "cors", then throw a TypeError\.
> 
> Set r’s request’s use\-CORS\-preflight flag\.
> 
> If inputBody is body and input is disturbed or locked, then throw a TypeError\.
> 
> If inputBody is body and inputBody i\.\.\.

yoichio:
How about introducing more constant variables?

annevk:
I'm not opposed to a refactoring of this. Might also help #933.

PR #1152, opened 2021-01-30
Editorial: rename keepalive flag to keepalive

annevk:
This simplifies some algorithms and helps with #954.

***
<a href="https://whatpr.org/fetch/1152.html" title="Last updated on Feb 2, 2021, 2:20 PM UTC (a75ffa6)">Preview</a> | <a href="https://whatpr.org/fetch/1152/987c3fd...a75ffa6.html" title="Last updated on Feb 2, 2021, 2:20 PM UTC (a75ffa6)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1136,8 +1136,9 @@ requests, but not service worker requests. It references an
 explicitly set <a for=/>request</a>'s
 <a for=request>window</a>.
 
-<p id=keep-alive-flag>A <a for=/>request</a> has an associated
-<dfn for=request export>keepalive flag</dfn>. Unless stated otherwise it is unset.
+<p id=keep-alive-flag>A <a for=/>request</a> has an associated boolean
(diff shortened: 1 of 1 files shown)


annevk:
@yoichio is the second commit here what you had in mind?

yoichio:
LGTM for me to fix #1061.

annevk:
Thanks @yoichio! Does the variable name refactoring help with #954?

yoichio:
Yes, but I have 2 requests.
1. Use `inputOrInitBody`  and `InitBody` rather than `ExtractedBody` to emphasize the later comes from `input`.
2. Avoid using `extractedBody` at the step 41 after introducing `inputOrExtractedBody` because it is "consumed". How about move this step into 38?

annevk:
We cannot move all of 41 to 38 as that would move a step that should happen after all throwing, before some of the throwing. I will look into the names though.

annevk:
@yoichio if you could take another look that'd be great.

yoichio:
Thanks. LGTM.

PR #1153, opened 2021-01-30
Change default of request's credentials mode

### LINKED ISSUE #804: "Set fallbackCredentials to "same-origin"."
There are two lines that seem to be at odds with each other.
By default a request should be using "omit" as the credentials mode  but in the Response Constructor, a fallback of "same-origin" should be used.

Should these two lines sync up to say the same thing?

https://fetch.spec.whatwg.org/commit-snapshots/4321ab66c7ecc73793229939eaaeb811333e6cab/#keep-alive-flag

> A request has an associated credentials mode, which is "omit", "same\-origin", or "include"\. Unless stated otherwise, it is "omit"\.

https://fetch.spec.whatwg.org/commit-snapshots/4321ab66c7ecc73793229939eaaeb811333e6cab/#request-class

> Set fallbackCredentials to "same\-origin"\.

annevk:
This only affects specifications that do not set credentials mode and I'm not aware of any.

Closes #804.

***
<a href="https://whatpr.org/fetch/1153.html" title="Last updated on Jan 30, 2021, 11:21 AM UTC (587b2aa)">Preview</a> | <a href="https://whatpr.org/fetch/1153/d070ea2...587b2aa.html" title="Last updated on Jan 30, 2021, 11:21 AM UTC (587b2aa)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1423,7 +1423,7 @@ one or more event listeners are registered on an {{XMLHttpRequestUpload}} object
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-credentials-mode>credentials mode</dfn>,
 which is "<code>omit</code>", "<code>same-origin</code>", or
-"<code>include</code>". Unless stated otherwise, it is "<code>omit</code>".
+"<code>include</code>". Unless stated otherwise, it is "<code>same-origin</code>".
 
 <div class="note no-backref">
  <dl>
@@ -5950,8 +5950,6 @@ constructor steps are:
 
  <li><p>Let <var>fallbackMode</var> be null.
 
- <li><p>Let <var>fallbackCredentials</var> be null.
-
  <li><p>Let <var>baseURL</var> be <a>this</a>'s <a>relevant settings object</a>'s
  <a for="environment settings object">API base URL</a>.
 
@@ -5974,8 +5972,6 @@ constructor steps are:
    <a for=request>url</a> is <var>parsedURL</var>.
 
    <li><p>Set <var>fallbackMode</var> to "<code>cors</code>".
-
-   <li><p>Set <var>fallbackCredentials</var> to "<code>same-origin</code>".
   </ol>
 
  <li>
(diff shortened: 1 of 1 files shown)


domenic:
Many parts of HTML fail to set credentials mode, I think? E.g. a bunch of navigation ones:

> Let resource be a new request whose url is url and whose referrer policy is the current state of element's referrerpolicy content attribute.

> If destination is not a request, then set destination to a new request whose URL is destination.

> Plan to navigate to a new request whose url is parsed action, method is method, header list consists of `Content-Type`/MIME type, and body is body.

> Let request be a new request whose URL is urlRecord.

(and more I skipped over)

importScripts():

> Let request be a new request whose url is url, client is settings object, destination is "script", parser metadata is "not parser-inserted", synchronous flag is set, and whose use-URL-credentials flag is set.

Downloads:

> Let request be a new request whose url is URL, client is entry settings object, initiator is "download", destination is the empty string, and whose synchronous flag and use-URL-credentials flag are set.

I also found service worker script imports:

> Let importRequest be a new request whose url is importUrl, client is job’s client, destination is "script", parser metadata is "not parser-inserted", synchronous flag is set, and whose use-URL-credentials flag is set.

annevk:
None of those should use "omit".

domenic:
Sure, I wasn't really questioning the change, just the commit message. Although it might be worth using this opportunity as a forcing function to fix HTML, at least.

annevk:
What would we want in HTML? Set credentials mode less I suppose as the default already handles it?

domenic:
Well navigate (and downloads?) seems like it should always be include? We could force that inside the navigate algorithm, but that seems hacky; I'd almost prefer updating all the call sites then adding an assert inside the navigate algorithm.

I'm honestly not sure what browsers do for importScripts(). If it's same-origin then omitting it seems OK I guess.

Alternately we could try to remove the default entirely and make all callers consider this? Hmm, not sure...

annevk:
Oh right, yeah, some of those need "include", at which point that would be a pretty significant undertaking with regards to tests. I don't want to take that on right now.

Issue #777, opened 2018-07-12
fetch() is unclear on which realm's Request it uses


domenic:
https://fetch.spec.whatwg.org/commit-snapshots/e6cbef27724dd6111d1646898ef4f3f9ad56740b/#fetch-method

> Let requestObject be the result of invoking the initial value of Request as constructor with input and init as arguments\. If this throws an exception, reject p with it and return p

Which realm does the Request constructor come from?

I think we want it to be the relevant realm of this WindowOrWorkerGlobalScope object. That seems most in line with our plan in https://github.com/heycam/webidl/issues/135, although that is about slightly different phrasing ("a new Response" vs. this line's "invoking the initial value of Request as constructor").

BTW, editorial issues with this sentence:

- "as constructor" -> "as a constructor"
- Missing period
- "Request" links to the class, but should probably link to the constructor.

Found by @benjamingr and @TimothyGu in IRC: https://freenode.logbot.info/whatwg/20180712#c1621394

TimothyGu:
(For those who don't want to read IRC logs, all of this is observable through the `Referer` header and using the `fetch()` function from an `iframe` for example.)

We probably don't want to use the word "invoking" here at all, since it can mean multiple things: the ES Call abstract operation; [invoke](https://heycam.github.io/webidl/#invoke-a-callback-function) a callback function algorithm (this is probably the closest, since it contains logic to convert IDL arguments to back ES, even though `Request` is not a callback function in the IDL sense); or just "execute the steps listed for that constructor"?

> I think we want it to be the relevant realm of this WindowOrWorkerGlobalScope object.

This would cause a slight discrepancy between `fetch()` and `Request()`. The [`Request()` constructor](https://fetch.spec.whatwg.org/#dom-request) explicitly uses the [current settings object](https://html.spec.whatwg.org/multipage/webappapis.html#current-settings-object) as the client of the new request it's creating; i.e., the realm for which the `Request` function was created.

I think we would want more testing before deciding if we want to use the relevant realm or the current realm.

domenic:
I mean, it has to be bound to some realm. And we generally want to pick the relevant realm of this, for methods. (For constructors the "relevant realm of this" is not super well-defined, since the this object is still being created, and is derived from the realm of new.target, which is usually-but-not-with-Reflect.construct the current realm.)

TimothyGu:
Right, I see.

[...]

PR #1157, opened 2021-02-01
Add abstractions for creating Request and Response objects

### LINKED ISSUE #771: Creating new request & response objects could be easier
In specs, if I want to create a new request, I'm creating the request object, then creating the headers object, then hooking the two up.

With responses it's slightly more complicated, as I need to deal with the trailer promise too.

Is this worth simplifying?

annevk:
Closes #771.

***
<a href="https://whatpr.org/fetch/1157.html" title="Last updated on Feb 2, 2021, 10:50 AM UTC (64f061a)">Preview</a> | <a href="https://whatpr.org/fetch/1157/98f23db...64f061a.html" title="Last updated on Feb 2, 2021, 10:50 AM UTC (64f061a)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5276,12 +5276,9 @@ of a <a for=/>request</a> as demonstrated by {{Request}}
 objects.</span>
 
 <p>A {{Headers}} object also has an associated
-<dfn export for=Headers id=concept-headers-guard>guard</dfn>, which is
-"<code>immutable</code>",
-"<code>request</code>",
-"<code>request-no-cors</code>",
-"<code>response</code>" or
-"<code>none</code>".
+<dfn export for=Headers id=concept-headers-guard>guard</dfn>, which is a <dfn>headers guard</dfn>. A
+<a for=/>headers guard</a> is "<code>immutable</code>", "<code>request</code>",
+"<code>request-no-cors</code>", "<code>response</code>" or "<code>none</code>".
 
 <p>To <dfn export for=Headers id=concept-headers-append>append</dfn> a
 <a for=header>name</a>/<a for=header>value</a>
@@ -5802,8 +5799,8 @@ omitted from <a enum><code>RequestMode</code></a> as it cannot be used nor obser
 <p>A {{Request}} object also has an associated <dfn for=Request export>headers</dfn> (null or a
 {{Headers}} object), initially null.
 
-<p>A {{Request}} object has an associated <dfn for=Request>signal</dfn> (an {{AbortSignal}} object),
-initially a new {{AbortSignal}} object.
+<p>A {{Request}} object has an associated <dfn for=Request>signal</dfn> (null or an {{AbortSignal}}
+object), initially null.
 
 <p>A {{Request}} object's <a for=Body>MIME type</a> is to return the result of
 <a for="header list">extracting a MIME type</a> from its <a for=Request>request</a>'s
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -6608,12 +6622,11 @@ method steps are:
   <a for=Request>signal</a>:
 
   <ol>
-    <li><p>Set <var>locallyAborted</var> to true.
+   <li><p>Set <var>locallyAborted</var> to true.
 
-    <li><p><a>Abort fetch</a> with <var>p</var>, <var>request</var>, and
-    <var>responseObject</var>.
+   <li><p><a>Abort fetch</a> with <var>p</var>, <var>request</var>, and <var>responseObject</var>.
```
annevk, line 6627: I'm not sure if this works as at this point _responseObject_ is null and later it might get a value at which point we'd want this to operate on that value I think. Is there a better way to tackle this?

fetch.bs
```diff
@@ -5938,6 +5938,25 @@ initially a new {{AbortSignal}} object.
 
 <hr>
 
+<p>To
+<dfn export lt="create a Request object|creating a Request object">create a <code>Request</code> object</dfn>,
+given a <a for=/>request</a> <var>request</var> and header guard <var>guard</var>, run these steps:
+<!-- Still need to deal with Realms; maybe create AbortSignal explicitly? -->
```
jakearchibald, line 5944: I think it's fine for the realm to be a parameter. I don't know if there's a way to determine a sensible default.

fetch.bs
```diff
@@ -5938,6 +5938,25 @@ initially a new {{AbortSignal}} object.
 
 <hr>
 
+<p>To
+<dfn export lt="create a Request object|creating a Request object">create a <code>Request</code> object</dfn>,
+given a <a for=/>request</a> <var>request</var> and header guard <var>guard</var>, run these steps:
```
jakearchibald, line 5943:
```suggestion
given a <a for=/>request</a> <var>request</var> and <a for=Header>guard</a> <var>guard</var>, run these steps:
```

Just so it's easy to discover the acceptable values

fetch.bs
annevk: See https://github.com/whatwg/fetch/issues/777#issuecomment-416225862 onward. I haven't really been able to make progress on Realm stuff. Not enough people seem to care...

fetch.bs
annevk: Unfortunately this is wrong as guard is a member of an object and not a standalone concept. I could make it a standalone concept so it could be used as type here though. Maybe that's worth it.

fetch.bs
```diff
@@ -6412,6 +6426,27 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
 <p>A {{Response}} object's <a for=Body>body</a> is its
 <a for=Response>response</a>'s <a for=response>body</a>.
 
+<hr>
+
+<p>To
+<dfn export lt="create a Response object|creating a Response object">create a <code>Response</code> object</dfn>,
+given a <a for=/>response</a> <var>response</var> and headers guard <var>guard</var>, run these
```
jakearchibald, line 6433:
```suggestion
given a <a for=/>response</a> <var>response</var> and <a for=Header>guard</a> <var>guard</var>, run these
```

fetch.bs
jakearchibald: Could you check the signal right before creating the response object?

jakearchibald (changes requested):
Only minor things, otherwise LGTM

fetch.bs
jakearchibald: Huh, I didn't know it was wrong to do that (given there's no `this` use in the definition).

jakearchibald: (having it be linkable to discover the expected values seems worth it though)

annevk:
For the eventual commit message:

* Clarifies some Realms, but more needs to be done in #777 and other issues.
* Ensures that Headers's header list is set correctly for Response.error() and redirect().

fetch.bs
```diff
@@ -6633,7 +6653,8 @@ method steps are:
    <li><p>If <var>response</var> is a <a>network error</a>, then <a for=/>reject</a> <var>p</var>
    with a {{TypeError}} and terminate these substeps.
 
-   <li><p>Associate <var>responseObject</var> with <var>response</var>.
+   <li><p>Set <var>responseObject</var> to the result of <a for=Response>creating</a> a {{Response}}
```
jakearchibald, line 6656: If you're worried about races between this and the "Abort fetch" line above, you could check the state `locallyAborted` and abort these steps if it's true.

annevk: That's step 1 here so I think we're good, no?

jakearchibald (approved):
LGTM!

PR #1163, opened 2021-02-08
fetch() does not need in parallel as fetch already does so

annevk:
There is no need for the fetch() method to go in parallel to run fetch as the main fetch algorithm already does that itself. This also prevents the issue of the main fetch algorithm grabbing global state at the wrong time.

***
<a href="https://whatpr.org/fetch/1163.html" title="Last updated on Feb 8, 2021, 1:44 PM UTC (f4bcad7)">Preview</a> | <a href="https://whatpr.org/fetch/1163/e5500e5...f4bcad7.html" title="Last updated on Feb 8, 2021, 1:44 PM UTC (f4bcad7)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6733,8 +6733,6 @@ method steps are:
   </ol>
 
  <li>
-  <p>Run the following <a>in parallel</a>:
-
   <p><a for=/>Fetch</a> <var>request</var>.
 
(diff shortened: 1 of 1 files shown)


domenic (approved):
LGTM although I found another problem (related to the stuff you're investigating) which I'll file a separate issue for.

PR #1170, opened 2021-02-16
Clarify cors requests need cors response tainting

sideshowbarker:
This change clarifies a (non-normative) statement that "cors" request mode is what makes a request into a cors request. In fact, for requests with "cors" request mode, "cors" response tainting is also necessary in order for the request to be considered a cors request. So this change refines the relevant statement to make that clear.

Otherwise, without this change, considering the case of a same-origin GET request whose mode is "cors", the spec is claiming that same-origin GET request is a cors request.

But because the spec defines a "cors request" as “an HTTP request that includes an `Origin` header”, a same-origin GET request cannot in fact be a cors request — because it doesn’t include an `Origin` header. (And that’s because for GET requests, the spec requires an `Origin` header to be appended only if the request’s response tainting is "cors"; but for same-origin requests, the request’s response tainting will be "basic".)

***
<a href="https://whatpr.org/fetch/1170.html" title="Last updated on Feb 17, 2021, 10:23 AM UTC (73a43d4)">Preview</a> | <a href="https://whatpr.org/fetch/1170/3c90676...73a43d4.html" title="Last updated on Feb 17, 2021, 10:23 AM UTC (73a43d4)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1390,7 +1390,8 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
   <a>network error</a> if the request is not made to a same-origin URL.
 
   <dt>"<code>cors</code>"
-  <dd>Makes the request a <a>CORS request</a>. Fetch will return a <a>network error</a> if the
+  <dd>For requests whose <a for=request>response tainting</a> is also "<code>cors</code>", makes the
+  request a <a>CORS request</a> — in which case, fetch will return a <a>network error</a> if the
   requested resource does not understand the <a>CORS protocol</a>.
 
   <dt>"<code>no-cors</code>"
--- a/fetch.bs
+++ b/fetch.bs
@@ -1392,7 +1392,8 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
   <dt>"<code>cors</code>"
   <dd>For requests whose <a for=request>response tainting</a> gets set to "<code>cors</code>", makes
   the request a <a>CORS request</a> — in which case, fetch will return a <a>network error</a> if the
-  requested resource does not understand the <a>CORS protocol</a>.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1390,7 +1390,8 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
   <a>network error</a> if the request is not made to a same-origin URL.
 
   <dt>"<code>cors</code>"
-  <dd>Makes the request a <a>CORS request</a>. Fetch will return a <a>network error</a> if the
+  <dd>For requests whose <a for=request>response tainting</a> is also "<code>cors</code>", makes the
```
annevk, line 1393: "gets set to" seems more accurate since response tainting is not set by those creating the request.

sideshowbarker: > "gets set to" seems more accurate since response tainting is not set by those creating the request.

Made it so

fetch.bs
```diff
@@ -1390,7 +1390,8 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
   <a>network error</a> if the request is not made to a same-origin URL.
 
   <dt>"<code>cors</code>"
-  <dd>Makes the request a <a>CORS request</a>. Fetch will return a <a>network error</a> if the
+  <dd>For requests whose <a for=request>response tainting</a> gets set to "<code>cors</code>", makes
+  the request a <a>CORS request</a> — in which case, fetch will return a <a>network error</a> if the
   requested resource does not understand the <a>CORS protocol</a>.
```
annevk, line 1395: Sorry, noticed this afterwards, but maybe you want to improve this as well since the resource might well understand it, but not wish to share.

sideshowbarker: Went with: “or if the requested resource is one that intentionally does not participate in the CORS protocol”

PR #1172, opened 2021-02-17
Formalize reading bodies

### LINKED ISSUE #661: Getting all bytes in a body
(This spans HTML, Encoding, Streams, and Fetch. Let's tag at least @ricea to help while we're here, and @jakearchibald since he was helpful in https://github.com/whatwg/infra/issues/181.)

I need help figuring out how at a spec level to get all bytes in a body. This is part of solving https://github.com/whatwg/html/issues/3316.

Right now specs are using a few patterns:

- https://fetch.spec.whatwg.org/#concept-body-consume-body or https://w3c.github.io/payment-method-manifest/#fetch-pmm 4.5.4: get reader, read all bytes into Uint8Array. Then treat it as a byte sequence.
- https://html.spec.whatwg.org/#fetch-a-classic-script step 8 (which I need to replace): treat "body" as if it was the same as Encoding's "byte stream" concept and just feed it into that.

In general the problem here is Streams use of the JS formalism, including promises, and how that interacts poorly with spec-level code and its different conventions and type system. The confusion between encoding's streams and Streams's streams is also tricky.

Looking at https://html.spec.whatwg.org/#fetch-a-classic-script my first thought is that it should be written as:

8. Let _body bytes_ be the byte sequence obtained by [reading all the bytes] of the [body] of _response_.
9. Let _source text_ be the result of [decoding] _body bytes_ to Unicode, using _character encoding_ as the fallback encoding.
10. ... muted errors ...
11. Let _script_ be the result of creating a classic script given _body bytes_, _source text_, ... the other stuff.

This doesn't quite work on a few levels:

- "reading all the bytes" would at the very least need to be asynchronous, if it goes through Streams's promise machinery.
- "reading all the bytes" might fail. This is currently unaccounted for, but if we go through Streams's machinery, it's more explicit.
- "reading all the bytes" will need to hand-wave to get from Uint8Array to byte sequence.
- "decoding" doesn't accept byte sequences, only "byte streams".

So here is my proposal, which is essentially trying for a minimal delta from today:

- We define "reading all the bytes", probably in whatwg/fetch. It returns either a byte sequence, or null to signal failure. Either its asynchronous, or we use "wait" (see https://github.com/whatwg/infra/issues/181) to make it synchronous. It wraps up all the promise/Uint8Array stuff so that spec authors don't have to worry about it, hand-waving as appropriate.
- We define in Encoding that "byte sequences" can be used as "byte streams" implicitly, when appropriate.

---

An alternate approach, which you might prefer, is to double-down on the spec-level concept of a byte stream. We'd provide some way of translating a ReadableStream, and thus a body, into a spec-level byte stream. It creates a reader, locks the ReadableStream, and then from then on, specs only manipulate the byte stream.

The hardest part of this, I think, is putting the asynchronicity of "reading" from the byte stream on solid ground. It seems very hand-wavey in Encoding, and most of Encoding's clients, right now. We could either:

- Say that you can read "synchronously", in that it waits for more data to come in before "read from a byte stream" returns
- or say that read needs to be asynchronous

This ties back to https://github.com/whatwg/infra/issues/181 again. The problem with synchronous reads is that you can only do them from in-parallel sections, and I think in most cases "decode" does not run in those sections.

Indeed, the larger issue where encode/decode often are used on strings/byte sequences, instead of on "character streams"/"byte streams", seems pretty prevalent: see e.g. https://html.spec.whatwg.org/#form-submission-algorithm:encode or https://html.spec.whatwg.org/#navigating-across-documents:utf-8-decode. So maybe we need to do something about that anyway. Those cases wouldn't need to run in parallel.

annevk:
Closes #661.

***
<a href="https://whatpr.org/fetch/1172.html" title="Last updated on Feb 24, 2021, 11:51 AM UTC (1b5a8cd)">Preview</a> | <a href="https://whatpr.org/fetch/1172/0f29092...1b5a8cd.html" title="Last updated on Feb 24, 2021, 11:51 AM UTC (1b5a8cd)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1311,7 +1311,7 @@ the empty string,
      * CSP: https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request
      * Mixed Content: https://w3c.github.io/webappsec-mixed-content/#should-block-fetch
      * Preload: https://w3c.github.io/preload/#processing
-     * SRI: https://w3c.github.io/webappsec-subresource-integrity/#apply-algorithm-to-request
+     * SRI
      * HTML -->
 
 <p>A <a for=/>request</a>'s <a for=request>destination</a> is
@@ -3632,8 +3632,7 @@ steps:
     <p>Let <var>processBody</var> given <var>bytes</var> be these steps:
 
     <ol>
-     <li><p>If <var>bytes</var> do not
-     <a href=https://w3c.github.io/webappsec-subresource-integrity/#does-response-match-metadatalist>match</a>
+     <li><p>If <var>bytes</var> do not <a lt="Do bytes match metadataList?">match</a>
      <var>request</var>'s <a for=request>integrity metadata</a>, then run
      <var>processBodyError</var> and abort these steps. [[!SRI]]
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -182,7 +182,7 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication). 
 steps:
 
 <ol>
- <li><p>If <var>taskDestination</var> is a <a>parallel queue</a>, then
+ <li><p>If <var>taskDestination</var> is a <a for=/>parallel queue</a>, then
  <a lt="enqueue steps" for="parallel queue">enqueue</a> <var>algorithm</var> to
  <var>taskDestination</var>.
 
@@ -1003,24 +1003,11 @@ worked on in <a href=https://github.com/whatwg/fetch/issues/1156>issue #1156</a>
 <ul>
  <li><p>A <dfn export for=body id=concept-body-stream>stream</dfn> (a {{ReadableStream}} object).
 
- <li><p>A <dfn export for=body id=concept-body-transmitted>transmitted bytes</dfn>
- (an integer), initially 0.
-
  <li><p>A <dfn export for=body id=concept-body-total-bytes>total bytes</dfn> (an
  integer), initially 0.
 
  <li><p>A <dfn export for=body id=concept-body-source>source</dfn>, initially null.
 </ul>
-
-<p>A <a for=/>body</a> <var>body</var> is said to be
-<dfn export for=body id=concept-body-done>done</dfn> if <var>body</var> is null or
-<var>body</var>'s <a for=body>stream</a> is
-<a for=ReadableStream>closed</a> or
-<a for=ReadableStream>errored</a>.
-
-<p>To <dfn export for=body id=concept-body-wait>wait</dfn> for a
-<a for=/>body</a> <var>body</var>, wait for <var>body</var> to be
-<a for=body>done</a>.
 
 <p>To <dfn export for=body id=concept-body-clone>clone</dfn> a
 <a for=/>body</a> <var>body</var>, run these steps:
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1037,6 +1037,131 @@ worked on in <a href=https://github.com/whatwg/fetch/issues/1156>issue #1156</a>
...
+<ol>
+ <li><p>If <var>taskDestination</var> is null, then set <var>taskDestination</var> to the result of
+ <a>starting a new parallel queue</a>.
+
+ <li><p>Let <var>promise</var> be the result of <a>fully reading body as promise</a> given
+ <var>body</var>.
+
+ <li><p>Let <var>fulfilledSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
+ <a>queue a fetch task</a> to run <var>processBody</var> given <var>bytes</var>, with
+ <var>taskDestination</var>.
```
annevk, line 1140: @domenic for instance, in theory this shouldn't have to queue a task as there's no way for these fulfilled steps to run without one, but the streams primitives this builds on are not exactly clear about that.

This is a clear case, but I have similar questions about the incremental read above. Thoughts?

Apart from that I think this is largely the correct setup that we can use both for request (to incrementally read and then invoke transmit in the callback) and response bodies (mostly to read). And we can also use this to create a proxy stream for SRI, etc.

domenic: fulfilledSteps and stream's chunk steps/close steps/error steps run as microtasks, on the stream object's relevant agent's event loop. So if you're happy with that event loop and that task priority, you don't need to queue a fetch task. You could do so if you want to switch event loops or change task priorities.

annevk: Well, you've been asking for this functionality. What do you need? 😊 I could use some help here.

domenic: Probably queuing a fetch task would be best, especially for HTML where a lot of the cases talk about the networking task source (e.g. "the last task that is queued by the networking task source once the resource has been fetched", "Queue an element task on the networking task source given el to process the linked resource", ...). It also seems more future-proof if we move to realm-independent "C++ streams".

annevk:
I think this is getting somewhere, but:

- [ ] Need to read through it all once again to make sure I haven't dropped anything. (I would still appreciate some review to ensure it's on the right track. And also for ideas on how to address the bits that are still awkward, e.g., request's done flag.)
- [ ] Create a PR for SRI as this changes the handshake between the specifications.

It makes things more verbose for sure, but overall I think that is probably worth it. Early takes from @yutakahirano and @jakearchibald would also be welcome with regards to that.

fetch.bs
```diff
@@ -1037,6 +1027,138 @@ worked on in <a href=https://github.com/whatwg/fetch/issues/1156>issue #1156</a>
...
+ <li><p><a for=ReadableStreamDefaultReader>Read a chunk</a> from <var>reader</var> given
+ <var>readRequest</var>.
+</ol>
+
+<p>To <dfn export for=body>fully read</dfn> a <a for=/>body</a> <var>body</var>, given an algorithm
+<var>processBody</var>, an optional null or algorithm
+<dfn export for="body/fully read"><var>processError</var></dfn> (default null), and an optional
+null, <a for=/>parallel queue</a>, or <a for=/>global object</a>
+<dfn export for="body/fully read">taskDestination</dfn> (default null), run these steps.
+<var>processBody</var> must be an algorithm accepting failure or a <a for=/>byte sequence</a>. If
```
domenic, line 1123: I think it doesn't need to handle the failure case; processError deals with that?

yutakahirano (commented):
Overall this looks good.

fetch.bs
```diff
@@ -1037,6 +1024,134 @@ worked on in <a href=https://github.com/whatwg/fetch/issues/1156>issue #1156</a>
...
+<p>To <dfn export for=body>incrementally read</dfn> a <a for=/>body</a> <var>body</var>, given an
+algorithm <var>processBodyChunk</var>, an algorithm <var>processEndOfBody</var>, an algorithm
+<var>processBodyError</var>, and an optional null, <a for=/>parallel queue</a>, or
+<a for=/>global object</a> <var>taskDestination</var> (default null), run these steps.
+<var>processBodyChunk</var> must be an algorithm accepting a <a for=/>byte sequence</a>.
+<var>processBodyError</var> must be an algorithm accepting an exception. <var>processEndOfBody</var>
+must be an algorithm accepting no arguments.
+
+<ol>
+ <li><p>Assert: <var>body</var>'s <a for=body>stream</a> is non-null.
```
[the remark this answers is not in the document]
annevk: In HTTP-network-or-cache fetch we create a body whose stream is null. If that's not correct, file an issue?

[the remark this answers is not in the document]
annevk: Thinking about it some more that does seem wrong. Maybe that should just clone request and we'll leave optimizations to implementers. Fixing that here might be a stretch though.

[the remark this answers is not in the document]
annevk: #1177. I'll rebase this on top once that lands.

fetch.bs
```diff
@@ -1037,6 +1024,124 @@ worked on in <a href=https://github.com/whatwg/fetch/issues/1156>issue #1156</a>
...
+
+<ol>
+ <li><p>If <var>taskDestination</var> is null, then set <var>taskDestination</var> to the result of
+ <a>starting a new parallel queue</a>.
+
+ <li>
+  <p>Let <var>reader</var> be the result of <a for=ReadableStream>getting a reader</a> for
+  <var>body</var>'s <a for=body>stream</a>.
+
+  <p class=note>This operation ought not to throw an exception.
```
domenic, line 1045: out not to -> will not?

fetch.bs
```diff
@@ -1037,6 +1024,124 @@ worked on in <a href=https://github.com/whatwg/fetch/issues/1156>issue #1156</a>
...
+     <li><p>If <var>chunk</var> is not a {{Uint8Array}} object, then set
+     <var>continueAlgorithm</var> to this step: run <var>processBodyError</var> given a
+     {{TypeError}}.
+
+     <li>
+      <p>Otherwise:
+
+      <ol>
+       <li><p>Let <var>bytes</var> be a <a lt="get a copy of the buffer source">copy of</a>
+       <var>chunk</var>.
```
domenic, line 1076: Maybe note that this copy can often be elided if the user agent is working on its own streams, or e.g. a stream supplied from a service worker where the copy has already been made while crossing the realm boundary? I dunno, maybe not worth it...

fetch.bs
```diff
@@ -3241,9 +3226,11 @@ an optional algorithm <dfn export for=fetch id=process-response><var>processResp
 optional algorithm
 <dfn export for=fetch id=process-response-end-of-body oldids=process-response-end-of-file><var>processResponseEndOfBody</var></dfn>,
 and an optional boolean <dfn export for=fetch><var>useParallelQueue</var></dfn> (default false), run
-the steps below. If given, <var>processRequestBody</var> and <var>processRequestEndOfBody</var> must
-be an algorithm accepting a <a for=/>request</a>; <var>processResponse</var> and
-<var>processResponseEndOfBody</var> must be an algorithm accepting a <a for=/>response</a>.
+the steps below. If given, <var>processRequestBody</var> must be an algorithm accepting an integer.
```
domenic, line 3229: "an integer representing the number of bytes transmitted"? I had to click around a lot to figure out what this was for.

fetch.bs
```diff
@@ -3241,9 +3226,11 @@ an optional algorithm <dfn export for=fetch id=process-response><var>processResp
...
+If given, <var>processRequestEndOfBody</var> must be an algorithm accepting no arguments.
+<var>processResponse</var> must be an algorithm accepting a <a for=/>response</a>.
```
domenic, line 3231: "If given" for both of these

fetch.bs
```diff
@@ -3284,15 +3271,9 @@ the request.
...
-  <p>If <var>request</var>'s <a for=request>body</a> is a <a for=/>byte sequence</a>, then:
-
-  <ol>
-   <li><p>Let <var>body</var> and <var ignore>ignoreType</var> be the result of
-   <a for=BodyInit>safely extracting</a> <var>request</var>'s <a for=request>body</a>.
-
-   <li><p>Set <var>request</var>'s <a for=request>body</a> to <var>body</var>.
-  </ol>
+ <li><p>If <var>request</var>'s <a for=request>body</a> is a <a for=/>byte sequence</a>, then set
+ <var>request</var>'s <a for=request>body</a> to the first return value of
```
domenic, line 3275: "the first return value" seems pretty unusual; why not just "the result of"? Here and below.

fetch.bs
```diff
@@ -3630,51 +3611,75 @@ steps:
...
+  <p>If <var>fetchParams</var>'s <a for="fetch params">process response end-of-body</a> is non-null,
+  then:
 
   <ol>
-   <li><p>Set <var>request</var>'s <a>done flag</a>.
+   <li><p>Let <var>processBody</var> given <var>nullOrFailureOrBytes</var> be this step: run
+   <var>fetchParams</var>'s <a for="fetch params">process response end-of-body</a> given
+   <var>response</var> and <var>nullOrFailureOrBytes</var>.
+
+   <li><p>Let <var>processBodyError</var> be to run <var>processBody</var> given failure.
```
domenic, line 3667: Although this is clever, I think it'd be easier to understand if it directly called to the output algorithm, i.e.

> Let _processBodyError_ be this step: run _fetchParam_'s process response end-of-body given _response_ and failure.

fetch.bs
```diff
@@ -3630,51 +3611,75 @@ steps:
...
 
-   <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response end-of-body</a> is
-   non-null, then run <var>fetchParams</var>'s
-   <a for="fetch params">process response end-of-body</a> given <var>response</var>.
+   <li><p>If <var>response</var>'s <a for=response>body</a> is null, then <a>queue a fetch task</a>
+   to <var>processBody</var> given null, with <var>fetchParams</var>'s
```
domenic, line 3670:
```suggestion
   to run <var>processBody</var> given null, with <var>fetchParams</var>'s
```

domenic (approved):
LGTM with editorial suggestions

fetch.bs
annevk: It returns two values. I'm aligning it with text elsewhere that uses this terminology.

annevk:
Thanks, I addressed those comments. I also updated the first commit to include a longer description of the changes.

PR #1174, opened 2021-02-19
Clarify: credentials param affect response too

sideshowbarker:
This change updates the descriptions of all “credentials mode” values, to clarify that the values affect whether browsers use credentials sent back in responses (e.g., any Set-Cookie response headers) — not just whether credentials are sent with the request. Related: https://github.com/mdn/content/pull/2468

***
<a href="https://whatpr.org/fetch/1174.html" title="Last updated on Feb 19, 2021, 2:58 PM UTC (8cc1ae8)">Preview</a> | <a href="https://whatpr.org/fetch/1174/c64f074...8cc1ae8.html" title="Last updated on Feb 19, 2021, 2:58 PM UTC (8cc1ae8)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1434,7 +1434,7 @@ which is "<code>omit</code>", "<code>same-origin</code>", or
 
   <dt>"<code>same-origin</code>"
   <dd>Include credentials with requests made to same-origin URLs, and use any credentials sent back
-  in responses from same-origin URLs
+  in responses from same-origin URLs.
 
   <dt>"<code>include</code>"
   <dd>Always includes credentials with this request, and always use any credentials sent back in the
--- a/fetch.bs
+++ b/fetch.bs
@@ -1429,13 +1429,16 @@ which is "<code>omit</code>", "<code>same-origin</code>", or
 <div class="note no-backref">
  <dl>
   <dt>"<code>omit</code>"
-  <dd>Excludes credentials from this request.
+  <dd>Excludes credentials from this request, and causes any credentials sent back in the response
+  to be ignored.
 
   <dt>"<code>same-origin</code>"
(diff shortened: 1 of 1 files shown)


annevk (approved):
I guess in practice there's only really `Set-Cookie` for responses, but maybe keeping this generic is fine. Thoughts?

sideshowbarker:
> I guess in practice there's only really `Set-Cookie` for responses, but maybe keeping this generic is fine. Thoughts?

For developers reading it, it seems best to be as explicit as possible — so if the Set-Cookie header is the only form or credentials the response can have which would be affected, then let’s say that.

annevk:
Sorry, I'm wrong. It also affects 401 responses (and connection pooling). (Search for includeCredentials.)

sideshowbarker:
> Sorry, I'm wrong. It also affects 401 responses (and connection pooling). (Search for includeCredentials.)

OK — stripped out the other commit

fetch.bs
```diff
@@ -1429,13 +1429,16 @@ which is "<code>omit</code>", "<code>same-origin</code>", or
...
  <dl>
   <dt>"<code>omit</code>"
-  <dd>Excludes credentials from this request.
+  <dd>Excludes credentials from this request, and causes any credentials sent back in the response
+  to be ignored.
 
   <dt>"<code>same-origin</code>"
-  <dd>Include credentials with requests made to same-origin URLs.
+  <dd>Include credentials with requests made to same-origin URLs, and use any credentials sent back
+  in responses from same-origin URLs
```
annevk, line 1437:
```suggestion
  in responses from same-origin URLs.
```

annevk (approved):

annevk:
Thanks @sideshowbarker!

Issue #1156, opened 2021-02-01
HTTP/1 parsing


annevk:
HTTP/1 was poorly tested for error handling and as a result implementations are not interoperable. This lack of interoperability reaches Fetch at times. Where HTTP/1 itself takes no stance or implementations cannot follow it, it might make sense to document it in Fetch or at least document it in this issue.

* status-line parsing: https://github.com/web-platform-tests/wpt/pull/5102
* status-code parsing: https://github.com/whatwg/fetch/issues/1142
* CR parsing: https://github.com/web-platform-tests/wpt/pull/13524
* 0x00 parsing: https://github.com/whatwg/xhr/issues/165#issuecomment-380447687 (apart from Firefox and a bug in Safari this is pretty close; agreement here is network error)
* field-value is "incorrect": https://fetch.spec.whatwg.org/#concept-header-value (there is agreement on this, modulo 0x00 above)

cc @whatwg/http 

(Previously there was https://wiki.whatwg.org/wiki/HTTP.)

MattMenke2:
Another fun one:  Chrome allows 4 bytes of random slop between requests, before the start of headers.  I believe this was copied from FireFox of IE's behavior at the time.  I don't believe all browsers currently do this, so it may be easier to remove.  My main concern is servers sending 0-length compressed gzip bodies with "Content-Length: 0", which aren't actually of length 0.  In this case, there's slop on the end, which I suspect this logic deals with.

JannisBush:
Several other differences I discovered:
1. Space (or tab) between `header-name` and `:`, e.g., `x-frame-options : DENY`:
- [HTTP spec](https://www.rfc-editor.org/rfc/rfc9112.html#name-field-line-parsing) says that servers MUST reject such responses and proxies must remove such whitespace. It does not say what clients should do.
- Firefox ignores only the invalid header
- Chromium and Safari ignore the space, even the invalid header gets parsed and in the example here: XFO is applied
2. [NULL bytes in header names](https://github.com/whatwg/fetch/issues/1747)
- Invalid according to the spec (token) but unclear what clients should do
- Firefox and Safari only ignore the invalid header
- Chromium treats the whole response as invalid (network error), similar to NULL in header values
3. NULL in header values
- Agreement seems to be network error
- Tests do not test for all kind of inclusions: e.g., no tests for [fetch](https://github.com/web-platform-tests/wpt/issues/45737) where currently browser behavior diverges
4. Lone LF (0x0A) in responses
[...]

Issue #604, opened 2017-09-18
Request's total bytes isn't set


jakearchibald:
https://fetch.spec.whatwg.org/#concept-body-total-bytes isn't set for requests, unless the input is a Request object.

Feels like it should be set in https://fetch.spec.whatwg.org/#concept-bodyinit-extract for non-stream sources.

annevk:
cc @yutakahirano

wanderview:
It would be nice, though, to acknowledge the size can be unknown.  Zero really isn't a good representation of this.

annevk:
I really want to remove this field and for responses we can look at the `Content-Length` header (at least for HTTP/1, guess we might have to write/look at some XMLHttpRequest tests to see how the progress events behave).

For non-stream requests the main difficulty will be getting a length out of https://fetch.spec.whatwg.org/#concept-bodyinit-extract when _object_ isn't a stream. That might also involve acknowledging that some serialization happens on the main thread. `FormData` is the trickiest as it can hold blobs so you would partially serialize it and then use the size of the blobs to calculate the remaining space that is required. Perhaps this can be implementation-defined until we have `FormData` serialization itself defined?

annevk:
Here's my plan:

* Define convert BodyInit object to bytes routine that returns a tuple of a byte sequence and null or a multipart/form-data boundary. This will synchronous but implementations can optimize.
* Define a size of BodyInit object routine that invokes convert and returns the length. We'll use this to set the Content-Type header in requests. (I kinda wanna use `<a for="BodyInit,byte sequence">length</a>` there, but if that's not a thing we'll just branch.)
* Make extract use convert as well instead of all the "actions".

Add a bunch of notes that encourage implementations to optimize this. All ears to better ideas though.

annevk:
My plan does not work for `Blob` and `FormData`: https://github.com/whatwg/html/issues/6424. The other types should probably still be eagerly converted to a byte sequence...

So maybe we should have convert and that returns a tuple of

* a byte sequence or a  read operation of sorts
* a length

but it's still rather inelegant to do extract and length (which would both use convert) determination separately. So maybe body should store a length after all. (Though we should also be clear that it's different between requests and responses as for responses it can lie, especially with service workers and maybe also with H/2 and H3/, which don't really have that concept although if you specify `Content-Length` maybe it's still required to match?)

PR #1183, opened 2021-03-01
Define Content-Length parser

annevk:
As the body concept is refactored for #604, XMLHttpRequest will need to use this algorithm for its events. We also want to require browsers to use this algorithm instead of the one defined by HTTP as part of #1156.

Tests: https://github.com/web-platform-tests/wpt/pull/10548.

   * Chrome
   * Firefox
   * See above.
   * Chrome: N/A
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1695931 & https://bugzilla.mozilla.org/show_bug.cgi?id=1696203
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=222596

@MattMenke2 I haven't attempted to actually require browsers use this yet. I suspect we want some kind of "HTTP monkey patching" section that outlines the various things we want browsers (and clients that emulate them) to implement for that. Ideas welcome, but it's not my main priority here (I'm trying to scope my yak shaving).

***
<a href="https://whatpr.org/fetch/1183.html" title="Last updated on Mar 3, 2021, 4:18 PM UTC (78d9116)">Preview</a> | <a href="https://whatpr.org/fetch/1183/24a8670...78d9116.html" title="Last updated on Mar 3, 2021, 4:18 PM UTC (78d9116)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2831,7 +2831,7 @@ from a <a for=/>header list</a> <var>headers</var>, run these steps:
    <li><p>If <var>candidateValue</var> is null, then set <var>candidateValue</var> to
    <var>value</var>.
 
-   <li><p>Otherwise, if <var>value</var> is not <var>candidateValue</var>, then return failure.
+   <li><p>Otherwise, if <var>value</var> is not <var>candidateValue</var>, return failure.
   </ol>
 
  <li><p>If <var>candidateValue</var> is the empty string or has a <a for=/>code point</a> that is
--- a/fetch.bs
+++ b/fetch.bs
@@ -2807,10 +2807,44 @@ consideration for the security consequences. New exceptions can be proposed by
 <a href=https://github.com/whatwg/fetch/issues/new>filing an issue</a>.
 
 
+<h3 id=content-length-header>`<code>Content-Length</code>` header</h3>
+
+<p>The `<code>Content-Length</code>` header is largely defined in HTTP. Its processing model is
+defined here as the model defined in HTTP is not compatible with web content. [[HTTP]]
+
+<p>To <dfn export for="header list" lt="extract a length|extracting a length">extract a length</dfn>
+from a <a for=/>header list</a> <var>headers</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>values</var> be the result of
+ <a for="header list">getting, decoding, and splitting</a> `<code>Content-Length</code>` from
+ <var>headers</var>.
+
+ <li><p>If <var>values</var> is null, then return null.
+
+ <li><p>Let <var>candidateValue</var> be null.
+
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>values</var>:
+
+  <ol>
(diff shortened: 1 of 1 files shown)


annevk:
One could argue that we should only define this for a response. I don't think I have found a place where browsers use this for requests.

MattMenke2 (approved):
This change looks good to me.

I'm not sure I'll have the time to fully dig through #1184, make sure it happily interacts with the chunked encoding bits, etc.

annevk:
I've decided to land this part and leave landing the additional tests and browser bugs to #1184 as this is fairly standalone.

mnot:
When I run http://wpt.live/fetch/content-length/content-length.html in both FF and Safari, my browser doesn't fetch `content-lengths.json`, and only runs one test; wpt.fyi shows only one test result at <https://wpt.fyi/results/fetch/content-length/content-length.html?label=master&label=experimental&aligned>. What am I missing?

MattMenke2:
Hrm...You're right.  Looks like there's nothing that loads parsing.window.js or too-long.window.js.

Anne:  Did you forget to upload a file?

annevk:
No, you want to run http://wpt.live/fetch/content-length/parsing.window.html and http://wpt.live/fetch/content-length/too-long.window.html for the added tests. `.window.js` (as well as `.any.js`) are shorthands to avoid having to write a bunch of boilerplate: https://web-platform-tests.org/writing-tests/testharness.html#without-html-boilerplate-window-js.

mnot:
Aha - they are suddenly showing up now. Weird, they weren't before, even with the same commit; caching, maybe?

annevk:
Ah, I think the results page is often behind a bit as it takes a while to run all the tests. So it only updates every so often.

mnot:
Considering that you're really only showing interoperability between Blink-derived browsers, it's a bit premature to be saying that 'the model defined in HTTP is not compatible with web content', don't you think?

annevk:
I suppose that Safari's results are closest to indicating that things could perhaps be more strict (although Safari is less strict than Chrome in many cases as well).

Perhaps what you're saying though is that https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.8.6 doesn't contain many requirements on recipients so anything is fine? But even then, the model where recipients get to choose how to behave does not seem compatible as duplicate `Content-Length` headers are a thing.

martinthomson:
The reporting here is just odd.  All the percent encoding, and no expectations.  Do these have to be like this?

annevk:
There's a variety of ways to do it, but this seemed nice in that you can see the exact input (albeit encoded).

martinthomson:
It doesn't say what the expectation is.  And all the percent encoding makes it very hard to parse.  Given that this test doesn't depend on the spacing, removing the encoding would be a start at making the output readable.

annevk:
Sure: https://github.com/web-platform-tests/wpt/pull/27894. (The expectations are also part of the Message column, but it can be a bit cryptic if you don't know what you're looking for.)

PR #1187, opened 2021-03-04
Editorial: remove some duplicate abort logic

annevk:
The fetch algorithm also errors this stream with an "AbortError" DOMException in HTTP-network fetch.

(I could see waiting with landing this until more is cleaned up around how abort/terminate works. My tentative plan is that we keep much of the current structure, but rather than "the ongoing fetch is terminated" we could check something like "fetchHandle's stop is not 'none'" with other values being 'abort' and 'terminate'.)

***
<a href="https://whatpr.org/fetch/1187.html" title="Last updated on Oct 19, 2022, 3:39 PM UTC (01bd945)">Preview</a> | <a href="https://whatpr.org/fetch/1187/e5f025a...01bd945.html" title="Last updated on Oct 19, 2022, 3:39 PM UTC (01bd945)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7854,21 +7854,19 @@ method steps are:
   <ol>
    <li><p>Set <var>locallyAborted</var> to true.
 
-   <li><p><a>Abort the <code>fetch()</code> call</a> with <var>p</var>, <var>request</var>, <var>responseObject</var>,
-   and <var>requestObject</var>'s <a for=Request>signal</a>'s <a for=AbortSignal>abort reason</a>.
+   <li><p><a for=/>Assert</a>: <var>controller</var> is non-null.
 
-   <li><p>If <var>controller</var> is not null, then <a for="fetch controller">abort</a>
-   <var>controller</var> with <var>requestObject</var>'s <a for=Request>signal</a>'s
-   <a for=AbortSignal>abort reason</a>.
+   <li><p><a for="fetch controller">Abort</a> <var>controller</var> with <var>requestObject</var>'s
+   <a for=Request>signal</a>'s <a for=AbortSignal>abort reason</a>.
   </ol>
 
  <li>
   <p><p>Set <var>controller</var> to the result of calling <a for=/>fetch</a> given
   <var>request</var> and <a for=fetch><i>processResponse</i></a> given <var>response</var> being
-  these substeps:
+  these steps:
 
   <ol>
-   <li><p>If <var>locallyAborted</var> is true, terminate these substeps.
+   <li><p>If <var>locallyAborted</var> is true, then abort these steps.
 
    <li>
     <p>If <var>response</var>'s <a for=response>aborted flag</a> is set, then:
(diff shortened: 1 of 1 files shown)


annevk:
See also https://github.com/whatwg/dom/issues/1031.

annevk:
@nidhijaju @domenic could you take a look at this? My inkling is that we want to take this even further and also apply it to request uploading. It actually seems that in the case where you hit an early network error, the request upload stream would not end up being cancelled at all. There's nothing nearby "To transmit request’s body _body_, run these steps" that would end up doing that. Do you agree?

fetch.bs
```diff
@@ -7906,14 +7906,6 @@ method steps are:
  <li><p>If <var>request</var>'s <a for=request>body</a> is not null and is
  <a for=ReadableStream>readable</a>, then <a for=ReadableStream>cancel</a> <var>request</var>'s
  <a for=request>body</a> with <var>error</var>.
-
- <li><p>If <var>responseObject</var> is null, then return.
```
domenic, line 7910: What about if we're fetching from cache, not network? What about if we're fetching a blob?

Also, if this did go through, we'd delete the _responseObject_ argument.

fetch.bs
```diff
@@ -7854,21 +7854,19 @@ method steps are:
...
 
-   <li><p><a>Abort the <code>fetch()</code> call</a> with <var>p</var>, <var>request</var>, <var>responseObject</var>,
-   and <var>requestObject</var>'s <a for=Request>signal</a>'s <a for=AbortSignal>abort reason</a>.
+   <li><p><a for=/>Assert</a>: <var>controller</var> is non-null.
 
-   <li><p>If <var>controller</var> is not null, then <a for="fetch controller">abort</a>
-   <var>controller</var> with <var>requestObject</var>'s <a for=Request>signal</a>'s
-   <a for=AbortSignal>abort reason</a>.
+   <li><p><a for="fetch controller">Abort</a> <var>controller</var> with <var>requestObject</var>'s
+   <a for=Request>signal</a>'s <a for=AbortSignal>abort reason</a>.
```
domenic, line 7860: Note to self: this does indeed work. Setting the controller's state will trigger various "abort when" conditions, both inside HTTP network-fetch, HTTP cache-fetch, and blob's part of scheme fetch. All those "abort when" conditions will eventually end up in processResponse with response set to a network error whose the aborted flag set, and that will call "abort the fetch() call".

domenic (commented):
Promise logic looks good, not so sure about stream logic.

fetch.bs
annevk: This was a good comment. It resulted in #1505, #1506, #1507, and #1508 thus far. I suspect I should reinstate this, but we might still want some follow-up to also create a stream for HTTP cache cases.

fetch.bs
annevk: It doesn't actually work for blobs I think because the reading from disk happens in parallel from that. And we don't abort that stream. So I suspect we need to keep all the original invocations. I should probably start a new PR that only does some minor cleanup and perhaps explains how this works.

We should probably also inline the "abort when" thing as a single check at the start of scheme fetch seems sufficient?

annevk: What we could do here maybe is branch on scheme, to keep the dream of https://github.com/whatwg/dom/issues/1031 mostly alive. Hmm.

PR #1188, opened 2021-03-05
Export "determine the network partition key"

yutakahirano:
Export "determine the network partition key", for external use.

It will be used by w3c/webtransport#217.

This doesn't change any behavior.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2222,7 +2222,7 @@ clearly stipulates that <a>connections</a> are keyed on
 an <a>implementation-defined</a> value.
 
 <p>To
-<dfn lt="determine the network partition key|determining the network partition key">determine the network partition key</dfn>,
+<dfn export lt="determine the network partition key|determining the network partition key">determine the network partition key</dfn>,
 given <var>request</var>, run these steps:
 
 <ol>


yutakahirano:
Hi @annevk, what do you think about this change?

Alternatively we can
 - Define a more primitive operation taking an origin and a settings object, and export it, or
 - Define a similar operation in w3c/webtransport.

annevk:
I don't like callers having to make up a fake request, so yeah, let's define the primitive operation.

I think the name of that primitive operation can be "determine the network partition key" and the current operation can be renamed to "determine the request partition key" which just ends up calling the other one (and we don't export this one and just keep it as a shorthand for the various callers inside Fetch). If that sounds good to you.

annevk:
Also, passing an environment settings object might be enough for the primitive?

yutakahirano:
Thank you! How about the latest change?

annevk (commented):
This looks good, although I'm less sure about the name of the internal algorithm than I was four days ago. (One alternative is that we use the same name with `for=request`, but I'm on the fence whether that is better.)

There are many more callers of the internal request algorithm though and they all need to be updated.

annevk:
@yutakahirano it wasn't entirely clear to me whether you are finished with this now, but if you are I'm happy to land this?

Issue #1201, opened 2021-03-25
New Resource Timing issues


annevk:
@noamr when looking at https://github.com/whatwg/xhr/pull/319 I started looking at `fetch()` again and I don't understand a couple things. Let's start with this bit:

> Let handleFetchDone be to finalize and report timing with response, globalObject, and "fetch".

Where does _response_ come from here? I guess https://fetch.spec.whatwg.org/#finalize-response has to pass it on.

How does this interact with callers of HTTP-network fetch that might take this response and determine to return a network error to the caller of fetch?

Another problem here is that this might not be the final response. We end up calling these things for redirects and such too.

One way to potentially solve this is that in "fetch finale" we declare that this response is the one we need to hear about in terms of a done callback.

I suppose if "fetch finale" got a network error we could at that point invoke "done" as well, if it should happen for network errors. Do you get an entry if there's a CORS failure? I'm not sure that's currently adequately covered.

noamr:
In "fetch finale" we attach the timing info to the response. Intermediate redirect responses don't get that, so their `timing info` is null, and gets discarded when the `fetch params` are gone, so `report and finalize` will be a noop.

But I do see what you're saying, that `responseDone` is reported for intermediate responses - it should only be reported for responses with an attached `timing info` (we can make that more explicit with a flag)

annevk:
From IRC:

> if I do `fetch("https://www.google.com")` and then check the last entry of `window.performance.getEntriesByType("resource")` it's there, so network errors do result in entries afaict

And note that in this case the network error is due to CORS. I think we should either report all network errors or none of them.

Doing this would also require some kind of integration with "fetch finale" as only there do we know if a response became a network error.

noamr:
OK this is actually my oversight, CORS errors *should* be reported as an RT entry, but with some of the metrics zeroed. I'll post a PR to that effect.

noamr:
https://github.com/whatwg/fetch/pull/1202 addresses both issues now.

annevk:
Part of this was split into #1215.

PR #1202, opened 2021-03-25
processResponseDone should receive a response

noamr:
Only call processResponseDone for responses that have their
timing info set, which are responses that have passed CORS
and redirects.

Closes https://github.com/whatwg/fetch/issues/1201

***
<a href="https://whatpr.org/fetch/1202.html" title="Last updated on Apr 21, 2021, 10:57 AM UTC (b455c91)">Preview</a> | <a href="https://whatpr.org/fetch/1202/6453947...b455c91.html" title="Last updated on Apr 21, 2021, 10:57 AM UTC (b455c91)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7281,7 +7281,7 @@ method steps are:
   </ol>
 
  <li><p>Let <var>handleFetchDone</var> given <a for=/>response</a> <var>response</var> be to
- <a>finalize and report timing</a> with <var>responseObject</var>, <var>globalObject</var>, and
+ <a>finalize and report timing</a> with <var>response</var>, <var>globalObject</var>, and
  "<code>fetch</code>".
 
  <li>
@@ -7293,16 +7293,16 @@ method steps are:
    <li><p>If <var>locallyAborted</var> is true, terminate these substeps.
 
    <li><p>If <var>response</var>'s <a for=response>aborted flag</a> is set, then <a>abort fetch</a>
-   with <var>p</var>, <var>request</var>, and <var>response</var>, and terminate these
+   with <var>p</var>, <var>request</var>, and <var>responseObject</var>, and terminate these
    substeps.
 
    <li><p>If <var>response</var> is a <a>network error</a>, then <a for=/>reject</a> <var>p</var>
    with a {{TypeError}} and terminate these substeps.
 
-   <li><p>Set <var>response</var> to the result of <a for=Response>creating</a> a {{Response}}
+   <li><p>Set <var>responseObject</var> to the result of <a for=Response>creating</a> a {{Response}}
    object, given <var>response</var>, "<code>immutable</code>", and <var>relevantRealm</var>.
 
-   <li><p><a for=/>Resolve</a> <var>p</var> with <var>response</var>.
+   <li><p><a for=/>Resolve</a> <var>p</var> with <var>responseObject</var>.
   </ol>
 
  <li><p>Return <var>p</var>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -3988,8 +3988,8 @@ steps:
  <a for=request>done flag</a>.
 
  <li><p>If <var>timingInfo</var> is null, then return.
- 
- <li><p>Set <var>timingInfo</var>'s <a for="fetch timing info">response end time</a> to
+
+ <li><p>Set <var>timingInfo</var>'s <a for="fetch timing info">end time</a> to
  the <a for=/>coarsened shared current time</a> given <var>fetchParams</var>'s
  <a for="fetch params">cross-origin isolated capability</a>.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -4011,13 +4011,20 @@ steps:
 
  <li><p>Let <var>timingInfo</var> be <var>response</var>'s <a for=response>timing info</a>.
 
+ <li><p>Let <var>cacheState</var> be <var>response</var>'s <a for=response>cache state</a>.
+
  <li><p>If <var>timingInfo</var> is null, then return.
 
- <li><p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then set
- <var>timingInfo</var> to a new <a for=/>fetch timing info</a> whose
- <a for="fetch timing info">start time</a> and
- <a for="fetch timing info">post-redirect start time</a> are <var>timingInfo</var>'s
- <a for="fetch timing info">start time</a>.
+ <li><p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then
+ perform the following steps:
+ <ol>
+  <li><p>Set <var>timingInfo</var> to a new <a for=/>fetch timing info</a> whose
+  <a for="fetch timing info">start time</a> and
+  <a for="fetch timing info">post-redirect start time</a> are <var>timingInfo</var>'s
+  <a for="fetch timing info">start time</a>.
+
+  <li><p>Set <var>cacheState</var> to an empty string.
+ </ol>
 
  <li><p>Set <var>timingInfo</var>'s <a for="fetch timing info">end time</a> to the
  <a for=/>coarsened shared current time</a> given <var>global</var>'s
@@ -4027,7 +4034,8 @@ steps:
  <li><p>Set <var>response</var>'s <a for="response">timing info</a> to <var>timingInfo</var>.
 
  <li><p><a>Mark resource timing</a> for
- <var>timingInfo</var>, <var>originalURL</var>, <var>initiatorType</var>, and <var>global</var>.
+ <var>timingInfo</var>, <var>originalURL</var>, <var>initiatorType</var>, <var>global</var> and
+ <var>cacheState</var>.
  <!-- TODO -->
 </ol>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -1932,6 +1932,9 @@ message as HTTP/2 does not support them.
 <dfn export for=response id=concept-response-cache-state>cache state</dfn> (the empty string or
 "<code>local</code>"). Unlesss stated otherwise, it is the empty string.
 
+<p class=note>This is intended for usage by <cite>Service Workers</cite> and
+<cite>Resource Timing</cite>.  [[SW]]
+
 <p>A <a for=/>response</a> has an associated
 <dfn export for=response id=concept-response-csp-list>CSP list</dfn>, which is a <a for=/>list</a>
 of <a>Content Security Policy objects</a> for the <a for=/>response</a>. The list is empty unless
--- a/fetch.bs
+++ b/fetch.bs
@@ -1933,7 +1933,9 @@ message as HTTP/2 does not support them.
 "<code>local</code>"). Unlesss stated otherwise, it is the empty string.
 
 <p class=note>This is intended for usage by <cite>Service Workers</cite> and
-<cite>Resource Timing</cite>.  [[SW]]
+<cite>Resource Timing</cite>.  [[SW]] [[RESOURCE-TIMING]]
+<!-- If we ever expand the utility of this we need to carefully consider whether filtered responses
+     need to mask it, whether the cache API needs to store it, etc. -->
 
 <p>A <a for=/>response</a> has an associated
 <dfn export for=response id=concept-response-csp-list>CSP list</dfn>, which is a <a for=/>list</a>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4009,8 +4009,7 @@ steps:
  <li><p>If <var>timingInfo</var> is null, then return.
 
  <li>
-  <p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then
-  perform the following steps:
+  <p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then:
 
   <ol>
    <li><p>Set <var>timingInfo</var> to a new <a for=/>fetch timing info</a> whose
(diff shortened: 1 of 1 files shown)


noamr:
@annevk

fetch.bs
```diff
@@ -4019,7 +4027,7 @@ steps:
 
  <li><p>Set <var>response</var>'s <a for="response">timing info</a> to <var>timingInfo</var>.
 
- <li><p><a href="https://github.com/w3c/resource-timing/pull/261">Mark resource timing</a> for
+ <li><p><span>Mark resource timing</span> for
```
annevk, line 4030: You want to use `<a>` here.

fetch.bs
```diff
@@ -4251,9 +4259,20 @@ these steps:
...
-    <a>network error</a>.
+    <a>CORS check</a> for <var>request</var> and <var>response</var> returns failure, then:
+    <ul>
+     <li><p>Let <var>response</var> be a <a>network error</a>.
 
+     <li><p>Set <var>response</var>'s <a for=response>timing info</a> to a new
+     <a for=/>fetch timing info</a> with its <a for="fetch timing info">start time</a> and its
+     <a for="fetch timing info">post-redirect start time</a> set to <var>fetchParams</var>'s
+     <a for="fetch params">timing info</a>'s <a for="fetch timing info">start time</a>, and
+     its <a for="fetch timing info">end time</a> set to the <a for=/>coarsened shared current
```
annevk, line 4270: No newlines inside `<a>` et al.

fetch.bs
```diff
@@ -4251,9 +4259,20 @@ these steps:
 
    <li>
     <p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>cors</code>" and a
-    <a>CORS check</a> for <var>request</var> and <var>response</var> returns failure, then return a
-    <a>network error</a>.
+    <a>CORS check</a> for <var>request</var> and <var>response</var> returns failure, then:
+    <ul>
```
annevk, line 4263: This should be an ordered list.

annevk (commented):
I think this is problematic as this would distinguish a CORS network error from other network errors. That they are indistinguishable is a desirable security property.

Can we still remove this from implementations?

I checked other network errors, such as a bad port, and we don't seem to be creating RT entries there, but I haven't done exhaustive testing for all possible network errors.

I'm also a bit concerned that somewhere we might do if "response is a network error, return a network error" which would drop this kind of information. I don't think we've audited that as there hasn't been a reason for it thus far.

cc @yoavweiss @npm1 @rniwa @bdekoz @sefeng211

noamr:
> I think this is problematic as this would distinguish a CORS network error from other network errors. That they are indistinguishable is a desirable security property.
> 

All network errors that come after the CORS check should go through `fetch finale` and receive the RT entry, which only has a start/end time.
It's true that only CORS-or-after errors would receive an RT entry. Is that still too exposing?
I can remove this bit for now so that we can proceed with the other issues.

noamr:
Added https://github.com/whatwg/fetch/issues/1215 to address the CORS/network error issue
This PR handles other issues.

npm1:
> I think this is problematic as this would distinguish a CORS network error from other network errors. That they are indistinguishable is a desirable security property.

We expose errors in RT to prevent an attacker from distinguishing HTTP status codes (success vs errors). Would something similar apply here for CORS or not really (distinguishing success vs failed)?

annevk:
I see where the confusion lies. A HTTP status code of 4xx, 5xx, or even 9xx is not a network error. That's a normal response as far as fetch is concerned. A network error is something more fundamental or the result of a security policy, e.g.:

* DNS error
* TLS error
* Broken H/2 framing
* CORS error
* CSP block
* Mixed Content block

Whether a response with a non-2xx HTTP status code results in an "error" depends on the endpoint. For `<object>` and `<script>` it does, for `` it does not. (I.e., unfortunately 2xx vs non-2xx is exposed to some extent.)

I think in principle we could do one of these:

* expose timing information for all of these, including network errors (using TAO as a policy and treating network errors as not having TAO).
* expose timing information for responses, but not network errors
* expose timing information for responses that result in a "load" event on their endpoint

But what we cannot do (in my opinion) is distinguish between network errors.

fetch.bs
```diff
@@ -4017,9 +4032,10 @@ steps:
 
  <li><p>Set <var>response</var>'s <a for="response">timing info</a> to <var>timingInfo</var>.
 
- <li><p><a href="https://github.com/w3c/resource-timing/pull/261">Mark resource timing</a> for
- <var>timingInfo</var>, <var>originalURL</var>, <var>initiatorType</var>, and <var>global</var>.
- <!-- TODO -->
+ <li><p>
+ <a href="https://w3c.github.io/resource-timing/#dfn-mark-resource-timing">Mark resource timing</a>
```
annevk, line 4036: Why is this needed, is it not exported? Let's export it first.

fetch.bs
```diff
@@ -3981,13 +3981,21 @@ steps:
...
+ <li><p>Let <var>timingInfo</var> be <var>response</var>'s <a for=response>timing info</a>.
+
  <li><p>Set <var>fetchParams</var>'s <a for="fetch params">request</a>'s
  <a for=request>done flag</a>.
 
+ <li><p>If <var>timingInfo</var> is null, then return.
+
+ <li><p>Set <var>timingInfo</var>'s <a for="fetch timing info">end time</a> to
+ the <a for=/>coarsened shared current time</a> given <var>fetchParams</var>'s
+ <a for="fetch params">cross-origin isolated capability</a>.
```
annevk, line 3993: Why do we need to set end time twice?

fetch.bs
```diff
@@ -7217,8 +7233,9 @@ method steps are:
    <li><p><a lt=terminated for=fetch>Terminate</a> the ongoing fetch with the aborted flag set.
   </ol>
 
- <li><p>Let <var>handleFetchDone</var> be to <a>finalize and report timing</a> with
- <var>response</var>, <var>globalObject</var>, and "<code>fetch</code>".
+ <li><p>Let <var>handleFetchDone</var> given <a for=/>response</a> <var>response</var> be to
+ <a>finalize and report timing</a> with <var>response</var>, <var>globalObject</var>, and
+ "<code>fetch</code>".
```
annevk, line 7238: I think you should rename the _response_ variable here as otherwise it's confusing with the other one, no?

fetch.bs
```diff
@@ -4017,9 +4032,10 @@ steps:
 
  <li><p>Set <var>response</var>'s <a for="response">timing info</a> to <var>timingInfo</var>.
 
- <li><p><a href="https://github.com/w3c/resource-timing/pull/261">Mark resource timing</a> for
- <var>timingInfo</var>, <var>originalURL</var>, <var>initiatorType</var>, and <var>global</var>.
- <!-- TODO -->
+ <li><p>
+ <a href="https://w3c.github.io/resource-timing/#dfn-mark-resource-timing">Mark resource timing</a>
+ for <var>timingInfo</var>, <var>originalURL</var>, <var>initiatorType</var>, <var>global</var> and
```
annevk, line 4037: Oxford comma.

fetch.bs
```diff
@@ -4002,13 +4010,20 @@ steps:
...
+ <li><p>Let <var>cacheState</var> be <var>response</var>'s <a for=response>cache state</a>.
+
  <li><p>If <var>timingInfo</var> is null, then return.
 
- <li><p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then set
- <var>timingInfo</var> to a new <a for=/>fetch timing info</a> whose
- <a for="fetch timing info">start time</a> and
- <a for="fetch timing info">post-redirect start time</a> are <var>timingInfo</var>'s
- <a for="fetch timing info">start time</a>.
+ <li><p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then
```
annevk, line 4017: As this `<li>` contains multiple flow elements the `<p>` needs to indented on a newline.

annevk:
(I'm still okay with landing this modulo the above comments and leaving the error question to #1215.)

fetch.bs
noamr: Sure, waiting for Resource-Timing to be published to TR (bikeshed vs. Respec stuff).

noamr: Removing this bit for now, will update once the refs are resolvable.

fetch.bs
noamr: We don't actually. Removing.

fetch.bs
noamr: ok, renaming to `responseObject`

fetch.bs
noamr: ack

fetch.bs
noamr: ack

noamr:
> I see where the confusion lies. A HTTP status code of 4xx, 5xx, or even 9xx is not a network error. That's a normal response as far as fetch is concerned. A network error is something more fundamental or the result of a security policy, e.g.:
> 
> * DNS error
> * TLS error
> * Broken H/2 framing
> * CORS error
> * CSP block
> * Mixed Content block
> 
> Whether a response with a non-2xx HTTP status code results in an "error" depends on the endpoint. For `<object>` and `<script>` it does, for `` it does not. (I.e., unfortunately 2xx vs non-2xx is exposed to some extent.)
> 
> I think in principle we could do one of these:
> 
> * expose timing information for all of these, including network errors (using TAO as a policy and treating network errors as not having TAO).
> * expose timing information for responses, but not network errors
> * expose timing information for responses that result in a "load" event on their endpoint
> 
> But what we cannot do (in my opinion) is distinguish between network errors.

Let's continue the discussion in https://github.com/whatwg/fetch/issues/1215.

fetch.bs
```diff
@@ -3981,13 +3981,17 @@ steps:
 <a for=/>response</a> <var>response</var>, run these steps:
 
 <ol>
+ <li><p>Let <var>timingInfo</var> be <var>response</var>'s <a for=response>timing info</a>.
+
  <li><p>Set <var>fetchParams</var>'s <a for="fetch params">request</a>'s
  <a for=request>done flag</a>.
 
+ <li><p>If <var>timingInfo</var> is null, then return.
```
annevk, line 3989: Why do we have this here and also below in finalize and report timing? It seems this doesn't need to be here and you should get process response done either way?

fetch.bs
```diff
@@ -4002,13 +4006,21 @@ steps:
...
 
- <li><p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then set
- <var>timingInfo</var> to a new <a for=/>fetch timing info</a> whose
- <a for="fetch timing info">start time</a> and
- <a for="fetch timing info">post-redirect start time</a> are <var>timingInfo</var>'s
- <a for="fetch timing info">start time</a>.
+ <li>
+  <p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then
+  perform the following steps:
+  <ol>
```
annevk, line 4016: Newline before this.

fetch.bs
```diff
@@ -4002,13 +4006,21 @@ steps:
...
+    <li><p>Set <var>timingInfo</var> to a new <a for=/>fetch timing info</a> whose
```
annevk, line 4017: Single-space indentation.

fetch.bs
```diff
@@ -4017,9 +4029,10 @@ steps:
 
  <li><p>Set <var>response</var>'s <a for="response">timing info</a> to <var>timingInfo</var>.
 
- <li><p><a href="https://github.com/w3c/resource-timing/pull/261">Mark resource timing</a> for
- <var>timingInfo</var>, <var>originalURL</var>, <var>initiatorType</var>, and <var>global</var>.
- <!-- TODO -->
```
annevk, line 4022: Let's restore the TODO.

fetch.bs
annevk: Oh sorry, this can be _response_ because _responseObject_ is used in the outer scope. And _response_ is used below, but only within the scope of those substeps.

fetch.bs
```diff
@@ -1932,10 +1932,6 @@ message as HTTP/2 does not support them.
 <dfn export for=response id=concept-response-cache-state>cache state</dfn> (the empty string or
 "<code>local</code>"). Unlesss stated otherwise, it is the empty string.
 
-<p class=note>This is intended solely for usage by service workers. [[SW]]
-<!-- If we ever expand the utility of this we need to carefully consider whether filtered responses
-     need to mask it, whether the cache API needs to store it, etc. -->
```
annevk, line 1937: I think we should keep this including the comment and instead say something like

> `This is intended for usage by <cite>Service Workers</cite> and <cite>Resource Timing</cite>.`

I guess for Resource Timing we also need opaque responses (that have TAO) to leak this information. I'll have a look if that needs more work.

fetch.bs
noamr: Done

fetch.bs
noamr: Done

annevk:
@noamr I think it would be good if there was a corresponding PR to Resource Timing the commit could link to for it acknowledging the new _cacheState_ argument.

noamr:
> @noamr I think it would be good if there was a corresponding PR to Resource Timing the commit could link to for it acknowledging the new _cacheState_ argument.

It’s already there in the editors draft.

annevk:
I found the link above, it's https://github.com/w3c/resource-timing/pull/266.

PR #1204, opened 2021-03-30
Add a limited size buffer for streaming upload in HTTP request.

### LINKED ISSUE #538: Request body streams that cannot be retried
Currently our design is such that request body streams cannot be retried and that any redirects (other than 303) and HTTP authentication end up failing the request. This is rather nice for memory usage, but according to @ekr this plays badly with TLSv13 and QUIC which both assume requests can be retried.

Sigh.

cc @jakearchibald @yutakahirano @tyoshino @domenic @mnot

yoichio:
   * Chrome
   * Firefox
   * https://github.com/web-platform-tests/wpt/issues/28568
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1077174
   * Firefox: N/A as streaming request body has not been implemented
   * Safari: N/A as streaming request body has not been implemented

This fixes #538.

***
<a href="https://whatpr.org/fetch/1204.html" title="Last updated on Apr 19, 2021, 9:34 AM UTC (ff67a7a)">Preview</a> | <a href="https://whatpr.org/fetch/1204/f361d9c...ff67a7a.html" title="Last updated on Apr 19, 2021, 9:34 AM UTC (ff67a7a)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5063,14 +5063,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
      <li>
       <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
-      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer of
-      <a>implementation-defined</a> size and store a part of <var>request</var>'s
-      <a for=request>body</a> to the buffer. The buffer is used to resend <var>request</var>'s
-      <a for=request>body</a> to the network. If the user agent sent more than the buffer size and
-      is required to resend, return a <a>network error</a>.
+      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer up
+      to 64k bytes and store a part of <var>request</var>'s <a for=request>body</a> to the buffer.
+      If the user agent reads from <var>request</var>'s <a for=request>body</a> beyond the buffer
+      size, <var>request</var> is no longer eligible for resending.
 
       <div class="note no-backref">
-       <p>The re-sending is needed when the connection is timed out, for example.
+       <p>The resending is needed when the connection is timed out, for example.
 
        <p>The buffer is not needed when request's <a for=request>body</a>'s
        <a for=body>source</a> is non-null, because the <a for=request>body</a> can be re-created
--- a/fetch.bs
+++ b/fetch.bs
@@ -5066,7 +5066,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
       <a for=request>body</a>'s <a for=body>source</a> is null, then the user agent may have a
       buffer up to 64 kibibytes and store a part of <var>request</var>'s <a for=request>body</a> in
       the buffer. If the user agent reads from <var>request</var>'s <a for=request>body</a> beyond
-      the buffer size, then <var>request</var> is no longer eligible for resending.
+      the buffer size and the user agent is required to resend <var>request</var>, return a
+      return a <a>network error</a>.
 
       <div class=note>
        <p>The resending is needed when the connection is timed out, for example.
--- a/fetch.bs
+++ b/fetch.bs
@@ -5064,19 +5064,19 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
      <li>
       <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
       <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer up
-      to 64k bytes and store a part of <var>request</var>'s <a for=request>body</a> to the buffer.
+      to 64 kibibytes and store a part of <var>request</var>'s <a for=request>body</a> to the buffer.
       If the user agent reads from <var>request</var>'s <a for=request>body</a> beyond the buffer
       size, <var>request</var> is no longer eligible for resending.
 
       <div class="note no-backref">
        <p>The resending is needed when the connection is timed out, for example.
 
-       <p>The buffer is not needed when request's <a for=request>body</a>'s
+       <p>The buffer is not needed when <var>request</var>'s <a for=request>body</a>'s
        <a for=body>source</a> is non-null, because the <a for=request>body</a> can be recreated
-       from <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
+       from it.
 
-       <p>When <var>request</var>'s <a for=request>body</a>'s source is null, it means
-       <a for=request>body</a> is created from {{ReadableStream}}, which means
+       <p>When <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a> is null, it
+       means <a for=request>body</a> is created from {{ReadableStream}}, which means
        <a for=request>body</a> cannot be recreated and that is why the buffer is needed.
       </div>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -5066,7 +5066,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
       <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer up
       to 64 kibibytes and store a part of <var>request</var>'s <a for=request>body</a> to the buffer.
       If the user agent reads from <var>request</var>'s <a for=request>body</a> beyond the buffer
-      size, <var>request</var> is no longer eligible for resending.
+      size and the user agent is required to resend <var>request</var>, return a
+      <a>network error</a>.
 
       <div class="note no-backref">
        <p>The resending is needed when the connection is timed out, for example.
--- a/fetch.bs
+++ b/fetch.bs
@@ -5064,14 +5064,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
      <li>
       <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
       <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer of
-      <a>implementation-defined</a> size and store the part of <var>request</var>'s
+      <a>implementation-defined</a> size and store a part of <var>request</var>'s
       <a for=request>body</a> to the buffer. The buffer is used to resend <var>request</var>'s
       <a for=request>body</a> to the network. If the user agent sent more than the buffer size and
       is required to resend, return a <a>network error</a>.
 
       <div class="note no-backref">
-       <p>The re-sending in resolving the HTTP request is needed when the
-       connection is timed out, for example.
+       <p>The re-sending is needed when the connection is timed out, for example.
 
        <p>The buffer is not needed when request's <a for=request>body</a>'s
        <a for=body>source</a> is non-null, because the <a for=request>body</a> can be re-created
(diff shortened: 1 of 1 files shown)


yoichio:
Ping?

fetch.bs
```diff
@@ -5061,6 +5061,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
     <ul>
      <li><p>Follow the relevant requirements from HTTP. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
 
+     <li><p> In resolving the HTTP request, there is an evitable situaion that the user agent must
+     resend the <var>request</var> body (e.g. socket time-out).
+     If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+      <a for=request>body</a>'s <a for=body>source</a> is null, there is an <a>implementation-defined</a>
```
yutakahirano, line 5067: -1 indent

fetch.bs
```diff
@@ -5061,6 +5061,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
     <ul>
      <li><p>Follow the relevant requirements from HTTP. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
 
+     <li><p> In resolving the HTTP request, there is an evitable situaion that the user agent must
```
yutakahirano, line 5064: > evitable

inevitable?

fetch.bs
```diff
@@ -5061,6 +5061,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+     resend the <var>request</var> body (e.g. socket time-out).
+     If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+      <a for=request>body</a>'s <a for=body>source</a> is null, there is an <a>implementation-defined</a>
+      cache limit not to cache the entire <a for=request>body</a>'s stream. If the cache runs out,
```
yutakahirano, line 5068: I think "cache" is not good here. It is a storage providing a faster way to read/write, but a cache shouldn't provide additional functionality - we need to be able to do everything without cache.

In this case, without the storage we cannot resend the request body, so there is a functional difference, and hence it's not cache.

I  suggest using "buffer" instead of "cache".

fetch.bs
```diff
@@ -5061,6 +5061,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+      return a <a>network error</a>.
```
yutakahirano, line 5069: This paragraph sounds like a mixture of normative text and non-normative notes. How about the following?

The user agent may have a buffer of an implementation-defined size and store the part of request's body to the buffer, when request's body is non-null and its source is null. The user agent may use the buffer to re-send request's body to the network.

(note) The re-sending is needed when the connection is timed out, for example. If request's body is sent more than the buffer size, it leads to a network error.

(note) The buffer is not needed when request's body's source is non-null, because the body can be re-created from request's body's source.

yoichio: Done.

fetch.bs
```diff
@@ -5061,6 +5061,27 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+     <li>
+      <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer of
+      <a>implementation-defined</a> size and store the part of <var>request</var>'s
+      <a for=request>body</a> to the buffer. The buffer is used to resend <var>request</var>'s
+      <a for=request>body</a> to the network. If the user agent sent more than the buffer size and
+      is required to resend, return a <a>network error</a>.
+
+      <div class="note no-backref">
+       <p>The re-sending in resolving the HTTP request is needed when the
```
yutakahirano, line 5073: Is "in resolving the HTTP request" needed?

fetch.bs
```diff
@@ -5061,6 +5061,27 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
     <ul>
      <li><p>Follow the relevant requirements from HTTP. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
 
+     <li>
+      <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer of
+      <a>implementation-defined</a> size and store the part of <var>request</var>'s
```
yutakahirano, line 5067: Sorry this was my suggestion, but "a part of" is better than "the part of" here.

yutakahirano:
LGTM. Can you update the PR title?

@annevk: do you want to take a look at this?

annevk:
I'm concerned about this not having an upper bound as I wrote in https://github.com/whatwg/fetch/issues/538#issuecomment-617566746. I think there should be an upper bound.

fetch.bs
```diff
@@ -5061,6 +5061,26 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
     <ul>
      <li><p>Follow the relevant requirements from HTTP. [[!HTTP]] [[!HTTP-SEMANTICS]] [[!HTTP-COND]] [[!HTTP-CACHING]] [[!HTTP-AUTH]]
 
+     <li>
+      <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer of
+      <a>implementation-defined</a> size and store a part of <var>request</var>'s
+      <a for=request>body</a> to the buffer. The buffer is used to resend <var>request</var>'s
+      <a for=request>body</a> to the network. If the user agent sent more than the buffer size and
+      is required to resend, return a <a>network error</a>.
```
annevk, line 5070: I think it could be slightly clearer that this buffer doesn't contain the entire body and that if the user agent reads from the stream beyond the buffer size, the request is no longer eligible for resending. Rather than saying it's about what was sent as that doesn't matter as much.

yoichio: I agree to emphasizing this buffer is not big as much as stream.
However this buffer is only needed if body is stream.
As https://bugs.chromium.org/p/chromium/issues/detail?id=1077174, Chrome adopts 64k bytes.
Does this sound good?

ddragana: 64k sound reasonable.

yoichio: Done.

fetch.bs
```diff
@@ -5063,14 +5063,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
      <li>
       <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
-      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer of
-      <a>implementation-defined</a> size and store a part of <var>request</var>'s
-      <a for=request>body</a> to the buffer. The buffer is used to resend <var>request</var>'s
-      <a for=request>body</a> to the network. If the user agent sent more than the buffer size and
-      is required to resend, return a <a>network error</a>.
+      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer up
+      to 64k bytes and store a part of <var>request</var>'s <a for=request>body</a> to the buffer.
```
annevk, line 5067: Please write this out as we do elsewhere, "64 kibibytes" (I'm assuming we don't want literal kilobytes in this instance).

fetch.bs
```diff
@@ -5061,6 +5061,25 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+      If the user agent reads from <var>request</var>'s <a for=request>body</a> beyond the buffer
+      size, <var>request</var> is no longer eligible for resending.
+
+      <div class="note no-backref">
+       <p>The resending is needed when the connection is timed out, for example.
+
+       <p>The buffer is not needed when request's <a for=request>body</a>'s
+       <a for=body>source</a> is non-null, because the <a for=request>body</a> can be re-created
```
annevk, line 5075:
```suggestion
       <a for=body>source</a> is non-null, because the <a for=request>body</a> can be recreated
```

fetch.bs
```diff
@@ -5061,6 +5061,25 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+       from <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
+
+       <p>When <var>request</var>'s <a for=request>body</a>'s source is null, it means
+       <a for=request>body</a> is created from {{ReadableStream}}, which means
+       <a for=request>body</a> can not be re-created and that's because why the buffer is needed.
```
annevk, line 5080:
```suggestion
       <a for=request>body</a> cannot be recreated and that is why the buffer is needed.
```

fetch.bs
```diff
@@ -5061,6 +5061,25 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+      size, <var>request</var> is no longer eligible for resending.
+
+      <div class="note no-backref">
+       <p>The resending is needed when the connection is timed out, for example.
+
+       <p>The buffer is not needed when request's <a for=request>body</a>'s
+       <a for=body>source</a> is non-null, because the <a for=request>body</a> can be re-created
+       from <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
+
+       <p>When <var>request</var>'s <a for=request>body</a>'s source is null, it means
```
annevk, line 5078: xref source

fetch.bs
```diff
@@ -5061,6 +5061,25 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+      <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
+      <a for=request>body</a>'s <a for=body>source</a> is null, the user agent may have a buffer up
+      to 64k bytes and store a part of <var>request</var>'s <a for=request>body</a> to the buffer.
+      If the user agent reads from <var>request</var>'s <a for=request>body</a> beyond the buffer
+      size, <var>request</var> is no longer eligible for resending.
+
+      <div class="note no-backref">
+       <p>The resending is needed when the connection is timed out, for example.
+
+       <p>The buffer is not needed when request's <a for=request>body</a>'s
```
annevk, line 5074: request in `<var>`

fetch.bs
```diff
@@ -5061,6 +5061,25 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+       <a for=body>source</a> is non-null, because the <a for=request>body</a> can be re-created
+       from <var>request</var>'s <a for=request>body</a>'s <a for=body>source</a>.
```
annevk, line 5076: from it* would also work here.

annevk (commented):
Thanks, this looks good modulo a couple tiny nits.

annevk:
We probably need to say explicitly that a network error is returned if you need to resend and the buffer became useless?

Also, let's file an issue against web-platform-tests outlining what would be needed to test this.

Other than that we're good I think.

yoichio:
Done both.

annevk (approved):
I made some final tweaks, does this still look good to you @yoichio?

yoichio:
LGTM.

PR #1228, opened 2021-05-02
Associate service worker timing with a response

noamr:
In preperation for https://github.com/w3c/ServiceWorker/pull/1575 and https://github.com/w3c/navigation-timing/pull/143, the response has to have an associated service worker timing. This would only affect navigation timing, where the actual start/ready timing of the service worker is a meaningful metric.


fetch.bs
```diff
@@ -1973,6 +1973,10 @@ this is also tracked internally using the request's <a for=request>timing allow
 <dfn for=response id=concept-response-timing-info>timing info</dfn> (null or a
 <a for=/>fetch timing info</a>), which is initially null.
 
+<p>A <a for=/>response</a> has an associated
+<dfn export for=response id=concept-response-service-worker-timing-info>service worker timing</dfn> (null or a
```
annevk, line 1977: Exceeds 100 columns.

annevk:
It would also be good to have a commit message that links the two service worker PRs to make it clear this is part of a larger whole.

fetch.bs
annevk: Also no need for the ID here and I wonder if we should name this field "service worker timing info" for consistency with "timing info".

noamr:
> @noamr I still think we should name fields on response with some consistency.

Sure, but can you be more specific?

PR #1229, opened 2021-05-03
Specify the behavior of `COEP: credentialless`,

ArthurSonzogni:
Originally described in: https://github.com/mikewest/credentiallessness

`credentialless` and `require-corp` are similar. One or the other is a requirements for the `window.crossOriginIsolated` capability.
They differ mostly in the fetch specification. `require-corp` requires a CORP header for cross-origin no-cors responses. `credentialless` doesn't, but omits credentials (Cookies, clients certificates, etc...) in the request.

* HTML (https://github.com/whatwg/html/pull/6638)
  * Define how to parse the `credentialless` value.
  * From the HTML spec point of view, `credentialless` and `require-corp` are equivalent. They have been grouped into `compatible with cross-origin isolation` and the HTML spec rewritten to use this concept.

* Fetch: (https://github.com/whatwg/fetch/pull/1229)
  * Define `Cross-Origin-Embedder-Policy allows credentials` algorithm. It omit credentials for no-cors, cross-origin, COEP:credentialless requests.
  * Define `response's` `request-include-credentials` flag.
  * In the `Cross-Origin-Resource-Policy check`, if `embedderPolicy` is `credentialless`, require CORP for navigational responses, and opaque responses with `request-include-credentials`.

See: https://github.com/whatwg/html/issues/6637

----

   * Chrome: https://chromestatus.com/feature/4918234241302528#details
   * Firefox: https://github.com/mozilla/standards-positions/issues/539  (worth prototyping)
   * Safari: https://lists.webkit.org/pipermail/webkit-dev/2021-June/031898.html (pending)
  
   * https://wpt.fyi/results/html/cross-origin-embedder-policy/credentialless

- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/1175099
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1731778
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=230550

***
<a href="https://whatpr.org/fetch/1229.html" title="Last updated on Nov 3, 2021, 7:54 AM UTC (60a8cf1)">Preview</a> | <a href="https://whatpr.org/fetch/1229/b2f04e2...60a8cf1.html" title="Last updated on Nov 3, 2021, 7:54 AM UTC (60a8cf1)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1892,8 +1892,8 @@ source of security bugs. Please seek security review for features that deal with
  <a for="URL serializer"><i>exclude fragment</i></a> set to true.
 </ol>
 
-<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a
-<a for=/>request</a> <var>request</var>, run theses steps:
+<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a <a
+for=/>request</a> <var>request</var>, run theses steps:
 
 <ol>
  <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", return
@@ -1902,12 +1902,11 @@ source of security bugs. Please seek security review for features that deal with
  <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
 
  <li><p>If <var>request</var>'s <a for=request>client</a>'s <a for="environment settings
- object">embedder policy</a> is not
- "<code><a for="embedder policy value">credentialless</a></code>", return true.</p>
+ object">embedder policy</a> is not "<code><a for="embedder policy
+ value">credentialless</a></code>", return true.</p>
 
- <li><p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a>
- with <var>request</var>'s <a for=request>current URL</a>'s <a
- for=url>origin</a>, return true.</p>
+ <li><p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
+ <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, return true.</p>
 
  <li><p>Return false.</p>
 </ol>
@@ -3536,8 +3535,7 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
    <dd> Do nothing.
 
    <dt>`<a for="embedder policy value">credentialless</a>`
-   <dd> Set <var>policy</var> to `<code>same-origin</code>` if one of the
-   following is true:
+   <dd> Set <var>policy</var> to `<code>same-origin</code>` if one of the following is true:
    <ul>
     <li><var>response</var>'s <a for="response">request-include-credentials</a> is true.
     <li><var>forNavigation</var> is true.
@@ -4705,9 +4703,8 @@ steps. They return a <a for=/>response</a>.
 
     <p>is true; otherwise false.
 
-   <li>
-    <p>If <a>Cross-Origin-Embedder-Policy allows credentials</a> with
-    <var>request</var> is false, set <var>includeCredentials</var> to false.</p>
+   <li><p>If <a>Cross-Origin-Embedder-Policy allows credentials</a> with <var>request</var> is
+   false, set <var>includeCredentials</var> to false.</p>
 
    <li><p>Let <var>contentLength</var> be <var>httpRequest</var>'s <a for=request>body</a>'s
    <a for=body>length</a>, if <var>httpRequest</var>'s <a for=request>body</a> is non-null;
--- a/fetch.bs
+++ b/fetch.bs
@@ -3533,6 +3533,8 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
  <li><p>If <var>policy</var> is null, switch on <var>embedderPolicyValue</var>:
   <dl class=switch>
    <dt>`<a for="embedder policy value">unsafe-none</a>`
+   <dd> Do nothing.
+
    <dt>`<a for="embedder policy value">credentialless</a>`
    <dd> Set <var>policy</var> to `<code>same-origin</code>` if one of the
    following is true:
--- a/fetch.bs
+++ b/fetch.bs
@@ -1905,7 +1905,8 @@ source of security bugs. Please seek security review for features that deal with
  is not "<code><a for="embedder policy value">credentialless</a></code>", return true.</p>
 
  <li><p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
- <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, return true.</p>
+ <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> and
+ <var>request</var>'s <a for=request>tainted origin flag</a> is not set, return true.</p>
 
  <li><p>Return false.</p>
 </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -1896,14 +1896,15 @@ source of security bugs. Please seek security review for features that deal with
 <a for=/>request</a> <var>request</var>, run theses steps:
 
 <ol>
- <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", return
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", then return
  true.</p>
 
  <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
 
  <li><p>If <var>request</var>'s <a for=request>client</a>'s
  <a for="environment settings object">policy container</a>'s
- <a for="policy container">embedder policy</a> is not
+ <a for="policy container">embedder policy</a>'s
+ <a for="embedder policy">value</a> is not
  "<code><a for="embedder policy value">credentialless</a></code>", return true.</p>
 
  <li><p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
@@ -3535,17 +3536,18 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
   <p>If <var>policy</var> is null, switch on <var>embedderPolicyValue</var>:</p>
   <dl class=switch>
    <dt>`<a for="embedder policy value">unsafe-none</a>`
-   <dd> Do nothing.
+   <dd><p>Do nothing.
 
    <dt>`<a for="embedder policy value">credentialless</a>`
-   <dd> Set <var>policy</var> to `<code>same-origin</code>` if one of the following is true:
-   <ul>
-    <li><var>response</var>'s <a for="response">request-include-credentials</a> is true.
-    <li><var>forNavigation</var> is true.
-   </ul>
+   <dd>
+    <p>Set <var>policy</var> to `<code>same-origin</code>` if one of the following is true:
+    <ul>
+     <li><var>response</var>'s <a for="response">request-include-credentials</a> is true.
+     <li><var>forNavigation</var> is true.
+    </ul>
 
    <dt>`<a for="embedder policy value">require-corp</a>`
-   <dd> Set <var>policy</var> to `<code>same-origin</code>`.
+   <dd><p>Set <var>policy</var> to `<code>same-origin</code>`.
   </dl>
  </li>
 
@@ -4707,7 +4709,7 @@ steps. They return a <a for=/>response</a>.
     <p>is true; otherwise false.
 
    <li><p>If <a>Cross-Origin-Embedder-Policy allows credentials</a> with <var>request</var> returns
-   false, set <var>includeCredentials</var> to false.</p>
+   false, then set <var>includeCredentials</var> to false.</p>
 
    <li><p>Let <var>contentLength</var> be <var>httpRequest</var>'s <a for=request>body</a>'s
    <a for=body>length</a>, if <var>httpRequest</var>'s <a for=request>body</a> is non-null;
--- a/fetch.bs
+++ b/fetch.bs
@@ -1892,6 +1892,25 @@ source of security bugs. Please seek security review for features that deal with
  <a for="URL serializer"><i>exclude fragment</i></a> set to true.
 </ol>
 
+<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a
+<a for=/>request</a> <var>request</var>, run theses steps:
+
+<ol>
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", return
+ true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a>'s <a for="environment settings
+ object">embedder policy</a> is not
+ "<code><a for="embedder policy value">credentialless</a></code>", return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>origin</a> is not <a>same origin</a> with
+ <var>request</var>'s <a for=request>client</a>'s <a for="environment settings object">origin</a>,
+ return true.</p>
+
+ <li><p>Return false.</p>
+</ol>
 
 <h4 id=responses>Responses</h4>
 
@@ -1977,6 +1996,10 @@ initially unset.
 <p class=note>This is used to ensure to prevent a partial response from an earlier ranged request
 being provided to an API that didn't make a range request. See the flag's usage for a detailed
 description of the attack.
+
+<p>A <a for=/>response</a> has an associated <dfn for=response
+id=concept-response-request-include-credentials>request-include-credentials</dfn>, which is
+initially set.
 
 <p>A <a for=/>response</a> has an associated
 <dfn for=response id=concept-response-timing-allow-passed>timing allow passed flag</dfn>, which is
@@ -3507,9 +3530,22 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
  <li><p>If <var>policy</var> is neither `<code>same-origin</code>`, `<code>same-site</code>`, nor
  `<code>cross-origin</code>`, then set <var>policy</var> to null.
 
- <li><p>If <var>policy</var> is null and <var>embedderPolicyValue</var> is
- "<code><a for="embedder policy value">require-corp</a></code>", then set <var>policy</var> to
- `<code>same-origin</code>`.
+ <li><p>If <var>policy</var> is null, switch on <var>embedderPolicyValue</var>:
+  <dl class=switch>
+   <dt>`<a for="embedder policy value">unsafe-none</a>`
+   <dt>`<a for="embedder policy value">credentialless</a>`
+   <dd> Set <var>policy</var> to `<code>same-origin</code>` if one of the
+   following is true:
+   <ul>
+    <li><var>response</var>'s <a for="response">request-include-credentials</a> is true and
+        <var>response</var>'s <a for="response">type</a> is "<code>opaque</code>".
+    <li><var>forNavigation</var> is true.
+   </ul>
+
+   <dt>`<a for="embedder policy value">require-corp</a>`
+   <dd> Set <var>policy</var> to `<code>same-origin</code>`.
+  </dl>
+ </li>
 
  <li>
   <p>Switch on <var>policy</var>:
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
I wrote page gathering the two PR in a single page if you want to have a broader view:
https://htmlpreview.github.io/?https://github.com/mikewest/credentiallessness/blob/master/index.html

Would you have some suggestion?

The warning I have:
```
LINK ERROR: No 'dfn' refs found for 'credentialless'.
&lt;a data-link-for="embedder policy value" data-link-type="dfn" data-lt="credentialless">credentialless&lt;/a>
```
will be resolved when landing the HTML PR.

fetch.bs
```diff
@@ -1892,6 +1892,25 @@ source of security bugs. Please seek security review for features that deal with
  <a for="URL serializer"><i>exclude fragment</i></a> set to true.
 </ol>
 
+<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a
```
yutakahirano, line 1895: This disallows credentials to same-origin requests with mode: no-cors. Is it OK?

fetch.bs
```diff
@@ -1892,6 +1892,25 @@ source of security bugs. Please seek security review for features that deal with
...
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", return
+ true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a>'s <a for="environment settings
+ object">embedder policy</a> is not
+ "<code><a for="embedder policy value">credentialless</a></code>", return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>origin</a> is not <a>same origin</a> with
```
yutakahirano, line 1908: Can you tell me why this is needed? (It might be good to leave a note).

ArthurSonzogni: I missed inverting the condition in the early return. This was meant to be the opposite, to skip same-origin requests.

yutakahirano:
Don't you want to compare request's current URL's origin and request's origin, rather than request's origin and request's client's origin?

ArthurSonzogni:
> Don't you want to compare request's current URL's origin and request's origin, rather than request's origin and request's client's origin?

Yes, this should be the `current URL` instead of the `URL`. Fixed in "Address Yutaka comment #2"
However the second origin should continue to be the client's origin. For instance, in the base case of a cross-origin no-cors request with no redirects, credentials must not be included. So the comparison is in between the client's origin and the request's current URL.

yutakahirano:
> However the second origin should continue to be the client's origin. For instance, in the base case of a cross-origin no-cors request with no redirects, credentials must not be included. So the comparison is in between the client's origin and the request's current URL.

Sorry, can you give me an example? request's origin and request's client's origin are usually the same, but other checks such as [CORS](https://fetch.spec.whatwg.org/#cors-check) use request's origin, so I want to understand why in this case we need to use the client's origin.

ArthurSonzogni:
> > However the second origin should continue to be the client's origin. For instance, in the base case of a cross-origin no-cors request with no redirects, credentials must not be included. So the comparison is in between the client's origin and the request's current URL.
> 
> Sorry, can you give me an example? request's origin and request's client's origin are usually the same, but other checks such as [CORS](https://fetch.spec.whatwg.org/#cors-check) use request's origin, so I want to understand why in this case we need to use the client's origin.

In the example above, a.com requests b.com in no-cors mode, with COEP:credentialless.
This gives:
- request’s mode == "no-cors"
- request's client is not null.
- request's client's embedderPolicy is "credentialless".
- request's client's origin is "a.com"
- request's currentURL's origin is b.com   (a.com != b.com)

That's the 4 condition in the "COEP allow credentials" algorithm and it return false.

yutakahirano:
> What is the main difference in between request's origin and request's client's origin?

For example, [process a nativate fetch](https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-fetch) manually sets request's origin (step 4.1).

> Do you foresee a case where the difference matter for COEP:credentialless?

Regarding this algorithm, no. For requests with mode: "no-cors", I believe they match. In any case, request's origin is the source of truth for other checks, so using it here is better for consistency.

ArthurSonzogni:
> Regarding this algorithm, no. For requests with mode: "no-cors", I believe they match. In any case, request's origin is the source of truth for other checks, so using it here is better for consistency.

Thanks! I addressed this in 9f0be6a.

fetch.bs
```diff
@@ -4585,6 +4620,9 @@ steps. They return a <a for=/>response</a>.
 
     <p>is true; otherwise false.
 
+   <li><p>If <a>Cross-Origin-Embedder-Policy allows credentials</a> with <var>request</var> is
```
yutakahirano, line 4623: maybe "returns" is better than "is" here.

ArthurSonzogni: Done.

ArthurSonzogni (commented):
Thanks Yutaka,
I addressed your 3 comments in the latest commit.

fetch.bs
```diff
@@ -1892,6 +1892,24 @@ source of security bugs. Please seek security review for features that deal with
  <a for="URL serializer"><i>exclude fragment</i></a> set to true.
 </ol>
 
+<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a <a
+for=/>request</a> <var>request</var>, run theses steps:
```
yutakahirano, line 1896: Let's avoid having a new line in an inline element.

fetch.bs
```diff
@@ -1978,6 +1996,10 @@ initially unset.
 being provided to an API that didn't make a range request. See the flag's usage for a detailed
 description of the attack.
 
+<p>A <a for=/>response</a> has an associated <dfn for=response
```
yutakahirano, line 1999: ditto

yutakahirano (approved):

fetch.bs
domenic: Note that this rule differs between whatwg/html and whatwg/fetch :(. We are sorry about that.

ArthurSonzogni: Done.

fetch.bs
ArthurSonzogni: Done.

fetch.bs
```diff
@@ -1892,6 +1892,24 @@ source of security bugs. Please seek security review for features that deal with
  <a for="URL serializer"><i>exclude fragment</i></a> set to true.
 </ol>
 
+<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a
+<a for=/>request</a> <var>request</var>, run theses steps:
+
+<ol>
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", return
```
annevk, line 1899: Missing quotation mark.

annevk: then return* (throughout)

fetch.bs
```diff
@@ -1892,6 +1892,24 @@ source of security bugs. Please seek security review for features that deal with
...
+ true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a>'s <a for="environment settings
```
annevk, line 1904: No newlines inside phrasing elements, including in tags.

fetch.bs
```diff
@@ -3421,9 +3443,22 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
  <li><p>If <var>policy</var> is neither `<code>same-origin</code>`, `<code>same-site</code>`, nor
  `<code>cross-origin</code>`, then set <var>policy</var> to null.
 
- <li><p>If <var>policy</var> is null and <var>embedderPolicyValue</var> is
- "<code><a for="embedder policy value">require-corp</a></code>", then set <var>policy</var> to
- `<code>same-origin</code>`.
+ <li><p>If <var>policy</var> is null, switch on <var>embedderPolicyValue</var>:
```
annevk, line 3446: `<li>` contains multiple children so `<p>` needs to be on a new line.

fetch.bs
```diff
@@ -1892,6 +1892,24 @@ source of security bugs. Please seek security review for features that deal with
...
+ true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a>'s <a for="environment settings
+ object">embedder policy</a> is not
+ "<code><a for="embedder policy value">credentialless</a></code>", return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>origin</a> is <a>same origin</a> with
+ <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>, return true.</p>
```
annevk, line 1909: This check seems sketchy. What about redirects? Should this use tainted origin flag instead?

fetch.bs
ArthurSonzogni: Done

fetch.bs
ArthurSonzogni: Done.

fetch.bs
ArthurSonzogni: Done:
> Missing quotation mark.

But, I have trouble understanding what means:
> then return* (throughout)

fetch.bs
ArthurSonzogni: It was described by: https://wicg.github.io/credentiallessness/#redirect

> The decision to include credentials is done independently for each request. The variable includeCredentials is set for the initial request, but also after each redirect.
>
> For example, credentials are not included for a cross-origin no-cors request, but they can be added in the next request if it redirects to a same-origin resource.

Tested by: [redirect.tentative.html](https://wpt.fyi/results/html/cross-origin-embedder-policy/credentialless/redirect.tentative.html?label=master&label=experimental&aligned&q=redirect.tentative.html)

The tainted origin flag, could be used to omit credentials whenever there is a single cross-origin URL in the redirect chain. However, that's not what was originally envisioned. Do you think there is an argument if favor of a different behavior?

annevk: It means that in A -> B -> A, the final A isn't considered cross-origin, which is generally a thing we consider a bug, due to confused deputy attacks. It's considered cross-origin for CORS, tainting of responses, etc.

ArthurSonzogni: It was made this way to avoid breaking more clients than necessary. Goal is to make COEP deployable easily.

In the threat model, the attacker wants to use SAB to read a non-public cross-origin resource. So, allowing credentials for same-origin resource doesn't go against it.

However, I agree doing something similar to other logic would bring some kind of purity here. That's good too.

Do you feel strongly against it? This probably deserve some attention from @mikewest or @arturjanc with whom we discussed it in the design doc.

mikewest: I don't remember discussing this in particular, so apologies if I disagree with my past self: at the moment, I think it's probably best to match our behavior for `Sec-Fetch-Site` which will treat a request as `cross-origin` for A->B->A transitions (https://w3c.github.io/webappsec-fetch-metadata/#:~:text=for%20each%20url%20in%20r%E2%80%99s%20url%20list%3A). We didn't use the tainted origin flag there, both because I didn't know about it (TIL!), and because we needed to distinguish `cross-origin` and `same-site`.

It might be reasonable to use here, since we don't care about the site boundary in this mechanism.

ArthurSonzogni: If you both feel this is reasonable thing to do, then I guess... let's do it.
It was actually way simpler to implement this way.

However, I haven't seen a real security reason to do it here. We can ship with this strict version and consider relaxing it later, if we discover this creates some difficulties with developers.

annevk: The security issue is explained here: https://en.wikipedia.org/wiki/Confused_deputy_problem. B gets to make a request against A with the authority of A if you do not consider it to be cross-origin.

ArthurSonzogni: annevk@, I addressed your suggestion. In the latest patchset, we take into account the tainted origin flag.

This gives:
Capture d’écran du 2021-09-20 15-11-03

Note: Thanks for the link about "Confused deputy problem", however I still don't see there are anything about `COEP:credentialless ` and the threat model about COI & spectre attacks where this really matter.

ArthurSonzogni: (The corresponding WPT/chrome [change](https://github.com/web-platform-tests/wpt/pull/30871/files))

annevk: It's not a Spectre-related threat. Say you adopt credentialless to avoid sending credentials to other origins. You might reasonably expect that only same-origin requests contain credentials. But you might not expect that those other origins can also make same-origin requests through redirects. And they really shouldn't be able to do that as they don't have your authority.

annevk:
From CI it seems like this needs rebasing or at least some modernizing as embedder policy is now part of policy container.

ArthurSonzogni:
I rebased over PolicyContainer. See the latest patchset for the [difference](https://github.com/whatwg/fetch/pull/1229/commits/3c486868fe1dfacbfefe88992856b106a80c9474).
As a result, this error disapeared:
```
LINK ERROR: No 'dfn' refs found for 'embedder policy' with for='['environment settings object']'.
```

However, this would still require the HTML specification to land, in order to understand the `credentialless` value:
```
LINK ERROR: No 'dfn' refs found for 'credentialless'.
```

(I double checked there was no other LINK error beside this one, by replacing `credentialless` occurence with `require-corp` and checking this was correctly linking)

fetch.bs
annevk: Ah sorry I missed this, I meant that instead of "return", write "then return". That's the convention this specification uses for if statements.

fetch.bs
```diff
@@ -1892,6 +1892,26 @@ source of security bugs. Please seek security review for features that deal with
...
+
+<ol>
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is not <code>no-cors</code>", return
+ true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a> is null, return true.</p>
+
+ <li><p>If <var>request</var>'s <a for=request>client</a>'s
+ <a for="environment settings object">policy container</a>'s
+ <a for="policy container">embedder policy</a> is not
```
annevk, line 1906: This should be embedder policy's value, no?

fetch.bs
```diff
@@ -3507,9 +3531,23 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
...
  `<code>cross-origin</code>`, then set <var>policy</var> to null.
 
- <li><p>If <var>policy</var> is null and <var>embedderPolicyValue</var> is
- "<code><a for="embedder policy value">require-corp</a></code>", then set <var>policy</var> to
- `<code>same-origin</code>`.
+ <li>
+  <p>If <var>policy</var> is null, switch on <var>embedderPolicyValue</var>:</p>
+  <dl class=switch>
+   <dt>`<a for="embedder policy value">unsafe-none</a>`
+   <dd> Do nothing.
```
annevk, line 3538:
```suggestion
   <dd><p>Do nothing.
```

fetch.bs
```diff
@@ -3507,9 +3531,23 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
...
+
+   <dt>`<a for="embedder policy value">credentialless</a>`
+   <dd> Set <var>policy</var> to `<code>same-origin</code>` if one of the following is true:
```
annevk, line 3541: You also need to indent the `<ul>` below accordingly.
```suggestion
   <dd>
    <p>Set <var>policy</var> to `<code>same-origin</code>` if one of the following is true:
```

fetch.bs
```diff
@@ -3507,9 +3531,23 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
...
+   <ul>
+    <li><var>response</var>'s <a for="response">request-include-credentials</a> is true.
+    <li><var>forNavigation</var> is true.
+   </ul>
+
+   <dt>`<a for="embedder policy value">require-corp</a>`
+   <dd> Set <var>policy</var> to `<code>same-origin</code>`.
```
annevk, line 3548:
```suggestion
   <dd><p>Set <var>policy</var> to `<code>same-origin</code>`.
```

fetch.bs
```diff
@@ -4668,6 +4706,9 @@ steps. They return a <a for=/>response</a>.
 
     <p>is true; otherwise false.
 
+   <li><p>If <a>Cross-Origin-Embedder-Policy allows credentials</a> with <var>request</var> returns
+   false, set <var>includeCredentials</var> to false.</p>
```
annevk, line 4710:
```suggestion
   false, then set <var>includeCredentials</var> to false.</p>
```

annevk (commented):
Modulo nits this looks okay to me.

fetch.bs
ArthurSonzogni: Done in https://github.com/whatwg/fetch/pull/1229/commits/89aaca1fdd377c5b10133cceb56a100ca618fd27

fetch.bs
ArthurSonzogni: Done in https://github.com/whatwg/fetch/pull/1229/commits/89aaca1fdd377c5b10133cceb56a100ca618fd27

fetch.bs
ArthurSonzogni: Done in https://github.com/whatwg/fetch/pull/1229/commits/89aaca1fdd377c5b10133cceb56a100ca618fd27

fetch.bs
ArthurSonzogni: Done in https://github.com/whatwg/fetch/pull/1229/commits/89aaca1fdd377c5b10133cceb56a100ca618fd27

fetch.bs
ArthurSonzogni: Done in https://github.com/whatwg/fetch/pull/1229/commits/89aaca1fdd377c5b10133cceb56a100ca618fd27

fetch.bs
```diff
@@ -1892,6 +1892,27 @@ source of security bugs. Please seek security review for features that deal with
  <a for="URL serializer"><i>exclude fragment</i></a> set to true.
 </ol>
 
+<p>To check <dfn export>Cross-Origin-Embedder-Policy allows credentials</dfn>, given a
+<a for=/>request</a> <var>request</var>, run theses steps:
```
domenic, line 1896:
```suggestion
<a for=/>request</a> <var>request</var>, run these steps:
```

fetch.bs
```diff
@@ -3507,9 +3532,24 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
  <li><p>If <var>policy</var> is neither `<code>same-origin</code>`, `<code>same-site</code>`, nor
  `<code>cross-origin</code>`, then set <var>policy</var> to null.
 
- <li><p>If <var>policy</var> is null and <var>embedderPolicyValue</var> is
- "<code><a for="embedder policy value">require-corp</a></code>", then set <var>policy</var> to
- `<code>same-origin</code>`.
+ <li>
+  <p>If <var>policy</var> is null, switch on <var>embedderPolicyValue</var>:</p>
+  <dl class=switch>
+   <dt>`<a for="embedder policy value">unsafe-none</a>`
```
domenic, line 3538: ", not backtick, for the `<dt>`s since these are strings, not byte sequences

fetch.bs
```diff
@@ -4668,6 +4708,9 @@ steps. They return a <a for=/>response</a>.
 
     <p>is true; otherwise false.
 
+   <li><p>If <a>Cross-Origin-Embedder-Policy allows credentials</a> with <var>request</var> returns
+   false, then set <var>includeCredentials</var> to false.</p>
```
domenic, line 4712:
```suggestion
   false, then set <var>includeCredentials</var> to false.
```

fetch.bs
```diff
@@ -7843,6 +7889,7 @@ Arkadiusz Michalski,
 Arne Johannessen,
 Artem Skoretskiy,
 Arthur Barstow,
+Arthur Sonzogni, <!-- ArthurSonzogni; GitHub -->,
```
domenic, line 7892:
```suggestion
Arthur Sonzogni, <!-- ArthurSonzogni; GitHub -->
```

fetch.bs
ArthurSonzogni: Done

fetch.bs
ArthurSonzogni: Done.

fetch.bs
ArthurSonzogni: Done

fetch.bs
ArthurSonzogni: Done.

ArthurSonzogni (commented):
Thanks! @domenic, I addressed your last comments.

annevk:
This can land tomorrow once HTML has been updated. Note that I made a couple last minute changes, but they are all editorial.

annevk:
It seems Shepherd has not run. (At least, as far as I can tell HTML was updated and the new term followed an existing pattern that does work and having checked all those things again I cannot find any errors.) @tabatkins @plinss, is that known?

plinss:
Shepherd has been blocked for the last 18 hours or so, it just ran. The last version of fetch it scanned was published on 2021-11-01 01:32:23 PDT

annevk:
Thanks @plinss!

annevk:
And thanks @ArthurSonzogni for the work on the PR, of course! And @domenic and @yutakahirano for the reviews! 😊

ArthurSonzogni:
Awesome! 🚀
Thanks @annevk, @domenic, @yutakahirano for the reviews!
I learnt a lot from you.

Issue #832, opened 2018-11-14
fetch+CSP end up touching globals in a parallel section on redirect


bzbarsky:
Say a fetch starts via <https://fetch.spec.whatwg.org/#concept-main-fetch>.  Starting in step 5, we go parallel, so should be able to execute on an arbitary thread/process without touching ambient state.  If the response is an HTTP redirect, we will end up in https://fetch.spec.whatwg.org/#concept-http-redirect-fetch which again calls into main fetch, but already running in parallel.  But main fetch step 2.4 calls https://w3c.github.io/webappsec-csp/#should-block-request which does:

> Let CSP list be request’s client’s global object’s CSP list.

This is doing an unsynchronized access to the CSP list, which could be in the middle of being mutated due to a meta tag insertion, for example.

We need to either snapshot the CSP list when the request is created or do the redirect fetch off a task that ensures it runs when the CSP list is stable or something.

bzbarsky:
@mikewest @annevk

bzbarsky:
Worth checking whether this situation is covered by tests, by the way.

mikewest:
@andypaicu, mind taking a look? I think this is something you wanted to address anyway for navigation requests?

annevk:
(I wish GitHub offered a sigh emoji. `<meta>` FTW. Thanks for reporting this though!)

yutakahirano:
I prefer using the snapshot. We may move the redirect checks to a completely different process.

andypaicu:
From the CSP spec point of view, I'm happy to make any changes to use a snapshotted version of the CSP list. I suppose, from my uninformed point of view, responses already have a [CSP list](https://fetch.spec.whatwg.org/#concept-response-csp-list), so a request could also have a snapshot of the CSP list (which could slightly be trimmed down, I doubt `base-uri` for example is relevant).

In my ideal world mutable CSP policies don't seem like something anyone would ever use in a legitimate manner, but there is a need to allow `<meta` tags to add CSP because there are servers that don't give control of HTTP headers of the pages they host. Assuming there is no legitimate use case for mutable CSP policies, another idea might be perhaps to "lock" the CSP policies as soon a request is made (or maybe as soon as a CSP check is performed). Thoughts on this?

bzbarsky:
@ckerschb might have some thoughts on that; we were just discussing it yesterday.

youennf:
I agree with @yutakahirano.
[...]

PR #1233, opened 2021-05-12
Use referrer policy from policy container

antosart:
This is a companion change to https://github.com/whatwg/html/pull/6677, which moves referrer policy into the policy container.

As a consequence of this change, the referrer policy used for the request is a snapshot of the relevant referrer policy when the request is being created, and won't change during the lifetime of the request, addressing a variant of https://github.com/whatwg/fetch/issues/832 for referrer policy.

***
<a href="https://whatpr.org/fetch/1233.html" title="Last updated on May 17, 2021, 8:44 AM UTC (c08fa08)">Preview</a> | <a href="https://whatpr.org/fetch/1233/3c0dd16...c08fa08.html" title="Last updated on May 17, 2021, 8:44 AM UTC (c08fa08)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1541,8 +1541,8 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
 <dfn export for=request id=concept-request-referrer-policy>referrer policy</dfn>, which is a
 <a for=/>referrer policy</a>. Unless stated otherwise it is the empty string. [[!REFERRER]]
 
-<p class="note no-backref">This can be used to override a referrer policy associated with an
-<a>environment settings object</a>.
+<p class="note no-backref">This can be used to override the referrer policy to be used for this
+<a for=/>request</a>.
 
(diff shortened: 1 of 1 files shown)


antosart:
This depends on https://github.com/whatwg/html/pull/6677

annevk (commented):
Looks good, but we should also update this note:

> This can be used to override a referrer policy associated with an environment settings object.

annevk (approved):
I guess we'll wait for @domfarolino to double check (as he's also signed up for the other PRs), but this looks good from my perspective. Thanks!

antosart:
> Looks good, but we should also update this note:

Done, although I am not sure about the formulation (in principle, this can be used to override policy container's referrer policy, which in turn can be used to override the environment settings objects' policy container's referrer policy).

domfarolino:
I've been traveling the last few days; I'll check this out within the next two days.

PR #1240, opened 2021-05-25
Editorial: modernize connection prose

annevk:
***
<a href="https://whatpr.org/fetch/1240.html" title="Last updated on May 25, 2021, 12:27 PM UTC (2cc1b55)">Preview</a> | <a href="https://whatpr.org/fetch/1240/b6c6283...2cc1b55.html" title="Last updated on May 25, 2021, 12:27 PM UTC (2cc1b55)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2261,7 +2261,7 @@ steps:
 <hr>
 
 <p>To <dfn export id=concept-connection-obtain>obtain a connection</dfn>, given a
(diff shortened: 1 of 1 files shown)


ricea (approved):
lgtm

Chromium's implementation is more complex than this. Connections are "late bound", meaning that which connection is assigned to which request is not decided until after the connection completes. This allows high priority requests to get a connection first. That's probably too big a topic for this PR, however.

Issue #1122, opened 2020-12-07
WebSocket connections need partitioning


annevk:
Currently WebSocket connections share no infrastructure with "HTTP" connections, this is somewhat wrong, at least when it comes to state partitioning.

(See also #1118.)

cc @MattMenke2 @shivanigithub @sleevi

MattMenke2:
In Chromium, we've already wired up partitioning to WebSockets (though are not yet shipping it), so we certainly agree with this.

MattMenke2:
While they're not covered in the fetch spec, there are other caches (TLS session resumption), websocket over shared H2/QUIC connections, ALPN data, and the DNS cache, that are partitioned, and WebSockets should presumably pull all of those from the correct partition.  If proxy connections and the HTTP auth cache are partitioned, needs to use the right partition for those as well.

annevk:
Ideally we would cover those in Fetch I think (or at least mention that they should be considered part of the connection and thus be partitioned).

If WebSocket over H2 (and QUIC? Is there an RFC for that?) can reuse a connection, Fetch needs some more complicated logic than I suggested in #1241. Not immediately clear to me what that would look like. forceDedicatedIfH1 or some such...

MattMenke2:
Unfortunately, I'm not familiar enough with WebSockets or H2/H3 to know where the details are specified.

annevk:
There's https://datatracker.ietf.org/doc/html/rfc8441 for H2.

ricea:
WebSocket over H2 does reuse connections. Currently you can't use WS/H2 at all unless you have an existing H2 connection.

WebSocket over H3 is not a thing yet although in principle it should be a straightforward application of WS/H2 semantics.

annevk:
Thanks everyone. I'm going to fold this into #1243 as that will truly solve this (fingers crossed).

PR #1241, opened 2021-05-25
Editorial: make WebSocket use obtain a connection

annevk:
On closer inspection "obtain a WebSocket connection" does not appear to do anything that "obtain a connection" does not do, especially now the latter has a dedicated argument.

This brings "obtain a connection" closer to the goal of covering all web platform connections and a being a policy-enforcement point for them.

cc @MattMenke2

***
<a href="https://whatpr.org/fetch/1241.html" title="Last updated on May 25, 2021, 1:12 PM UTC (0798b0f)">Preview</a> | <a href="https://whatpr.org/fetch/1241/b6c6283...0798b0f.html" title="Last updated on May 25, 2021, 1:12 PM UTC (0798b0f)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1571,8 +1571,7 @@ to not have to set <a for=/>request</a>'s <a for=request>referrer</a>.
   <dd>This is a special mode used only when <a>navigating</a> between documents.
 
   <dt>"<code>websocket</code>"
-  <dd>This is a special mode used only when <a lt="establish a WebSocket connection">establishing
-  a WebSocket connection</a>.
+  <dd>This is a special mode used for the WebSocket API.
  </dl>
 
  <p>Even though the default <a for=/>request</a> <a for=request>mode</a> is "<code>no-cors</code>",
@@ -5074,21 +5073,14 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
  <li><p>Let <var>networkPartitionKey</var> be the result of
  <a for=request>determining the network partition key</a> given <var>request</var>.
 
- <li>
-  <p>Switch on <var>request</var>'s <a for=request>mode</a>:
+ <li><p>Let <var>dedicatedConnection</var> be true if <var>request</var>'s <a for=request>mode</a>
+ is "<code>websocket</code>"; otherwise false.
 
-  <dl>
-   <dt>"<code>websocket</code>"
-   <dd><p>Let <var>connection</var> be the result of
-   <a lt="obtain a WebSocket connection">obtaining a WebSocket connection</a>, given
-   <var>request</var>'s <a for=request>current URL</a>.
-
-   <dt>Otherwise
-   <dd><p>Let <var>connection</var> be the result of
-   <a lt="obtain a connection">obtaining a connection</a>, given <var>networkPartitionKey</var>,
-   <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>,
-   <var>includeCredentials</var>, and <var>forceNewConnection</var>.
-  </dl>
+ <li><p>Let <var>connection</var> be the result of
+ <a lt="obtain a connection">obtaining a connection</a>, given <var>networkPartitionKey</var>,
+ <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>,
+ <var>includeCredentials</var>, <var>forceNewConnection</var>, and with
+ <a for="obtain a connection"><i>dedicated</i></a> set to <var>dedicatedConnection</var>.
 
  <li>Set <var>timingInfo</var>'s <a for="fetch timing info">final connection timing info</a> to the
  result of calling <a>clamp and coarsen connection timing info</a> with <var>connection</var>'s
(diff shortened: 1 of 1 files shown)


annevk:
As per #1122 forcing a new connection should only happen for H1.

fetch.bs
```diff
@@ -7425,32 +7417,6 @@ fetch("https://www.example.com/")
...
-<ol>
- <li><p>Let <var ignore>host</var> be <var>url</var>'s <a for=url>host</a>.
-
- <li><p>Let <var ignore>port</var> be <var>url</var>'s <a for=url>port</a>.
-
- <li><p>Let <var ignore>secure</var> be false, if <var>url</var>'s <a for=url>scheme</a> is
- "<code>http</code>", and true otherwise.
-
- <li><p>Follow the requirements stated in step 2 to 5, inclusive, of the first set of steps in
- <a href=http://tools.ietf.org/html/rfc6455#section-4.1>section 4.1</a> of The WebSocket Protocol
```
ricea, line 7442: It appears that this loses the WebSocket-specific connection throttling in section 4.1.2?

fetch.bs
```diff
@@ -5074,21 +5073,14 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
 
- <li>
-  <p>Switch on <var>request</var>'s <a for=request>mode</a>:
-
-  <dl>
-   <dt>"<code>websocket</code>"
-   <dd><p>Let <var>connection</var> be the result of
-   <a lt="obtain a WebSocket connection">obtaining a WebSocket connection</a>, given
-   <var>request</var>'s <a for=request>current URL</a>.
+ <li><p>Let <var>dedicatedConnection</var> be true if <var>request</var>'s <a for=request>mode</a>
```
ricea, line 5076: This might change with WebSocket over HTTP/2?

annevk: Yeah, so what should we do. Add a _dedicatedIfHTTP1_ parameter?

fetch.bs
annevk: I guess we could try to reinstate that somehow although it seems that would require all connections to be in some kind of collection, right?

Also, how does that work with partitioning?

fetch.bs
ricea: It's not pretty, but it would work.

fetch.bs
ricea: I think the only choice is to make it work per-partition. That's unfortunate if an attacker can manage to use many partitions at the same time, but I think we generally assume that is unlikely.

Moving the logic here rather than delegating to the RFC would let us make that unambiguous.

How it essentially works is you have a per-_ip:port_ queue, where only the WebSocket at the front of the queue is permitted to be in a connecting/handshake state. WebSockets are removed from the queue when the handshake succeeds, fails, or is cancelled.

What's troublesome is that you need to do a DNS lookup to work out which queue a WebSocket belongs in. And if there are multiple IP addresses you might need to try multiple queues before you find one that connects.

Dealing with proxies is awful and Chromium does it completely wrong.

Now that I say all this I think there's definite value is specifying this in the Fetch standard (or in the WebSocket standard, whichever comes first).

annevk: @ddragana convinced me that the win from partitioning might be small, as it might be indistinguishable from network jitter (as it's only blocking during connection creation). So maybe that's not needed.

But I also reached the conclusion that this brings up the define "DNS lookup" question again. @sleevi would like to keep everything in "obtain a connection", but this seems to make that rather hard. I suppose we could wave our hands a bunch and describe it implicitly as the current RFC is doing, but ugh.

(I'll try to clean this up in due course by moving this discussion into its own issue, but for now this suffices I think.)

sleevi: @annevk Yeah, entirely fair, as you also mentioned in https://github.com/WICG/private-network-access/issues/51 

No need to rehash the discussion here, but mostly just wanted to acknowledge that I’m not in principle opposed to trying to tackle it, just that it’s going to be complex either way to do so.

sleevi: @ricea Could you expand on the following comment?

> Dealing with proxies is awful and Chromium does it completely wrong.

I’m having trouble making sense of it in the context of this PR, or in general for Chrome, and just trying to sort through if this was a remark specific to WebSockets and the throttling not being able to adhere fully to the spec (because the spec is wrong), or if this was something else.

annevk:
Let's continue the discussion in #1243. Really appreciate your feedback @ricea (and @ddragana via Matrix)!

fetch.bs
ricea: > @ricea Could you expand on the following comment?
> 
> > Dealing with proxies is awful and Chromium does it completely wrong.

The problem at hand is the following text from RFC6455:

>        If the client cannot determine the IP address of the remote host
>        (for example, because all communication is being done through a
>        proxy server that performs DNS queries itself), then the client
>        MUST assume for the purposes of this step that each host name
>        refers to a distinct remote host, and instead the client SHOULD
>        limit the total number of simultaneous pending connections to a
>        reasonably low number (e.g., the client might allow simultaneous
>        pending connections to a.example.com and b.example.com, but if
>        thirty simultaneous connections to a single host are requested,
>        that may not be allowed).  

Problems:

1. On most networks we *can* determine the IP address, even if we're behind a proxy, However, for privacy reasons we shouldn't look up the IP address unless we already looked it up in the process of determining the proxy to use (ie. for proxy.pac). But AFAIK in Chromium we don't have a concept of "I already looked up this IP address", nor do we have a concept of "check the cache for this IP address, but don't hit the network".
2. Does `MUST assume ... that each host name refers a distinct remote host` imply that we throttle by _hostname:port_ the way we usually throttle by _ip:port_? That's what I've been assuming up until now, but the rest of the sentence maybe implies a completely different throttling strategy?
3. In practice what we do in Chromium is throttle by the IP address of the proxy server. This is completely wrong, but causes surprisingly few problems in practice.

MattMenke2: > I think the only choice is to make it work per-partition. That's unfortunate if an attacker can manage to use many partitions at the same time, but I think we generally assume that is unlikely.

Note that chrome uses top-level-frame schemeful site and innermost iframe schemeful site, so an attacker could, within the context of a frame, open as many iframes as it wants (with different sites it controls) to get as many NIKs as it needs for any putative attack.

ricea: > Note that chrome uses top-level-frame schemeful site and innermost iframe schemeful site, so an attacker could, within the context of a frame, open as many iframes as it wants (with different sites it controls) to get as many NIKs as it needs for any putative attack.

I see. So it's a potential DDoS vector. However, because the attacker needs to load HTML from a whole bunch of different sites to make it work, the attack magnification looks small.

Issue #1243, opened 2021-05-26
Refactoring WebSocket connections


annevk:
The objective is to make WebSocket connections hook into "obtain a connection" to make it clear what logic is shared.

There are two problems:

1. Step 2 of https://datatracker.ietf.org/doc/html/rfc6455#section-4.1.
   1. Defining this properly requires an abstract notion of turning a domain into an IP address. We might need that anyway for features such as `dns-prefetch`, but I recall @sleevi hoping we could avoid it.
   2. The side table might also have to be partitioned to be completely safe. This would not significantly widen the DOS scope as the attacker would be constrained by the number of tabs the user has open where the top-level site is unique and the attacker is embedded. An argument could be made that the privacy attack (not the DOS attack) is nearly indistinguishable from network jitter however and not worth the effort.
2. Requiring a dedicated connection, but only for HTTP/1. There are two approaches here:
   1. Introduce a new parameter.
   2. Turn the dedicated parameter into an enum: "not", "if-http1", "always". This would require slight tweaks to WebTransport. I slightly prefer this, but the alternative is acceptable to me as well.

We could side step some of the first problem by changing the monkey patching in "obtain a WebSocket connection", but that does not seem like a good long term solution.

This came out of a discussion in #1241 which had a rather optimistic-and-wrong take on what sharing more logic would mean. As I envision it now we would keep the "obtain a WebSocket connection" operation, but we'd pass it a partition key as well, inline the relevant bits from the RFC as outlined above, and make it use "obtain a connection".

If this seems like a good plan I will make time to work on this. (I would like to solve this before #889 and #1118.)

cc @ddragana @MattMenke2 @ricea @yutakahirano

sleevi:
CC’ing @letitz because the DNS discussion is also related to https://github.com/WICG/private-network-access/issues/51

My hope in avoiding is largely a proximity of the complexity it introduces, and because of the handwave “magic happens here” part of the composability of different protocol layers.
* How to account for Happy Eyeballs behaviours (which vary across implementations, and may extend to racing connections and not just racing resolutions)
* How to account for the different resolution layers and behaviours (DNS vs WINS/NETBIOS vs mDNS, etc)
* How to account for other protocol races (e.g. for alternate protocols returned via DNS)

[...]

PR #1244, opened 2021-06-02
Draft changes to "obtain a websocket connection"

ricea:
Modify "obtain a websocket connection" section to explicitly describe
throttling behaviour rather than delegating it to RFC6455.

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1244.html" title="Last updated on Jun 8, 2021, 5:56 AM UTC (938cb06)">Preview</a> | <a href="https://whatpr.org/fetch/1244/43a0c94...938cb06.html" title="Last updated on Jun 8, 2021, 5:56 AM UTC (938cb06)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7451,11 +7451,57 @@ fetch("https://www.example.com/")
  <li><p>Let <var ignore>secure</var> be false, if <var>url</var>'s <a for=url>scheme</a> is
  "<code>http</code>", and true otherwise.
 
- <li><p>Follow the requirements stated in step 2 to 5, inclusive, of the first set of steps in
- <a href=https://datatracker.ietf.org/doc/html/rfc6455#section-4.1>section 4.1</a> of The WebSocket
- Protocol to establish a <a lt="obtain a WebSocket connection">WebSocket connection</a>. [[!WSP]]
+ <li><p>If <var>host</var> is a domain and a proxy is not in use, then set <var>hosts</var> to the
+ results of resolving a domain with <var>key</var> and <var>host</var>.
 
- <li><p>If that established a connection, return it, and return failure otherwise.
+ <li><p>Otherwise, set <var>hosts</var> to « <var>host</var> ».
+
+<!-- TODO: Need to limit the total number of simultaneous pending connections for proxies. -->
+
+ <li><p>For each <var>host</var> in <var>hosts</var>,
+
+<!-- TODO: Happy eyeballs means we actually try IPv6 and IPv4 hosts in parallel. -->
+
+ <ol>
+  <li><p>If the user agent's connecting WebSocket connections[(<var>key</var>, <var>host</var>, <var>port</var>)] exists, then:
+
+  <ol>
+    <li><p>Let connectingWebSocketConnection be that connection.
+
+    <li><p>Wait for that connection to established or have failed. (Note: this removes it from the
+    map.)
+
+    <li><p>Wait an additional 10ms. (Note: this slows down denial-of-service attacks a bit without
+    significantly impacting legitimate use).
+    <!-- This is Chromium behaviour, not currently standardised -->.
+  </ol>
+
+  <li><p>Assert: the user agent's connecting WebSocket connections[(<var>key</var>, <var>host</var>,
+  <var>port</var>)] does not exist.
+
+  <li><p>Set the user agent's connecting WebSocket connections[(<var>key</var>, <var>host</var>,
+  <var>port</var>)] to the result of obtaining a connection with <a>network partition key</a>
+  <var>key</var>, <a for=connection>origin</a> <var>origin</var>, <var>credentials</var> set to
+  true, and <a>dedicated</a> set to true.
+
+<!-- TODO: Need to do the equivalent of section 4.1 step 3 for proxies.
+
+Note: Chromium and AFAIK other browsers don't obey this paragraph as they use "http" and "https"
+instead of "ws" and "wss" when executing proxy autoconfiguration scripts.
+
+       For the purpose of proxy autoconfiguration scripts, the URI to
+       pass the function MUST be constructed from /host/, /port/,
+       /resource name/, and the /secure/ flag using the definition of a
+       WebSocket URI as given in Section 3.
+
+-->
+
+  <li><p>If establishing connection failed with a TCP error, then remove (<var>key</var>,
+  <var>host</var>, <var>port</var>) from connecting WebSocket connections and <a>continue</a>.
+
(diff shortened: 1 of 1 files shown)


ricea:
Ah, this isn't quite right, because multiple connections can be waiting for a single (key, host, port) to be available. We need to keep a queue of waiting connections instead.

annevk:
I fixed that (mostly). Now #1245 needs to land at which point we can draft the changes needed to "obtain a connection". For the eventual commit message, this will fix #1243.

(Not necessarily a fan of the naming, e.g., "inner obtain a WebSocket connection". Suggestions appreciated.)

fetch.bs
```diff
@@ -7440,27 +7439,69 @@ fetch("https://www.example.com/")
...
+<a for=/>set</a> of unique identifiers.
+
 <p>To <dfn id=concept-websocket-connection-obtain>obtain a WebSocket connection</dfn>, given a
-<var>url</var>, run these steps:
+<a>network partition key</a> <var>key</var>, <a for=/>origin</a> <var>origin</var>, and boolean
+<var>credentials</var>, run these steps:
 
 <ol>
- <li><p>Let <var ignore>host</var> be <var>url</var>'s <a for=url>host</a>.
+ <li><p>Let <var>hosts</var> be the result of running resolve an origin given <var>key</var> and
```
ricea, line 7451: Do we need language here to explicitly enable aborting the connection?

ricea:
> (Not necessarily a fan of the naming, e.g., "inner obtain a WebSocket connection". Suggestions appreciated.)

I thought of "obtain a WebSocket connection to a single host" but it's a bit verbose.

ricea:
938cb06 is a strawman sketch of proxy support. Surprisingly little of RFC6455 section 4.1 step 3 seems to actually be needed.

The "total pending proxy WebSocket connections" map is fiction, but I don't want to standardise Chromium's embarrassing behaviour. I would be interested to learn what other browsers do here.

ricea:
Upon further investigation, it appears that Chromium doesn't always convert "ws" and "wss" to "http" and "https" when executing proxy.pac. It depends on whether we're using the system implementation of proxy resolution or not. Maybe it needs to be implementation-defined.

fetch.bs
annevk: This is a problem with #1245 as well. I can add a sentence there to make it clear they may be closed.

fetch.bs
```diff
@@ -7440,27 +7439,100 @@ fetch("https://www.example.com/")
...
+<var>credentials</var>, run these steps:
 
 <ol>
- <li><p>Let <var ignore>host</var> be <var>url</var>'s <a for=url>host</a>.
+ <li><p>Let <var>proxies</var> be the result of performing proxy resolution for <var>url</var> in an
+ <a>implementation-defined</a> manner. If a proxy preference order has not been explicitly
+ configured then socks proxies should be preferred to https proxies which should be preferred to
+ http proxies.
+
+ <!-- Since this depends on the url it probably needs to be hoisted out of this algorithm. -->
```
annevk, line 7459: Why does this depend on the URL?

fetch.bs
```diff
@@ -7440,27 +7439,100 @@ fetch("https://www.example.com/")
...
-<var>url</var>, run these steps:
+<a>network partition key</a> <var>key</var>, <a for=/>origin</a> <var>origin</var>, and boolean
+<var>credentials</var>, run these steps:
 
 <ol>
- <li><p>Let <var ignore>host</var> be <var>url</var>'s <a for=url>host</a>.
+ <li><p>Let <var>proxies</var> be the result of performing proxy resolution for <var>url</var> in an
+ <a>implementation-defined</a> manner. If a proxy preference order has not been explicitly
+ configured then socks proxies should be preferred to https proxies which should be preferred to
+ http proxies.
```
annevk, line 7457: Need to capitalize SOCKS, HTTPS, and HTTP.

annevk (commented):
Why is there so much proxy logic here? Couldn't that be moved to obtain a connection?

fetch.bs
ricea: The URL is passed to the `FindProxyForURL()` function in the proxy.pac file. proxy.pac can in principle choose a different proxy depending on the URL path, although apparently this [didn't work properly in IE](https://en.wikipedia.org/wiki/Proxy_auto-config#Old_Microsoft_problems).

annevk: Ah, but that sounds like a problem for "obtain a connection" in general. I think we should add proxy support there instead and then layer this on top. Do you agree?

ricea: Yes. I am planning to draft a change which moves most of the proxy parts to "obtain a connection". Unless you get to it first.

fetch.bs
```diff
@@ -7440,27 +7439,100 @@ fetch("https://www.example.com/")
...
+
+ <!-- Since this depends on the url it probably needs to be hoisted out of this algorithm. -->
+
+ <p class=note>The scheme of the URL used in proxy autoconfiguration scripts has been converted from
+ "<code>ws</code>" or "<code>wss</code>" to "<code>http</code>" or "<code>https</code>". This
```
annevk, line 7462: Does this mean that when the user agent invokes `FindProxyForURL` this is the case or something else? (Just want to make sure I understand.)

fetch.bs
```diff
@@ -7440,27 +7439,100 @@ fetch("https://www.example.com/")
...
+<a for=/>map</a> of <a for=/>network partition key</a> to an integer, starting at 0.
+
 <p>To <dfn id=concept-websocket-connection-obtain>obtain a WebSocket connection</dfn>, given a
-<var>url</var>, run these steps:
+<a>network partition key</a> <var>key</var>, <a for=/>origin</a> <var>origin</var>, and boolean
+<var>credentials</var>, run these steps:
 
 <ol>
- <li><p>Let <var ignore>host</var> be <var>url</var>'s <a for=url>host</a>.
+ <li><p>Let <var>proxies</var> be the result of performing proxy resolution for <var>url</var> in an
```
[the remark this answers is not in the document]
annevk: Apparently "obtain a connection" needs to be passed a URL for proxies. It seems like this shouldn't be a problem for WebTransport if we make that change. Is that correct? (Note that this patch has several issues at the moment.)

annevk:
Closing this due to #1516. I filed https://github.com/whatwg/websockets/issues/38 against WebSockets to ensure this is still tracked and #1243 will remain open as well.

PR #1245, opened 2021-06-02
Elaborate on obtaining a connection

annevk:
In particular, define resolving domains and allow connection creation to be a race.

As a result this also inlines some of the time capture moments to be directly inside the obtain a connection algorithm.

This helps with #1243.

***
<a href="https://whatpr.org/fetch/1245.html" title="Last updated on Jun 14, 2021, 3:06 PM UTC (df2a2c0)">Preview</a> | <a href="https://whatpr.org/fetch/1245/43a0c94...df2a2c0.html" title="Last updated on Jun 14, 2021, 3:06 PM UTC (df2a2c0)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2204,6 +2204,39 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
 <a for="fetch group">fetch record</a>'s <a for="fetch record">fetch</a>.
 
 
+<h3 id=resolving-domains>Resolving domains</h3>
+
+<p>To <dfn>resolve a domain</dfn>, given a <a for=/>domain</a> <var>domain</var> and a
+<a for=/>network partition key</a> <var>key</var>, perform an <a>implementation-defined</a>
+operation to turn <var>domain</var> into a <a for=/>set</a> of one or more
+<a for=/>IP addresses</a>. If this operation succeeds, return the <a for=/>set</a> of
+<a for=/>IP addresses</a>. If it fails, return failure. The results of this operation may be cached.
+If they are cached, <var>key</var> must be used as part of the cache key.
+
+<p class=note>Typically this operation would involve DNS. The particulars (apart from the cache key)
+are not tied down as they are not pertinent to the system the Fetch Standard establishes. Other
+documents ought not to build on this primitive without having a considered discussion with the Fetch
+Standard community first. [[RFC1035]]
+
+<p>To <dfn>resolve an origin</dfn>, given an <a for=/>origin</a> <var>origin</var> and a
+<a for=/>network partition key</a> <var>key</var>:
+<!-- Should we assert the scheme here to be an HTTP(S) scheme or a WebRTC scheme? -->
+
+<ol>
+ <li><p>If <var>origin</var>'s <a for=origin>host</a> is an <a for=/>IP address</a>, then return
+ « <var>origin</var>'s <a for=origin>host</a> ».
+
+ <li><p>If the user agent is configured to use a proxy, then return
+ « <var>origin</var>'s <a for=origin>host</a> ».
+
+ <li><p>Return the result of <a lt="resolve a domain">resolving</a> <var>origin</var>'s
+ <a for=origin>host</a>.
+</ol>
+
+<p class=note>The same caveat applies. Do not build on this without having a considered discussion
+with the Fetch Standard community first.
+
+
 <h3 id=connections>Connections</h3>
 
 <p>A user agent has an associated <dfn export id=concept-connection-pool>connection pool</dfn>. A
@@ -2274,7 +2307,7 @@ steps:
 <hr>
 
 <p>To <dfn export id=concept-connection-obtain>obtain a connection</dfn>, given a
-<a>network partition key</a> <var>key</var>, <a for=/>origin</a>, <var>origin</var>, boolean
+<a>network partition key</a> <var>key</var>, <a for=/>origin</a> <var>origin</var>, boolean
 <var>credentials</var>, an optional boolean <var>forceNew</var> (default false), an optional boolean
 <dfn export for="obtain a connection"><var>http3Only</var></dfn> (default false), and an optional
 boolean <dfn export for="obtain a connection"><var>dedicated</var></dfn> (default false), run these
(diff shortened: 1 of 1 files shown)


annevk:
Hot takes appreciated from @yutakahirano @domenic @sleevi @ricea @MattMenke2 @ddragana and more.

fetch.bs
```diff
@@ -2204,6 +2204,39 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
...
+documents ought not to build on this primitive without having a considered discussion with the Fetch
+Standard community first. [[RFC1035]]
+
+<p>To <dfn>resolve an origin</dfn>, given an <a for=/>origin</a> <var>origin</var> and a
+<a for=/>network partition key</a> <var>key</var>:
+<!-- Should we assert the scheme here to be an HTTP(S) scheme or a WebRTC scheme? -->
+
+<ol>
+ <li><p>If <var>origin</var>'s <a for=origin>host</a> is an <a for=/>IP address</a>, then return
+ « <var>origin</var>'s <a for=origin>host</a> ».
```
MattMenke2, line 2227: Note that IP Address is defined as a 32-bit or 128-bit unsigned integer.  Neither of those are a valid host.  Is there some host to IP address mapping we should reference in the line above and here?

fetch.bs
```diff
@@ -2204,6 +2204,39 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
...
+
+ <li><p>If the user agent is configured to use a proxy, then return
+ « <var>origin</var>'s <a for=origin>host</a> ».
```
MattMenke2, line 2230: In the case of a proxy, do we still need to resolve IP addresses?  I don't believe Chrome currently does (in, e.g., the RFC 1918 case)

annevk: Do you mean resolving the domain? This avoids doing that, right?

fetch.bs
annevk: In spec-theory an origin's host that is an IP address is a 32-bit or 128-bit integer. As we don't define the actual connection-creation process, how that integer is serialized isn't referenced either (though we could reference it I suppose, https://url.spec.whatwg.org/#concept-ipv4-serializer and https://url.spec.whatwg.org/#concept-ipv6-serializer defines these.

MattMenke2: Maybe just add "as an IP address", then?  Or perhaps I'm alone in thinking pseudo-code as text should be type safe.  :)

fetch.bs
```diff
@@ -2204,6 +2204,39 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
...
 
 
+<h3 id=resolving-domains>Resolving domains</h3>
+
+<p>To <dfn>resolve a domain</dfn>, given a <a for=/>domain</a> <var>domain</var> and a
+<a for=/>network partition key</a> <var>key</var>, perform an <a>implementation-defined</a>
+operation to turn <var>domain</var> into a <a for=/>set</a> of one or more
+<a for=/>IP addresses</a>. If this operation succeeds, return the <a for=/>set</a> of
+<a for=/>IP addresses</a>. If it fails, return failure. The results of this operation may be cached.
+If they are cached, <var>key</var> must be used as part of the cache key.
```
sleevi, line 2214: I'm not sure this line makes sense, or at least, not normative sense.

The issue here is that unless you're writing your own recursive resolver implementation, the resolvers you communicate with will ignore `key` and cache. E.g. if you use an OS API and the OS API caches, it's not going to consider `key`. And even if you write your own recursive resolver, you can't be sure that the name servers themselves won't cache based on `(host, recordtype)` tuples, ignoring `key` (which is never on the wire)

While this is definitely more of @MattMenke2 's realm for wording, I wanted to flag this, because it seems like this might involve some form of note or form of [RFC 6919 language](https://datatracker.ietf.org/doc/html/rfc6919#section-1).

fetch.bs
MattMenke2: No, I mean resolve the IP address.  e.g., mapping https://127.0.0.1/ to 0xFF000001 (sorry, may have the order reversed).  Chrome, at least, only tracks the proxy's IP in this case, and only provides that to consumers.  Consumers are careful not to treat it as the server IP.  So this would require a rearchitecture a number of Chrome layers.  Consumers would either have to extract the IP from the URL in the proxy case, or Chrome's network stack would have to provide both IPs, and all consumers audited.

sleevi: It depends on the type of proxy being used. We do resolve hosts for some proxies (e.g. SOCKSv4), but not for others (e.g. HTTP CONNECT). And we may still need to resolve prior to a proxy being involved (e.g. a PAC script)

I think we should strip the proxy-particular language here with regards to resolving; it's more closely tied to the connection establishment logic, and I think we'd have to tackle it there. That said, I'm not 100% sure on that - I think it really depends on how we end up using this "method" here that dictates which parts are captured in the pseudocode.

MattMenke2: Note that while we resolve server IPs for SOCKS4, we don't expose them to consumers.  My main concern here is the impact on any features that want the server IP from a socket (RFC1918 is the only one I'm aware of, but there may be others).

fetch.bs
MattMenke2: I noticed this too, but wasn't sure it was worth commenting on, given the style here tends to be not to explain things.  Anyhow, I'm not sure if it's reasonable to note that upstream DNS servers may still cache across keys, leave as-is, or remove the section.

fetch.bs
sleevi: Gotcha. Yeah, I believe you're talking about https://github.com/WICG/private-network-access/issues/52 . 

I don't believe Chromium's behaviour here were particularly opinionated, so much as it being an implementation detail, but tagging @letitz to review :) You're right, though, that the current logic prefers the IP address of the proxy server, at least when deciding Private Network Access resolution, even though it may use the IP address of the host to connect through a proxy.

MattMenke2: I think we have an "unknown" value in the proxy case, with its own special handling, rather than using the proxy's IP, or the server's IP.  Anyhow, guess that's a bit of a tangent.

fetch.bs
annevk: That does seem good to point out. And it seems that even for local lookup this can at best be a SHOULD for now.

fetch.bs
annevk: What part is not type safe here though? An origin's host is a host, which is roughly speaking a domain or an IP address. Asking if it's one or the other should be okay?

MattMenke2: Well, we're returning, e.g., "127.0.0.1" or "[::1]".  Neither of those fit the linked definition of "IP address" of an unsigned integer.

annevk: But the spec doesn't, right? That you implement them as a different type is okay, as long as the results are indistinguishable.

MattMenke2: I'm talking about this spec, not Chrome's code.  It says return the host.

MattMenke2: Oh, host is defined as a domain or an IP address, as opposed to as the raw representation from the URL...ok, the magic is behind extracting the host from the URL, not the IP address from the host.

annevk: Defined parser magic, yes :-)

fetch.bs
```diff
@@ -2204,6 +2204,39 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
 <a for="fetch group">fetch record</a>'s <a for="fetch record">fetch</a>.
 
 
+<h3 id=resolving-domains>Resolving domains</h3>
+
+<p>To <dfn>resolve a domain</dfn>, given a <a for=/>domain</a> <var>domain</var> and a
+<a for=/>network partition key</a> <var>key</var>, perform an <a>implementation-defined</a>
+operation to turn <var>domain</var> into a <a for=/>set</a> of one or more
```
ricea, line 2211: Pedantic note: the set is ordered, although the order may be different every time. Given the language below, I don't think it makes any difference.

fetch.bs
```diff
@@ -2298,55 +2331,103 @@ steps:
...
-
-    <p>If establishing a connection does not succeed (e.g., a DNS, TCP, or TLS error), then return
-    failure.
-  </ol>
+ <li><p>Set <var>timingInfo</var>'s <a for="connection timing info">domain lookup end time</a> to
+ the <a for=/>unsafe shared current time</a>.
 
  <li>
-  <p><a>If aborted</a>, then:
+  <p>Let <var>connection</var> be the result of running this step: run <a>create a connection</a>
```
ricea, line 2348: This is probably a bad time to mention that Chromium starts racing a second connection if the first one takes more than 250ms. Maybe this can be treated as a kind of preconnect for standardisation purposes?

ricea (commented):
I think it looks okay.

fetch.bs
annevk: I guess to allow that kind of behavior I shouldn't put an upper limit on the number of times create a connection can be invoked. I think that's reasonable.

fetch.bs
annevk: We don't have a data structure available that's not ordered, but I'll add this to the ever growing note. :-)

fetch.bs
annevk: I think this is good now.

fetch.bs
annevk: I struggle the most with this series of comments. As I understand it, if we have a proxy that takes care of resolving domains, we don't want to resolve them locally for the purposes of determining 1) what to create a connection with 2) WebSocket's "do we have a connection to this host" side table. These are the two use cases that "resolve an origin" is targeting and there's potentially dns-prefetch as a third use case in the future.

letitz: It seems there are actually two concepts here to distinguish: 1) the set of IP addresses resolved from a hostname, and 2) the remote IP address of a network socket. This operation defines 1), but Private Network Access operates on 2).

Currently, even in the case of SOCKSv4 proxies where Chrome resolves hosts itself, Private Network Access checks apply to the proxy's IP address because that is the remote endpoint of the socket. If we are to specify the current behavior,  then I agree with @sleevi that this special case might be better handled at the connection establishment layer rather than at the resolution layer.

Now, I would be happy to improve the situation in Chrome by using the actually-resolved IP address in cases where it is available, because that would reduce the extent of proxy weirdness. I don't think I will tackle this in the near term - given the corner case it is, it seems low priority to me.

annevk: Ah, that's the 4th use case, PNA. I haven't addressed that yet. For that we indeed need to add a field on connection and decide what value it should have. I suggest we try tackling that separately from this PR.

letitz: I agree! If this language about proxies is useful for others, I'm happy to let it be :)

fetch.bs
```diff
@@ -2204,6 +2204,47 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
...
+<a for=/>IP addresses</a>. If this operation succeeds, return the <a for=/>set</a> of
+<a for=/>IP addresses</a>. If it fails, return failure. The results of this operation may be cached.
+If they are cached, <var>key</var> should be used as part of the cache key.
+
+<div class=note>
+ <p>Typically this operation would involve DNS and as such caching can happen on DNS servers without
+ <var>key</var> being taken into account. Depending on the implementation it might also not be
+ possible to take <var>key</var> into account locally. [[RFC1035]]
+
+ <p>Callers ought to not rely on the order of the <a for=/>set</a> <a>resolve a domain</a> can
```
ricea, line 2221: We take it as a hint for what order to try connecting to the IP addresses in. It's okay to use it in that way, even though the order may end up being different from what the server intended.

A couple of traditional ways of setting the order that results are returned in:
* Return the "main" web server as the first item in the list, and the "backup" web server second.
* Round-robin the results to balance load over a cluster of servers.

ricea (commented):
I think this is a good improvement in concreteness that nicely skirts implementation-specific issues.

yutakahirano (approved):
LGTM. Unrelated to this change but it may be good to turn _forceNew_ and _dedicated_ into a three-value enum.

fetch.bs
annevk: @MattMenke2 FWIW, I'd be happy to explain more or accept patches that explain more of it, but as I understand it from @sleevi that would be quite a bit of effort and it's not clear it'll help solve a problem. But we do need some basics written down for the rest of the things to make sense and fit together. In general though I'd say that the style is the opposite of what you suggest, in that we do try to cover all the details.

MattMenke2: The style tends to be purely descriptive of what browsers do, with no additional detail (e.g., nowhere is it mentioned why network partition keys exist, or what the "essence" of a MIME type is intended for, merely how its calculated and used).

annevk: Ah, I see what you mean. That's not really intentional, more of a side effect of prioritizing the normative description and having limited time. 😟

annevk:
@ricea can you take a final look? I'll merge this tomorrow.

@yutakahirano yeah, happy to make changes there if that works for WebTransport. See also the discussion in #1243.

ricea:
lgtm

PR #1246, opened 2021-06-02
Make response body streams readable byte streams

domenic:
Depends on https://github.com/whatwg/streams/pull/1130. Closes https://github.com/whatwg/fetch/issues/267.

   * I think most implementers have expressed _interest_, but I'm not sure it's on anyone's near-term plan to implement this, so we may want to leave this PR open for a while...
   * TODO
   * Chrome: TODO
   * Firefox: TODO
   * Safari: TODO

***
<a href="https://whatpr.org/fetch/1246.html" title="Last updated on Jul 20, 2021, 9:24 AM UTC (3f03a81)">Preview</a> | <a href="https://whatpr.org/fetch/1246/613aad9...3f03a81.html" title="Last updated on Jul 20, 2021, 9:24 AM UTC (3f03a81)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5397,9 +5397,9 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
       <p>While true:
 
       <ol>
-       <li><p>User agents should use the <a for=ReadableStream>current requested number of bytes</a>
-       for <var>stream</var> to inform the transmission process about how many bytes are being
-       requested for this stream, if it is non-null.
+       <li><p>If <var>stream</var>'s <a for=ReadableStream>current BYOB request view</a> is
+       non-null, then user agents should use its <a for=BufferSource>byte length</a> to inform the
+       transmission process about how many bytes are being requested for this stream.
 
        <li>
         <p>If one or more bytes have been transmitted from <var>response</var>'s message body, then:
@@ -5426,8 +5426,20 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>If <var>bytes</var> is failure, then <a lt=terminated for=fetch>terminate</a> the
          ongoing fetch.
 
+         <li><p>Let <var>view</var> be null.
+
+         <li><p>If <var>stream</var>'s <a for=ReadableStream>current BYOB request view</a> is
+         non-null, and <var>bytes</var>'s <a for="byte sequence">length</a> is less than
+         <var>stream</var>'s <a for=ReadableStream>current BYOB request view</a>'s
+         <a for=BufferSource>byte length</a>, then set <var>view</var> to <var>stream</var>'s
+         <a for=ReadableStream>current BYOB request view</a>, and <a for=ArrayBufferView>write</a>
+         <var>bytes</var> into <var>view</var>.
+
+         <li><p>Otherwise, set <var>view</var> to the result of <a for=ArrayBufferView>creating</a>
+         a {{Uint8Array}} from <var>bytes</var>.
+
          <li><p><a for=ReadableStream>Enqueue bytes</a> into <var>stream</var> given
-         <var>bytes</var>.
+         <var>view</var>.
 
          <li><p>If <var>stream</var> is <a for=ReadableStream>errored</a>, then
          <a lt=terminated for=fetch>terminate</a> the ongoing fetch.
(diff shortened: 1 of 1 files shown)


MattiasBuelens:
We're also blocked by whatwg/streams#1114. When ["clone a body" tees a stream](https://github.com/whatwg/fetch/blob/43a0c94508fb72f7f56a9c3969b15801de4a5829/fetch.bs#L1080), the result must *also* be a readable byte stream.

fetch.bs
```diff
@@ -5356,8 +5359,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>If <var>bytes</var> is failure, then <a lt=terminated for=fetch>terminate</a> the
          ongoing fetch.
 
-         <li><p><a for=ReadableStream>Enqueue</a> a {{Uint8Array}} wrapping an {{ArrayBuffer}}
-         containing <var>bytes</var> into <var>stream</var>.
+         <li><p><a for=ReadableStream>Enqueue bytes</a> into <var>stream</var> given
+         <var>bytes</var>.
```
annevk, line 5363: There is an existing problem that this needs to queue a task. I can PR that, but this will have to be rebased then.

domenic: Re-reading this section it feels like most of the section needs to take place in a task. (I.e., everything dealing with _stream_.) Just the reading of the bytes from the network needs to be done in parallel.

fetch.bs
```diff
@@ -5423,8 +5426,19 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>If <var>bytes</var> is failure, then <a lt=terminated for=fetch>terminate</a> the
          ongoing fetch.
 
-         <li><p><a for=ReadableStream>Enqueue</a> a {{Uint8Array}} wrapping an {{ArrayBuffer}}
-         containing <var>bytes</var> into <var>stream</var>.
+         <li><p>Let <var>view</var> be null.
+
+         <li><p>If <var>stream</var>'s <a for=ReadableStream>current BYOB request view</a> is
+         non-null, and <var>bytes</var>'s <a for="byte sequence">length</a> is less than
```
annevk, line 5432:
```suggestion
         non-null and <var>bytes</var>'s <a for="byte sequence">length</a> is less than
```

fetch.bs
```diff
@@ -5423,8 +5426,19 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+         <var>stream</var>'s <a for=ReadableStream>current BYOB request view</a>'s
+         <a for=BufferSource>byte length</a>, then set <var>view</var> to <var>stream</var>'s
+         <a for=ReadableStream>current BYOB request view</a>, and <a for=ArrayBufferView>write</a>
```
annevk, line 5435:
```suggestion
         <a for=ReadableStream>current BYOB request view</a> and <a for=ArrayBufferView>write</a>
```

annevk (commented):
I personally prefer not using commas before and if there are only two "simple" things.

What I don't understand here is that you still get a fresh view if the BYOB view cannot be written into due to its size. Wouldn't an error or request for a new view be more appropriate? If the idea is to avoid allocations presumably you would not want such allocations to nevertheless happen silently sometimes?

domenic:
> What I don't understand here is that you still get a fresh view if the BYOB view cannot be written into due to its size. Wouldn't an error or request for a new view be more appropriate? If the idea is to avoid allocations presumably you would not want such allocations to nevertheless happen silently sometimes?

Interesting. This is the first time we're hearing this feedback.

I'm not sure how these ideas would work. If you ask for a 1 KiB buffer, and Fetch is unable to restrict what it pulls from the network to less than 16 KiB, you'd get an error... and then what would you do? At that point Fetch has 16 KiB in memory in a new buffer. The only things we can do are:

- Give you the first 1 KiB of the 16 KiB buffer, and next time you ask give you (the requested portion of) the remaining 15 KiB; or
- Give you the entire 16 KiB buffer anyway.

The first option is more friendly toward people who are using BYOB readers to read specific numbers of bytes, e.g. when parsing a protocol, and so is what we chose.

I don't quite understand what you mean by "request for a new view" as opposed to error.

annevk:
We discussed this further in https://matrixlogs.bakkot.com/WHATWG/2021-07-19#L105-L119. I'm a bit surprised at all the transfers, but I suppose that is to avoid exposing races.

domenic:
Superseded by #1593; somehow I forgot this existed when reviewing that :)

Issue #1215, opened 2021-04-13
Clarify which network errors create a resource timing entry


noamr:
It's not clear from the spec/implementation which errors should or shouldn't create an RT entry.
CORS errors explicitly don't, though in the original Resource Timing spec they should create an RT entry. However, the discrepancy between different error types might create a case where it could be detected whether an error is of CORS nature, which should be avoided. 

See https://github.com/whatwg/fetch/pull/1202#pullrequestreview-634571341

noamr:
See https://github.com/w3c/resource-timing/issues/12

noamr:
See comment: https://github.com/whatwg/fetch/pull/1202#issuecomment-819286114

noamr:
One simplified option would be that the caller is responsible to report the RT entry in case of a network/CORS error.
The caller (e.g. a `script` tag) knows when the call to FETCH was made and when it has received an error, so it could report those timing as the start/end time to resource timing, without exposing anything about the connection info.

Not sure if we want to expose connection info and other things when there was a network error, probably not.

annevk:
Yeah, if we report network errors we want to act as if they are cross-origin and lack TAO. We could store the relevant information on the network error for the caller (it's a type of response after all, albeit rather special). That's my option 1.

Option 2 would be reporting all responses that are not network errors.

Option 3 would be reporting all responses that the caller considers actionable (i.e., would exclude non-2xx in case of `<script>`).

They are not in a preference order. I don't really know what would help web developers the most.

Whichever option we decide to go with should not provide a new information channel.

cc @yoavweiss @npm1 @rniwa @bdekoz @sefeng211

annevk:
I noticed another problem, due to network errors having an empty URL list, finalize and report timing always returns early. So as currently specified option 1 does not work.

npm1:
We briefly discussed this in the WebPerf WG call today, it wasn't clear to me if there's a use case for entries from network errors. Also @nicjansma pointed out that there's already overlap with the Reporting API.

annevk:
You mean with https://w3c.github.io/network-error-logging/? That's true. If everyone is fine to align on no entries for network errors, let's do that.

noamr:
@yoavweiss what's needed to reach consensus/conclusions on this?

nicjansma:
[...]

PR #1248, opened 2021-06-03
Set final connection timing info after failure check

annevk:
@noamr @yoavweiss could you review?

***
<a href="https://whatpr.org/fetch/1248.html" title="Last updated on Jun 3, 2021, 1:57 PM UTC (2a7e355)">Preview</a> | <a href="https://whatpr.org/fetch/1248/43a0c94...2a7e355.html" title="Last updated on Jun 3, 2021, 1:57 PM UTC (2a7e355)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5103,18 +5103,17 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
    <var>includeCredentials</var>, and <var>forceNewConnection</var>.
   </dl>
 
- <li>Set <var>timingInfo</var>'s <a for="fetch timing info">final connection timing info</a> to the
- result of calling <a>clamp and coarsen connection timing info</a> with <var>connection</var>'s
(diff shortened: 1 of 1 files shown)


yoavweiss:
Seems reasonable. IIUC, this would means that we drop all connection time reporting in case of a network error. Is that correct? Would this resolve https://github.com/whatwg/fetch/issues/1215 or is there more work involved?

We'd also want to add related tests, or at least keep track of the fact we need to add ones.

annevk:
No, doing things accurately for network errors is an outstanding issue still. This just fixes an existing problem whereby it tries to obtain final connection timing info from a failure value (i.e., not a connection).

yoavweiss (approved):
LGTM

yoavweiss:
> No, doing things accurately for network errors is an outstanding issue still. This just fixes an existing problem whereby it tries to obtain final connection timing info from a failure value (i.e., not a connection).

OK, that makes sense!

PR #1255, opened 2021-06-17
Create AbortSignal earlier in new Request()

### LINKED ISSUE #1249: `Request`: this's signal is null when it is set to follow parent signal
Might just be misunderstanding something, but I think there is a spec inconsistency. Here is my thinking:

1. this's signal of `Request` is set to `null` initially
2. step 28. of the `Request` constructor algorithm reads: "If signal is not null, then make this’s signal follow signal."
3. as per [follow a signal](https://dom.spec.whatwg.org/#abortsignal-follow), following doesn't create a signal, it only sets a child signal to follow a parent signal
4. at the time when step 28 of `Request` constructor is run, this's signal is still null.

Shouldn't either the signal get initialized to an AbortSignal, or have step 28 create a new signal before setting it to follow the parent?

Additionally as per the current spec wording, step 41 of `Request` constructor would reset any existing this's signal to a new `AbortSignal`. Maybe this should just be moved up before step 28?

annevk:
This regressed in 987c3fd3def02b630cce174cc33b827e64203b19.

Fixes #1249.

***
<a href="https://whatpr.org/fetch/1255.html" title="Last updated on Jun 17, 2021, 4:21 PM UTC (055acb9)">Preview</a> | <a href="https://whatpr.org/fetch/1255/acef461...055acb9.html" title="Last updated on Jun 17, 2021, 4:21 PM UTC (055acb9)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6880,6 +6880,9 @@ constructor steps are:
 
  <li><p>Set <a>this</a>'s <a for=Request>request</a> to <var>request</var>.
 
+ <li><p>Set <a>this</a>'s <a for=Request>signal</a> to a <a for=/>new</a> {{AbortSignal}} object
+ with <a>this</a>'s <a>relevant Realm</a>.
+
  <li><p>If <var>signal</var> is not null, then make <a>this</a>'s <a for=Request>signal</a>
  <a for=AbortSignal>follow</a> <var>signal</var>.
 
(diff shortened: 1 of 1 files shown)


annevk:
@lucacasonato want to review as well?

lucacasonato (approved):
LGTM!

PR #1258, opened 2021-06-22
Adapt to embedder policy being in policy container

antosart:
This is an editorial PR which adapts fetch to embedder policy moving to the policy container (https://github.com/whatwg/html/pull/6793).

As a side effect, this fixes possible inconsistencies caused by concurrent accessing an environment settings object's embedder policy during fetch, since now fetch uses instead the embedder policy of the request's policy container (snapshotted at the beginning of fetch) for the checks.

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3413,15 +3413,13 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
 </code></pre>
 
 <p>To perform a <dfn export>cross-origin resource policy check</dfn>, given an <a for=url>origin</a>
-<var>origin</var>, an <a for=/>environment settings object</a> <var>settingsObject</var>, a string
-<var>destination</var>, a <a for=/>response</a> <var>response</var>, and an optional boolean
-<var>forNavigation</var>, run these steps:
+<var>origin</var>, an <a for=/>environment settings object</a> <var>settingsObject</var>, an
+<a for=/>embedder policy</a> <var>embedderPolicy</var>, a string <var>destination</var>, a
+<a for=/>response</a> <var>response</var>, and an optional boolean <var>forNavigation</var>,
+run these steps:
 
 <ol>
  <li><p>Set <var>forNavigation</var> to false if it is not given.
-
- <li><p>Let <var>embedderPolicy</var> be <var>settingsObject</var>'s
- <a for="environment settings object">embedder policy</a>.
 
  <li>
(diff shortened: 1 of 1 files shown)


antosart:
This depends on https://github.com/whatwg/html/pull/6793.

fetch.bs
```diff
@@ -3413,16 +3413,14 @@ Cross-Origin-Resource-Policy     = %s"same-origin" / %s"same-site" / %s"cross-or
...
+<var>origin</var>, an <a for=/>environment settings object</a> <var>settingsObject</var>, an
+<a for=/>embedder policy</a> <var>embedderPolicy</var>, a string <var>destination</var>, a
+<a for=/>response</a> <var>response</var>, and an optional boolean <var>forNavigation</var>,
+run these steps:
 
 <ol>
  <li><p>Set <var>forNavigation</var> to false if it is not given.
 
- <li><p>Let <var>embedderPolicy</var> be <var>settingsObject</var>'s
- <a for="environment settings object">embedder policy</a>.
```
annevk, line 3424: Can we not do settingsObject -> policy container -> embedder policy and avoid the additional argument?

antosart: Right. In fact, we can completely abandon this PR, everything still works.

I like slightly better to take policies from the request's policy container (which is copied and set when starting the fetch) than from the settings object in the middle of fetching, since that ensures consistency. Embedder policy, however, cannot change, so we can also just leave this.

annevk: I see, I agree with that. It's unfortunate we need to keep the _settingsObject_ argument for communication purposes as it creates the opportunity for subtle mistakes to creep in. Perhaps if we keep this as-is we should at least add a note here that this is fine because embedder policy cannot change over time.

Anyone else have an opinion one way or another?

annevk: @antosart it seems we cannot abandon this PR as _settingsObject_ does not hold an embedder policy directly (and now #1259 fails to build, for instance).

antosart: Right, sorry! So are you fine with this PR? Otherwise, I can prepare a one-liner PR to replace settingsObject's embedder policy with settingsObject's policy container's embedder policy to fix the build.

antosart: @domenic maybe you have an opinion here? It would be good to decide here, since the reference is currently undefined.

domenic: A one-line PR makes more sense to me.

antosart: Sounds good. I opened https://github.com/whatwg/fetch/pull/1264. It ended up being 6 lines though :)

PR #1259, opened 2021-06-24
Add proxy support to obtain a connection

annevk:
This uses some of the language added to #1244 but ports it to "obtain a connection" instead. The one thing that's missing a bit is how _proxy_ is supposed to be used. @ricea can you help with that?

@yutakahirano this also includes your suggested change to the "obtain a connection" parameters by merging _forceNew_ and _dedicated_.

***
<a href="https://whatpr.org/fetch/1259.html" title="Last updated on Jul 19, 2021, 5:37 PM UTC (29c7bda)">Preview</a> | <a href="https://whatpr.org/fetch/1259/613aad9...29c7bda.html" title="Last updated on Jul 19, 2021, 5:37 PM UTC (29c7bda)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2349,15 +2349,15 @@ steps:
   </ol>
 
  <li>
-  <p>Let <var>proxies</var> be the result of performing proxy resolution for <var>url</var> in an
-  <a>implementation-defined</a> manner. If a proxy preference order has not been explicitly
-  configured, then SOCKS proxies should be preferred to HTTPS proxies which should be preferred to
-  HTTP proxies. If there is no proxy, let <var>proxies</var> be « "<code>DIRECT</code>" ».
+  <p>Let <var>proxies</var> be the result of finding proxies for <var>url</var> in an
+  <a>implementation-defined</a> manner. If there are no proxies, let <var>proxies</var> be
+  « "<code>DIRECT</code>" ».
 
   <p class=note>This is where non-standard technology such as
   <a href="https://en.wikipedia.org/wiki/Web_Proxy_Auto-Discovery_Protocol">Web Proxy Auto-Discovery Protocol (WPAD)</a>
   and <a href="https://en.wikipedia.org/wiki/Proxy_auto-config">proxy auto-config (PAC)</a> come
-  into play.
+  into play. The "<code>DIRECT</code>" value means to not use a proxy for this particular
+  <var>url</var>.
 
  <li><p>Let <var>timingInfo</var> be a new <a for=/>connection timing info</a>.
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2349,38 +2348,59 @@ steps:
    <a>connection</a>.
   </ol>
 
+ <li>
+  <p>Let <var>proxies</var> be the result of performing proxy resolution for <var>url</var> in an
+  <a>implementation-defined</a> manner. If a proxy preference order has not been explicitly
+  configured, then SOCKS proxies should be preferred to HTTPS proxies which should be preferred to
```
ricea, line 2354: This part is WebSocket-specific.

For HTTP and HTTPS we prefer to match the proxy type.

The reasoning is that the proxy configured for "HTTPS" may be optimised for the "CONNECT" method (ie. low latency, no cache) and the proxy configured for "HTTP" may be optimised oppositely. Which protocol they use is orthogonal to what protocol they are configured for. In fact, it might be perfectly reasonable to use `https://proxy.local/` for the HTTP proxy and `http://proxy.local:3128/` for the HTTPS proxy.

SOCKS works like a fallback in the non-WebSocket case, and will only be used if a matching proxy type is not there.

This only applies to static configurations. If proxy configuration happens via a JavaScript file (`proxy.pac` or `wpad.dat`), it decides the priority.

The way that manual proxy configuration works varies by OS, and we probably shouldn't say anything about the priority order except in the WebSocket case where the underlying RFC6455 attempts to dictate it.

ricea (commented):
I'm not sure what if anything we need to write to handle complex cases like when `FindProxyForURL()` returns something like `"PROXY proxy:3128; PROXY proxy2:8080; DIRECT"`. In this case 

1. Try `proxy:3128` for "example.com"
2. If connection failed, try `proxy2:8080` for "example.com"
3. If connection failed, do a host lookup on "example.com"
4. Attempt connection to one or more of the IPs returned by the host lookup.

fetch.bs
```diff
@@ -2349,38 +2348,59 @@ steps:
...
+  HTTP proxies. If there is no proxy, let <var>proxies</var> be « "<code>DIRECT</code>" ».
+
+  <p class=note>This is where non-standard technology such as
+  <a href="https://en.wikipedia.org/wiki/Web_Proxy_Auto-Discovery_Protocol">Web Proxy Auto-Discovery Protocol (WPAD)</a>
+  and <a href="https://en.wikipedia.org/wiki/Proxy_auto-config">proxy auto-config (PAC)</a> come
+  into play.
```
mnot, line 2360: I don't know if it's a good thing to mention WPAD here; it's widely considered to be a huge security issue.

fetch.bs
```diff
@@ -2349,38 +2348,59 @@ steps:
...
+   <li><p>Set <var>timingInfo</var>'s <a for="connection timing info">domain lookup start time</a>
+   to the <a for=/>unsafe shared current time</a>.
 
- <li><p>Let <var>hosts</var> be the result of running <a>resolve an origin</a> given <var>key</var>
- and <var>origin</var>.
+   <li><p>Let <var>hosts</var> be « <var>origin</var>'s <a for=origin>host</a> ».
 
- <li><p>If <var>hosts</var> is failure, then return failure.
+   <li><p>If <var>proxy</var> is "<code>DIRECT</code>", then set <var>hosts</var> to the result of
+   running <a>resolve an origin</a> given <var>key</var> and <var>url</var>'s <a for=url>origin</a>.
```
mnot, line 2374: `DIRECT` isn't really defined anywhere...

fetch.bs
ricea: Yes, but everyone implements it. Ignoring it won't make it go away.

ricea:
> I'm not sure what if anything we need to write to handle complex cases like when `FindProxyForURL()` returns something like `"PROXY proxy:3128; PROXY proxy2:8080; DIRECT"`. In this case

I was wrong, this is already handled.

fetch.bs
annevk: I would be open to pointing out the security issues.

annevk:
I've removed the WebSocket bits from this PR. I suggest we add those in #1244 by introducing a new optional boolean websocket parameter to obtain a connection that handles proxy preference sorting and dedicated HTTP/1 or (shared) HTTP/2+ connection.

I would be open to adding more details/notes around proxies, including pointing out pitfalls and dangers. I'm also okay with doing that as a follow-up. (Also very amenable to someone trying to standardize them more properly. It's somewhat astonishing there's essentially no standard for 25yo technology that works everywhere.)

redransom13 (approved):

ricea (approved):
I think that covers everything we need.

Rajakashi420:
> I'm not sure what if anything we need to write to handle complex cases like when `FindProxyForURL()` returns something like `"PROXY proxy:3128; PROXY proxy2:8080; DIRECT"`. In this case
> 
> 1. Try `proxy:3128` for "example.com"
> 2. If connection failed, try `proxy2:8080` for "example.com"
> 3. If connection failed, do a host lookup on "example.com"
> 4. Attempt connection to one or more of the IPs returned by the host lookup.

PR #1273, opened 2021-07-23
Support setting websocket credential mode

lucacasonato:
Accompanying change for https://github.com/whatwg/html/pull/6896

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7533,7 +7533,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <h3 id=websocket-opening-handshake>Opening handshake</h3>
 
 <p>To <dfn id=concept-websocket-establish>establish a WebSocket connection</dfn>, given a
-<var>url</var>, <var>protocols</var>, and <var>client</var>, run these steps:
+<var>url</var>, <var>protocols</var>, <var>credentials mode</var>, and <var>client</var>, run these steps:
 
 <ol>
  <li>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -7533,7 +7533,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <h3 id=websocket-opening-handshake>Opening handshake</h3>
 
 <p>To <dfn id=concept-websocket-establish>establish a WebSocket connection</dfn>, given a
-<var>url</var>, <var>protocols</var>, and <var>client</var>, run these steps:
+<var>url</var>, <var>protocols</var>, <var>credentials mode</var>, and <var>client</var>, run these steps:
```
annevk, line 7536:
```suggestion
<var>url</var>, <var>protocols</var>, <var>credentialsMode</var>, and <var>client</var>, run these steps:
```
and the same below. That makes it slightly clearer it's a single variable and not two.

fetch.bs
```diff
@@ -7533,7 +7533,7 @@ therefore not shareable, a WebSocket connection is very close to identical to an
 <h3 id=websocket-opening-handshake>Opening handshake</h3>
 
 <p>To <dfn id=concept-websocket-establish>establish a WebSocket connection</dfn>, given a
-<var>url</var>, <var>protocols</var>, and <var>client</var>, run these steps:
+<var>url</var>, <var>protocols</var>, <var>credentialsMode</var>, and <var>client</var>, run these steps:
```
annevk, line 7536: Nit: exceeds a 100 columns.

annevk (commented):
Missed a nit, but looks good to me apart from that. I guess the main thing here is getting implementer interest and tests, probably best managed as part of the HTML change.

PR #1309, opened 2021-09-23
Expose a mechanism for hiding internal timing info

noamr:
This is needed for https://github.com/whatwg/html/issues/7104
and later on for https://github.com/whatwg/fetch/issues/1215.

Navigation timing reports the timing info from the HTML spec,
so it needs a mechanism to obfuscate the internals. So far that
obfuscation was internal to fetch and was done upon reporting.

***
<a href="https://whatpr.org/fetch/1309.html" title="Last updated on Sep 28, 2021, 2:33 PM UTC (d6889b0)">Preview</a> | <a href="https://whatpr.org/fetch/1309/b2f04e2...d6889b0.html" title="Last updated on Sep 28, 2021, 2:33 PM UTC (d6889b0)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -233,7 +233,7 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
 
 <p>To <dfn export lt="create an opaque timing info|creating an opaque timing info">create an
 opaque timing info</dfn>, given a <a for=/>fetch timing info</a> <var>timingInfo</var>,
- create a new <a for=/>fetch timing info</a> whose <a for="fetch timing info">start time</a> and
+ return a new <a for=/>fetch timing info</a> whose <a for="fetch timing info">start time</a> and
  <a for="fetch timing info">post-redirect start time</a> are <var>timingInfo</var>'s
  <a for="fetch timing info">start time</a>.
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -231,6 +231,12 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
  <dd>Null or a <a for=/>connection timing info</a>.
 </dl>
 
+<p>To <dfn export lt="create an opaque timing info|creating an opaque timing info">create an
+opaque timing info</dfn>, given a <a for=/>fetch timing info</a> <var>timingInfo</var>,
+ create a new <a for=/>fetch timing info</a> whose <a for="fetch timing info">start time</a> and
```
annevk, line 236:
```suggestion
 return a new <a for=/>fetch timing info</a> whose <a for="fetch timing info">start time</a> and
```

fetch.bs
```diff
@@ -231,6 +231,12 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
...
+ <a for="fetch timing info">post-redirect start time</a> are <var>timingInfo</var>'s
+ <a for="fetch timing info">start time</a>.
```
annevk, line 238: Indentation is wrong in this block. Also has newlines inside phrasing elements.

annevk (commented):
Looks good modulo nits. If we end up creating fetch timing info instances in more places, we should perhaps consider an abstraction for that too. In the two places where we do that today in Fetch we always set start time and post-redirect start time to the same value and adjust nothing else.

fetch.bs

PR #1311, opened 2021-09-26
Attach timing info and URL to network errors, and report for fetch API

### LINKED ISSUE #1215: Clarify which network errors create a resource timing entry

noamr:
The timing info attached to a network error is always "opaque",
containing only start/end time and the original request URL.

This change currently only applies to the fetch API, and should be
applied to the different callers of FETCCH as part of [this](whatwg/html#6542) and [this](whatwg/xhr#319) work.

Closes #1215
- [x ] At least two implementers are interested (and none opposed):
   * See discussion at #1215.

   * Currently WIP, see https://github.com/web-platform-tests/wpt/issues/30968
   * Chrome: …
   * Firefox: …
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=193902

***
<a href="https://whatpr.org/fetch/1311.html" title="Last updated on Dec 13, 2021, 12:17 PM UTC (78010e3)">Preview</a> | <a href="https://whatpr.org/fetch/1311/6257e22...78010e3.html" title="Last updated on Dec 13, 2021, 12:17 PM UTC (78010e3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4213,6 +4213,25 @@ steps:
   </ol>
  </li>
 
+ <li>
+  <p>Otherwise, if <var>response</var>'s <a for=response>body</a> is not null and is
+  <a for=ReadableStream>readable</a>, then:</o>
+
+  <ol>
+   <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
+
+   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
+   <a for=TransformStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
+
+   <li><p><a for=TransformStream lt="setting up">Set up</a> <var>transformStream</var> with
+   <var>identityTransformAlgorithm</var> and <var>finalize</var>.
+
+   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
+   <a for=ReadableStream lt="piping through">piping</a> <var>response</var>'s
+   <a for=response>body</a> through <var>transformStream</var>.
+  </ol>
+ </li>
+
  <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response</a> is non-null, then
  <a>queue a fetch task</a> to run <var>fetchParams</var>'s
  <a for="fetch params">process response</a> given <var>response</var>, with <var>fetchParams</var>'s
@@ -4254,26 +4273,8 @@ steps:
   </ol>
  </li>
 
- <li>
-  <p>Otherwise, if <var>response</var>'s <a for=response>body</a> is not null and is
-  <a for=ReadableStream>readable</a>, then:</o>
-
-  <ol>
-   <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
-
-   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
-   <a for=TransformStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
-
-   <li><p><a for=TransformStream lt="setting up">Set up</a> <var>transformStream</var> with
-   <var>identityTransformAlgorithm</var> and <var>finalize</var>.
-
-   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
-   <a for=ReadableStream lt="piping through">piping</a> <var>response</var>'s
-   <a for=response>body</a> through <var>transformStream</var>.
-  </ol>
- </li>
-
- <li><p>Otherwise, run <var>finalize</var>.</p></li>
+ <li><p>If <var>response</var>'s <a for=response>body</a> is null or
+ not <a for=ReadableStream>readable</a>, then run <var>finalize</var>.</p></li>
 </ol>
 
 <p>To <dfn>finalize response</dfn> given a <a for=/>fetch params</a> <var>fetchParams</var> and a
--- a/fetch.bs
+++ b/fetch.bs
@@ -203,8 +203,6 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication). 
  <dd>Null, a <a for=/>global object</a>, or a <a for=/>parallel queue</a>.
 
  <dt><dfn for="fetch params">cross-origin isolated capability</dfn> (default false)
- <dt><dfn for="fetch params">ready for clients</dfn> (default false).
- <dt><dfn for="fetch params">network read complete</dfn> (default false).
  <dd>A boolean.
 
  <dt><dfn for="fetch params">timing info</dfn>
@@ -4212,30 +4210,39 @@ steps:
    <li><p>Set <var>response</var>'s <a for=response>timing info</a> to the result of
    <a>creating an opaque timing info</a> for <var>fetchParams</var>'s
    <a for="fetch params">timing info</a>.</p></li>
-
-   <li><p>Set <var>fetchParams</var>'s <a for="fetch params">network read complete</a> to true.
   </ol>
  </li>
-
- <li><p>If <var>response</var>'s <a for="response">cache state</a> is not the empty string, then
- set <var>fetchParams</var>'s <a for="fetch params">network read complete</a> to true.</p></li>
 
  <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response</a> is non-null, then
  <a>queue a fetch task</a> to run <var>fetchParams</var>'s
  <a for="fetch params">process response</a> given <var>response</var>, with <var>fetchParams</var>'s
  <a for="fetch params">task destination</a>.
 
+ <li><p>Let <var>finalize</var> be this step: <a>finalize response</a> given <var>fetchParams</var>
+ and <var>response</var>.
+
  <li>
   <p>If <var>fetchParams</var>'s <a for="fetch params">process response end-of-body</a> is non-null,
   then:
 
   <ol>
-   <li><p>Let <var>processBody</var> given <var>nullOrBytes</var> be this step: run
-   <var>fetchParams</var>'s <a for="fetch params">process response end-of-body</a> given
-   <var>response</var> and <var>nullOrBytes</var>.
+   <li>
+    <p>Let <var>processEndOfBody</var> given <var>nullBytesOrFailure</var> be the following
+    steps:
 
-   <li><p>Let <var>processBodyError</var> be this step: run <var>fetchParams</var>'s
-   <a for="fetch params">process response end-of-body</a> given <var>response</var> and failure.
+    <ol>
+     <li><p>Run <var>finalize</var>.
+
+     <li><p>Run <var>fetchParams</var>'s <a for="fetch params">process response end-of-body</a>
+     given <var>response</vaR> and <var>nullBytesOrFailure</var>.
+    </ol>
+   </li>
+
+   <li><p>Let <var>processBody</var> given <var>nullOrBytes</var> be this step: run
+   <var>processEndOfBody</var> with <var>nullOrBytes</var>.
+
+   <li><p>Let <var>processBodyError</var> be this step: run <var>processEndOfBody</var>given
+   failure.
 
    <li><p>If <var>response</var>'s <a for=response>body</a> is null, then <a>queue a fetch task</a>
    to run <var>processBody</var> given null, with <var>fetchParams</var>'s
@@ -4248,41 +4255,35 @@ steps:
  </li>
 
  <li>
-  <p><a>Queue a fetch task</a> to run the following steps with <var>fetchParams</var>'s
-  <a for="fetch params">task destination</a>:</p>
+  <p>Otherwise, if <var>response</var>'s <a for=response>body</a> is not null and is
+  <a for=ReadableStream>readable</a>, then:</o>
 
   <ol>
-   <li><p>Set <var>fetchParams</var>'s <a for="fetch params">ready for clients</a> to true.</p></li>
+   <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
 
-   <li><p><a>Finalize response</a> given <var>fetchParams</var> and <var>response</var>.</p></li>
+   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
+   <a for=ReadableStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
+
+   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
+   <a for=TransformStream>setting up</a> <var>transformStream</var> with
+   <var>identityTransformAlgorithm</var> and <var>finalize</var>.
   </ol>
-</ol>
+ </li>
 
-<p>To <dfn>fetch network finale</dfn> given a <a for=/>fetch params</a> <var>fetchParams</var>
-and a <a for=/>response</a> <var>response</var>, <a>queue a fetch task</a>
-given the following steps with <var>fetchParams</var>'s
-<a for="fetch params">task destination</a>:</p>
-
-<ol>
- <li><p>Set <var>fetchParams</var>'s
- <a for="fetch params">network read complete</a> to true.</p></li>
-
- <li><p><a>Finalize response</a> given <var>fetchParams</var> and <var>response</var>.</p></li>
+ <li><p>Otherwise, run <var>finalize</var>.</p></li>
 </ol>
 
 <p>To <dfn>finalize response</dfn> given a <a for=/>fetch params</a> <var>fetchParams</var> and a
 <a for=/>response</a> <var>response</var>, run these steps:
 
 <ol>
- <li><p>If <var>fetchParams</var>'s <a for="fetch params">ready for clients</a> is false or
- <var>fetchParams</var>'s <a for="fetch params">network read complete</a> is false, return.</p></li>
-
  <li><p>Set <var>fetchParams</var>'s <a for="fetch params">request</a>'s
  <a for=request>done flag</a>.
 
  <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response done</a> is not null,
- then run <var>fetchParams</var>'s <a for="fetch params">process response done</a> with
- <var>response</var>.
+ then <a>queue a fetch task</a> to run <var>fetchParams</var>'s
+ <a for="fetch params">process response done</a> given <var>response</var> with
+ <var>fetchParams</var>'s <a for="fetch params">task destination</a>.
 </ol>
 
 <p>To <dfn export>finalize and report timing</dfn> given a <a for=/>response</a>
@@ -5636,8 +5637,7 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
        <li><p>Otherwise, if the bytes transmission for <var>response</var>'s message body is done
        normally and <var>stream</var> is <a for=ReadableStream>readable</a>, then
-       <a for=ReadableStream>close</a> <var>stream</var>, run <a for=/>fetch network finale</a> for
-       <var>fetchParams</var> and <var>response</var>, and abort these in-parallel steps.
+       <a for=ReadableStream>close</a> <var>stream</var>, and abort these in-parallel steps.
       </ol>
     </ol>
 
@@ -5645,7 +5645,7 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
     <p><a>If aborted</a>, then:
 
     <ol>
-     <li>Run <a for=/>fetch network finale</a> for <var>fetchParams</var> and <var>response</var>.
+     <li><p><a>Finalize response</a> given <var>fetchParams</var> and <var>response</var>.</p></li>
 
      <li><p>Let <var>aborted</var> be the termination's aborted flag.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -4200,7 +4200,7 @@ steps:
   <ol>
    <li>
     <p>Set <var>response</var>'s <a for=response>URL list</a> to « <var>fetchParams</var>'s
-    <a for="fetch params">request</a>'s <a for=request>URL list</a> ≫.
+    <a for="fetch params">request</a>'s <a for=request>URL list</a>[0] ≫.
 
     <p class=note>This is needed as after <a for=list>cloning</a> <var>fetchParams</var>'s
     <a for="fetch params">request</a>'s <a for=request>URL list</a> earlier, <var>response</var>
@@ -4212,9 +4212,6 @@ steps:
    <a for="fetch params">timing info</a>.</p></li>
   </ol>
  </li>
-
- <li><p>Let <var>hasReadableStream</var> be true if <var>response</var>'s <a for=response>body</a>
- is not null and is <a for=ReadableStream>readable</a>; otherwise false.
 
  <li>
   <p>Let <var>processResponseDone</var> be the following steps:
@@ -4230,8 +4227,11 @@ steps:
   </ol>
  </li>
 
+ <li><p>If <var>response</var>'s <a for=response>body</a> is null, then run
+ <var>processResponseDone</var>.
+
  <li>
-  <p>If <var>hasReadableStream</var> is true, then:
+  <p>Otherwise:</p>
 
   <ol>
    <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -7417,8 +7479,14 @@ method steps are:
...
-   <li><p>If <var>response</var> is a <a>network error</a>, then <a for=/>reject</a> <var>p</var>
-   with a {{TypeError}} and terminate these substeps.
+   <li><p>If <var>response</var> is a <a>network error</a>, then:
+   <ol>
+    <li><p><a for=/>Reject</a> <var>p</var> with a {{TypeError}}.
+
+    <li><p>Call <var>handleFetchDone</var> with <var>response</var>.
+
+    <li><p>terminate these substeps.
+   </ol>
```
annevk, line 7489: I think this is wrong. At least, in my mental model processResponseDone should run when you get a network error.

fetch.bs
```diff
@@ -2011,6 +2017,14 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
...
 <a for=response>body</a> is always null.
 
+<p>To <dfn for="network error" lt="create|creating">create</dfn> a <a>network error</a> given
+<a for=/>fetch params</a> <var>fetchParams</var>, return a new <a>network error</a> with its
+<a for=response>timing info</a> set to the result of <a>creating an opaque timing info</a> with
+<var>fetchParams</var>'s <a for="fetch params">timing info</a>, and its
+<a for=response>URL list</a> set to a <a for=/>list</a> containing <var>fetchParams</var>'s
+<a for="fetch params">request</a>'s <a for=request>URL</a>.
+
+
```
annevk, line 2027: Nit: single newline here.

fetch.bs
```diff
@@ -2011,6 +2017,14 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
 <a for=response>header list</a> is always empty, and
 <a for=response>body</a> is always null.
 
+<p>To <dfn for="network error" lt="create|creating">create</dfn> a <a>network error</a> given
+<a for=/>fetch params</a> <var>fetchParams</var>, return a new <a>network error</a> with its
+<a for=response>timing info</a> set to the result of <a>creating an opaque timing info</a> with
+<var>fetchParams</var>'s <a for="fetch params">timing info</a>, and its
+<a for=response>URL list</a> set to a <a for=/>list</a> containing <var>fetchParams</var>'s
+<a for="fetch params">request</a>'s <a for=request>URL</a>.
```
annevk, line 2025: Nit: it seems that the network error is an argument to this algorithm currently although not an explicit one. E.g., sometimes this is passed an aborted network error rather than a network error. As such, it seems more like this algorithm initializes a network error rather than creating one.

annevk (commented):
Generally this seems to be heading in the right direction. Thanks for tackling this!

I haven't checked in detail whether this overlooks any network error creation.

I think the creation algorithm needs some slight tweaks.

And I think we want to change things such that processResponseDone also runs in the event of a network error.

annevk:
Looking at this again I think I missed something initially and this could be a lot simpler. All these network errors end up in "fetch finale" and it seems that at that point we could initialize them.

That leaves "finalize response". I suppose that for some network errors (in particular CORS) there could be a successful body stream so this would come later. For others that might not happen. Either way, we probably want to report them as soon as possible (i.e., as part of "fetch finale") to avoid leaking timing information. (Which also means "finalize response" (or its callers) has to filter out network errors or some such.)

Does that make sense or am I missing something here?

noamr:
> Looking at this again I think I missed something initially and this could be a lot simpler. All these network errors end up in "fetch finale" and it seems that at that point we could initialize them.

Yes, that would simplify things!

> That leaves "finalize response". I suppose that for some network errors (in particular CORS) there could be a successful body stream so this would come later. For others that might not happen. Either way, we probably want to report them as soon as possible (i.e., as part of "fetch finale") to avoid leaking timing information. (Which also means "finalize response" (or its callers) has to filter out network errors or some such.)
> 
> Does that make sense or am I missing something here?

We're already reporting them in the fetch finale, by calling `process response done` which would schedule the caller to run some steps, one of which would be to report timing for the response with the correct `initiatorType`, and allowing the caller to decide the order of events / reporting.

annevk:
As far as I can tell "fetch finale" does not invoke "process response done"; that's "finalize response".

noamr:
> As far as I can tell "fetch finale" does not invoke "process response done"; that's "finalize response".

It does now for the network error case (see new commit, a lot simpler indeed)

fetch.bs
```diff
@@ -4103,6 +4103,23 @@ steps:
    <var>processBody</var>, <var>processBodyError</var>, and <var>fetchParams</var>'s
    <a for="fetch params">task destination</a>.
   </ol>
+
+ <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response done</a> is non-null, and
```
annevk, line 4107: Nit: li has multiple children so p needs to be on a newline and indented. ol below also needs indentation.

fetch.bs
```diff
@@ -4103,6 +4103,23 @@ steps:
...
+ <var>response</var> is a <a>network error</a>, then:
+
+ <ol>
+  <li><p>Set <var>response</var>'s <a for=response>URL list</a> to a <a for=/>list</a> containing
+  <var>fetchParams</var>'s <a for="fetch params">request</a>'s <a for=request>URL</a>.
+
+  <li><p>Set <var>response</var>'s <a for=response>timing info</a> to the result of
+  <a>creating an opaque timing info</a> for <var>fetchParams</var>'s
+  <a for="fetch params">timing info</a>.
```
annevk, line 4116: This would mean that only a network error through this callback is initialized this way. I think we want to do this at the top of "fetch finale".

annevk (commented):
As I tried to point out, this doesn't prevent "finalize response" from being invoked and running "process response done" again. So you need to early exit for network errors there I think.

I also think we want to do something similar for "process response end-of-body". Where we check if the response's body is null, we should also check if it's a network error and do the same thing.

fetch.bs
noamr: Right, fixed.

fetch.bs
noamr: It doesn't have multiple children in the new commit.

noamr:
> That leaves "finalize response". I suppose that for some network errors (in particular CORS) there could be a successful body stream so this would come later. For others that might not happen. Either way, we probably want to report them as soon as possible (i.e., as part of "fetch finale") to avoid leaking timing information. (Which also means "finalize response" (or its callers) has to filter out network errors or some such.)

Are you suggesting to omit the timing entry if the network error comes after the body stream has been opened?
I don't think it leaks any information. The fetch client knows when they get the `error` event, and the timing entry should correspond to that.

noamr:
> As I tried to point out, this doesn't prevent "finalize response" from being invoked and running "process response done" again. So you need to early exit for network errors there I think.
> 

I've added a check for the `done` flag which should cover the case of double reporting.

annevk:
@noamr maybe that one isn't needed. I was thinking that step 3 needed to start with "If response is a network error or response’s body is null" to ensure we don't end up reading a network error, but by definition a network error has no body so it should already do the right thing. In the case I was concerned with (CORS error) I forgot that we end up returning a new network error) and basically forget (and don't invoke callbacks) about the network response. If you could double check that I'd appreciate it though.

annevk:
We don't know that progressResponseDone is only used for reporting and callers might well rely on the other; it also seems wrongish for it to happen before the error event as that would be a new timing channel.

noamr:
> > * Finalize response is called too soon in fetch finale. The other callbacks need to happen first.
> 
> It's there on purpose. I want to make sure `processResposeDone` is always called before the other ones for network errors, so that reporting etc. would happen before any `onerror` events.

Though I guess perhaps it would more sense to call `done` after, and to use the `process response` and `process response end-of-body` callbacks in the HTML spec callers.

noamr:
> We don't know that progressResponseDone is only used for reporting and callers might well rely on the order; it also seems wrongish for it to happen before the error event as that would be a new timing channel.

In any case, setting the URL list and timing info for the error response should be done before calling the other callbacks, in case they do the reporting, so I should split this condition into two.

noamr:
> That's not what I meant. Basically when there's a request we get a response from the server. Then we start processing the response while also processing its body _in parallel_. Processing its body in parallel eventually calls "process response done" (through finalize response), but if processing the response turns into a network error that will also want to call "process response done" (through finalize response). The question is whether the latter always happens before the former.
> 
> And I don't think that's necessarily the case when integrity is involved as that also awaits the body.
> 
> Two questions then:
> 
> 1. Is that the only problematic case?
> 2. How do we solve this? The fundamental problem here is the finalize response call from the place where we are receiving bytes from the network as we are not certain at that point whether that is the final response. Here is a rough idea:
>    
>    * Fetch finale adds a listener to response if the response is not a network error. (If it is a network error it can invoke finalize response itself.)
>    * Where we are receiving bytes from the network we check if a listener was added and only then do we invoke it. (I suppose this could also be a flag of sorts.)
>    * To avoid race conditions, where we are receiving bytes from the network also sets the done flag.
>    * Fetch finale also checks the done flag and if the done flag is already set it doesn't add a listener and invokes finalize response directly. It also sets the done flag for network errors.
>    * Finalize response no longer deals in the done flag.
> 
> I'd love review from @yutakahirano and maybe @ricea on this, as well as you @noamr and @yoavweiss.

OK now I understand what you meant.

I tried something similar in the new PR amendment.
`finalize` becomes a callback associated with `response`, that is only set during finale for the non network-error case. This should cover the integrity case, and any potential future case where the body is read before the finale.

fetch.bs
```diff
@@ -4165,19 +4188,43 @@ steps:
...
+  <ol>
+   <li><p>Set <var>fetchParams</var>'s <a for="fetch params">ready for clients</a> to true.</p></li>
+
+   <li><p><a>Finalize response</a> given <var>fetchParams</var> and <var>response</var>.</p></li>
+  </ol>
+</ol>
+
+<p>To <dfn>fetch network finale</dfn> given a <a for=/>fetch params</a> <var>fetchParams</var>
+and a <a for=/>response</a> <var>response</var>, <a>queue a fetch task</a>
+with to run the following steps with <var>fetchParams</var>'s
```
yoavweiss, line 4206: Spurious "with"?

yoavweiss:
> See [web-platform-tests/wpt#30970](https://github.com/web-platform-tests/wpt/pull/30970)
> 
> For now, reporting RT entries for fetch() network errors. In later PRs, for other resources.

Seems like there are a lot of failures, so fixing this would require work on the Chromium side. @npm1 & @yutakahirano - let's discuss off-thread to figure out an implementation plan.

npm1:
I'm not super familiar with this, but it sounds like this could be non-trivial to implement; is that right? In this case, are other implementors are also willing to implement this change? I don't recall very strong use cases for the error entries on the call (and there is already NEL). Don't mean to ask to revisit this but we should make sure browsers are going to align on whichever fix we choose to spec.

noamr:
> I'm not super familiar with this, but it sounds like this could be non-trivial to implement; is that right? In this case, are other implementors are also willing to implement this change? I don't recall very strong use cases for the error entries on the call (and there is already NEL). Don't mean to ask to revisit this but we should make sure browsers are going to align on whichever fix we choose to spec.

I think browsers should align on either this or on also not reporting RT entries for CORS failures, which is something that's currently different between implementations and exposes cross-domain information.

noamr:
> > I'm not super familiar with this, but it sounds like this could be non-trivial to implement; is that right? In this case, are other implementors are also willing to implement this change? I don't recall very strong use cases for the error entries on the call (and there is already NEL). Don't mean to ask to revisit this but we should make sure browsers are going to align on whichever fix we choose to spec.
> 
> I think browsers should align on either this or on also not reporting RT entries for CORS failures, which is something that's currently different between implementations and exposes cross-domain information.

@npm1 I checked Chrome & Firefox now and it does report network errors as RT entries for a lot of common network errors (CORS, time out, connection refused, etc). Safari does not report any of those.

One difference from this spec PR is that errors (and successes) are only reported for HTTP(S) fetches, which is perhaps a good idea to add to the spec and remove non-HTTP(S) tests from the WPT PR.

So to summarize there is no current security issue in implementations, but implementations are incompatible. Currently Safari is conforming to the spec (not reporting network errors to timeline) and Firefox/Chrome are not, until this PR and subsequent reporting PRs are accepted.

/cc @yoavweiss

noamr:
After making an exception for aborts and non-HTTP(s) URLs, this now matches the Chrome/Firefox behavior + reported bugs like [this one](https://bugzilla.mozilla.org/show_bug.cgi?id=1523275). Corresponding [WPT PR](https://github.com/web-platform-tests/wpt/pull/30970) is also ready to be merged.

fetch.bs
```diff
@@ -4165,19 +4188,43 @@ steps:
...
+ <li><p>Set <var>fetchParams</var>'s <a for="fetch params">network read complete</a>.</p></li>
+
+ <li><p><a>Finalize response</a> given <var>fetchParams</var> and <var>response</var>.</p></li>
 </ol>
 
 <p>To <dfn>finalize response</dfn> given a <a for=/>fetch params</a> <var>fetchParams</var> and a
 <a for=/>response</a> <var>response</var>, run these steps:
 
 <ol>
+ <li><p>If<var>fetchParams</var>'s <a for="fetch params">ready for clients</a> is false or
```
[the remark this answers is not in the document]
noamr: Fixed

yutakahirano:
Chrome is going to implement this once this and tests are ready.

fetch.bs
```diff
@@ -4225,6 +4258,25 @@ steps:
...
+  <a for=ReadableStream>readable</a>, then:</o>
+
+  <ol>
+   <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
+
+   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
+   <a for=ReadableStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
+
+   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
+   <a for=TransformStream>setting up</a> <var>transformStream</var> with
```
MattiasBuelens, line 4274: [Set up a `TransformStream`](https://streams.spec.whatwg.org/#transformstream-set-up) doesn't return anything, so you can't set *response*'s body to its result. (Looks like "HTTP fetch" also does this wrong.) You want to do this in two steps:

1. Set up *transformStream* with *identityTransformAlgorithm* and *finalize*.
2. Set *response*'s body to *transformStream*'s `readable`.

However, it looks like Streams doesn't yet expose helpers for other specs to access the `readable` and `writable` properties of a `TransformStream`. (We should probably fix this in Streams. 🤔) 

Luckily, we don't really need to access it directly. We need to pipe *response*'s body through *transformStream* anyway, so we can use the [pipe through](https://streams.spec.whatwg.org/#readablestream-pipe-through) helper, which does this for us:

1. Set up *transformStream* with *identityTransformAlgorithm* and *finalize*.
2. Set *response*'s body to the result of piping *response*'s body through *transformStream*.

fetch.bs
```diff
@@ -203,6 +203,8 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication).
  <dd>Null, a <a for=/>global object</a>, or a <a for=/>parallel queue</a>.
 
  <dt><dfn for="fetch params">cross-origin isolated capability</dfn> (default false)
+ <dt><dfn for="fetch params">ready for clients</dfn> (default false).
+ <dt><dfn for="fetch params">network read complete</dfn> (default false).
```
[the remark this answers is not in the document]
noamr: Fixed, using a TransformStream made these flags superfluous.

fetch.bs
```diff
@@ -4200,6 +4200,60 @@ steps:
 <a for=/>response</a> <var>response</var>, run these steps:
 
 <ol>
+ <li>
+  <p>If <var>response</var> is a <a>network error</a>, then:
+
+  <ol>
+   <li>
+    <p>Set <var>response</var>'s <a for=response>URL list</a> to « <var>fetchParams</var>'s
+    <a for="fetch params">request</a>'s <a for=request>URL list</a> ≫.
```
annevk, line 4209: There's two problems here. The list closing code point is incorrect and the contents of the list is another list? I thought we only wanted to reveal the initial URL?

fetch.bs
```diff
@@ -4200,6 +4200,60 @@ steps:
...
+  <p>If <var>hasReadableStream</var> is true, then:
+
+  <ol>
+   <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
+
+   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
+   <a for=TransformStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
+
+   <li><p><a for=TransformStream lt="setting up">Set up</a> <var>transformStream</var> with
+   <var>identityTransformAlgorithm</var> and <var>processResponseDone</var>.
```
annevk, line 4249: Nit: let's use the named arguments here for clarity. It wasn't immediately clear to me the second algorithm would only be run once. (https://infra.spec.whatwg.org/#algorithm-params explains this.)

fetch.bs
```diff
@@ -4210,12 +4264,23 @@ steps:
   then:
 
   <ol>
+   <li>
+    <p>Let <var>processEndOfBody</var> given <var>nullBytesOrFailure</var> be the following
+    steps:
+
+    <ol>
+     <li><p>Run <var>processResponseDone</var>.
```
annevk, line 4272: Why is this needed? Wouldn't this run as part of reading the body?

fetch.bs
```diff
@@ -4200,6 +4200,60 @@ steps:
...
+   </li>
+
+   <li><p>Set <var>response</var>'s <a for=response>timing info</a> to the result of
+   <a>creating an opaque timing info</a> for <var>fetchParams</var>'s
+   <a for="fetch params">timing info</a>.</p></li>
+  </ol>
+ </li>
+
+ <li><p>Let <var>hasReadableStream</var> be true if <var>response</var>'s <a for=response>body</a>
+ is not null and is <a for=ReadableStream>readable</a>; otherwise false.
```
annevk, line 4223: I think it should always be a stream if body is non-null.

fetch.bs
```diff
@@ -4225,26 +4290,18 @@ steps:
...
-<ol>
- <li><p>Set <var>fetchParams</var>'s <a for="fetch params">request</a>'s
- <a for=request>done flag</a>.
+ </li>
 
- <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response done</a> is not null,
- then <a>queue a fetch task</a> to run <var>fetchParams</var>'s
- <a for="fetch params">process response done</a> given <var>response</var>,
- with <var>fetchParams</var>'s <a for="fetch params">task destination</a>.
+  <li><p>If <var>hasReadableStream</var> is false, then run <var>processResponseDone</var>.</p></li>
```
annevk, line 4295: We already run this above when response's body is null (through _processBody_), so it's not clear to me this is needed.

fetch.bs
```diff
@@ -4200,6 +4200,60 @@ steps:
...
+   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
+   <a for=TransformStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
+
+   <li><p><a for=TransformStream lt="setting up">Set up</a> <var>transformStream</var> with
+   <var>identityTransformAlgorithm</var> and <var>processResponseDone</var>.
+
+   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
+   <a for=ReadableStream lt="piping through">piping</a> <var>response</var>'s
+   <a for=response>body</a> through <var>transformStream</var>.
+  </ol>
```
annevk, line 4254: Let's add a note after this `</ol>` explaining that we are doing this to get a notification after at end-of-stream.

annevk (commented):
Using streams in this way seems really nice and reduces the overall complexity a lot. Great work! I do think there are some logic errors here and there, but they seem fixable.

fetch.bs
noamr: OK

fetch.bs
noamr: Done

fetch.bs
noamr: I was trying to fiddle with the order of things, but will revise to make it cleaner.

noamr (commented):
Revised in the spirit of the comments.

fetch.bs
```diff
@@ -4194,6 +4194,64 @@ steps:
...
+  <p>Otherwise:</p>
+
+  <ol>
+   <li><p>Let <var>transformStream</var> be a new a {{TransformStream}}.
+
+   <li><p>Let <var>identityTransformAlgorithm</var> be an algorithm which, given <var>chunk</var>,
+   <a for=TransformStream lt=enqueue>enqueues</a> <var>chunk</var> in <var>transformStream</var>.
+
+   <li><p><a for=TransformStream lt="setting up">Set up</a> <var>transformStream</var> with
+   [=TransformStream/set up/transformalgorithm=] set to <var>identityTransformAlgorithm</var> and
```
annevk, line 4243: These parameter names also need to be wrapped in `<i>`. And for consistency let's continue to use `<a>` here.

fetch.bs
```diff
@@ -4194,6 +4194,64 @@ steps:
...
+   [=TransformStream/set up/flushalgorithm=] set to <var>processResponseDone</var>.
+
+   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
+   <a for=ReadableStream lt="piping through">piping</a> <var>response</var>'s
+   <a for=response>body</a> through <var>transformStream</var>.
+  </ol>
+
+  <p class=note>This {{TransformStream}} is needed for the purpose of receiving a notification when
+  the stream reaches its end, and is otherwise an [=identity transform stream=].
```
annevk, line 4252: Nit: can we use `<a>` here as well? I'd rather not introduce new syntax here in only a few new places until we decided more generally to go for that.

fetch.bs
```diff
@@ -4219,26 +4277,16 @@ steps:
...
-
-<ol>
- <li><p>Set <var>fetchParams</var>'s <a for="fetch params">request</a>'s
- <a for=request>done flag</a>.
-
- <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response done</a> is not null,
- then <a>queue a fetch task</a> to run <var>fetchParams</var>'s
- <a for="fetch params">process response done</a> given <var>response</var>,
- with <var>fetchParams</var>'s <a for="fetch params">task destination</a>.
+ </li>
```
annevk, line 4280: Nit: no need for this closing tag.

fetch.bs
```diff
@@ -4194,6 +4194,64 @@ steps:
...
+   [=TransformStream/set up/flushalgorithm=] set to <var>processResponseDone</var>.
+
+   <li><p>Set <var>response</var>'s <a for=response>body</a> to the result of
+   <a for=ReadableStream lt="piping through">piping</a> <var>response</var>'s
+   <a for=response>body</a> through <var>transformStream</var>.
+  </ol>
+
+  <p class=note>This {{TransformStream}} is needed for the purpose of receiving a notification when
+  the stream reaches its end, and is otherwise an [=identity transform stream=].
+ </li>
```
annevk, line 4253: Nit: no need for this closing tag.

fetch.bs
```diff
@@ -4194,6 +4194,64 @@ steps:
...
+
+   <li><p>If <var>fetchParams</var>'s <a for="fetch params">process response done</a> is not null,
+   then <a>queue a fetch task</a> to run <var>fetchParams</var>'s
+   <a for="fetch params">process response done</a> given <var>response</var> with
+   <var>fetchParams</var>'s <a for="fetch params">task destination</a>.
+  </ol>
+ </li>
+
+ <li><p>If <var>response</var>'s <a for=response>body</a> is null, then run
+ <var>processResponseDone</var>.
```
annevk, line 4231: This makes it run before "process response", which seems bad?

noamr: Sure, will re-order

fetch.bs

fetch.bs

fetch.bs

annevk (approved):

Issue #1265, opened 2021-07-04
Range header 'helper' produces incorrect header value


whymarrh:
Snapshot: [d20e8c9](https://fetch.spec.whatwg.org/commit-snapshots/d20e8c932a639d04fb8251eb6d647fcd54ef47fd/#concept-request-add-range-header)

The spec currently outlines the process for adding a `Range` header to a request:

> To add a range header to a request _request_, with an integer _first_, and an optional integer _last_, run these steps:​
> 
> 1. Let _rangeValue_ be \`<code>bytes </code>\`\.
> 2. Serialize and isomorphic encode first, and append the result to _rangeValue_\.
> 3. Append 0x2D \(\-\) to rangeValue\.
> 4. If _last_ is given, then serialize and isomorphic encode it, and append the result to _rangeValue_\.
> 5. Append \`<code>Range</code>\`/_rangeValue_ to _request_’s header list\.

As an example, if _first_ and _last_ had values of 3 and 5, I think this would result in a `Range` header value of `bytes 3-5` when it should be `bytes=3-5`, with 0x3D (=) as the separator.

Per RFC 7233 §2.1:<sup>[\[1\]][1]</sup>

> The "bytes" range unit is defined for expressing subranges of the data's octet sequence.
>
>     bytes-unit       = "bytes"
>
> […]
>
>     byte-ranges-specifier = bytes-unit "=" byte-range-set
>     byte-range-set  = 1#( byte-range-spec / suffix-byte-range-spec )

> Examples of byte-ranges-specifier values:
>
> - The first 500 bytes (byte offsets 0-499, inclusive): `bytes=0-499`
> - The second 500 bytes (byte offsets 500-999, inclusive): `bytes=500-999`

It should be a simple enough fix, unless I'm missing something:

```diff
- <li><p>Let <var>rangeValue</var> be `<code>bytes </code>`.
+ <li><p>Let <var>rangeValue</var> be `<code>bytes=</code>`.
```

I'm willing to open a PR for this, if that would be useful.

  [1]:https://httpwg.org/specs/rfc7233.html#byte.ranges

annevk:
If you can also sign https://participate.whatwg.org/agreement, that would be great!

cc @jakearchibald

annevk:
This was fixed as part of #1312 in 6f37b51ecf23f974d08578ba07088e3faee634ac, but we forgot to note it in the final commit message.

PR #1312, opened 2021-09-28
Allow range header without preflight

### LINKED ISSUE #1310: Add `range` to CORS-safelisted request-headers
We allow no-cors requests to have `Range` headers as long as it's set by an internal API, and the request is unmodified.

It feels like CORS requests should be able to use `Range` in particular ways without triggering a preflight. If the format matches what https://fetch.spec.whatwg.org/#concept-request-add-range-header can create, that's ok right?

jakearchibald:
Fixes #1310.

   * Chromium
   * Mozilla
   * https://github.com/web-platform-tests/wpt/pull/31058
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1255711
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1733981
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=231174
- [x] Security review

***
<a href="https://whatpr.org/fetch/1312.html" title="Last updated on Oct 4, 2021, 4:11 PM UTC (ed0a9ca)">Preview</a> | <a href="https://whatpr.org/fetch/1312/b2f04e2...ed0a9ca.html" title="Last updated on Oct 4, 2021, 4:11 PM UTC (ed0a9ca)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -825,6 +825,9 @@ fetch("https://victim.example/naïve-endpoint", {
 </code></pre>
     </div>
 
+   <dt>`<code>Range</code>`
+   <dd><p>If <var>value</var> is not a <a>simple range header value</a>, then return false.
+
    <dt>Otherwise
    <dd><p>Return false.
   </dl>
@@ -1037,6 +1040,75 @@ run these steps:
 
  <li><p>Return <var>values</var>.
 </ol>
+
+<p>To determine if a <a>byte sequence</a> <var>value</var> is a
+<dfn>simple range header value</dfn>, perform the following steps. They return a <a>boolean</a>.
+
+<ol>
+ <li><p>If <var>value</var> does not <a for="byte sequence" lt="starts with">start with</a>
+ `<code>bytes </code>`, then return false.
+
+ <li><p>If the 6th <a>byte</a> of <var>value</var> is not in the range 0x30 (0) to 0x39 (9),
+ inclusive, then return false.
+
+ <li><p>Let <var>rangeStartBytes</var> be the 6th <a>byte</a> of <var>value</var>.
+
+ <li><p>Let <var>rangeEndBytes</var> be an empty <a>byte sequence</a>.
+
+ <li><p>Let <var>state</var> be "<code>rangeStart</code>".
+
+ <li><p>Let <var>i</var> be 7.
+
+ <li>
+  <p>While <var>i</var> is less than <var>value</var>'s <a for="byte sequence">length</a>:
+
+  <ol>
+   <li>
+    <p>Switch on <var>state</var>:
+
+    <dl class="switch">
+     <dt>"<code>rangeStart</code>"
+     <dd>
+      <ol>
+       <li><p>If the <var>i</var>th <a>byte</a> of <var>value</var> is 0x2D (-), then set
+       <var>state</var> to "<code>rangeEnd</code>".
+
+       <li>
+        <p>Otherwise:
+
+        <ol>
+         <li><p>If the <var>i</var>th <a>byte</a> of <var>value</var> is not in the range 0x30 (0)
+         to 0x39 (9), inclusive, then return false.
+
+         <li><p>Append the <var>i</var>th <a>byte</a> of <var>value</var> to
+         <var>rangeStartBytes</var>.
+        </ol>
+      </ol>
+
+     <dt>"<code>rangeEnd</code>"
+     <dd>
+      <ol>
+       <li><p>If the <var>i</var>th <a>byte</a> of <var>value</var> is not in the range 0x30 (0) to
+       0x39 (9), inclusive, then return false.
+
+       <li><p>Append the <var>i</var>th <a>byte</a> of <var>value</var> to <var>rangeEndBytes</var>.
+      </ol>
+    </dl>
+
+    <li><p>Increment <var>i</var>.
+  </ol>
+
+ <li>If <var>rangeEndBytes</var>'s <a for="byte sequence">length</a> is 0, then return true.
+
+ <li><p>If <var>rangeStartBytes</var>, parsed as an integer, is greater than
+ <var>rangeEndBytes</var>, parsed as an integer, then return false.
+
+ <li><p>Return true.
+</ol>
+
+<p class="note">A <a>simple range header value</a> is a subset of allowed range header values, but
+it's the most common form used by user agents when requesting media or resuming downloads. This
+format of range header value can be set using <a>add a range header</a>.
 
 <hr>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -825,7 +825,7 @@ fetch("https://victim.example/naïve-endpoint", {
 </code></pre>
     </div>
 
-   <dt>`<code>Range</code>`
+   <dt>`<code>range</code>`
    <dd><p>If <var>value</var> is not a <a>simple range header value</a>, then return false.
 
    <dt>Otherwise
--- a/fetch.bs
+++ b/fetch.bs
@@ -1047,7 +1047,8 @@ run these steps:
 <ol>
  <li><p>Let <var>data</var> be the <a>isomorphic decoding</a> of <var>value</var>.
 
- <li><p>If <var>data</var> does not <a for=string>start with</a> "<code>bytes </code>", then return false.
+ <li><p>If <var>data</var> does not <a for=string>start with</a> "<code>bytes=</code>", then return
+ false.
 
  <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially pointing at the 6th <a>code point</a> of <var>data</var>.
 
@@ -1886,7 +1887,7 @@ is to return the result of <a>serializing a request origin</a> with <var>request
  <li><p>Assert: <var>last</var> is not given, or <var>first</var> is less than or equal to
  <var>last</var>.
 
- <li><p>Let <var>rangeValue</var> be `<code>bytes </code>`.
+ <li><p>Let <var>rangeValue</var> be `<code>bytes=</code>`.
 
  <li><p><a lt="serialize an integer">Serialize</a> and <a>isomorphic encode</a> <var>first</var>,
  and append the result to <var>rangeValue</var>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -1054,8 +1054,7 @@ run these steps:
  pointing at the 6th <a>code point</a> of <var>data</var>.
 
  <li><p>Let <var>rangeStart</var> be the result of <a>collecting a sequence of code points</a> that
- are in the range U+0030 (0) to U+0039 (9), inclusive, from <var>data</var> given
- <var>position</var>.
+ are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
 
  <li><p>If the <a>code point</a> at <var>position</var> within <var>data</var> is not U+002D (-),
  then return false.
--- a/fetch.bs
+++ b/fetch.bs
@@ -1064,7 +1064,7 @@ run these steps:
  <li><p>Let <var>rangeEnd</var> be the result of <a>collecting a sequence of code points</a> that
  are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
 
- <li><p>If <var>position</var> is not past the end of <var>data</var>, return false.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -770,6 +770,9 @@ production as
 <ol>
  <li><p>Let <var>value</var> be <var>header</var>'s <a for=header>value</a>.
 
+ <li><p>If <var>value</var>'s <a for="byte sequence">length</a> is greater than 128, then return
+ false.
```
jakearchibald, line 774: Feels like this simple condition should go before the more complex conditions. I don't feel strongly about it though.

fetch.bs
```diff
@@ -1038,6 +1041,75 @@ run these steps:
...
+      </ol>
+    </dl>
+
+    <li><p>Increment <var>i</var>.
+  </ol>
+
+ <li>If <var>rangeEndBytes</var>'s <a for="byte sequence">length</a> is 0, then return true.
+
+ <li><p>If <var>rangeStartBytes</var>, parsed as an integer, is greater than
+ <var>rangeEndBytes</var>, parsed as an integer, then return false.
```
jakearchibald, line 1104: "parsed as an integer" feels hand-wavey, but I wasn't sure of the 'right' way.

fetch.bs
```diff
@@ -1038,6 +1041,77 @@ run these steps:
  <li><p>Return <var>values</var>.
 </ol>
 
+<p>To determine if a <a>byte sequence</a> <var>value</var> is a
+<dfn>simple range header value</dfn>, perform the following steps. They return a <a>boolean</a>.
+
+<ol>
```
yutakahirano, line 1047: How about using regexp?

annevk:
Can we consolidate with https://wicg.github.io/background-fetch/#extract-content-range-values-algorithm perhaps? (I realize there are delimiter differences, but perhaps they can be abstracted somehow. See also https://github.com/WICG/background-fetch/issues/157.)

The other thing that might help is to use isomorphic decode, so you can operate on a string. That removes the need for state and allows for using Infra's "collect a sequence" algorithms.

fetch.bs
```diff
@@ -1038,6 +1041,77 @@ run these steps:
...
+ <li><p>If <var>value</var> does not <a for="byte sequence">start with</a> `<code>bytes </code>`,
+ then return false.
+
+ <li><p>If the 6th <a>byte</a> of <var>value</var> is not in the range 0x30 (0) to 0x39 (9),
+ inclusive, then return false.
+
+ <li><p>Let <var>rangeStartBytes</var> be the 6th <a>byte</a> of <var>value</var>.
```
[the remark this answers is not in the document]
jakearchibald: https://infra.spec.whatwg.org/#byte-sequence-prefix was my inspiration here, which uses _i_th, and starts at 0.

fetch.bs
jakearchibald: I haven't seen regexp used in a spec before (and it raises questions about the flavour of regexp). I'm going to switch to "collect a sequence". Hopefully that'll be more readable.

jakearchibald:
@annevk I'm not sure how useful it is to consolidate with `Content-Range`, since the format is pretty different.

- `^bytes=\d+-\d+/(\*|\d+)$` - simple `Content-Range`
- `^bytes \d+-\d*$` - simple `Range`

I think it'd be confusing to try and combine the two.

I'll look at defining this using "collect a sequence", or would you rather I used the [ABNF style from background-fetch](https://wicg.github.io/background-fetch/#single-byte-content-range)?

annevk:
Fair about combining. They are a bit more different than I thought. Implementers (at least Matt Menke) seem to prefer algorithms over ABNF and that's also the style I've used for the other header parsers thus far so I guess we should stick with that. Also, ABNF often leads to implementations that approximate the parser requirements in my experience.

jakearchibald:
@annevk cool, I've rewritten it using "collecting a sequence of code points", it's much easier to read now.

fetch.bs
```diff
@@ -822,13 +825,13 @@ fetch("https://victim.example/naïve-endpoint", {
 </code></pre>
     </div>
 
+   <dt>`<code>Range</code>`
```
annevk, line 828:
```suggestion
   <dt>`<code>range</code>`
```
(it's lowercased above)

fetch.bs
```diff
@@ -1038,6 +1041,39 @@ run these steps:
  <li><p>Return <var>values</var>.
 </ol>
 
+<p>To determine if a <a>byte sequence</a> <var>value</var> is a
+<dfn>simple range header value</dfn>, perform the following steps. They return a <a>boolean</a>.
+
+<ol>
+ <li><p>Let <var>data</var> be the <a>isomorphic decoding</a> of <var>value</var>.
+
+ <li><p>If <var>data</var> does not <a for=string>start with</a> "<code>bytes </code>", then return false.
```
annevk, line 1050: While I agree with you on wrapping, until we change it I do think we should maintain consistency.

fetch.bs
```diff
@@ -1038,6 +1041,39 @@ run these steps:
...
+
+ <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially pointing at the 6th <a>code point</a> of <var>data</var>.
+
+ <li><p>Let <var>rangeStart</var> be the result of <a>collecting a sequence of code points</a> that are in the range U+0030 (0) to U+0039 (9), inclusive, from <var>data</var> given <var>position</var>.
+
+ <li><p>If the <a>code point</a> at <var>position</var> within <var>data</var> is not U+002D (-), then return false.
+
+ <li><p>Advance <var>position</var> by 1.
+
+ <li><p>Let <var>rangeEnd</var> be the result of <a>collecting a sequence of code points</a> that are in the range U+0030 (0) to U+0039 (9), inclusive, from <var>data</var> given <var>position</var>.
```
annevk, line 1060: Either you need to collect all code points here or you need to do another check at the end. At the moment `bytes 0-0garbage` is valid I think.

fetch.bs
```diff
@@ -1038,6 +1041,39 @@ run these steps:
...
+  <p class="note">The range end can be omitted, e.g., `<code>bytes 0-</code>` is valid.
+
+ <li><p>If <var>rangeStart</var>, parsed as an integer, is greater than <var>rangeEnd</var>, parsed
+ as an integer, then return false.
+
+ <li><p>Return true.
+</ol>
+
+<p class="note">A <a>simple range header value</a> is a subset of allowed range header values, but
+it's the most common form used by user agents when requesting media or resuming downloads. This
```
annevk, line 1074: it is*

fetch.bs
```diff
@@ -1038,6 +1041,39 @@ run these steps:
...
+ <li><p>Advance <var>position</var> by 1.
+
+ <li><p>Let <var>rangeEnd</var> be the result of <a>collecting a sequence of code points</a> that are in the range U+0030 (0) to U+0039 (9), inclusive, from <var>data</var> given <var>position</var>.
+
+ <li>
+  <p>If <var>rangeEnd</var>'s <a for=string>length</a> is 0, then return true.
+
+  <p class="note">The range end can be omitted, e.g., `<code>bytes 0-</code>` is valid.
+
+ <li><p>If <var>rangeStart</var>, parsed as an integer, is greater than <var>rangeEnd</var>, parsed
```
annevk, line 1067: interpreted as decimal number is what I used in Content-Length.

annevk (commented):
Thanks, this looks very readable indeed.

fetch.bs
jakearchibald: Yeah, just forgot 😄

jakearchibald:
Ugh, I just realised it's `bytes=` rather than `bytes `. It's `Content-Range` that uses `bytes `. I've fixed up the range-setting algorithm too, but I'm happy to make that a different PR.

jakearchibald:
This PR aims to allow, without requiring a preflight, CORS requests with a `Range` header similar to those the browser would send when downloading media or a resumable download.

When requesting media, browsers will typically request ranges like `bytes=123456-789012`, to target things like metadata in the resource, and `bytes=0-` for the initial media download.

When resuming the download of files, browsers will typically just use the initial range, like `bytes=123-`.

However, there are a few novel requests that this PR enables:

- Ranges outside the `Content-Length` of the resource.
- Tiny ranges, eg `bytes=123-123`. I don't think browsers do this.
- `Range` along with a `POST` request. Downloads can be initiated via a `POST` request, but browsers will not send a `Range` header in these cases.
- `Range` and `Origin` headers on a `GET` request. We already include `Origin` with `POST` requests, but not `GET` unless it's a CORS request.
- Combining `Range` with unusual values in other [safe-listed headers](https://fetch.spec.whatwg.org/#cors-safelisted-request-header), such as `accept: whatever`.

Only "ranges outside the `Content-Length` of the resource" makes me raise an eyebrow, so I'd like a second opinion on that.

fetch.bs
```diff
@@ -1038,6 +1041,48 @@ run these steps:
...
+
+ <li><p>If <var>data</var> does not <a for=string>start with</a> "<code>bytes=</code>", then return
+ false.
+
+ <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially
+ pointing at the 6th <a>code point</a> of <var>data</var>.
+
+ <li><p>Let <var>rangeStart</var> be the result of <a>collecting a sequence of code points</a> that
+ are in the range U+0030 (0) to U+0039 (9), inclusive, from <var>data</var> given
+ <var>position</var>.
```
annevk, line 1058:
```suggestion
 are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
```

fetch.bs
```diff
@@ -1038,6 +1041,48 @@ run these steps:
...
+
+ <li><p>If the <a>code point</a> at <var>position</var> within <var>data</var> is not U+002D (-),
+ then return false.
+
+ <li><p>Advance <var>position</var> by 1.
+
+ <li><p>Let <var>rangeEnd</var> be the result of <a>collecting a sequence of code points</a> that
+ are in the range U+0030 (0) to U+0039 (9), inclusive, from <var>data</var> given
+ <var>position</var>.
```
annevk, line 1067:
```suggestion
 are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
```

fetch.bs
```diff
@@ -1038,6 +1041,48 @@ run these steps:
...
+
+ <li><p>If <var>position</var> is not past the end of <var>data</var>, return false.
```
annevk, line 1069:
```suggestion
 <li><p>If <var>position</var> is not past the end of <var>data</var>, then return false.
```

annevk (commented):
Thanks for the analysis! Modulo some final nits this looks good to me now. I wonder if @mikewest, @yutakahirano, and @youennf can help out with additional security review.

(This also ends up fixing #1265 now. It seems okay to include the fix for that here, as long as we indicate it in the final commit message.)

jakearchibald:
> (This also ends up fixing #1265 now. It seems okay to include the fix for that here, as long as we indicate it in the final commit message.)

Ah, sorry I missed that original issue.

yutakahirano (approved):
LGTM

jakearchibald:
@rayankans over to you for the tests

rayankans:
I created a PR for the tests here: https://github.com/web-platform-tests/wpt/pull/31058

annevk:
I'm comfortable with merging this once we have implementation bugs.

jakearchibald:
Implementation bugs added.

Commit message: Allow particular Range header values without a preflight.

The allowed format aligns with the values the browser uses when requesting media and resuming downloads.
Spec discussion: https://github.com/whatwg/fetch/issues/1310
Spec PR: https://github.com/whatwg/fetch/pull/1312
Tests PR: https://github.com/web-platform-tests/wpt/pull/31058

annevk:
Thanks all!

anson0370:
I tried this code in both Chrome and Safari, but it doesn't seem to work:
```js
fetch('https://pscrb.fm/rss/p/media.transistor.fm/db73bbc2/d569819b.mp3', {
  mode: 'no-cors',
  headers: {
    Range: 'bytes=0-10240'
  }
})
```

When I check the Network tab in DevTools, the request headers don't show the Range header. I'm not sure if I'm misunderstanding something.

PR #1314, opened 2021-09-29
Using fetch in other standards

### LINKED ISSUE #1306: What's the difference/intended usage of the response* algorithms in fetch?
The fetch algo takes five algorithms as input, of which three deal with the response: `processResponse`, `processResponseEndOfBody`, and `processResponseDone`. No explanation is given as to when each of these run or what each is for.

The two examples I've found of correct usage of the fetch algo ([fetch() itself](https://fetch.spec.whatwg.org/#dom-global-fetch) and [XHR.send()](https://xhr.spec.whatwg.org/#the-send()-method)) each use processResponse, and fetch also uses processResponseDone. 

From reading the fetch algorithm in detail, I can infer that processResponse is called early, as soon as the request finishes, so you can do things like handle progress events or immediately handle errors. It's not clear to me if this is an idiom to be followed, or if it's okay to do all the processing in one of the later-occurring algos.

I don't know what the difference is between processResponseEndOfBody and processResponseDone, other than that, again from reading the fetch algo in detail, they're called in that order, with some degree of processing happening between them that I'm not sure of the details of. So I'm not sure what sort of tasks are appropriate to do in one vs the other.

Could we get some explanation of these inputs in <https://fetch.spec.whatwg.org/#fetch-params-process-request-body> maybe?

annevk:
Closes #1306.

@tabatkins @ricea @yutakahirano care to review?

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7828,6 +7828,62 @@ resource — for non-<a>CORS requests</a> as well as <a>CORS
 requests</a> — and do not use `<code>Vary</code>`.
 
 
+<h2 id=fetch-elsewhere class=no-num>Fetch elsewhere</h2>
+
+<p>In its essence <a for=/>fetching</a> is an exchange of a <a for=/>request</a> for a
+<a for=/>response</a>. In reality it is rather complex mechanism for standards to adopt and use
+correctly. This section aims to give some advice.
+
+<p class=warning>Always ask domain experts for review.
+
+<p class=XXX>This is a work in progress.
+
+
+<h3 id=fetch-elsewhere-fetch class=no-num>Invoking fetch</h3>
+
+<p>Aside from a <a for=/>request</a> the <a for=/>fetch</a> operation takes several optional
+arguments. For those arguments that take an algorithm: the algorithm will be called from a task (or
+in a <a for=/>parallel queue</a> if <a for=fetch><i>useParallelQueue</i></a> is true).
+
+<dl>
+ <dt><a for=fetch><i>processRequestBody</i></a>
+ <dd><p>Takes an algorithm that will be passed the number of bytes that have been transmitted from
+ the <a for=/>request</a>'s <a for=request>body</a>. Most standards will not need this.
+
+ <dt><a for=fetch><i>processRequestEndOfBody</i></a>
+ <dd><p>Takes an algorithm that will be passed nothing. Indicates <a for=/>request</a>'s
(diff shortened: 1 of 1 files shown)


tabatkins:
Oh wow, I haven't run into this bug before. I've got it reproduced locally, fix incoming asap.

tabatkins:
Okay I'm still figuring out how the error is happening, but it's being caused by a markup error - you have a `</p>` instead of `</a>` on lines 7865 and 7877. Fixing those makes the error stop happening.

I'll continue working on the error itself, since crash errors aren't acceptable.

fetch.bs
```diff
@@ -7828,6 +7828,62 @@ resource — for non-<a>CORS requests</a> as well as <a>CORS
...
+<h3 id=fetch-elsewhere-fetch class=no-num>Invoking fetch</h3>
+
+<p>Aside from a <a for=/>request</a> the <a for=/>fetch</a> operation takes several optional
+arguments. For those arguments that take an algorithm: the algorithm will be called from a task (or
+in a <a for=/>parallel queue</a> if <a for=fetch><i>useParallelQueue</i></a> is true).
+
+<dl>
+ <dt><a for=fetch><i>processRequestBody</i></a>
+ <dd><p>Takes an algorithm that will be passed the number of bytes that have been transmitted from
+ the <a for=/>request</a>'s <a for=request>body</a>. Most standards will not need this.
```
tabatkins, line 7851: Is this algo potentially called repeatedly during sending of the rquest, or just once?

fetch.bs
```diff
@@ -7828,6 +7828,62 @@ resource — for non-<a>CORS requests</a> as well as <a>CORS
...
+
+ <dt><a for=fetch><i>processRequestEndOfBody</i></a>
+ <dd><p>Takes an algorithm that will be passed nothing. Indicates <a for=/>request</a>'s
+ <a for=request>body</a> has been transmitted. Most standards will not need this.
+
+ <dt><a for=fetch><i>processResponse</i></a>
+ <dd><p>Takes an algorithm that will be passed a <a for=/>response</a>. Indicates
+ <a for=/>response</a>'s <a for=response>header list</a> has been transmitted and initialized. This
```
tabatkins, line 7859: s/transmitted/received/, since this is from the POV of the browser?

tabatkins (approved):
A good starting point! We'll eventually want usage examples, but this description of inputs and rough meaning/usage is already enormously helpful.

PR #1325, opened 2021-10-07
Acknowledge interaction with draft-ietf-dnsop-svcb-https scheme redirect

### LINKED ISSUE #1315: Acknowledge interaction with draft-ietf-dnsop-svcb-https scheme redirect
Chrome is preparing to launch the feature from draft-ietf-dnsop-svcb-https-07 Section 8.5 that http:// requests get redirected to https:// on discovery of a matching HTTPS RR in DNS.  Firefox seems to have already recently lauched this feature (see https://www.mozilla.org/en-US/firefox/92.0/releasenotes/).

I think Fetch needs to be updated to acknowledge this behavior, similar to the references to HSTS.  Particularly, I think Section 4.9, point 9, (the text currently describing the HSTS redirection) needs a simple sentence or two of "or" explanation citing draft-ietf-dnsop-svcb-https and explaining when that causes the scheme to be changed to "https".

I can take a stab at a PR for this next week if there's general agreement that this is the right way to go to account for this new behavior.

ericorth:
Closes #1315

Adding some vague mostly-implementation-defined language to acknowledge that scheme upgrade can occur, similarly to HSTS, on discovery of a DNS HTTPS RR.  Also added a (similarly vague and implementation-defined) note that origin resolution might query DNS for HTTPS and return more than addresses.

***
<a href="https://whatpr.org/fetch/1325.html" title="Last updated on Oct 11, 2021, 6:12 PM UTC (38a35c3)">Preview</a> | <a href="https://whatpr.org/fetch/1325/6f37b51...38a35c3.html" title="Last updated on Oct 11, 2021, 6:12 PM UTC (38a35c3)">Diff</a>

***
<a href="https://whatpr.org/fetch/1325.html" title="Last updated on Apr 7, 2022, 6:23 PM UTC (9e2c56d)">Preview</a> | <a href="https://whatpr.org/fetch/1325/8ebf2fd...9e2c56d.html" title="Last updated on Apr 7, 2022, 6:23 PM UTC (9e2c56d)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2305,36 +2305,6 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
 
 <h3 id=resolving-domains>Resolving domains</h3>
 
-<p tracking-vector>To <dfn>resolve a domain</dfn>, given a <a for=/>network partition key</a>
-<var>key</var> and a <a for=/>domain</a> <var>domain</var>:
-
-<ol>
- <li><p>If <var>domain</var> is a <a for=/>host</a> whose <a for=host>public suffix</a> is
- "<code>localhost</code>", then return « <code>::1</code>, <code>127.0.0.1</code> ».
-
- <li><p>Perform an <a>implementation-defined</a> operation to turn <var>domain</var> into a
- <a for=/>set</a> of one or more <a for=/>IP addresses</a>. If this operation succeeds, return the
- <a for=/>set</a> of <a for=/>IP addresses</a>.
-
- <li><p>Return failure.
-</ol>
-
-<p>The results of <a>resolve a domain</a> may be cached. If they are cached, <var>key</var> should
-be used as part of the cache key.
-
-<div class=note>
- <p>Typically this operation would involve DNS and as such caching can happen on DNS servers without
- <var>key</var> being taken into account. Depending on the implementation it might also not be
- possible to take <var>key</var> into account locally. [[RFC1035]]
-
- <p>The order of the <a for=/>IP addresses</a> <a>resolve a domain</a> can return return can differ
- between invocations.
-
- <p>The particulars (apart from the cache key) are not tied down as they are not pertinent to the
- system the Fetch Standard establishes. Other documents ought not to build on this primitive without
- having a considered discussion with the Fetch Standard community first.
-</div>
-
 <p>To <dfn>resolve an origin</dfn>, given a <a for=/>network partition key</a> <var>key</var> and an
 <a for=/>origin</a> <var>origin</var>:
 <!-- Should we assert the scheme here to be an HTTP(S) scheme or a WebRTC scheme? -->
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -3873,11 +3884,16 @@ steps:
    <li>Matching <var>request</var>'s <a for=request>current URL</a>'s <a for=url>host</a> per
    <a href=https://datatracker.ietf.org/doc/html/rfc6797#section-8.2>Known HSTS Host Domain Name Matching</a>
    results in either a superdomain match with an asserted <code>includeSubDomains</code> directive
-   or a congruent match (with or without an asserted <code>includeSubDomains</code> directive).
-   [[!HSTS]]
+   or a congruent match (with or without an asserted <code>includeSubDomains</code> directive) or
+   DNS resolution for the request finds a matching HTTPS RR per
+   <a href=https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https#section-8.5>HTTP Strict Transport Security</a>.
+   [[!HSTS]][[!SVCB]]
```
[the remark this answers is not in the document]
ericorth: And finally coming back to this after a lot of completely unrelated delays on my time.  Sorry, all, for leaving this hanging so long.

I added the extra bit of text I proposed long ago to further acknowledge that the DNS queries may not occur until connection time and to acknowledge that there's a lot of wishy-washing "implementation-defined" going on here.

Is this looking like an overall reasonable PR to people, or is there a strong desire for HTTPS RRs to be more completely specified within Fetch? I personally don't think there's much value in specifying it more completely, and I don't see how it could be done without either way over-specifying it or significantly reworking huge portions of Fetch way beyond what I feel I am personally comfortable or capable of writing up myself.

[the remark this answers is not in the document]
annevk: Could you please rebase against main and force push? The PR seems to contain many unrelated commits at the moment.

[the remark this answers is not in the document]
ericorth: Sorry about that.  Tried doing my resyncing a different way from how I normally do it, and I clearly messed it up.  :)

Should now be fixed up.

fetch.bs
```diff
@@ -2341,8 +2347,15 @@ be used as part of the cache key.
...
-<p class=note>The same caveat applies. Do not build on this without having a considered discussion
-with the Fetch Standard community first.
+<div class=note>
+ <p>The same caveat applies. Do not build on this without having a considered discussion
+ with the Fetch Standard community first.
+
+ <p>It is <a>implementation-defined</a> whether other DNS queries might be performed to get
+ connection information beyond just <a for=/>IP addresses</a>. For example, if <var>origin</var>'s
+ <a for=origin>scheme</a> is "http" or "https", the implementation might perform a DNS query for
+ HTTPS RRs. [[SVCB]]
```
annevk, line 2357: This shouldn't be a note. It seems this should be part of the step that turns a domain into an IP address, no? Although I suppose this does highlight that the "resolve a domain" algorithm maybe ought to take an origin and not a domain.

annevk (commented):
This looks somewhat reasonable, modulo reshuffling "resolve a domain". Should we also have a follow-up on reconsidering HSTS integration?

fetch.bs
ericorth: Moved outside the note into the procedure.  And good idea on "resolve a domain" taking the origin instead of domain.  Since there were already two sections here to take origin or domain, and all the origin one did was call the other with the origin's host, and nothing else seems to directly depend on the domain one, I did some trivial merging to give a single "resolve an origin" procedure.

fetch.bs
```diff
@@ -2299,51 +2305,49 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
...
- <a for=/>set</a> of one or more <a for=/>IP addresses</a>. If this operation succeeds, return the
- <a for=/>set</a> of <a for=/>IP addresses</a>.
+ <li>
+  <p>Perform an <a>implementation-defined</a> operation to turn <var>origin</var> into a
+  <a for=/>set</a> of one or more <a for=/>IP addresses</a>.
+
+  <p>It is also <a>implementation-defined</a> whether other operations might be performed to get
+  connection information beyond just <a for=/>IP addresses</a>. For example, if <var>origin</var>'s
+  <a for=origin>scheme</a> is "http" or "https", the implementation might perform a DNS query for
+  HTTPS RRs. [[SVCB]]
```
annevk, line 2326:
```suggestion
  <a for=origin>scheme</a> is an <a>HTTP(S) scheme</a>, the implementation might perform a DNS query
  for HTTPS RRs. [[SVCB]]
```

annevk (commented):
@mikewest and @davidben could you have another look as well? I think this is probably reasonable for now. Though perhaps we should have a follow-up for reworking HSTS.

fetch.bs
```diff
@@ -3989,11 +3993,20 @@ steps:
    <li>Matching <var>request</var>'s <a for=request>current URL</a>'s <a for=url>host</a> per
    <a href=https://datatracker.ietf.org/doc/html/rfc6797#section-8.2>Known HSTS Host Domain Name Matching</a>
    results in either a superdomain match with an asserted <code>includeSubDomains</code> directive
-   or a congruent match (with or without an asserted <code>includeSubDomains</code> directive).
-   [[!HSTS]]
+   or a congruent match (with or without an asserted <code>includeSubDomains</code> directive); or
+   DNS resolution for the request finds a matching HTTPS RR per
+   <a href=https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https#section-8.5>HTTP Strict Transport Security</a>.
```
davidben, line 3998: The link "HTTP Strict Transport Security" should be to section 9.5, but it's also a little confusing because "HTTP Strict Transport Security" is the name of the other feature. Perhaps...

```suggestion
   or a congruent match (with or without an asserted <code>includeSubDomains</code> directive) [[!HSTS]]; or
   DNS resolution for the request finds a matching HTTPS RR per
   <a href=https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https#section-9.5>section 9.5</a>
   of [[!SVCB]].
```

davidben (approved):
Agreed this is reasonable for now. Reworking HSTS to be a redirect would be good, but we can do that separately. The boundary between Fetch and various RFCs is also a little odd (having to respecify stuff, etc.), but I think that's a broader problem to work out.

ericorth:
Accepted all suggested changes.  Any remaining concerns?

fetch.bs
```diff
@@ -2299,51 +2305,50 @@ unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>termi
...
 be used as part of the cache key.
 
 <div class=note>
  <p>Typically this operation would involve DNS and as such caching can happen on DNS servers without
  <var>key</var> being taken into account. Depending on the implementation it might also not be
  possible to take <var>key</var> into account locally. [[RFC1035]]
 
- <p>The order of the <a for=/>IP addresses</a> <a>resolve a domain</a> can return return can differ
+ <p>The order of the <a for=/>IP addresses</a> <a>resolve an origin</a> can return return can differ
  between invocations.
```
mikewest, line 2345: Nit: I stumbled over this while reading. Since you're here, would you mind making a small change along the lines of the following? (The double "return" should be fixed regardless.)

```suggestion
 <p>The order of the <a for=/>IP addresses</a> that the <a>resolve an origin</a> algorithm can return 
can differ between invocations.
```

mikewest (approved):
Apologies for the delay. I've read through the diff, and I'm happy with where this ended up. I have only one dumb nit. Otherwise LGTM.

annevk:
I think we are good, modulo a commit message. Proposal:
```
Define interaction with DNS HTTPS resource record

This also makes everything go through "resolve an origin" and removes "resolve a domain" as we need the scheme sometimes.

Closes #1315.
```
If that looks good I'll land it. Thanks for pushing this forward!

ericorth:
> 



> I think we are good, modulo a commit message. Proposal:
> 
> ```
> Define interaction with DNS HTTPS resource record
> 
> This also makes everything go through "resolve an origin" and removes "resolve a domain" as we need the scheme sometimes.
> 
> Closes #1315.
> ```
> 
> If that looks good I'll land it. Thanks for pushing this forward!

That looks good!

Issue #1208, opened 2021-04-01
When response comes from a worker, some of the timing-info should pass through


noamr:
Right now when the response comes from a worker, it comes with timing info, however FETCH later overrides the timing info with the timing info of the fetch, which would lose some information. Some of the info such as  the connection info should come from the pass-through response, while other information such as the start time should come from the original request.

annevk:
This looks reasonable to me. One tricky aspect might be the time coarsening. If the SW has the capability but we don't, we need to coarsen the SW times here. Otherwise they can be kept as-is.

(That would be in line with what I suggested at https://github.com/w3c/ServiceWorker/pull/1575#discussion_r610467138 so if that changes we'd need to adjust for that here too.)

noamr:
As decided in the WG meeting, the info from the SW would remain opaque, but accessible from inside the service worker's performance timeline. The RT entry would stand for a particular `fetch` rather than a `response`.
[This work](https://github.com/whatwg/fetch/pull/1329) is the first step to make this explicit.

PR #1329, opened 2021-10-12
Return a "controller" instance from fetch

noamr:
Use the controller for external timing reporting and aborts.

This allows removing the 1:1 dependency between timing info and a response, as a response can be shared across fetch
instances (e.g. when restoring a response from cache, or when forwarding a response through a service worker).

See https://github.com/whatwg/fetch/issues/1208

   * Should reflect current implementation.

   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1329.html" title="Last updated on Mar 11, 2022, 10:12 AM UTC (dc7f07c)">Preview</a> | <a href="https://whatpr.org/fetch/1329/435c5ec...dc7f07c.html" title="Last updated on Mar 11, 2022, 10:12 AM UTC (dc7f07c)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2134,7 +2134,7 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
 <var>fetchParams</var>:
 
 <ol>
- <li><p>Assert: <var>fetchParams</var> is  <a for="fetch params">canceled</a>.</li>
+ <li><p>Assert: <var>fetchParams</var> is <a for="fetch params">canceled</a>.</li>
 
  <li><p>Return an <a>aborted network error</a> if <var>fetchParams</var> is
  <a for="fetch params">aborted</a>; otherwise return a <a>network error</a>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -4523,7 +4523,7 @@ these steps:
    <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
    <var>requestForServiceWorker</var>, with <var>fetchParams</var>'s
    <a for="fetch params">controller</a> and <var>fetchParams</var>'s
-   <a for="fetch params">cross-origin isolated capability</a> and . [[!HTML]] [[!SW]]
+   <a for="fetch params">cross-origin isolated capability</a>. [[!HTML]] [[!SW]]
 
    <li>
     <p>If <var>response</var> is not null, then:
--- a/fetch.bs
+++ b/fetch.bs
@@ -210,9 +210,8 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication). 
  <dt><dfn for="fetch params">cross-origin isolated capability</dfn> (default false)
  <dd>A boolean.
 
- <dt><dfn for="fetch params">terminated flag</dfn>
- <dt><dfn for="fetch params">aborted flag</dfn>
- <dd>A flag, initially unset.
+ <dt><dfn for="fetch params">fetch state</dfn>
+ <dd>"<code>ongoing</code>", "<code>terminated</code>", or "<code>aborted</code>".
 
  <dt><dfn for="fetch params">timing info</dfn>
  <dd>A <a for=/>fetch timing info</a>.
@@ -264,6 +263,15 @@ given a <a for=/>fetch timing info</a> <var>timingInfo</var>, return a new
  <var>storedTimingInfo</var>'s <a for="fetch timing info">decoded body size</a>.
 </ol>
 
+<p>To <dfn for="fetch params">abort</dfn> a <a for=/>fetch params</a> <var>fetchParams</var>,
+set <var>fetchParams</var>'s <a for="fetch params">fetch state</a> to "<code>aborted</code>".
+
+<p>To <dfn for="fetch params">terminate</dfn> a <a for=/>fetch params</a> <var>fetchParams</var>,
+set <var>fetchParams</var>'s <a for="fetch params">fetch state</a> to "<code>terminated</code>".
+
+<p>A <a for=/>fetch params</a> <var>fetchParams</var> is <a for="fetch params">terminated</a> if
+its <a for="fetch params">fetch state</a> is "<code>aborted</code>" or "<code>terminated</code>".
+
 <p>A <dfn export>fetch controller</dfn> is an object used to enable clients of an ongoing fetch
 to perform certain actions in the context of that fetch instance.
 
@@ -271,11 +279,12 @@ to perform certain actions in the context of that fetch instance.
 <dfn for="fetch controller">ongoing fetch params</dfn>.</p>
 
 <p>To <dfn for="fetch controller">terminate</dfn> a <a for=/>fetch controller</a>
-<var>controller</var> given a flag <var>aborted</var>, which is unset unless otherwise specified,
-set <var>controller</var>'s <a for="fetch controller">ongoing fetch params</a>
-<a for="fetch params">terminated flag</a>, and set <var>controller</var>'s
-<a for="fetch controller">ongoing fetch params</a> <a for="fetch params">aborted flag</a> if
-<var>aborted</var> is set.
+<var>controller</var>, <a for="fetch params">terminate</a> <var>controller</var>'s
+<a for="fetch controller">ongoing fetch params</a>.
+
+<p>To <dfn for="fetch controller">abort</dfn> a <a for=/>fetch controller</a>
+<var>controller</var>, <a for="fetch params">abort</a> <var>controller</var>'s
+<a for="fetch controller">ongoing fetch params</a>.
 
 <p>To <dfn>queue a fetch task</dfn>, given an algorithm <var>algorithm</var>, a
 <a for=/>global object</a> or a <a for=/>parallel queue</a> <var>taskDestination</var>, run these
@@ -2123,7 +2132,8 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
 
 <p>To create the <dfn>appropriate network error</dfn> given <a for=/>fetch params</a>
 <var>fetchParams</var>, return an <a>aborted network error</a> if <var>fetchParams</var>'s
-<a for="fetch params">aborted flag</a> is set, otherwise return a <a>network error</a>.
+<a for="fetch params">fetch state</a> is "<code>aborted</code>"; otherwise return a
+<a>network error</a>.
 
 <hr>
 
@@ -2307,14 +2317,14 @@ functionality.
 
 <p>A <a for="fetch group">fetch record</a> has an associated
 <dfn export for="fetch record" id=concept-fetch-record-fetch>fetch</dfn> (a
-<a for=/>fetch</a> algorithm or null).
+<a for=/>fetch controller</a> or null).
 
 <hr>
 
 <p>When a <a for=fetch>fetch group</a> is
 <dfn export for="fetch group" id=concept-fetch-group-terminate>terminated</dfn>, for each associated
 <a for="fetch group">fetch record</a> whose <a for="fetch record">request</a>'s <a>done flag</a> is
-unset or <a for=request>keepalive</a> is false, <a lt=terminated for=fetch>terminate</a> the
+unset or <a for=request>keepalive</a> is false, <a for="fetch controller">terminate</a>
 <a for="fetch group">fetch record</a>'s <a for="fetch record">fetch</a>.
 
 
@@ -3839,7 +3849,13 @@ the request.
  <p>If this instance of <a for=/>fetch</a> is
  <dfn export for=fetch id=concept-fetch-terminate>terminated</dfn> given flag <var>aborted</var>,
  which is unset unless otherwise specified, <a for="fetch controller">terminate</a>
- <var>controller</var> with <var>aborted</var>.
+ <var>controller</var> if <var>aborted</var> is set, otherwise <a for="fetch controller">abort</a>
+ <var>controller</var>.
+
+ <p class="note">The above exported function is to be replaced by
+ <a for="fetch controller" lt=terminate>terminating</a> or
+ <a for="fetch controller" lt=abort>aborting</a> the <a for=/>fetch controller</a>
+ returned by this algorithm.
 
  <li><p>If <var>request</var>'s <a for=request>body</a> is a <a for=/>byte sequence</a>, then set
  <var>request</var>'s <a for=request>body</a> to the first return value of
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -255,6 +258,29 @@ given a <a for=/>fetch timing info</a> <var>timingInfo</var>, return a new
...
+set <var>fetchParams</var>'s <a for="fetch params">fetch state</a> to "<code>aborted</code>".
+
+<p>To <dfn for="fetch params">terminate</dfn> a <a for=/>fetch params</a> <var>fetchParams</var>,
+set <var>fetchParams</var>'s <a for="fetch params">fetch state</a> to "<code>terminated</code>".
+
+<p>A <a for=/>fetch params</a> <var>fetchParams</var> is <dfn for="fetch params">canceled</dfn> if
+its <a for="fetch params">fetch state</a> is "<code>aborted</code>" or "<code>terminated</code>".
+
+<p>A <dfn export>fetch controller</dfn> is an object used to enable clients of an ongoing fetch
+to perform certain actions in the context of that fetch instance.
```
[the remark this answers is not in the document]
annevk: It seems a bit weird for state to be terminated. But I do like this idea. Can it instead be the controller that holds the state? And fetch params / fetch record have a reference to the controller?

And if we need to expose more things on the controller in due course, we could do that this way? And only expose it at the appropriate time, etc.

That does seem somewhat cleaner than having an internal slot and hiding it by not exporting it.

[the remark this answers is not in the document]
noamr: > It seems a bit weird for state to be terminated. But I do like this idea. Can it instead be the controller that holds the state? And fetch params / fetch record have a reference to the controller?
> 
> 
> 
> And if we need to expose more things on the controller in due course, we could do that this way? And only expose it at the appropriate time, etc.
> 
> 
> 
> That does seem somewhat cleaner than having an internal slot and hiding it by not exporting it.

Sure, I will revise 
It might feel clunky with `timing info` but we can cross the bridge when we get there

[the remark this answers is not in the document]
annevk: I suspect that might be something we need to keep on both. But once it's put on controller it isn't modified again, or some such.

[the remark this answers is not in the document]
noamr: Gotcha 
Let me put out the PR for this one and I will tinker with how this works for timing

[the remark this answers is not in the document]
jakearchibald: @annevk 

> It seems a bit weird for state to be terminated. But I do like this idea. Can it instead be the controller that holds the state? And fetch params / fetch record have a reference to the controller?

fwiw, that's what I was trying to say, although I then renamed the controller since it was just a state holder.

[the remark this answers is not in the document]
noamr: It's a state holder now, but when I put "finalize and report timing" there it will do a bit more.
The controller is what allows you to terminate/finalize the fetch after it's started, the fact that it holds a state is less relevant since that state is not readable from the outside, only writable with terminate/abort/finalize.

annevk:
As noted on Matrix we need to account for service workers and in particular "handle fetch". Either we make it so that "handle fetch" returns a controller for the service worker fetch event, or we pass the controller into "handle fetch" so the service worker can observe it. The former seems a bit nicer as it reuses the pattern introduced here and could be used for other kinds of information exchange in the future. It would require "handle fetch" to take a callback.

@yutakahirano @jakearchibald @noamr thoughts?

noamr:
> As noted on Matrix we need to account for service workers and in particular "handle fetch". Either we make it so that "handle fetch" returns a controller for the service worker fetch event, or we pass the controller into "handle fetch" so the service worker can observe it. The former seems a bit nicer as it reuses the pattern introduced here and could be used for other kinds of information exchange in the future. It would require "handle fetch" to take a callback.
> 
> @yutakahirano @jakearchibald @noamr thoughts?

I like the idea of returning a controller from SW's fetch rather than having SW query its caller fetch for "is terminated"... 
Does it affect this PR?

annevk:
Yeah, I think we need to tackle it as otherwise fetch termination in service workers ends up in UB territory.

noamr:
> Yeah, I think we need to tackle it as otherwise fetch termination in service workers ends up in UB territory.

It is currently in undefined behavior territory, this doesn't change much about it. can't we fix this iteratively?

noamr:
> As noted on Matrix we need to account for service workers and in particular "handle fetch". Either we make it so that "handle fetch" returns a controller for the service worker fetch event, or we pass the controller into "handle fetch" so the service worker can observe it. The former seems a bit nicer as it reuses the pattern introduced here and could be used for other kinds of information exchange in the future. It would require "handle fetch" to take a callback.
> 
> @yutakahirano @jakearchibald @noamr thoughts?

See https://github.com/w3c/ServiceWorker/pull/1612 and latest revision of this PR.
- Instead of returning a controller, we emit it in a new `onFetchStart` callback
- Termination of the service worker and of fetches initiated by the service worker are done explicitly

fetch.bs
```diff
@@ -3756,6 +3804,9 @@ the request.
  <a for="fetch params">cross-origin isolated capability</a> is
  <var>crossOriginIsolatedCapability</var>.
 
+ <li><p>If <var>processFetchStart</var> is not null, invoke <var>processFetchStart</var> with
+ <var>fetchParams</var>'s <a for="fetch params">controller</a>.
```
annevk, line 3808: Why the change to this rather than returning the controller?

fetch.bs
```diff
@@ -4377,8 +4425,14 @@ these steps:
...
 
+   <li><p>Let <var>onServiceWorkerFetchStarted</var> given <a>service worker fetch controller</a>
+   <var>controller</var>: Set <var>fetchParams</var>'s <a for="fetch params">controller</a>'s
+   <a for="fetch controller">service worker controller</a> to <var>controller</var>. [[!SW]]
+
    <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
-   <var>requestForServiceWorker</var>. [[!HTML]] [[!SW]]
+   <var>requestForServiceWorker</var> with <var>fetchParams</var>'s
+   <a for="fetch params">crossOriginIsolatedCapability</a>, and
+   <var>onServiceWorkerFetchStarted</var>. [[!SW]]
```
annevk, line 4435: Here too I thought the setup was going to be that handle fetch returns a controller that we abort if our controller is aborted. And that handle fetch would run in parallel after that, invoking a callback with a response at some point.

fetch.bs
noamr: The problem was that HTTP fetch is synchronous and returns a response. So changing SW fetch to be asynchronous would require a big refactor and I try to be consistent. Unless there is some await-style language I can use?

annevk: HTTP fetch always runs in parallel so in principle we can indeed "wait" for something to happen. Or we could split HTTP fetch and invoke the subsequent part from the callback.

However, I think a simpler version could be that we share the fetch controller with the service worker. That way the service worker can observe the state and act accordingly. That would require less refactoring and less data structures. What do you think?

noamr: That would make things simpler

But it would mean that the SW terminating its internal fetch would be a "side effect" of termination ('when the controller is terminated'), and I tried to make it an explicit call from a termination algorithm where it's clear what calls what. I can go either way

annevk: I see. I think if we use https://infra.spec.whatwg.org/#abort-when that should be okay. That would also make it clear what the scope is.

fetch.bs
noamr: Revised based on our conversation back to this simple way of working, with "when aborted" as a side effect inside service workers.

fetch.bs
```diff
@@ -4419,7 +4452,9 @@ these steps:
    given <var>fetchParams</var>'s <a for="fetch params">cross-origin isolated capability</a>.
 
    <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
-   <var>requestForServiceWorker</var>. [[!HTML]] [[!SW]]
+   <var>requestForServiceWorker</var>, with <var>fetchParams</var>'s
+   <a for="fetch params">cross-origin isolated capability</a> and <var>fetchParams</var>'s
+   <a for="fetch params">controller</a>. [[!HTML]] [[!SW]]
```
annevk, line 4457: What do you think about making the order controller, request, capability? That would match rough order of importance I think.

annevk (commented):
Are you working on a new service worker PR? I think the Fetch side looks good, though I'd like to do another detailed read once we have the full story. @jakearchibald what do you think?

fetch.bs
noamr: Sure

noamr: Actually I think it should be `request`, `controller` and `capability`. Request is the direct thing being passed, and the other two are more like context.

noamr:
@annevk I was working on the corresponding SW PR: https://github.com/whatwg/fetch/pull/1329 and I see that it's closed for some reason, I must have closed it by mistake and I don't seem to be able to re-open. Opened a new one: https://github.com/w3c/ServiceWorker/pull/1620

fetch.bs
```diff
@@ -4419,7 +4452,9 @@ these steps:
    given <var>fetchParams</var>'s <a for="fetch params">cross-origin isolated capability</a>.
 
    <li><p>Set <var>response</var> to the result of invoking <a for=/>handle fetch</a> for
-   <var>requestForServiceWorker</var>. [[!HTML]] [[!SW]]
+   <var>requestForServiceWorker</var>, with <var>fetchParams</var>'s
+   <a for="fetch params">controller</a> and <var>fetchParams</var>'s
+   <a for="fetch params">cross-origin isolated capability</a> and . [[!HTML]] [[!SW]]
```
annevk, line 4457:
```suggestion
   <a for="fetch params">cross-origin isolated capability</a>. [[!HTML]] [[!SW]]
```

fetch.bs
```diff
@@ -2247,14 +2278,14 @@ functionality.
 
 <p>A <a for="fetch group">fetch record</a> has an associated
 <dfn export for="fetch record" id=concept-fetch-record-fetch>fetch</dfn> (a
-<a for=/>fetch</a> algorithm or null).
+<a for=/>fetch controller</a> or null).
```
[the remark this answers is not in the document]
annevk: I suggest we tackle this in a follow-up.

annevk:
@jakearchibald @yutakahirano could you please review this PR?

fetch.bs
```diff
@@ -2104,6 +2130,16 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
 <a for=response>header list</a> is always empty, and
 <a for=response>body</a> is always null.
 
+<p>To create the <dfn>appropriate network error</dfn> given <a for=/>fetch params</a>
+<var>fetchParams</var>:
+
+<ol>
+ <li><p>Assert: <var>fetchParams</var> is  <a for="fetch params">canceled</a>.</li>
```
annevk, line 2137:
```suggestion
 <li><p>Assert: <var>fetchParams</var> is <a for="fetch params">canceled</a>.</li>
```

fetch.bs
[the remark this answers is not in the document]
annevk: Can you file an issue for this?

fetch.bs
annevk: This does not appear to be resolved?

fetch.bs
```diff
@@ -5521,20 +5532,16 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+   <li><p>Return the <a for=/>appropriate network error</a> for <var>fetchParams</var>.
   </ol>
 
  <li><p>Let <var>pullAlgorithm</var> be an action that <a lt=resumed for=fetch>resumes</a> the
  ongoing fetch if it is <a lt=suspend for=fetch>suspended</a>.
 
- <li><p>Let <var>cancelAlgorithm</var> be an action that <a lt=terminated for=fetch>terminates</a>
- the ongoing fetch with the aborted flag set.
+ <li><p>Let <var>cancelAlgorithm</var> be an algorithm that
+ <a for="fetch controller">abort</a>s <var>fetchParams</var>'s <a for="fetch params">controller</a>.
```
annevk, line 5544:
```suggestion
 <a for="fetch controller">aborts</a> <var>fetchParams</var>'s <a for="fetch params">controller</a>.
```

fetch.bs
```diff
@@ -8074,6 +8085,11 @@ in a <a for=/>parallel queue</a> if <a for=fetch><i>useParallelQueue</i></a> is
  as arguments will be invoked. Hopefully most standards will not need this.
 </dl>
 
+<p>When invoked, the <a for=/>fetch</a> operation returns a <a for=/>fetch controller</a>.
+The controller is used for performing actions on a fetch operation that has already started,
+such as <a for="fetch controller" lt=abort>aborting</a> the operation by the user or page logic, or
+<a for="fetch controller" lt=abort>terminating</a> it due to a browser-internal circumstance.
```
annevk, line 8091: No lt=abort here, right?

annevk (commented):
Found a couple minor remaining issues. Also pinging @jakearchibald and @yutakahirano again for review.

fetch.bs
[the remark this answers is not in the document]
noamr: https://github.com/whatwg/fetch/issues/1403

noamr:
> Found a couple minor remaining issues. Also pinging @jakearchibald and @yutakahirano again for review.

@jakearchibald ping?

fetch.bs
```diff
@@ -210,10 +210,36 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication).
...
  <dd>A <a for=/>fetch timing info</a>.
 </dl>
 
+<p>A <dfn export>fetch controller</dfn> is a <a for=/>struct</a> used to enable callers of
+<a for=/>fetch</a> to perform certain operations on it after it has started. It has the following
+<a for=struct>items</a>:
+
+<dl>
+ <dt><dfn for="fetch controller">state</dfn> (default "<code>ongoing</code>")
+ <dd>"<code>ongoing</code>", "<code>terminated</code>", or "<code>aborted</code>"
```
jakearchibald, line 226: Should there be a state that represents a fetch that can no longer be aborted or terminated? I guess that would be "complete".

It seems weird that a fully completed fetch would have a state of "ongoing".

jakearchibald (approved):
LGTM. One comment, but it can be a followup

fetch.bs
noamr: Yes! I will add it when "Finalize and report timing" would move to the fetch controller, otherwise no one would set that value :)

annevk:
This looks good to me too. One thing we'll need is a tentative commit message that summarizes the changes and links the various follow-up issues. Is that something you can put together @noamr? In a new comment would be best.

noamr:
> This looks good to me too. One thing we'll need is a tentative commit message that summarizes the changes and links the various follow-up issues, the resolved issues, and the related service worker PR. Is that something you can put together @noamr? In a new comment would be best.

Sure, I will post follow up issues and comment with a commit message.

noamr:
Commit message:
---

Return a "controller" instance from the fetch algorithm, which allows clients of fetch to perform certain actions on an ongoing fetch, initially to abort or terminate it, and later on to conclude it.

This will eventually remove the need of using `response` as the object that retains `timing info`, as that creates certain ambiguities.

Followup action items:
https://github.com/whatwg/fetch/issues/1412

PR #1330, opened 2021-10-12
CORS: more information for server developers

### LINKED ISSUE #1102: No documentation on how a server must response to a wrong CORS-headers-request.
The specification states that a client must send `Access-Control-Request-Method` and `Access-Control-Request-Headers` headers to the server to "indicate which method/headers a future CORS request to the same resource might use".

But there is nowhere to be found explicitly, why a server needs them and how it should react if these client-sent headers are not included in the response.

What I've found:
- A client must send `Access-Control-Request-Method` and `Access-Control-Request-Headers`.
- A server must respond with available `Access-Control-Allow-Methods` and `Access-Control-Allow-Headers`
- A server must respond with 2xx status if it allows CORS requests.
- A server must respond with 403 status if it doesn't allow any CORS requests.

But nothing states about what server should respond with if it allows CORS requests, but requested method or headers are not available.

Example request/response:

REQUEST:
Access-Control-Request-Method: BADMETHOD
Access-Control-Request-Headers: Bad-Header

RESPONSE: 200 OK (204 NO CONTENT)
Access-Control-Allow-Methods: OPTIONS,HEAD,GET,PUT,POST,DELETE
Access-Control-Allow-Headers: Origin,X-Requested-With,Content-Type,Accept

This seems illogical because these headers are no use for the server: the server has a static response no matter what requested headers were.

Thank you in advance.

annevk:
Fixes #1102.

***
<a href="https://whatpr.org/fetch/1330.html" title="Last updated on Oct 13, 2021, 9:45 AM UTC (b123f27)">Preview</a> | <a href="https://whatpr.org/fetch/1330/6f37b51...b123f27.html" title="Last updated on Oct 13, 2021, 9:45 AM UTC (b123f27)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2965,7 +2965,7 @@ to an <a for=/>ok status</a>, e.g., 200 or 204.
 
 <p>Any other kind of HTTP response is not successful and will either end up not being shared or fail
 the <a>CORS-preflight request</a>. Be aware that any work the server performs might nonetheless leak
-through side channels such as timing. If server developers wish to denote this explicitly the 403
+through side channels, such as timing. If server developers wish to denote this explicitly, the 403
 <a for=/>status</a> can be used, coupled with omitting the relevant <a for=/>headers</a>.
 
 <p class=note>If desired, “failure” could also be shared, but that would make it a successful HTTP


annevk:
Thanks, that's very useful feedback!

And yeah, that kind of implementation should be fine and I should make that more clear.

We indeed require a 2xx response for CORS preflights (other CORS requests can have any kind of response code), but that's not written down in this section. It should be. 204 should always be okay though.

annevk:
I revised the text quite significantly to account for the feedback from @rossabaker. I also downplayed the 403 suggestion, but only slightly, and gave a hint how 403 could be used while still making debugging nice.

annevk:
If @rossabaker could take another look that would be appreciated. Perhaps @jakearchibald too? I suggest we take further discussion of 403 to a new issue, if needed, but I think we need to give some advice on this one way or another.

Issue #590, opened 2017-08-30
Define the "Preload Cache"


yoavweiss:
We need to define the preload cache, as currently it is not defined and different implementations are doing [observably different things](https://bugzilla.mozilla.org/show_bug.cgi?id=1222633#c38).

This issue was [opened](https://github.com/w3c/preload/issues/97) on the Preload spec, but as the cache would mostly sit inside Fetch and it's not clear what action would be needed on the preload spec side, I'm "moving" it here.

Related previous discussion here was at https://github.com/whatwg/fetch/issues/354 where @jakearchibald made a [proposal](https://github.com/whatwg/fetch/issues/354#issuecomment-237508188) to address this. I open this as a separate issue as I think the H2 push cache and the preload cache are inherently different things in different layers.

I think that anything we define here should, to some extent, look at how the different implementations tackle this today. The logic that they apply for this (e.g. in [Blink](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/loader/fetch/ResourceFetcher.cpp?sq=package:chromium&l=997) or in [WebKit](https://github.com/WebKit/webkit/blob/master/Source/WebCore/loader/cache/CachedResourceLoader.cpp#L946)) seems a bit complex, but was created for the "memory cache" case. We need to think if it can be safely simplified for the preload case. (and if memory cache itself should be standardized)

wanderview:
What preload cache behavior is desired when `document.open()` is called?  It seems the document is preserved, but "reset" in this case.  The spec and firefox also create a new global.  Should preload cache be cleared or preserved across `document.open()`?

annevk:
I think we should just pick something and make sure it's tested. If we can get away with not creating a new global for `document.open()` (which would simplify matters quite a bit, as a document can then only ever be associated with a single global) it would probably be most natural that the cache is kept.

igrigorik:
> I open this as a separate issue as I think the H2 push cache and the preload cache are inherently different things in different layers.

From a developer ergonomics perspective, it would be nice if that wasn't the case. As it stands, there are subtle (potentially) observable differences in behavior between pushing and preloading a resource — one ends up in h2 push cache, other in renderer's memory cache, at least for Chrome. It would be nice to normalize all this, for everyone's sanity.. if we can.

> (and if memory cache itself should be standardized)

Given that, afaik, every implementation has one. Yes? Which is what we're after here.. :)

wanderview:
[...]

PR #1342, opened 2021-10-26
Integration with preload

noamr:
Before any particular fetch steps are performed, see if there is a
matching request already in the preload store and consume it.

This is called from the main fetch to avoid race conditions.

Depends on https://github.com/whatwg/html/pull/7260, and together they fix https://github.com/whatwg/fetch/issues/590.

   * …
   * …
- [x ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/31539
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1342.html" title="Last updated on Dec 13, 2021, 11:47 AM UTC (4c66e58)">Preview</a> | <a href="https://whatpr.org/fetch/1342/6257e22...4c66e58.html" title="Last updated on Dec 13, 2021, 11:47 AM UTC (4c66e58)">Diff</a>

***
<a href="https://whatpr.org/fetch/1342.html" title="Last updated on Feb 11, 2022, 1:03 PM UTC (77aa0d9)">Preview</a> | <a href="https://whatpr.org/fetch/1342/6257e22...77aa0d9.html" title="Last updated on Feb 11, 2022, 1:03 PM UTC (77aa0d9)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3868,7 +3868,7 @@ the request.
    <span>consume a preloaded resource</span> for <var>req</var>'s <a for=request>window</a>,
    given <var>req</var>'s <a for=request>URL</a>, <var>req</var>'s <a for=request>destination</a>,
    <var>req</var>'s <a for=request>mode</a>, <var>req</var>'s <a for=request>credentials mode</a>,
-   <var>req</var>'s <a for=request>integrity metadata</a>, and 
+   <var>req</var>'s <a for=request>integrity metadata</a>, and
    <var>onPreloadedResponseAvailable</var>.
 
    <li><p>If <var>foundPreloadedResource</var> is true, then return.
--- a/fetch.bs
+++ b/fetch.bs
@@ -3828,9 +3828,7 @@ the request.
   <p>If all of the following conditions are true:</p>
 
   <ul class=brief>
-   <li><p><var>request</var>'s <a for=request>unsafe-request flag</a> is not set, or
-   <var>request</var>'s <a for=request>headers</a>'s <a for=Headers>header list</a> is
-   <a for="list">empty</a>
+   <li><p><var>request</var>'s <a for=request>destination</a> is not "<code>document</code>"
 
    <li><p><var>request</var>'s <a for=request>URL</a>'s <a for=url>scheme</a> is an
    <a>HTTP(S) scheme</a>
@@ -3841,6 +3839,10 @@ the request.
    <li><p><var>request</var>'s <a for=request>window</a> is not null
 
    <li><p><var>request</var>'s <a for=request>method</a> is `<code>GET</code>`
+
+   <li><p><var>request</var>'s <a for=request>unsafe-request flag</a> is not set, or
+   <var>request</var>'s <a for=request>headers</a>'s <a for=Headers>header list</a> is
+   <a for="list">empty</a>
   </ul>
 
   <p>then perform the following steps:
--- a/fetch.bs
+++ b/fetch.bs
@@ -3828,8 +3828,6 @@ the request.
   <p>If all of the following conditions are true:</p>
 
   <ul class=brief>
-   <li><p><var>request</var>'s <a for=request>destination</a> is not "<code>document</code>"
-
    <li><p><var>request</var>'s <a for=request>URL</a>'s <a for=url>scheme</a> is an
    <a>HTTP(S) scheme</a>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -1823,9 +1823,6 @@ otherwise, it is unset.
 <a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
-<p>A <a for=/>request</a> has an associated boolean <dfn for=request>allow preloads</dfn>.
-Unless stated otherwise it is true.
-
 <hr>
 
 <p>A <dfn export>subresource request</dfn> is a <a for=/>request</a>
@@ -3831,7 +3828,9 @@ the request.
   <p>If all of the following conditions are true:</p>
 
   <ul class=brief>
-   <li><p><var>request</var>'s <a for=request>allow preloads</a> is true
+   <li><p><var>request</var>'s <a for=request>unsafe-request flag</a> is not set, or
+   <var>request</var>'s <a for=request>headers</a>'s <a for=Headers>header list</a> is
+   <a for="list">empty</a>
 
    <li><p><var>request</var>'s <a for=request>URL</a>'s <a for=url>scheme</a> is an
    <a>HTTP(S) scheme</a>
@@ -7138,10 +7137,6 @@ constructor steps are:
 
    <li><p>Otherwise, <a for=Headers>fill</a> <a>this</a>'s <a for=Request>headers</a> with
    <var>headers</var>.
-
-   <li><p>If <a>this</a>'s <a for=Request>headers</a>'s <a for=Headers>header list</a> is not
-   <a for="list">empty</a>, then set <a>this</a>'s <a for=Request>request</a>'s
-   <a for=request>allow preloads</a> to false.
   </ol>
  </li>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -1823,6 +1823,9 @@ otherwise, it is unset.
 <a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
+<p>A <a for=/>request</a> has an associated boolean <dfn for=request>allow preloads</dfn>.
+Unless stated otherwise it is true.
+
 <hr>
 
 <p>A <dfn export>subresource request</dfn> is a <a for=/>request</a>
@@ -3823,6 +3826,50 @@ the request.
  <li><p>If <var>request</var>'s <a for=request>origin</a> is "<code>client</code>", then set
  <var>request</var>'s <a for=request>origin</a> to  <var>request</var>'s <a for=request>client</a>'s
  <a for="environment settings object">origin</a>.
+
+ <li><p>Let <var>hasCustomHeaders</var> be false.
+
+ <li><p><a for=list>For each</a> <var>header</var> in <var>request</var>'s
+ <a for=request>header list</a>, if <var>header</var>'s <a for=header>name</a> is
+ not `<code>If-Modified-Since</code>`, `<code>If-None-Match</code>`, `<code>If-Unmodified-Since</code>`,
+ `<code>If-Match</code>`, `<code>User-Agent</code>`, or `<code>Cache-Control</code>`, set
+ <var>hasCustomHeaders</var> to true.
+
+ <li>
+  <p>If all of the following conditions are true:</p>
+
+  <ul class=brief>
+   <li><p><var>request</var>'s <a for=request>allow preloads</a> is true
+
+   <li><p><var>request</var>'s <a for=request>origin</a> is <var>request</var>'s
+   <a for=request>client</a>'s <a for="environment settings object">origin</a>
+
+   <li><p><var>request</var>'s <a for=request>URL</a>'s <a for=url>scheme</a> is an
+   <a>HTTP(S) scheme</a>
+
+   <li><p><var>request</var>'s <a for=request>mode</a> is "<code>same-origin</code>",
+   "<code>cors</code>", or "<code>no-cors</code>"
+
+   <li><p><var>hasCustomHeaders</var> is false
+
+   <li><p><var>request</var>'s <a for=request>window</a> is not null
+
+   <li><p><var>request</var>'s <a for=request>method</a> is `<code>GET</code>`
+  </ul>
+
+  <p>then perform the following steps:
+
+  <ol>
+   <li><p>Let <var>preloadedResourceConsumed</var> be the result of invoking
+   <span>consume a preloaded resource</span> for <var>req</var>'s <a for=request>window</a>,
+   given <var>req</var>'s <a for=request>URL</a>, <var>req</var>'s <a for=request>destination</a>,
+   <var>req</var>'s <a for=request>mode</a>, <var>req</var>'s <a for=request>credentials mode</a>,
+   <var>req</var>'s <a for=request>integrity metadata</a>, and the next step <a>in parallel</a>: Run
+   <a>fetch finale</a> given <var>response</var> and <var>fetchParams</var>
+
+   <li><p>If <var>preloadedResourceConsumed</var> is true, then return.
+  </ol>
+ </li>
 
  <li>
   <p>If <var>request</var>'s <a for=request>policy container</a> is "<code>client</code>", then:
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -3848,11 +3848,18 @@ steps:
...
  <li><p>Let <var>request</var> be <var>fetchParams</var>'s <a for="fetch params">request</a>.
 
- <li><p>Let <var>response</var> be null.
+ <li>
+  <p>Let <var>response</var> be the result of invoking <span>consume a preloaded resource</span>
+  for <var>request</var>.
+
+  <p class="note">This is done in main fetch and not in HTTP-fetch to avoid blocking or race
+  conditions between the networking queue and the document queue. The implication is that if a
+  fetch redirects to a URL that was preloaded, it would not be consumed.</p>
```
annevk, line 3857: I don't understand this comment. HTTP-redirect fetch invokes main fetch at the end.

fetch.bs
```diff
@@ -3848,11 +3848,18 @@ steps:
...
+ </li>
 
- <li><p>If <var>request</var>'s <a>local-URLs-only flag</a> is set and <var>request</var>'s
- <a for=request>current URL</a> is not <a lt="is local">local</a>, then set <var>response</var> to a
- <a>network error</a>.
+ <li><p>If <var>response</var> is null, <var>request</var>'s <a>local-URLs-only flag</a> is set and
+ <var>request</var>'s  <a for=request>current URL</a> is not <a lt="is local">local</a>, then set
+ <var>response</var> to a <a>network error</a>.
```
annevk, line 3862: This seems to give the idea that a network URL could return something in a local-URLs-only environment, e.g., SVG-as-image. That seems wrong.

fetch.bs
noamr: Hmm looking at this again, this patch is wrong. Will resubmit

noamr:
@annevk I've revised this, https://github.com/whatwg/html/pull/7260 would also need some changes to match, but I wanted to see that this is on the right track first.

The idea is that preloads are matched *before*  the fetch becomes parallel and overrides all the HTTP specific check as it's assumed they were done for the preloaded request.

An alternative to doing things this way is to have the preload list inside fetch, and have the HTML spec add to it / clear it. This might make things easier in terms of future potential things like supporting `Link` headers in workers, but I'm not sure if that should be a concern right now (also doing that is not difficult inside the HTML spec).

fetch.bs
```diff
@@ -3811,6 +3814,44 @@ the request.
  <var>request</var>'s <a for=request>origin</a> to  <var>request</var>'s <a for=request>client</a>'s
  <a for="environment settings object">origin</a>.
 
+ <li><p>Let <var>hasCustomHeaders</var> be false.
+
+ <li><p><a for=list>For each</a> <var>header</var> in <var>request</var>'s
+ <a for=request>header list</a>, if <var>header</var>'s <a for=header>name</a> is
+ not `<code>If-Modified-Since</code>`, `<code>If-None-Match</code>`, `<code>If-Unmodified-Since</code>`,
+ `<code>If-Match</code>`, `<code>Origin</code>`, `<code>User-Agent</code>`, `<code>Referer</code>`,
+ or `<code>Cache-Control</code>`, set <var>hasCustomHeaders</var> to true.
```
annevk, line 3823: It cannot contain some of these headers at this point. E.g., Origin and Referer are forbidden.

annevk (commented):
Generally I think this can work, but it will depend a bit on the other details.

I'm also not sure where the cache itself should be defined. I can see arguments either way.

annevk:
Actually, I'm not sure. Doing a cache lookup synchronously seems bad, but perhaps that is what implementations do? That probably warrants a bit more investigation. One way to do it later in "main fetch" would be to check the value of _recursive_ I suppose, although that might still run into problems with HTTP authentication requests.

noamr:
> Actually, I'm not sure. Doing a cache lookup synchronously seems bad, but perhaps that is what implementations do? That probably warrants a bit more investigation. One way to do it later in "main fetch" would be to check the value of _recursive_ I suppose, although that might still run into problems with HTTP authentication requests.

Cache lookup *should* be synchronous IMO as it relies on something controlled by the main thread, making it parallel would create either locks or unnecessary races.

Maybe keeping it where I put it is correct?

noamr:
See https://github.com/whatwg/html/pull/7260#issuecomment-962600126

fetch.bs
```diff
@@ -3811,6 +3814,54 @@ the request.
...
+
+   <li><p><var>request</var>'s <a for=request>window</a> is not null
+
+   <li><p><var>request</var>'s <a for=request>method</a> is `<code>GET</code>`
+  </ul>
+
+  <p>then perform the following steps:
+
+  <ol>
+   <li><p>Let <var>onPreloadedResponseAvailable</var> be to run the following step
```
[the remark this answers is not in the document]
noamr: Not sure how, this seems to be the consistent form. Suggestions?

noamr:
Signed and attached. Looks good!

fetch.bs
```diff
@@ -3811,6 +3814,50 @@ the request.
...
+  </ul>
+
+  <p>then perform the following steps:
+
+  <ol>
+   <li><p>Let <var>preloadedResourceConsumed</var> be the result of invoking
+   <span>consume a preloaded resource</span> for <var>req</var>'s <a for=request>window</a>,
+   given <var>req</var>'s <a for=request>URL</a>, <var>req</var>'s <a for=request>destination</a>,
+   <var>req</var>'s <a for=request>mode</a>, <var>req</var>'s <a for=request>credentials mode</a>,
+   <var>req</var>'s <a for=request>integrity metadata</a>, and the next step <a>in parallel</a>: Run
```
yoavweiss: It'd be clearer IMO to pass a helper algorithm here.

fetch.bs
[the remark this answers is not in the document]
yoavweiss: Maybe "Let onPreloadedResponseAvailabe be an algorithm that runs the following steps..."?

fetch.bs
yoavweiss: No longer relevant

fetch.bs
noamr: This part has been deleted

fetch.bs
```diff
@@ -3811,6 +3814,54 @@ the request.
  <var>request</var>'s <a for=request>origin</a> to  <var>request</var>'s <a for=request>client</a>'s
  <a for="environment settings object">origin</a>.
 
+ <li><p>Let <var>hasCustomHeaders</var> be false.
+
+ <li><p><a for=list>For each</a> <var>header</var> in <var>request</var>'s
+ <a for=request>header list</a>, if <var>header</var>'s <a for=header>name</a> is
+ not `<code>If-Modified-Since</code>`, `<code>If-None-Match</code>`, `<code>If-Unmodified-Since</code>`,
+ `<code>If-Match</code>`, `<code>User-Agent</code>`, or `<code>Cache-Control</code>`, set
+ <var>hasCustomHeaders</var> to true.
```
annevk, line 3823: This would always end up with _hasCustomHeaders_ being true.

fetch.bs
```diff
@@ -3811,6 +3814,54 @@ the request.
...
+
+ <li>
+  <p>If all of the following conditions are true:</p>
+
+  <ul class=brief>
+   <li><p><var>request</var>'s <a for=request>allow preloads</a> is true
+
+   <li><p><var>request</var>'s <a for=request>origin</a> is <var>request</var>'s
+   <a for=request>client</a>'s <a for="environment settings object">origin</a>
```
annevk, line 3832: When would this not be true?

fetch.bs
```diff
@@ -3811,6 +3814,54 @@ the request.
...
+
+   <li><p><var>request</var>'s <a for=request>method</a> is `<code>GET</code>`
+  </ul>
+
+  <p>then perform the following steps:
+
+  <ol>
+   <li><p>Let <var>onPreloadedResponseAvailable</var> be to an algorithm that runs the following
+   step <a>in parallel</a> given <a for=/>response</a> <var>response</var>:
+   Run <a>fetch finale</a> given <var>response</var> and <var>fetchParams</var>.
```
annevk, line 3852: Wrapping needs to be adjusted here. Run should be lowercase.

annevk (commented):
Where does this end up invoking finalize response?

fetch.bs
noamr: It's theoretically possible to call fetch with a different origin, the only place where I found this to happen is [Process a navigate fetch](https://html.spec.whatwg.org/#process-a-navigate-fetch) which is not relevant for preloads. Maybe this can be an assert?

fetch.bs
noamr: How come?
Anyway, maybe we should avoid `hasCustomHeaders` at all, and Let `fetch()` set `allow preloads` to false if there are user-defined headers.

fetch.bs
noamr: OK

noamr:
> Where does this end up invoking finalize response?

[Here](https://github.com/whatwg/fetch/pull/1342/files#diff-b4827ee42aceaf9dc3e7216f72262d81c32217e9996fb70a657f259e3ac92604R3852). When the response is available, `fetch finale` is called and should run `finalize response` (once https://github.com/whatwg/fetch/pull/1311 is in that part would be a lot cleaner)

fetch.bs
```diff
@@ -3811,6 +3814,44 @@ the request.
...
+  <ol>
+   <li><p>Assert: <var>request</var>'s <a for=request>origin</a> is <var>request</var>'s
+   <a for=request>client</a>'s <a for="environment settings object">origin</a>.
+
+   <li><p>Let <var>onPreloadedResponseAvailable</var> be to an algorithm that runs the following
+   step <a>in parallel</a> given <a for=/>response</a> <var>response</var>: run
+   <a>fetch finale</a> given <var>response</var> and <var>fetchParams</var>.
+
+   <li><p>Let <var>foundPreloadedResource</var> be the result of invoking
+   <span>consume a preloaded resource</span> for <var>req</var>'s <a for=request>window</a>,
```
annevk, line 3845: This should use `<a>`.

fetch.bs
annevk: I had misread it, but it also seems the requirement is different now. Any author-defined headers triggers it now. But doesn't that also apply to `XMLHttpRequest`?

annevk (commented):
When would we expect to use "allow preloads"? If it's just XHR and `fetch()` I think we have enough state on request to make that determination already.

noamr:
> When would we expect to use "allow preloads"? If it's just XHR and `fetch()` I think we have enough state on request to make that determination already.

Custom headers currently should disallow preloads, but also things like "synchronous XHR", which I don't know if we have the knowledge for at this point.

fetch.bs
```diff
@@ -3824,6 +3824,47 @@ the request.
  <var>request</var>'s <a for=request>origin</a> to  <var>request</var>'s <a for=request>client</a>'s
  <a for="environment settings object">origin</a>.
 
+ <li>
+  <p>If all of the following conditions are true:</p>
+
+  <ul class=brief>
+   <li><p><var>request</var>'s <a for=request>destination</a> is not "<code>document</code>"
```
annevk, line 3831: Should this use https://fetch.spec.whatwg.org/#navigation-request instead? The only overlap I can see with mode is for embed/object. Otherwise this would be redundant.

fetch.bs
```diff
@@ -3824,6 +3824,47 @@ the request.
...
+
+   <li><p><var>request</var>'s <a for=request>mode</a> is "<code>same-origin</code>",
+   "<code>cors</code>", or "<code>no-cors</code>"
+
+   <li><p><var>request</var>'s <a for=request>window</a> is not null
+
+   <li><p><var>request</var>'s <a for=request>method</a> is `<code>GET</code>`
+
+   <li><p><var>request</var>'s <a for=request>unsafe-request flag</a> is not set, or
+   <var>request</var>'s <a for=request>header list</a> is <a for="list">empty</a>
```
annevk, line 3844:
```suggestion
   <var>request</var>'s <a for=request>header list</a> <a for="list">is empty</a>
```

annevk (commented):
To clarify, this depends on the other PR that merges everything into "fetch finale", right?

noamr:
> To clarify, this depends on the other PR that merges everything into "fetch finale", right?

I don't think it needs to be dependent. Fetch finale clarifies things but doesn't change the behavior of this.
Because the response is already fully loaded into a byte stream by the time it's returned from the "consume" algorithm (marking the response "done"), the current version of fetch finale would work just the same.

fetch.bs
noamr: Actually this line is wrong. It's ok to preload a document.

noamr:
@annevk: another review?

noamr:
@annevk: another review?

fetch.bs
```diff
@@ -3824,6 +3824,45 @@ the request.
...
+  <ol>
+   <li><p>Assert: <var>request</var>'s <a for=request>origin</a> is <var>request</var>'s
+   <a for=request>client</a>'s <a for="environment settings object">origin</a>.
+
+   <li><p>Let <var>onPreloadedResponseAvailable</var> be to an algorithm that runs the following
+   step <a>in parallel</a> given <a for=/>response</a> <var>response</var>: run
+   <a>fetch finale</a> given <var>response</var> and <var>fetchParams</var>.
+
+   <li><p>Let <var>foundPreloadedResource</var> be the result of invoking
+   <a>consume a preloaded resource</a> for <var>req</var>'s <a for=request>window</a>,
```
yoavweiss, line 3856: Is "consume a preloaded resource" defined elsewhere? I'm failing to find it..

noamr: https://html.spec.whatwg.org/#consume-a-preloaded-resource

yoavweiss (commented):
Non-authoritative LGTM

annevk (approved):
Do the changes I made look okay to you?

noamr:
> Do the changes I made look okay to you?

Sure!

annevk:
@noamr where is the work tracked on the "memory cache"? To some extent it was tracked as part of the preload issue before.

noamr:
> @noamr where is the work tracked on the "memory cache"? To some extent it was tracked as part of the preload issue before.

What was tracked before is the relationship between preload and the memory cache. The way it is spec'ed now, the relationship with requests and preloads is simple - if the CORS-mode and URL match then the preloaded resource should be consumed, otherwise ignored.
On WebKit and Gecko this is not the case and I submitted a WPT and opened relevant bugs.

I do believe the memory cache should be better defined, but it's not currently tracked.

annevk:
@noamr could you link the bugs from OP?

noamr:
> @noamr could you link the bugs from OP?

Not sure what OP is.
https://bugzilla.mozilla.org/show_bug.cgi?id=1743764
https://bugs.webkit.org/show_bug.cgi?id=233854

annevk:
It stands for original post(er) in a thread.

PR #1343, opened 2021-10-27
Add abort reason to abort fetch

nidhijaju:
Following on from https://github.com/whatwg/dom/pull/1027, this PR uses the AbortSignal's abort reason in the [abort fetch](https://fetch.spec.whatwg.org/#abort-fetch) algorithm.

@annevk Does this look okay to you?
/cc @yutakahirano, @domenic 

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/35374
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1371359
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1793736
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=246069

***
<a href="https://whatpr.org/fetch/1343.html" title="Last updated on Nov 1, 2021, 5:51 AM UTC (c0833a5)">Preview</a> | <a href="https://whatpr.org/fetch/1343/e01ba65...c0833a5.html" title="Last updated on Nov 1, 2021, 5:51 AM UTC (c0833a5)">Diff</a>

***
<a href="https://whatpr.org/fetch/1343.html" title="Last updated on Oct 5, 2022, 7:24 AM UTC (01a87c0)">Preview</a> | <a href="https://whatpr.org/fetch/1343/1fbc40c...01a87c0.html" title="Last updated on Oct 5, 2022, 7:24 AM UTC (01a87c0)">Diff</a>


ricea:
> > What happens if the signal enters a service worker? We'd have to serialize/deserialize the reason, right?
> 
> That seems ideal, and kind of nice as a developer communication channel. But, the plumbing might get pretty complicated, and it might not be worth the extra spec/implementation/test work. The alternative is just saying that cross-realm aborting is special and we censor any abort reasons to "`AbortError`" `DOMException`s.

What we do for transferring abort reasons in streams is that we try to serialise them, but if the platform doesn't know how to serialise it we end up with an empty object. Since ECMAScript knows how to serialise Error types this works reasonably well for streams. But somehow it seems worse to end up with an empty object as an abort reason in fetch. So I have a preference for censoring it to an `AbortError` on transfer in fetch. I don't know what that would look like in standards language.

annevk:
I guess when either serialization or deserialization ends up throwing we can pass on undefined and then do the undefined -> new "AbortError" DOMException trick. It would be nice if this was aligned with Streams though.

I'm not entirely sure how the plumbing for this would work and that's probably due to the message passing that needs to happen being rather implicit. Thoughts:

1. I guess we need to store it on request. This would also allow the fetch algorithm itself to act on it better. (Ideally we can refactor XMLHttpRequest's timeout stuff to use this underneath.)
2. Then when the request is cloned for the service worker, we need some version of https://dom.spec.whatwg.org/#abortsignal-follow that does cross-thread following. Cross-thread following would be the same as same-thread following, except that it does the serialize/deserialize dance described above. And we'll not worry about the message passing aspects and how they get a handle to each other across threads...

cc @MattiasBuelens

MattiasBuelens:
I think the cleanest way to handle this is making `Request` as a whole `[Transferable]`:
* `Request.signal` would be transferred using `AbortSignal`'s transfer steps.
* `Request.body` would be transferred using `ReadableStream`'s transfer steps.

To transfer an `AbortSignal`, we could use some sort of cross-thread message channel. For example, we could construct a `MessageChannel`, attach an abort algorithm to send a message containing the abort reason through one end, and then call "signal abort" with that reason when we receive it on the other end. This is similar to how we currently transfer streams. (And, also similar to streams, we'll have to be careful around "double-transferring" an `AbortSignal`. 😛)

Making `Request` transferable also allows us to clear up some vagueness around how we pass streams through service workers:
* Right now, [Service Worker's handle fetch](https://w3c.github.io/ServiceWorker/#handle-fetch) says:
    > 24.3.2 Let _requestObject_ be a new `Request` object associated with _request_ and a new associated `Headers` object whose guard is `"immutable"`.

    However, _request_'s body may be a `ReadableStream` created in a different realm. It's never specified how this stream should be accessed from within the _workerRealm_. Really, we should *transfer* the entire `Request` to the worker's realm, including its signal and body.

* When [Fetch's HTTP fetch](https://fetch.spec.whatwg.org/commit-snapshots/09acae88ed9e16b842b20b3d464da22f78443367/#http-fetch) receives a non-null response from the service worker (in step 5.5), it should transfer it from the worker's realm to the current realm.

---

> But somehow it seems worse to end up with an empty object as an abort reason in fetch. So I have a preference for censoring it to an `AbortError` on transfer in fetch. I don't know what that would look like in standards language.

I would find it weird if transferring `Request` and then accessing `.signal` would behave different from transferring `Request.signal` directly. 😕

I guess we could *not* make `Request` and `AbortSignal` transferable from user-land code, and only provide dedicated algorithms to "transfer a request" and to "transfer-receive a request". Then we could do some Fetch-specific censoring in there.

> I guess when either serialization or deserialization ends up throwing we can pass on undefined and then do the undefined -> new "AbortError" DOMException trick. It would be nice if this was aligned with Streams though.

It won't necessarily throw. You may still get back an empty object, because it's merely serializing all the own properties.
```javascript
class MyError {
  get message() {
    return "Oops!";
  }
}
const original = new MyError();
original.message; // -> "Oops!"
const clone = structuredClone(original);
clone.message; // -> undefined
```

annevk:
I think in the case where serialization and deserialization succeed we shouldn't temper with `signal.reason` (i.e., it should be possible for it to be an "empty" object). But when they don't I think we have to temper with it in some form, so we might as well use "`AbortError`" `DOMException`. That seems better than "`DataCloneError`" `DOMException`. (I guess we should test an explicit "`DataCloneError`" `DOMException`, that ought to work. 😊)

I'm not sure about adding public APIs for serializing/transferring `Request`, but anything we do should be compatible with that happening at some point.

yutakahirano:
> What happens if the signal enters a service worker? We'd have to serialize/deserialize the reason, right?

Is this (i.e., aborting without reason) currently specified? https://fetch.spec.whatwg.org/#http-fetch handles a [request](https://fetch.spec.whatwg.org/#requests), but AbortSignal is not on it so I don't know how to tell the fetch is aborted to the service worker.

annevk:
It's not well-defined at the moment, but it's intended to "work". See https://github.com/web-platform-tests/wpt/pull/6484#issuecomment-315775251 for a great explanation by @jakearchibald. There's additional context in #523 and https://github.com/w3c/ServiceWorker/pull/1178. The gist of it is that it is based upon the hand-wavy termination concept and the "aborted flag".

Now, #1329 should improve that setup, but looking at this again it seems that service workers need more attention there. That is, how do we pass the state on to service workers? Perhaps "handle fetch" should get a controller as well?

yutakahirano:
In that case can we keep being hand-wavy about service workers for now, and fix it altogether?

annevk:
Note that currently you can infer from the text what is supposed to happen, albeit not in a great way. E.g., it is somewhat clear the `AbortSignal` that is exposed in the service worker ends up aborted due to a document invoking `abort()`. To what extent would it be clear what that `AbortSignal`'s abort reason ends up being if we don't change the language at all? Only tests?

annevk:
@yutakahirano also note that @noamr is making great progress on #1329 and per our discussion on Matrix we have a general idea of how to forward the termination to service workers. After which this should be relatively straightforward.

yutakahirano:
> Note that currently you can infer from the text what is supposed to happen, albeit not in a great way. E.g., it is somewhat clear the AbortSignal that is exposed in the service worker ends up aborted due to a document invoking abort(). To what extent would it be clear what that AbortSignal's abort reason ends up being if we don't change the language at all? Only tests?

My mental model is for the service worker interception is that you are transferring the request from the initiator context to the service worker. Hence it feels natural that `request.signal` is aborted when `abortController.abort()` is called, and the abort reason is also cloned or transferred, with some edge cases. I'm not sure if this matches with your definition of inference though.

> @yutakahirano also note that @noamr is making great progress on #1329 and per our discussion on Matrix we have a general idea of how to forward the termination to service workers. After which this should be relatively straightforward.

If #1329 will be landed soon I'm fine with waiting for the change.

fetch.bs
```diff
@@ -7494,11 +7499,12 @@ method steps are:
...
 
-<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>, and
-<var>responseObject</var>, run these steps:
+<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>,
+<var>responseObject</var>, and an optional <var>error</var>, run these steps:
 
 <ol>
- <li><p>Let <var>error</var> be an "<code><a exception>AbortError</a></code>" {{DOMException}}.
+ <li><p>If <var>error</var> is not given, let <var>error</var> be an
+ "<code><a exception>AbortError</a></code>" {{DOMException}}.
```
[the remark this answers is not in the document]
saschanaz: But can it ever be undefined? It seems both `AbortController#abort` and static `AbortSignal.abort` sets the reason if undefined.

annevk:
@nidhijaju I think most dependent PRs have landed now. Are you willing to push this over the finish line? Perhaps @noamr can help as he worked on the underlying infrastructure quite a bit.

noamr:
Now that #1329 is merged, I guess we can put the abort reason on the fetch controller, which is available in [handle fetch](https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm) when it aborts its descendant fetches. It would then have to be serialized to the new realm as discussed in comments on this issue. I don't think the whole request needs to be serialized for this reason, though maybe it does for other reasons.

noamr:
Regarding serialization, what makes sense to me is that when the reason has to be sent across realms (when handle fetch aborts its internal fetches), perform the equivalent of a [structuredClone](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserializeinternal) on `reason` since it can be any JS object.

nidhijaju:
@annevk I've made some changes to this PR and started a service worker PR as well (https://github.com/w3c/ServiceWorker/pull/1655). Would you be able to take a look?

fetch.bs
```diff
@@ -231,6 +231,9 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication).
 
  <dt><dfn for="fetch controller">report timing steps</dfn> (default null)
  <dd>Null or an algorithm accepting a <a for=/>global object</a>.
+
+ <dt><dfn for="fetch controller">cloned error</dfn> (default "<code><a exception>AbortError</a></code>" {{DOMException}})
```
noamr, line 235: Maybe `abort reason`, and null as a default?

fetch.bs
```diff
@@ -7793,15 +7802,21 @@ method steps are:
    <li><p><a for=/>Resolve</a> <var>p</var> with <var>responseObject</var>.
   </ol>
 
+ <li><p>Let <var>serializedError</var> be the result of [$StructuredSerialize$]
```
noamr, line 7805: the result of calling. Not sure you can use [$] in the fetch spec

fetch.bs
nidhijaju: Done. I made the default `null`.

fetch.bs
domenic: `[$` should be fine, it's only used for abstract ops, but StructuredSerialize is indeed an abstract op.

noamr: OK then!

fetch.bs
```diff
@@ -231,6 +231,9 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication).
 
  <dt><dfn for="fetch controller">report timing steps</dfn> (default null)
  <dd>Null or an algorithm accepting a <a for=/>global object</a>.
+
+ <dt><dfn for="fetch controller">cloned error</dfn> (default null)
+ <dd>Null, "<code><a exception>AbortError</a></code>" {{DOMException}}, or a Javascript value.
```
domenic, line 236:
```suggestion
 <dd>Null, an "{{AbortError}}" {{DOMException}}, or a JavaScript value.
```

fetch.bs
```diff
@@ -254,8 +257,10 @@ given a <a>fetch controller</a> <var>controller</var>:
 </ol>
 
 <p>To <dfn export for="fetch controller">abort</dfn> a <a for=/>fetch controller</a>
-<var>controller</var>, set <var>controller</var>'s <a for="fetch controller">state</a> to
-"<code>aborted</code>".
+<var>controller</var> with optional <var>error</var>, set <var>controller</var>'s
+<a for="fetch controller">state</a> to "<code>aborted</code>", and set <var>controller</var>'s
+<a for="fetch controller">cloned error</a> to <var>error</var> if given, and otherwise to
+"<code><a exception>AbortError</a></code>" {{DOMException}}.
```
domenic, line 263:
```suggestion
an "{{AbortError}}" {{DOMException}}.
```

fetch.bs
```diff
@@ -7731,12 +7736,15 @@ method steps are:
 
  <li><p>Let <var>request</var> be <var>requestObject</var>'s <a for=Request>request</a>.
 
+ <li><p>Let <var>error</var> be <var>requestObject</var>'s <a for=Request>signal</a>'s
+ <a for=AbortSignal>abort reason</a>.
```
domenic, line 7740: I don't think the modifications to fetch() work as written. Remember, they are running synchronously, while the fetch proceeds in parallel. So consider code like this:

```js
const controller = new AbortController();
const p = fetch(url, { signal: controller.signal });
setTimeout(() => controller.abort(new Error("boo!")), 1000);
```

In your inserted step 4 here, you set _error_ to a clone of `controller.signal.reason`. But `controller.signal.reason` is undefined, because `controller.abort()` is not called yet. So when we get down to step 14+15, we will serialize undefined and store it, which is pretty useless.

What happens if we just remove steps 14+15?

domenic (commented):
So this PR seems to be halfway between the simple version and the complex version.

The simple version here ignores service workers. For that, I think you don't need any serialization/deserialization; in fact, I don't think you need the "cloned error" concept at all. You just need the changes you made to fetch(), where if the signal is already aborted, or if the abort steps for the signal are triggered, you reject the returned promise with the signal's reason, instead of with an "AbortError" DOMException.

The complex version involves updating the `request.signal.reason` seen in a service worker, when the main page does `controller.abort(reason)` from outside the service worker. This requires more plumbing than what you have here; in particular, you need to specify how the message gets sent across realms.

I don't understand @noamr's recommendations about how to use the controller for this. In particular I don't understand how we go from "abort a fetch controller with a `reason` JavaScript object" to "post a message to the service worker realm containing the `Request` representing this request, to signal abort on that `Request`'s [signal](https://fetch.spec.whatwg.org/#request-signal) using a deserialization of that request". Maybe he can suggest how this would best be done...

noamr:
> So this PR seems to be halfway between the simple version and the complex version.
> 
> The simple version here ignores service workers. For that, I think you don't need any serialization/deserialization; in fact, I don't think you need the "cloned error" concept at all. You just need the changes you made to fetch(), where if the signal is already aborted, or if the abort steps for the signal are triggered, you reject the returned promise with the signal's reason, instead of with an "AbortError" DOMException.
> 
> The complex version involves updating the `request.signal.reason` seen in a service worker, when the main page does `controller.abort(reason)` from outside the service worker. This requires more plumbing than what you have here; in particular, you need to specify how the message gets sent across realms.

Why is that necessary? The aborted error is a result of serialization, so by definition it's transferrable between realms, and the main question is when is it serialized. We don't specify how fetch messages the service worker realm to abort in more detail than that today.

> 
> I don't understand @noamr's recommendations about how to use the controller for this. In particular I don't understand how we go from "abort a fetch controller with a `reason` JavaScript object" to "post a message to the service worker realm containing the `Request` representing this request, to signal abort on that `Request`'s [signal](https://fetch.spec.whatwg.org/#request-signal) using a deserialization of that request". Maybe he can suggest how this would best be done...

I think I might have been or still am confused about how this is observable. @nidhijaju, as I said WPTs would go a long way here to explain. Maybe I was proposing a solution for only part of the problem - which is aborting the navigation preload and I was missing a totally separate aspect.

domenic:
> Why is that necessary? The aborted error is a result of serialization, so by definition it's transferrable between realms, and the main question is when is it serialized. We don't specify how fetch messages the service worker realm to abort in more detail than that today.

I'm unsure what you're asking about. Are you questioning the goal of updating `request.signal.reason` for the service worker realm's `Request` object? That seems to be what people were indicating a desire for in, e.g., https://github.com/whatwg/fetch/pull/1343#issuecomment-960474944 and preceding messages.

Or are you asking why you need plumbing to do this? Well, there's nothing in the current PR which updates that value, so that's why I suggested adding spec text to do so...

noamr:
> > Why is that necessary? The aborted error is a result of serialization, so by definition it's transferrable between realms, and the main question is when is it serialized. We don't specify how fetch messages the service worker realm to abort in more detail than that today.
> 
> I'm unsure what you're asking about. Are you questioning the goal of updating `request.signal.reason` for the service worker realm's `Request` object? That seems to be what people were indicating a desire for in, e.g., [#1343 (comment)](https://github.com/whatwg/fetch/pull/1343#issuecomment-960474944) and preceding messages.
> 
> Or are you asking why you need plumbing to do this? Well, there's nothing in the current PR which updates that value, so that's why I suggested adding spec text to do so...

You answered my question, I indeed guided @nidhijaju to solve the wrong problem, though the fetch part of the solution holds.

I believe what we need to do something like the following:
- In fetch, serialize the reason immediately when abort(reason) is called, and save realm-agnostic result in the controller as this PR does
- In [handle fetch](https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm), create an AbortController, and when we create the `Request` object, associate it with that controller's signal
- If the fetch is aborted, also abort the created controller with the fetch controller's serialized abort reason ("cloned error" in this patch).
WDYT?

domenic:
I think that's about right! I didn't realize the fetch controller was shared with the handle fetch algorithm; indeed, with that fact, it becomes clearly the right coordination point.

Minor comment:

> In [handle fetch](https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm), create an AbortController, and when we create the Request object, associate it with that AbortController's signal

Right now in handle fetch, when we create a Request object in step 24.3.2, we never set its signal. But later in 24.3.19, we treat the signal as if it exists. So that's bad. Probably we should update 24.3.2 to use https://fetch.spec.whatwg.org/#request-create so that an AbortSignal is properly created.

Also, in spec land, we don't need to create an AbortController. We can just mess with the signals directly. So just that update should suffice.

So in summary, I think the changes look like this:

- Change "abort fetch" to accept an error (like this patch)
- Change "abort" for fetch controllers to accept an error (like this patch)
- In the "Add the following abort steps", grab signal's reason, serialize it, and store it on the controller. (Similar to this patch's steps 4, 14, and 15, but nested under step 12 instead. Also if serializing throws an exception, probably we should censor to "AbortError" DOMException instead of rejecting p.)
- Rename this patch's "cloned error" to "serialized abort reason" for clarity, and change its type to "null or a serialization from StructuredSerialize".
- Update Service Worker spec:
  - Change 24.3.2 to use "create a Request"
  - Change 24.3.19 to deserialize and use controller's serialized abort reason if it is non-null and deserializing doesn't throw; otherwise a "AbortError" DOMException.

nidhijaju:
Thank you, Noam and Domenic!
I wasn't quite sure how to get the abort reason for cancelAlgorithm, but I've tried to incorporate all other feedback into the latest changes. The ServiceWorker spec PR (https://github.com/w3c/ServiceWorker/pull/1655) has also been updated, and the WPTs (https://github.com/web-platform-tests/wpt/pull/35374) should also be complete now.

Could you take a look to see if this looks better?

fetch.bs
```diff
@@ -7797,15 +7815,24 @@ method steps are:
    <li><p><a for=/>Resolve</a> <var>p</var> with <var>responseObject</var>.
   </ol>
 
+ <li><p>Let <var>error</var> be <var>requestObject</var>'s <a for=Request>signal</a>'s
+ <a for=AbortSignal>abort reason</a> if it not undefined. Otherwise, let <var>error</var> be an
+ "{{AbortError}}" {{DOMException}}.
+
+ <li><p>Let <var>serializedError</var> be the result of calling [$StructuredSerialize$]
```
noamr, line 7822: I don't believe this is the right place to serialize. It should be done in `abort fetch`, not when the fetch is started.

noamr:
> Thank you, Noam and Domenic! I wasn't quite sure how to get the abort reason for cancelAlgorithm, but I've tried to incorporate all other feedback into the latest changes. The ServiceWorker spec PR ([w3c/ServiceWorker#1655](https://github.com/w3c/ServiceWorker/pull/1655)) has also been updated, and the WPTs ([web-platform-tests/wpt#35374](https://github.com/web-platform-tests/wpt/pull/35374)) should also be complete now.
> 
> Could you take a look to see if this looks better?

To me it does look better, thanks! Added comments to both PRs.

fetch.bs
```diff
@@ -254,8 +257,20 @@ given a <a>fetch controller</a> <var>controller</var>:
...
+<var>controller</var> with optional <var>error</var>:
+
+<ol>
+ <li><p>Set <var>controller</var>'s <a for="fetch controller">state</a> to "<code>aborted</code>".
+
+ <li><p>Let <var>error</var> be an "{{AbortError}}" {{DOMException}}, if it is not given.
+
+ <li><p>Let <var>serializedError</var> be the result of calling [$StructuredSerialize$] with
+ <var>error</var>. If that threw and exception, let <var>serializedError</var> be the result of calling
+ [$StructuredSerialize$] with "{{AbortError}}" {{DOMException}}.
```
domenic, line 269:
```suggestion
 <li><p>Let <var>serializedError</var> be [$StructuredSerialize$](<var>error</var>). If that threw an exception, let <var>serializedError</var> be [$StructuredSerialize$](a new  "{{AbortError}}" {{DOMException}}).
```

fetch.bs
```diff
@@ -7785,8 +7804,8 @@ method steps are:
    <li><p>If <var>locallyAborted</var> is true, terminate these substeps.
 
    <li><p>If <var>response</var>'s <a for=response>aborted flag</a> is set, then <a>abort fetch</a>
-   with <var>p</var>, <var>request</var>, and <var>responseObject</var>, and terminate these
-   substeps.
+   with <var>p</var>, <var>request</var>, <var>responseObject</var>, and <var>error</var>, and
```
domenic, line 7807: _error_ is not declared anywhere.

fetch.bs
```diff
@@ -7800,11 +7819,16 @@ method steps are:
...
+<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>,
+<var>responseObject</var>, and an <var>error</var>, run these steps:
 
 <ol>
- <li><p>Let <var>error</var> be an "<code><a exception>AbortError</a></code>" {{DOMException}}.
+ <li><p>Let <var>serializedError</var> be the result of calling [$StructuredSerialize$]
+ with <var>error</var>. If that threw an exception, let <var>serializedError</var> be the result of
+ calling [$StructuredSerialize$] with "{{AbortError}}" {{DOMException}}.
+
+ <li><p>Set <var>controller</var>'s <a for="fetch controller">serialized abort reason</a> to
```
domenic, line 7830: _controller_ is not declared.

Overall, I don't *think* these steps are necessary:

- If this is called from fetch() 4.1, then we are doing an early-abort; nothing will ever reach the server worker, and no controller will be involved.

- If this is called from fetch 11.2, then step 11.3 will take care of the serialization and controller-storage.

- If this is called from fetch 12.2, then... well, probably this case we do need to modify the controller, but maybe we should do it like step 11.2 + 11.3 and keep them separate? I.e. if response's aborted flag is set, then: 1. create a new "AbortError" DOMException; 2. "abort fetch"; 3. "abort controller"?

fetch.bs
```diff
@@ -7800,11 +7819,16 @@ method steps are:
  <li><p>Return <var>p</var>.
 </ol>
 
-<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>, and
-<var>responseObject</var>, run these steps:
+<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>,
```
domenic, line 7822: It would help a lot to rename this something like "abort a fetch() call". I keep getting it confused with "abort" as used on fetch controllers, which is more general; this is about messing with the JavaScript objects specifically vended by fetch().

domenic (commented):
Sorry for the long delay. I think this is on the right path; the overall structure makes sense to me now, including the service worker PR.

I will endeavor to do future reviews within 24 hours :).

fetch.bs
nidhijaju: That makes sense to me! I've added some substeps to fetch 12.2 to address that.

fetch.bs
nidhijaju: Renamed :)

fetch.bs
```diff
@@ -254,8 +257,20 @@ given a <a>fetch controller</a> <var>controller</var>:
...
-"<code>aborted</code>".
+<var>controller</var> with optional <var>error</var>:
+
+<ol>
+ <li><p>Set <var>controller</var>'s <a for="fetch controller">state</a> to "<code>aborted</code>".
+
+ <li><p>Let <var>error</var> be an "{{AbortError}}" {{DOMException}}, if it is not given.
+
+ <li><p>Let <var>serializedError</var> be [$StructuredSerialize$](<var>error</var>).
+ If that threw an exception, let <var>serializedError</var> be
```
domenic, line 268:
```suggestion
 If that threw an exception, catch it, and let <var>serializedError</var> be
```

fetch.bs
```diff
@@ -7800,12 +7831,10 @@ method steps are:
  <li><p>Return <var>p</var>.
 </ol>
 
-<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>, and
-<var>responseObject</var>, run these steps:
+<p>To <dfn lt="Abort the fetch() call" export id=abort-fetch-algorithm>abort a fetch() call</dfn> with a
```
domenic, line 7834:
```suggestion
<p>To <dfn lt="abort the fetch() call" export id=abort-fetch-algorithm>abort a fetch() call</dfn> with a
```

domenic (approved):
LGTM with nits, very nice!

nidhijaju:
Hi @annevk, would you be able to take a look at this when you have time?

fetch.bs
```diff
@@ -231,6 +231,9 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication).
 
  <dt><dfn for="fetch controller">report timing steps</dfn> (default null)
  <dd>Null or an algorithm accepting a <a for=/>global object</a>.
+
+ <dt><dfn for="fetch controller">serialized abort reason</dfn> (default null)
+ <dd>Null or a serialization from [$StructuredSerialize$].
```
annevk, line 236: Note to self: might want to add a source comment here about this being a Record.

fetch.bs
```diff
@@ -254,8 +257,20 @@ given a <a>fetch controller</a> <var>controller</var>:
 </ol>
 
 <p>To <dfn export for="fetch controller">abort</dfn> a <a for=/>fetch controller</a>
-<var>controller</var>, set <var>controller</var>'s <a for="fetch controller">state</a> to
-"<code>aborted</code>".
+<var>controller</var> with optional <var>error</var>:
```
annevk, line 260:
```suggestion
<var>controller</var> with an optional <var>error</var>:
```

fetch.bs
```diff
@@ -254,8 +257,20 @@ given a <a>fetch controller</a> <var>controller</var>:
...
+
+<ol>
+ <li><p>Set <var>controller</var>'s <a for="fetch controller">state</a> to "<code>aborted</code>".
+
+ <li><p>Let <var>error</var> be an "{{AbortError}}" {{DOMException}}, if it is not given.
```
annevk, line 265:
```suggestion
 <li><p>Set <var>error</var> to an "{{AbortError}}" {{DOMException}} if it is not given.
```

fetch.bs
```diff
@@ -7800,12 +7831,10 @@ method steps are:
  <li><p>Return <var>p</var>.
 </ol>
 
-<p>To <dfn>abort fetch</dfn> with a <var>promise</var>, <var>request</var>, and
-<var>responseObject</var>, run these steps:
+<p>To <dfn lt="abort the fetch() call" export id=abort-fetch-algorithm>abort a fetch() call</dfn> with a
```
annevk, line 7834: We should keep the existing ID `abort-fetch` here. Also, if we rename this to include the name of the method, `fetch()` should be inside `<code>`.

fetch.bs
```diff
@@ -7784,9 +7803,21 @@ method steps are:
   <ol>
    <li><p>If <var>locallyAborted</var> is true, terminate these substeps.
 
-   <li><p>If <var>response</var>'s <a for=response>aborted flag</a> is set, then <a>abort fetch</a>
-   with <var>p</var>, <var>request</var>, and <var>responseObject</var>, and terminate these
-   substeps.
+   <li><p>If <var>response</var>'s <a for=response>aborted flag</a> is set, then:
```
annevk, line 7806: In the actual fetch algorithm we also have this check and there is this line:

> If stream is [readable](https://streams.spec.whatwg.org/#readablestream-readable), [error](https://streams.spec.whatwg.org/#readablestream-error) stream with an "[AbortError](https://webidl.spec.whatwg.org/#aborterror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

I suspect that needs to be updated as well to use the abort reason on the controller?

It seems that the flow of this should be something like:

1. Website invokes `fetch()` with a signal.
2. Website aborts signal.
3. Signal is used in the fetch algorithm to terminate the actual fetch and cancel streams and such using the abort reason.
4. `fetch()` API observes the fetch algorithm being terminated and exposes the abort reason.

More concretely, why don't we use the abort reason on the controller in the steps below? I don't think we should end up in a situation where a response is aborted but the controller isn't yet.

annevk (commented):
Thanks so much for pushing this forward! I think there are still some things to improve with how `fetch()` and the fetch algorithm interact, but overall this is looking rather good.

fetch.bs
nidhijaju: Good point, I've updated the erroring stream bit.

With regards to the latter part, I guess the reason why we don't use the abort reason on the controller here is because we need to abort the controller first which we do in Step 12.2.4 which in turn sets the serialized abort reason on the controller. Do you mean more that the controller does not need to be aborted since it should already be aborted at that point?

fetch.bs
```diff
@@ -5789,8 +5805,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
        <li><p>Set <var>response</var>'s <a for=response>aborted flag</a>.
 
        <li><p>If <var>stream</var> is <a for=ReadableStream>readable</a>,
-       <a for=ReadableStream>error</a> <var>stream</var> with an
-       "<code><a exception>AbortError</a></code>" {{DOMException}}.
+       <a for=ReadableStream>error</a> <var>stream</var> with <var>fetchParams</var>'s
+       <a for="fetch params">controller</a>'s <a for="fetch controller">serialized abort reason</a>.
```
annevk, line 5809: We'll need to deserialize this first.

fetch.bs
annevk: Yeah, once the aborted flag is set the controller should also be aborted. I hope those cannot get out-of-sync... cc @noamr

nidhijaju: I see, in that case, I have updated the steps to use the controller's abort reason now. Hope it looks better!

annevk (commented):
Thank you! I'd appreciate another look from @domenic and @noamr (if available) and I think we should address the null-vs-undefined issue and we should probably introduce a shared algorithm given that we do the same thing three times, but otherwise this seems good to go.

fetch.bs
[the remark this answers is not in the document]
nidhijaju: Fair point, I've changed it to be undefined by default, instead of null, like AbortSignal's abort reason.

fetch.bs
```diff
@@ -7796,10 +7824,12 @@ method steps are:
   <ol>
    <li><p>Set <var>locallyAborted</var> to true.
 
-   <li><p><a>Abort fetch</a> with <var>p</var>, <var>request</var>, and <var>responseObject</var>.
+   <li><p><a>Abort the fetch() call</a> with <var>p</var>, <var>request</var>, <var>responseObject</var>, and
```
domenic, line 7827:
```suggestion
   <li><p><a>Abort the <code>fetch()</code> call</a> with <var>p</var>, <var>request</var>, <var>responseObject</var>, and
```

here and elsewhere

fetch.bs
```diff
@@ -254,8 +257,20 @@ given a <a>fetch controller</a> <var>controller</var>:
...
+<var>controller</var> with an optional <var>error</var>:
+
+<ol>
+ <li><p>Set <var>controller</var>'s <a for="fetch controller">state</a> to "<code>aborted</code>".
+
+ <li><p>Set <var>error</var> to an "{{AbortError}}" {{DOMException}} if it is not given.
+
+ <li><p>Let <var>serializedError</var> be [$StructuredSerialize$](<var>error</var>).
+ If that threw an exception, catch it, and let <var>serializedError</var> be
+ [$StructuredSerialize$](a new "{{AbortError}}" {{DOMException}}).
```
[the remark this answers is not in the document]
domenic: I'm not sure what this refers to... how would StructuredSerialize() return undefined?

domenic (approved):
LGTM with nit!

annevk:
Apologies, my last feedback round was misguided. There was no reason to switch from null to undefined as the null is not exposed to JS. And as @domenic noted serialization and deserialization are different operations so at most two algorithms could share more code.

So I've reverted the null to undefined change. I have implemented some code sharing for deserialization and as part of that I found a bug. The networking layer doesn't have a "this" so for now I've gone with an implementation-defined realm and a comment that we really need to clean all of that up in the future.

I think it's good to go now though. Anyone any final feedback on this? It seems the tests are ready to land and I think we can assume implementation support as this was always part of the plan.

nidhijaju:
> I think it's good to go now though. Anyone any final feedback on this? It seems the tests are ready to land and I think we can assume implementation support as this was always part of the plan.

Looks good to me -- thank you, @annevk!
Yes, the service worker and test PRs are ready to land, and I've also filed and linked the implementation bugs to the PR description.

noamr:
I found some time to look at it again, another LGTM From me.

annevk:
Thanks all!

Issue #1335, opened 2021-10-15
propagate request tainted origin flag and/or URL list through service workers


wanderview:
Similar to #1321, we also need to propagate the request [URL list](https://fetch.spec.whatwg.org/#concept-request-url-list) through a service worker `evt.respondWith(fetch(evt.request))`.  This is necessary to send the proper values for origin, sec-fetch-site, samesite cookies, etc when a request has redirected through a cross-origin url.

This is related to the collection of topics discussed at w3c/ServiceWorker#1604.

annevk:
I think turning it into an algorithm that operates on request's origin and URL list, and returns "same-origin", "same-site", or "cross-site" is probably reasonable. This would require adjusting the logic in "CORS-preflight fetch" some to clone the URL list.

wanderview:
So I started working on a pull request to remove the `tainted origin flag` and replace it with an algorithm that operates on the `URL list`.  I realized, though, we can't propagate the `URL list` to a CORS preflight without side effects.  Mainly, it would reduce the redirect count permitted for preflights by however large the original request's `URL list` is.

So I guess for now the easiest thing to do is propagate both the `tainted origin flag` and the `URL list` when constructing an unmodified copy of another request.

annevk:
@wanderview I don't think that matters as a preflight response isn't allowed to be a redirect. It has to be a 2xx.

wanderview:
Hmm, I looked to see if it used a redirect mode of `error`, but didn't see it.  I see, though, that its done as a separate check in step 7 of:

https://fetch.spec.whatwg.org/#cors-preflight-fetch-0

For some reason I thought http-network-or-cache-fetch would follow the redirect.

Ok, I'll try changing back to removing the tainted origin flag.

wanderview:
Note, I think we will also need a change to html spec to make navigation populate the request `URL list` when it manually follows redirects.

annevk:
HTML invokes https://fetch.spec.whatwg.org/#navigate-redirect-fetch which invokes https://fetch.spec.whatwg.org/#concept-http-redirect-fetch which appends to request's URL list.

PR #1345, opened 2021-10-29
Propagate origin and URL list through the request constructor. (Fixes #1321, #1335)

### LINKED ISSUE #1321: Request constructor should propagate origin internal field
Currently, the Request constructor always resets the request `origin` internal field to "client".  See step 12 in:

https://fetch.spec.whatwg.org/#dom-request

This is problematic because a pass-through FetchEvent SW handler will end up using the service worker's origin for things like:

* Origin header
* Sec-Fetch-Site header
* SameSite cookies

It would be safer for sites if we propagated the internal `origin` field from the original request.  We also need to propagate the `tainted origin flag` as well in order to properly handle redirects.

This was discussed at the virtual F2F (w3c/ServiceWorker#1604).  [Notes](https://docs.google.com/document/d/1q74ATke1oG2jEval1yBTVIrXZGItFi6DR5l8n_8ybgs/edit#bookmark=id.bg7ds8878foi) are available.

wanderview:
   * Chromium
   * Gecko
   * [Navigation headers test](https://github.com/web-platform-tests/wpt/blob/master/service-workers/service-worker/navigation-headers.https.html)
   * [SameSite cookies test](https://github.com/web-platform-tests/wpt/blob/master/service-workers/service-worker/same-site-cookies.https.html)
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1241188
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1658869
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=233640

***
<a href="https://whatpr.org/fetch/1345.html" title="Last updated on Dec 1, 2021, 10:09 AM UTC (8d0d199)">Preview</a> | <a href="https://whatpr.org/fetch/1345/5dc54a7...8d0d199.html" title="Last updated on Dec 1, 2021, 10:09 AM UTC (8d0d199)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6954,10 +6954,8 @@ constructor steps are:
    <li><p>Set <var>request</var>'s <a for=request>URL</a> to <var>request</var>'s
    <a for=request>current URL</a>.
 
-   <li><p>Set <var>request</var>'s <a for=request>URL list</a> to an empty <a for=/>list</a>.
-
-   <li><p><a for=list>Append</a> <var>request</var>'s <a for=request>URL</a> to
-   <var>request</var>'s <a for=request>URL list</a>.
+   <li><p>Set <var>request</var>'s <a for=request>URL list</a> to « <var>request</var>'s
+   <a for=request>URL</a> ».
   </ol>
 
   <p class=note>This is done to ensure that when a service worker "redirects" a request, e.g., from


annevk:
I don't think we should do this without addressing #1335 as we'd introduce a new hole.

wanderview:
Sorry, can you clarify how that would be a new hole?  It seems like the existing hole is just not fully closed without #1335.

annevk:
Well, the current model is somewhat self-consistent with the authority being the service worker. This makes the authority the initiator of the fetch, but it's not a complete fix for that.

wanderview:
I understand that at a conceptual level, but working incrementally to fix an issue has practical advantages.  Anyway, happy to just let this sit until the rest is done if you are.

wanderview:
Ok, I've updated this to also do the propagation of the origin tainted flag and URL list.  PTAL @annevk.  Thanks.

wanderview:
See #1357 for the removal of the tainted origin flag.  I'll update this PR if/when that one lands.

wanderview:
This has been updated now that `tainted origin flag` has been removed.  PTAL.

fetch.bs
```diff
@@ -6925,6 +6925,9 @@ constructor steps are:
 
    <dt><a for=request>history-navigation flag</a>
    <dd><var>request</var>'s <a for=request>history-navigation flag</a>.
+
+   <dt><a for=request>URL list</a>
+   <dd>A <a for=list>clone</a> of <var>request</var>'s <a for=request>URL list</a>.
```
annevk, line 6930: Because of this we should remove the setting of request's URL above, right?

fetch.bs
```diff
@@ -6939,10 +6942,19 @@ constructor steps are:
...
    <li><p>Set <var>request</var>'s <a for=request>referrer</a> to
    "<code>client</code>"
 
    <li><p>Set <var>request</var>'s <a for=request>referrer policy</a> to the empty string.
+
+   <li><p>Set <var>request</var>'s <a for=request>URL list</a> to an empty
+   <a for=/>list</a>.
+
+   <li><p><a for=list>Append</a> <var>request</var>'s <a for=request>URL</a> to
+   <var>request</var>'s <a for=request>URL list</a>.
```
annevk, line 6957: I'm not sure how this works. Request's URL is a pointer to the first item of request's URL list. It seems you would have to store it in a variable before you reset request's URL list, right? But also, you probably want the last item in request's URL list? I.e., request's current URL.

fetch.bs
```diff
@@ -6894,7 +6894,7 @@ constructor steps are:
    <dd><var>request</var>'s <a for=request>priority</a>.
 
    <dt><a for=request>origin</a>
-   <dd>"<code>client</code>".
+   <dd><var>request</var>'s <a for=request>origin</a>.
```
annevk, line 6897: Please add a note here about how particular the situation is where this is relevant. Probably with `<span class=note>` as we don't seem to have paragraphs here.

fetch.bs
wanderview: Request's URL is not defined as a pointer into the URL list:

https://fetch.spec.whatwg.org/#concept-request-url

Instead URL list is defined to hold a copy of the request's URL for its first entry:

https://fetch.spec.whatwg.org/#concept-request-url-list

So I think we need to still set request's url, but it should be the first url in the url list.  We can do this by setting the value as a clone of the input's request url (since that is the same as the first entry of url list).

fetch.bs
wanderview: As I linked above, its not defined that way.  The first entry in the URL list contains a copy of the request URL.

You are right, though, I need to use the current URL instead.

fetch.bs
wanderview: Done.

fetch.bs
wanderview: I handled this by storing the current URL in the request's URL and then appending to the new list.

wanderview:
Updated based on previous feedback.  PTAL.  Thanks.

annevk (approved):
@wanderview any thoughts on my nits? I think this is good to go, modulo bugs.

Thanks a lot for driving this! I think we ended up with a pretty decent solution all things considered.

wanderview:
Webkit bug: https://bugs.webkit.org/show_bug.cgi?id=233640

Issue #973, opened 2019-11-28
Use case for Headers getAll


xtuc:
Since https://github.com/whatwg/fetch/commit/42464c8c3d2fd3437a19fc6afd2438a0fd42dde8 `Headers.prototype.getAll` has been deprecated/removed from the spec and implementations.

I understand that in browsers (including service workers) filtered responses don't include headers that could benefit from the getAll method.
However, some JavaScript environment doesn't need to filter response/request headers, like serverless plateforms (for instance Cloudflare Workers) or maybe Nodejs at some point.

For example editing Cookie headers:
```js
const h = new Headers;
h.append("Set-Cookie", "a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT");
h.append("Set-Cookie", "b=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT");

h.get("Set-Cookie")
// a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT, b=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT
```

Spliting the combined header value by `, ` will give an invalid result. Instead `getAll` could be used to retrieve the individual headers.

annevk:
`Set-Cookie` is the only such header, see also HTTP and #506.

It might make sense to expose it properly for code that wants to use this class generically, but I think we ought to allow for implementations of this class to eagerly combine headers generally and store `Set-Cookie` separately as a special case. That would suggest something like `h.getSetCookie()` to me.

cc @youennf @ddragana @yutakahirano

annevk:
It's probably also worth investigating if https://www.npmjs.com/package/set-cookie-parser#splitcookiestringcombinedsetcookieheader is a reasonable alternative to expose. Though if that actually works HTTP would not have to consider cookies a special case...

xtuc:
Maybe to clarify the environment a bit, Node.js and Cloudflare Workers both boils down to a proxy written in JavaScript.

> Set-Cookie is the only such header

For such a proxy, it make sense to not filter headers in the response and let the user define its own headers. Any custom headers may contain `, `.

annevk:
I disagree, HTTP doesn't guarantee intermediaries won't combine such headers so we shouldn't either. I'd want HTTP to call out more than `Set-Cookie` as requiring a special data model before going there.

domenic:
[...]

PR #1346, opened 2021-10-29
Add special handling of `set-cookie` to Headers

lucacasonato:
Towards #973

   * Firefox (https://github.com/whatwg/fetch/issues/973#issuecomment-954815921)
   * Deno (https://github.com/whatwg/fetch/issues/973#issuecomment-902578584)
   * Chrome (https://github.com/whatwg/fetch/pull/1346#issuecomment-1404577750)
   * Cloudflare Workers
   * Node.js
   * https://github.com/web-platform-tests/wpt/pull/31442
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/1409512
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1812511
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=251194
   * Deno: https://github.com/denoland/deno/pull/13542#issuecomment-1419946439

***
<a href="https://whatpr.org/fetch/1346.html" title="Last updated on Feb 6, 2023, 11:39 PM UTC (9e950f1)">Preview</a> | <a href="https://whatpr.org/fetch/1346/093c622...9e950f1.html" title="Last updated on Feb 6, 2023, 11:39 PM UTC (9e950f1)">Diff</a>


fetch.bs
```diff
@@ -5960,18 +5977,24 @@ interface Headers {
   undefined append(ByteString name, ByteString value);
   undefined delete(ByteString name);
   ByteString? get(ByteString name);
+  sequence&ltByteString> getSetCookie();
```
annevk, line 5980:
```suggestion
  sequence&lt;ByteString> getSetCookie();
```

fetch.bs
```diff
@@ -732,12 +732,29 @@ a <a for=/>header list</a> <var>list</var>, run these steps:
   <p><a for=list>For each</a> <var>name</var> in <var>names</var>:
 
   <ol>
-   <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var>
-   from <var>list</var>.
+   <li><p>If <var>name</var> is "<code>set-cookie</code>", then:
+    
+    <ol>
+     <li><p>Let <var>values</var> be a list of all <a for=/>headers</a> in <var>list</var> whose
+      <a for=header>name</a> is a <a>byte-case-insensitive</a> match for <var>name</var>, in order.
```
annevk, line 739: be a list of all values of headers*?

fetch.bs
```diff
@@ -732,12 +732,29 @@ a <a for=/>header list</a> <var>list</var>, run these steps:
   <p><a for=list>For each</a> <var>name</var> in <var>names</var>:
 
   <ol>
-   <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var>
-   from <var>list</var>.
+   <li><p>If <var>name</var> is "<code>set-cookie</code>", then:
```
annevk, line 735:
```suggestion
   <li><p>If <var>name</var> is `<code>set-cookie</code>`, then:
```

fetch.bs
```diff
@@ -6018,6 +6041,9 @@ new Headers(meta2);
  <dd><p>Returns as a string the values of all headers whose name is <var>name</var>, separated by a
  comma and a space.
 
+ <dt><code><var>headers</var> . <a method for=Headers lt=getSetCookie()>getSetCookie</a>()</code>
+ <dd><p>Returns a list of the values for all headers whose name is <code>set-cookie</code>.
```
annevk, line 6045:
```suggestion
 <dd><p>Returns a list of the values for all headers whose name is `<code>Set-Cookie</code>`.
```

fetch.bs
```diff
@@ -6160,6 +6186,14 @@ method steps are to <a for=Headers>append</a> (<var>name</var>, <var>value</var>
  <a for=Headers>header list</a>.
 </ol>
 
+<p>The <dfn export for=Headers method><code>getSetCookie()</code></dfn> method steps are:
+
+<ol>
+ <li><p>Return <a lt=value for=header>values</a> of all <a for=/>headers</a> in <a>this</a>'s
+ <a for=Headers>header list</a> whose <a for=header>name</a> is a <a>byte-case-insensitive</a> match
+ for `<code>set-cookie</code>`, in order.
+</ol>
```
annevk, line 6195: This needs a step that returns an empty list when there are no such headers.

The second step should say "Return the values ... Set-Cookie ..." (in particular "the" and casing).

fetch.bs
```diff
@@ -5960,18 +5977,24 @@ interface Headers {
...
+`<code>Set-Cookie</code>` header, as unlike all other headers it can not be safely combined. For
+client-side JavaScript this is not very problematic, as the platform never exposes
+`<code>Set-Cookie</code>`. This can be problematic for server-side runtimes that do expose this
+header. Because of this there is some special handling for the `<code>Set-Cookie</code>` header in
+the implementation of the {{Headers}} interface: `<code>Set-Cookie</code>` headers are not
+concatenated when they are returned from the {{Headers}} iterator. This means that specifically for
+this header, multiple tuples of the same name can be returned. Implementations could choose the more
+efficient {{Headers}} object representation even for a <a for=/>header list</a>, as long as they
+also support an associated data structure for
 `<code>Set-Cookie</code>` headers.
```
annevk, line 5998: It's not true that client-side JavaScript never exposes Set-Cookie. That's only true in the general case.

So I don't think we need the client-side vs server-side exposition here.

We should note that we special-case Set-Cookie, just like HTTP. And that this alternative representation could be used for header lists in theory.

annevk:
Thanks for working on this, this looks quite promising to me. @yutakahirano @ricea @youennf thoughts?

fetch.bs
lucacasonato: > This needs a step that returns an empty list when there are no such headers.

Why? `[].filter((header) => header.name === "set-cookie").length === 0`. Does filtering on lists in specs not return an empty list for no matches?

annevk: That's an interesting idea, but I'm not sure from what it would follow. The way I read the prose is that it's not accounted for what happens when there is nothing named Set-Cookie.

fetch.bs
lucacasonato: > It's not true that client-side JavaScript never exposes Set-Cookie

Well the _platform_ never exposes a set-cookie header (because it is a forbidden header) if I understand correctly. Only ever user code.

I've updated the prose and removed the client-side & server-side comparison, and mentioned the HTTP spec does the same..

fetch.bs
```diff
@@ -6160,6 +6188,17 @@ method steps are to <a for=Headers>append</a> (<var>name</var>, <var>value</var>
  <a for=Headers>header list</a>.
 </ol>
 
+<p>The <dfn export for=Headers method><code>getSetCookie()</code></dfn> method steps are:
+
+<ol>
+ <li><p>If <var>list</var> <a for="header list">does not contain</a> `<code>Set-Cookie</code>`, then
```
andreubotella, line 6194:
```suggestion
 <li><p>If <a>this</a>'s <a for=Headers>header list</a> <a for="header list">does not contain</a> `<code>Set-Cookie</code>`, then
```

lucacasonato:
@jasnell @kentonv thoughts?

fetch.bs
```diff
@@ -6160,6 +6188,17 @@ method steps are to <a for=Headers>append</a> (<var>name</var>, <var>value</var>
  <a for=Headers>header list</a>.
 </ol>
 
+<p>The <dfn export for=Headers method><code>getSetCookie()</code></dfn> method steps are:
+
+<ol>
+ <li><p>If <a>this</a>'s <a for=Headers>header list</a> <a for="header list">does not contain</a> `<code>Set-Cookie</code>`, then
```
yutakahirano, line 6194: Do we need this step?

fetch.bs
```diff
@@ -732,12 +732,30 @@ a <a for=/>header list</a> <var>list</var>, run these steps:
...
-   <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var>
-   from <var>list</var>.
+   <li><p>If <var>name</var> is `<code>set-cookie</code>`, then:
+    
+    <ol>
+     <li><p>Let <var>values</var> be a list of all <a lt=value for=header>values</a> of
+      <a for=/>headers</a> in <var>list</var> whose <a for=header>name</a> is a
+      <a>byte-case-insensitive</a> match for <var>name</var>, in order.
+
+     <li><p><a for=list>For each</a> <var>value</var> of <var>values</var>:
```
yutakahirano, line 742: There needs to be a line break between `<li>` and `<p>`. Please see https://github.com/whatwg/fetch#formatting.

fetch.bs
```diff
@@ -732,12 +732,30 @@ a <a for=/>header list</a> <var>list</var>, run these steps:
   <p><a for=list>For each</a> <var>name</var> in <var>names</var>:
 
   <ol>
-   <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var>
-   from <var>list</var>.
+   <li><p>If <var>name</var> is `<code>set-cookie</code>`, then:
+    
+    <ol>
+     <li><p>Let <var>values</var> be a list of all <a lt=value for=header>values</a> of
+      <a for=/>headers</a> in <var>list</var> whose <a for=header>name</a> is a
```
yutakahirano, line 739: https://github.com/whatwg/fetch#formatting

"Do not indent for anything except a new "block" element."

yutakahirano (approved):
LGTM.

jasnell (approved):
LGTM. For Workers we do currently take a different approach using `getAll()` that we'll need to maintain for a while but there's nothing here that should interfere with us being able to support this as well.

fetch.bs
lucacasonato: @annevk says so (I don't understand why though). https://github.com/whatwg/fetch/pull/1346#discussion_r740067545

kentonv:
I'm fine with this.

I have a weak opinion that, for server-side vendors, `getAll()` is slightly more flexible: If we have some big customer come along and say "We really need to access this legacy API, but it has a header that misuses commas, which we can't change," we can easily say "No problem, we'll just add your special weird header to the `getAll()` allow list." On the server side, implementing special requests for high-paying customers are not uncommon. Obviously, on the browser side, such special requests would be complete nonsense, since web sites don't get to choose which browser is used to open them.

But I have seen no evidence that any such headers exist in the wild, so it seems unlikely we'll ever get this exact sort of special request. So, whatever. :)

annevk:
It's more flexible, but it would also violate HTTP, as would any header that needs it (with the exception of `Set-Cookie`).

jimmywarting:
I'm not opposed to adding this into node-fetch either. I will only be glad that we can unite on one unified api that works the same across all platform

mcollina:
Node.js is ok in implementing this too.

fetch.bs
annevk: I don't think it's defined behavior what the next step ends up returning otherwise. E.g., things like "in order" expect a list. At the very least we'd have to add "if any" somewhere in there, but I think this is much clearer.

However, this step does need some rewrapping to account for a 100 columns. I also think this PR in general could do with some rebasing.

annevk (commented):
I just realized that an alternative way of solving the request `Headers` representation issue would be to always return an empty list for `getSetCookie()` there. Then we don't really have to forbid adding it, but I think what we ended up with is fine.

shadiramadan:
bump - I upgraded node and realized I can't because `.raw` was removed and I need to query multiple set cookie headers and that was the only way I was able to.

fetch.bs
```diff
@@ -5960,18 +5979,25 @@ interface Headers {
...
+`<code>Set-Cookie</code>` header, as unlike all other headers, it cannot be safely combined. Because
+of this, there is some special handling for the `<code>Set-Cookie</code>` header in
+the implementation of the {{Headers}} interface, just like in the HTTP spec.
+`<code>Set-Cookie</code>` headers are not concatenated when they are returned from the {{Headers}}
+iterator, and there is a
+<code><var>headers</var> . <a method for=Headers lt=getSetCookie()>getSetCookie</a>()</code> method
+to retrieve the list of all `<code>Set-Cookie</code>`, ordered and uncombined. This means that
+specifically for this header, multiple tuples of the same name can be returned from the {{Headers}}
+iterator. Implementations could choose the more efficient {{Headers}} object representation even for
+a <a for=/>header list</a>, as long as they also support an associated data structure for
```
andreubotella, line 6000: Before this PR, header lists had to store `Set-Cookie` header values separately, since cookie parsing is a step in the HTTP-network fetch algorithm which uses header lists, but `Headers` objects could combine them. Now there is no difference in the handling of `Set-Cookie` between header lists and `Headers` objects.

This note should therefore be rewritten, since now there is no "more efficient" representation for `Headers` that doesn't work for header lists. It should also probably be moved to the definition of header list.

andreubotella:
Hi. We're trying to bring this PR back to life and get it landed as part of the WinterCG effort. It seems like the semantics are clear and no more work is needed there. So it comes down to implementer interest, I suppose.

@yutakahirano, @youennf. Is Firefox still interested, @smaug----?

yutakahirano:
cc: @ricea for Chrome

ricea:
Chrome supports if there is consensus. I don't want to be the only browser implementing this.

andreubotella:
I've filed implementation bugs for all browsers:
- Chromium: https://crbug.com/1409512
- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1812511
- WebKit: https://bugs.webkit.org/show_bug.cgi?id=251194

I've also submitted implementations for Chromium and Firefox, and I'm currently working on one for WebKit.

youennf:
I won't object to this addition but I do not see a lot of value for web browsers to support this.
Are there shims around `Headers` that have been deployed already?

jasnell:
With the case of the `keys()` iterator here, would the intention of this PR be to make it so that `set-cookie` could appear multiple times? Or should the `keys()` iterator always only produce a unique set? The way this is currently defined, it would appear that the keys iterator could have `set-cookie` multiple times.

andreubotella:
> With the case of the `keys()` iterator here, would the intention of this PR be to make it so that `set-cookie` could appear multiple times? Or should the `keys()` iterator always only produce a unique set? The way this is currently defined, it would appear that the keys iterator could have `set-cookie` multiple times.

That seems intentional in the way pair iteration works in WebIDL. That way, `headers.keys().length === headers.values().length` and the indices match.

Do you think this would break some existing code, though?

andreubotella:
> I do think there's a risk of breaking existing code that may be relying on the iterator producing a unique set of keys.

Deno's current implementation of `headers.keys()` will yield `"set-cookie"` repeatedly. Although @lucacasonato can speak more on it, this doesn't seem to have caused any breakage in practice.

andreubotella:
I have fixed this PR's merge conflicts and rewritten and move the `Headers` implementation note to be about header lists. PTAL.

annevk (commented):
Thanks, this looks good. I like the improved note and its placement. Remaining nits are all editorial.

Given the ranges of agreement expressed above I suggest we merge this February 8 and see how it goes.

PR #1351, opened 2021-11-09
Send "null" Origin header on cross-origin .onion requests

### LINKED ISSUE #1350: Send "null" Origin headers on cross-origin requests from an RFC7686 address
The special-use `.onion` domain name, defined in [RFC7686](https://datatracker.ietf.org/doc/rfc7686/), receives special treatment in the Tor browser and in Firefox: https://searchfox.org/mozilla-central/rev/6c8d325e61b0b445ed2e04899da38c3a4c266cba/netwerk/protocol/http/nsCORSListenerProxy.cpp#979-984

It seems like this behavior should be standardized since any browser could be setup to proxy traffic over the Tor SOCKS5 proxy.

I also filed an [issue against the Referrer Policy spec](https://github.com/w3c/webappsec-referrer-policy/issues/155) to address the Referrer header.

fmarier:
Fixes #1350.

   * [Firefox (off by default) / Tor Browser (on by default)](https://searchfox.org/mozilla-central/rev/6c8d325e61b0b445ed2e04899da38c3a4c266cba/netwerk/protocol/http/nsCORSListenerProxy.cpp#979-984)
   * [Brave](https://github.com/brave/brave-browser/issues/18071)
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1351.html" title="Last updated on Nov 10, 2021, 11:10 PM UTC (b4aebfd)">Preview</a> | <a href="https://whatpr.org/fetch/1351/3ecab20...b4aebfd.html" title="Last updated on Nov 10, 2021, 11:10 PM UTC (b4aebfd)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -111,6 +111,9 @@ urlPrefix:https://w3c.github.io/hr-time/#;spec:hr-time
         "href": "https://datatracker.ietf.org/doc/html/draft-ietf-masque-h3-datagram",
         "publisher": "IETF",
         "title": "Using QUIC Datagrams with HTTP/3"
+    },
+    "ONION": {
+        "aliasOf": "RFC7686"
     }
 }
 </pre>
@@ -2819,6 +2822,11 @@ given a <a for=/>request</a> <var>request</var>, run these steps:
 <ol>
  <li><p>Let <var>serializedOrigin</var> be the result of <a>byte-serializing a request origin</a>
  with <var>request</var>.
+
+ <li><p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>'s
+ <a for=origin>host</a> ends with "<code>.onion</code>" or "<code>.onion.</code>", and
+ is not <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then set
+ <var>serializedOrigin</var> to `<code>null</code>`. [[ONION]]
 
  <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>cors</code>" or
  <var>request</var>'s <a for=request>mode</a> is "<code>websocket</code>", then


fetch.bs
```diff
@@ -2820,6 +2823,11 @@ given a <a for=/>request</a> <var>request</var>, run these steps:
  <li><p>Let <var>serializedOrigin</var> be the result of <a>byte-serializing a request origin</a>
  with <var>request</var>.
 
+ <li><p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> uses the
+ <code>.onion</code> special-use domain name [[ONION]] and is not <a>same origin</a> with
+ <var>request</var>'s <a for=request>origin</a>, then set <var>serializedOrigin</var> to
+ `<code>null</code>`.
```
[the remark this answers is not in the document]
fmarier: Is this latest revision what you had in mind?

fetch.bs
```diff
@@ -2820,6 +2823,11 @@ given a <a for=/>request</a> <var>request</var>, run these steps:
  <li><p>Let <var>serializedOrigin</var> be the result of <a>byte-serializing a request origin</a>
  with <var>request</var>.
 
+ <li><p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>'s
+ <a for=origin>host</a> ends with <code>.onion</code> or <code>.onion.</code> [[ONION]], and
+ is not <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then set
+ <var>serializedOrigin</var> to `<code>null</code>`.
```
annevk, line 2829:
```suggestion
 <li><p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>'s
 <a for=origin>host</a> ends with "<code>.onion</code>" or "<code>.onion.</code>", and
 is not <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then set
 <var>serializedOrigin</var> to `<code>null</code>`. [[ONION]]
```

annevk (commented):
One thing remaining here is that if we want to have this check in at least three places, we should probably have an exported "is an onion URL" predicate that takes a URL and returns true or false instead. (Other things are tests, Mike's feedback, whether this should be optional somehow or conditional upon `.onion` support...)

domenic:
The other big remaining thing here is making sure this would be on track to become a part of the interoperable web, via implementation (on by default) in WebKit/Gecko/Chromium.

fmarier:
> making sure this would be on track to become a part of the interoperable web

Yes and trying to see whether there's agreement on the correct way to handle these was the reason I created these PRs in the first place.

In the short term, Brave is going to [match what the Tor Browser is doing](https://github.com/brave/brave-core/pull/10760), but ideally we'd like to see if we can align with the browsers that don't bundle Tor directly.

fmarier:
> whether this should be optional somehow or conditional upon .onion support

On this specific point, I would suggest the answer is no, it should be always ON because there's no easy way for the browser to know that the SOCKS5 proxy it's using is the Tor daemon.

fetch.bs
```diff
@@ -2820,6 +2823,11 @@ given a <a for=/>request</a> <var>request</var>, run these steps:
  <li><p>Let <var>serializedOrigin</var> be the result of <a>byte-serializing a request origin</a>
  with <var>request</var>.
 
+ <li><p>If <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a>'s
+ <a for=origin>host</a> ends with "<code>.onion</code>" or "<code>.onion.</code>", and
+ is not <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then set
+ <var>serializedOrigin</var> to `<code>null</code>`. [[ONION]]
```
annevk, line 2829: Due to "byte-serializing a request origin" we'll also end up with null if they are same origin but there's a cross-origin redirect in between. Ideally we'd test that, but I guess `.onion` will be hard to test.

annevk (approved):
This looks good editorially. I guess there's still the question of whether we want to abstract "is an origin URL", but that could happen later as well.

So this mainly hinges on implementer interest and resolving the question as to whether we should expose more information when CORS is used.

fetch.bs
fmarier: I've got these two no-cors manual test cases on my [test page](http://ixrdj3iwwhkuau5tby5jh3a536a2rdhpbdbu6ldhng43r47kim7a3lid.onion/referrer/onion.html) (you can also see it at http://fmarier.com/referrer/onion.html though it's meant to run from a `.onion`) :
- `example.onion` --307--> `example.onion` --307--> `example.com`
- `example.onion` --307--> `example.com` --307--> `example.onion`

For the first one, both Brave and Tor Browser send a `null` origin (and omit the referrer for that matter).

For the second one, however, Brave sends a `null` origin but the Tor browser sends the full Origin in that case.

I think a `null` Origin is the desired behavior here?

annevk: I'm not sure I understand the examples, what's the starting origin? As long as you are same-origin I would expect a referrer to be included, but once you've gone cross-origin once, it should no longer be possible.

fmarier: The starting origin in all cases is `example.onion`. It issues two 307 redirects.

In the first example, it goes from `example.onion` to `example.onion/anotherpage` before ending on `example.com`.

In the second example, it goes from `example.onion` to `example.com` before ending on `example.onion/anotherpage`.

annevk: So the user navigates to the starting origin? I guess then the scenarios would work. (Typically with fetching there's a starting document with some kind of authority and that would do the fetching, which could then result in a redirect. The initial 307 threw me off.)

Assuming the user does the navigation I'd expect:

1. example.onion 307 to example.onion (Origin could be included here, e.g., in case of a POST); example.onion 307 to example.com (it would be null here)
2. example.onion 307 to example.com (it would be null here); example.com to example.onion (it would continue to be null here, as it's been tainted by example.com)

fmarier: > So the user navigates to the starting origin?

Yes, that's correct. However, I just realized that I made a mistake in my description of the test cases, which probably threw you off.

What I meant was:

1. Navigate to `example.onion/index.html`.
2. Submit form (POST) to `example.onion/another.html`.
3. `example.onion/another.html` is a 307 redirect to `example.com`.
4. You end up on `example.com` with a `null` `Origin` header.

The second test case is:

1. Navigate to `example.onion/index.html`.
2. Submit form (POST) to `example.com`.
3. `example.com` is a 307 redirect to `example.onion/another.html`.
4. You end up on `example.onion/another.html` with a `null` `Origin` header.

I had the same expectations as you (regarding the `example.com` tainting due to the redirect) and that's what I implemented in Brave, but Tor Browser returns an `Origin` of `example.onion` in the second case.

annevk: That seems like a pretty serious bug. @KershawChang @valenting ^^

KershawChang: I've filed [bug 1742784](https://bugzilla.mozilla.org/show_bug.cgi?id=1742784) for this.

ghost (changes requested):
Remove

PR #1357, opened 2021-11-16
Remove 'tainted origin flag'. (#1335)

wanderview:
   * Should be a no-op without behavioral changes.
   * Should be a no-op without behavioral changes.
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1357.html" title="Last updated on Nov 18, 2021, 12:00 PM UTC (d2392ee)">Preview</a> | <a href="https://whatpr.org/fetch/1357/3ecab20...d2392ee.html" title="Last updated on Nov 18, 2021, 12:00 PM UTC (d2392ee)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1790,10 +1790,6 @@ Unless stated otherwise, it is false.
 <hr>
 
 <p>A <a for=/>request</a> has an associated
-<dfn for=request id=concept-request-tainted-origin>tainted origin flag</dfn>. Unless stated
-otherwise, it is unset.
-
-<p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-url-list>URL list</dfn> (a <a for=/>list</a> of one or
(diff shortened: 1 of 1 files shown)


wanderview:
@annevk PTAL.  This removes the tainted origin flag and I believe is a no-op.

Unfortunately I don't know how to find downstream consumer specs.  I've been told there is no tool to do that.

annevk:
@wanderview thoughts on my changes? I don't think this should impact downstream much if at all as the definition isn't exported.

(I was thinking this could be combined with https://w3c.github.io/webappsec-fetch-metadata/#abstract-opdef-set-site but that does not actually seem feasible. A request from A to B is not tainted, but is cross-site. Combining them does not seem like it would improve matters either.)

wanderview:
Your changes LGTM.  Thanks.

PR #1361, opened 2021-11-26
Editorial: scheme is not a standalone type

annevk:
Remove more instances where a concept member is used as a type.

***
<a href="https://whatpr.org/fetch/1361.html" title="Last updated on Nov 29, 2021, 9:41 AM UTC (837a840)">Preview</a> | <a href="https://whatpr.org/fetch/1361/5dc54a7...837a840.html" title="Last updated on Nov 29, 2021, 9:41 AM UTC (837a840)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2819,14 +2819,8 @@ request <a for=/>header</a> indicates where a
 <a lt=fetch for=/>fetches</a>.
 <!-- Ian Hickson told me Adam Barth researched that -->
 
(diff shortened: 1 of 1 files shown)


domenic (commented):
I think the existing text is OK as-is, as I think it's basically saying "local scheme is a predicate you can apply to the scheme member of URL records (or something which is on its way to being put into that field), which returns true if the value is...". But I can understand not wanting that kind of implicitness, editorially, so either way seems fine to me.

domenic (approved):

PR #1366, opened 2021-12-07
Remove "WebSocket protocol alterations"

yutakahirano:
The section is being moved to whatwg/websockets. There is no behiavor change.

***
<a href="https://whatpr.org/fetch/1366.html" title="Last updated on Oct 25, 2022, 9:44 AM UTC (09e9524)">Preview</a> | <a href="https://whatpr.org/fetch/1366/b56692a...09e9524.html" title="Last updated on Oct 25, 2022, 9:44 AM UTC (09e9524)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -169,7 +169,7 @@ exposes most of the networking functionality at a fairly low level of abstractio
 <p>This specification depends on the Infra Standard. [[!INFRA]]
 
 <p>This specification uses terminology from the ABNF, Encoding, HTML, HTTP, IDL, MIME Sniffing,
-Streams, and URL Standards.
+Streams, URL, and WebSockets Standards.
 [[!ABNF]]
 [[!ENCODING]]
 [[!HTML]]
@@ -178,6 +178,7 @@ Streams, and URL Standards.
 [[!MIMESNIFF]]
 [[!STREAMS]]
 [[!URL]]
+[[!WEBSOCKETS]]
 
 <p><dfn>ABNF</dfn> means ABNF as augmented by HTTP (in particular the addition of <code>#</code>)
 and RFC 7405. [[!RFC7405]]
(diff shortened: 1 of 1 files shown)


yutakahirano:
Hi @annevk, is this something you mentioned sometime ago?

@ricea said it'd be better to keep section links (e.g., https://fetch.spec.whatwg.org/#websocket-protocol) somehow. Do you know how to do that?

annevk:
Let's keep the section and do something similar to https://github.com/whatwg/xhr/pull/338/files? I'm open to completely removing it at some point, but for now we should probably have some forward references and keep the existing IDs working.

yutakahirano:
How about the latest one?

annevk:
That looks pretty good, but it seems the new WebSockets document isn't indexed yet so the terms do not end up linking. Filed https://github.com/tabatkins/bikeshed/issues/2193.

yutakahirano:
I'm exporting two definitions at https://github.com/whatwg/websockets/pull/9.

yutakahirano:
I'm not sure what to do with other two definitions, "fail the WebSocket connection" and "the WebSocket connection is established". https://websockets.spec.whatwg.org/ directly links to terms in RFC6465.

domenic:
An alternative to exporting those definitions (since we don't *really* want other specs calling them) is to use the "override" syntax, `<a spec=WEBSOCKETS>obtain a WebSocket connection</a>`, which ignores exported or not-exported.

Also I think this is using `<span>` whereas `<a>` is more idiomatic.

annevk:
Obtain a WebSocket connection should be exported for the usage in https://fetch.spec.whatwg.org/#concept-http-network-fetch. I'm okay with not exporting the others and using direct links there or some such.

annevk:
Closing this in favor of #1516.

Issue #1387, opened 2022-01-27
TAO support for iframes


noamr:
Currently TAO (Timing-Allow-Origin) is checked as part of the CORS mechanism.
However, for iframes it needs to be checked at the end, relative to the containing document's origin.
Implementations already do this.

I wonder how to go about it spec-wise:
- To create a variant of `Finalize and Report timing` that does this, receiving an origin
- Export some internals of the TAO check and let HTML do this when reporting an iframe (I think I prefer this, and it's closer to implementations)

noamr:
@annevk, any thoughts?

annevk:
I think I agree with your take that this is something HTML's navigate algorithm needs to solve, relying on infrastructure provided by Fetch.

We should also explore the relation to navigation timing as an `iframe` element is essentially a nested navigation.

PR #1388, opened 2022-01-27
Allow timing reporting with origin (for iframes)

noamr:
This will allow checking TAO for an iframe, which doesn't go via CORS - the check is done upon creation of the RT entry rather than during fetch (this is aligned with implementations).

Closes https://github.com/whatwg/fetch/issues/1387 
See https://github.com/whatwg/html/pull/7531No new behavior.

***
<a href="https://whatpr.org/fetch/1388.html" title="Last updated on Feb 14, 2022, 1:31 PM UTC (a89f4ea)">Preview</a> | <a href="https://whatpr.org/fetch/1388/8ebf2fd...a89f4ea.html" title="Last updated on Feb 14, 2022, 1:31 PM UTC (a89f4ea)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4337,18 +4337,6 @@ steps:
  and set <var>response</var>'s <a for="response">timing info</a> to the result.
 </ol>
 
-<p>To <dfn export>report timing</dfn> given a <a for=/>response</a>
-<var>response</var>, a <a for=/>global object</a> <var>global</var>, and a <a for=/>string</a>
-<var>initiatorType</var> (default "<code>other</code>"), run these steps:
-
-<ol>
- <li><p>Let <var>allowTiming</var> be true if <var>response</var>'s
- <a for=response>timing allow passed flag</a> is set; otherwise false.
-
- <li><a>Report timing</a> for <var>response</var>, <var>global</var>, and <var>initiatorType</var>,
- and set <var>response</var>'s <a for="response">timing info</a> to the result.
-</ol>
-
 <p>To <dfn export>report timing with origin</dfn> given a <a for=/>response</a>
 <var>response</var>, a <a for=/>global object</a> <var>global</var>, an <a for=/>origin</a>
 <var>origin</var>, and a <a for=/>string</a> <var>initiatorType</var> (default
--- a/fetch.bs
+++ b/fetch.bs
@@ -4346,9 +4346,9 @@ steps:
 <var>response</var> and the result of <a lt="ASCII serialization of an origin">serializing</a>
 <var>origin</var>.
 
-<p class=note>This algorithm is used by <cite>HTML</cite> to report resource timing for
-iframes, as iframes report both to the iframe's loaded document as navigation timing and to the
-containing document as resource timing. [[HTML]]
+<p class=note>This algorithm is used by <cite>HTML</cite> to report resource timing for documents
+nested through <code>iframe</code> elements, as iframes report both to the <code>iframe</code>'s
+loaded document as navigation timing and to the containing document as resource timing. [[HTML]]
 
 <p>To <dfn>report timing</dfn> given a <a for=/>response</a> <var>response</var>, a
 <a for=/>global object</a> <var>global</var>, a <a for=/>string</a>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4343,7 +4343,8 @@ steps:
 <var>origin</var>, and a <a for=/>string</a> <var>initiatorType</var> (default
 "<code>other</code>"), <a>report timing</a> for <var>response</var>,
 <var>global</var>, <var>initiatorType</var>, and the result of calling <a>TAO origin check</a> with
-<var>origin</var> and <var>response</var>.
+<var>response</var> and the result of <a lt="ASCII serialization of an origin">serializing</a>
+<var>origin</var>.
 
 <p>To <dfn>report timing</dfn> given a <a for=/>response</a> <var>response</var>, a
 <a for=/>global object</a> <var>global</var>, a <a for=/>string</a>
@@ -6031,13 +6032,12 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
  <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
  <var>request</var>.
 
- <li><p>Return the result of calling <a>TAO origin check</a> with <var>origin</var> and
- <var>response</var>.
+ <li><p>Return the result of calling <a>TAO origin check</a> with <var>response</var> and
+ <var>origin</var>.
 </ol>
 
-<p>To perform a <dfn id=concept-tao-origin-check>TAO origin check</dfn> for an
-<a for=/>origin</a> <var>origin</var> and a <a for=/>response</a> <var>response</var>, run these
-steps:
+<p>To perform a <dfn id=concept-tao-origin-check>TAO origin check</dfn> for a <a for=/>response</a>
+<var>response</var> and an <a for=/>ASCII string</a> <var>origin</var> and, run these steps:
 
 <ol>
  <li><p>Let <var>values</var> be the result of
--- a/fetch.bs
+++ b/fetch.bs
@@ -6008,14 +6008,25 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
  <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>basic</code>", then
  return success.
 
+ <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
+ <var>request</var>.
+
+ <li><p>Return the result of calling <a>TAO origin check</a> with <var>origin</var> and
+ <var>response</var>.
+</ol>
+
+<p>To perform a <dfn export id=concept-tao-origin-check>TAO origin check</dfn> for an
+<a for=/>origin</a> <var>origin</var> and a <a for=/>response</a> <var>response</var>, run these
+steps:
+
+<ol>
  <li><p>Let <var>values</var> be the result of
  <a for="header list">getting, decoding, and splitting</a> `<code>Timing-Allow-Origin</code>` from
  <var>response</var>'s <a for=response>header list</a>.
 
  <li><p>If <var>values</var> <a for=list>contains</a> "<code>*</code>", then return success.
 
- <li><p>If <var>values</var> <a for=list>contains</a> the result of
- <a>serializing a request origin</a> with <var>request</var>, then return success.
+ <li><p>If <var>values</var> <a for=list>contains</a> <var>origin</var>, then return success.
 
  <li><p>Return failure.
 </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4346,6 +4346,10 @@ steps:
 <var>response</var> and the result of <a lt="ASCII serialization of an origin">serializing</a>
 <var>origin</var>.
 
+<p class=note>This algorithm is used by <cite>HTML</cite> to report resource timing for
+iframes, as iframes report both to the iframe's loaded document as navigation timing and to the
+containing document as resource timing.
+
 <p>To <dfn>report timing</dfn> given a <a for=/>response</a> <var>response</var>, a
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4289,6 +4289,26 @@ steps:
 <var>response</var>, a <a for=/>global object</a> <var>global</var>, and a <a for=/>string</a>
 <var>initiatorType</var> (default "<code>other</code>"), run these steps:
 
+<ol>
+ <li><p>Let <var>taoPass</var> be true <var>response</var>'s
```
yoavweiss, line 4293: "if" missing?

fetch.bs
```diff
@@ -4289,6 +4289,26 @@ steps:
...
+
+ <li>Set <var>response</var>'s <a for="response">timing info</a> to the
+ result of calling <a>report timing</a> given <var>response</var>, <var>global</var>,
+ <var>initiatorType</var>, and <var>taoPass</var>.
+</ol>
+
+<p>To <dfn export>report timing with origin</dfn> given a <a for=/>response</a>
+<var>response</var>, a <a for=/>global object</a> <var>global</var>, an <a for=/>origin</a>
+<var>origin</var>, and a <a for=/>string</a> <var>initiatorType</var> (default
+"<code>other</code>"), <a>report timing</a> for <var>response</var>,
```
yoavweiss, line 4304: Adding a "run" before `report timing` here would make this more readable.

noamr: That's inconsistent with the rest of the fetch spec though

yoavweiss: ok, fair enough.

fetch.bs

fetch.bs
```diff
@@ -5968,14 +5988,25 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+<ol>
  <li><p>Let <var>values</var> be the result of
  <a for="header list">getting, decoding, and splitting</a> `<code>Timing-Allow-Origin</code>` from
  <var>response</var>'s <a for=response>header list</a>.
 
  <li><p>If <var>values</var> <a for=list>contains</a> "<code>*</code>", then return success.
 
- <li><p>If <var>values</var> <a for=list>contains</a> the result of
- <a>serializing a request origin</a> with <var>request</var>, then return success.
+ <li><p>If <var>values</var> <a for=list>contains</a> <var>origin</var>, then return success.
```
annevk, line 6009: If _origin_ is an origin this doesn't work, right? It seems elsewhere you are invoking this with a string, so maybe _origin_ is supposed to be an origin?

fetch.bs
```diff
@@ -5968,14 +5988,25 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
 
+ <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
+ <var>request</var>.
+
+ <li><p>Return the result of calling <a>TAO origin check</a> with <var>origin</var> and
+ <var>response</var>.
+</ol>
+
+<p>To perform a <dfn id=concept-tao-origin-check>TAO origin check</dfn> for an
+<a for=/>origin</a> <var>origin</var> and a <a for=/>response</a> <var>response</var>, run these
```
annevk, line 5999: Nit: flip these arguments.

fetch.bs
```diff
@@ -4289,6 +4289,26 @@ steps:
...
+<ol>
+ <li><p>Let <var>taoPass</var> be true if <var>response</var>'s
+ <a for=response>timing allow passed flag</a> is set; otherwise false.
+
+ <li>Set <var>response</var>'s <a for="response">timing info</a> to the
+ result of calling <a>report timing</a> given <var>response</var>, <var>global</var>,
+ <var>initiatorType</var>, and <var>taoPass</var>.
+</ol>
+
+<p>To <dfn export>report timing with origin</dfn> given a <a for=/>response</a>
```
annevk, line 4301: with an origin*?

annevk: This algorithm should also have a note that indicates it's called by HTML, ideally with a reference to the algorithm that does it.

fetch.bs
noamr: Changing it to be an ASCII string, as I need to use `serialize origin from request`

fetch.bs
```diff
@@ -6008,14 +6033,24 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
 
+ <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
+ <var>request</var>.
+
+ <li><p>Return the result of calling <a>TAO origin check</a> with <var>response</var> and
+ <var>origin</var>.
+</ol>
+
+<p>To perform a <dfn id=concept-tao-origin-check>TAO origin check</dfn> for a <a for=/>response</a>
+<var>response</var> and an <a for=/>ASCII string</a> <var>origin</var> and, run these steps:
```
annevk, line 6044: spurious "and"

fetch.bs
```diff
@@ -4329,6 +4329,31 @@ steps:
...
+<var>response</var>, a <a for=/>global object</a> <var>global</var>, an <a for=/>origin</a>
+<var>origin</var>, and a <a for=/>string</a> <var>initiatorType</var> (default
+"<code>other</code>"), <a>report timing</a> for <var>response</var>,
+<var>global</var>, <var>initiatorType</var>, and the result of calling <a>TAO origin check</a> with
+<var>response</var> and the result of <a lt="ASCII serialization of an origin">serializing</a>
+<var>origin</var>.
+
+<p class=note>This algorithm is used by <cite>HTML</cite> to report resource timing for
+iframes, as iframes report both to the iframe's loaded document as navigation timing and to the
+containing document as resource timing.
```
annevk, line 4351: This needs [[HTML]] or some such at the end. It should also use more formal language to describe these concepts. Perhaps "report resource timing documents nested through `iframe` elements"?

fetch.bs

fetch.bs
annevk: It seems the language comment wasn't addressed/pushed?

noamr: Oops, oversight. Sorry for the noise.

noamr:
Actually there's something missing here I believe... the iframe and the embedder might have a different cross-origin isolated capability, so the timing might need to be coarsened if the embedder is more strict than the frame. The same will be needed for early hints, as the timestamps are marked before the document is created.

I wonder if entries should be coarsened again when accessed (inside the `Resource Timing` spec) - which is what implementations do - it makes sure that the global object for coarsening is the one relevant to the entry regardless of how it was created.

fetch.bs
```diff
@@ -4329,6 +4329,31 @@ steps:
 <var>response</var>, a <a for=/>global object</a> <var>global</var>, and a <a for=/>string</a>
 <var>initiatorType</var> (default "<code>other</code>"), run these steps:
 
+<ol>
+ <li><p>Let <var>taoPass</var> be true if <var>response</var>'s
+ <a for=response>timing allow passed flag</a> is set; otherwise false.
+
+ <li>Set <var>response</var>'s <a for="response">timing info</a> to the
+ result of calling <a>report timing</a> given <var>response</var>, <var>global</var>,
+ <var>initiatorType</var>, and <var>taoPass</var>.
```
annevk, line 4338:
```suggestion
 <li><p>Set <var>response</var>'s <a for="response">timing info</a> to the result of calling 
 <a>report timing</a> given <var>response</var>, <var>global</var>, <var>initiatorType</var>, and 
 <var>taoPass</var>.
```

fetch.bs
```diff
@@ -4329,6 +4329,31 @@ steps:
...
+</ol>
+
+<p>To <dfn export>report timing with origin</dfn> given a <a for=/>response</a>
+<var>response</var>, a <a for=/>global object</a> <var>global</var>, an <a for=/>origin</a>
+<var>origin</var>, and a <a for=/>string</a> <var>initiatorType</var> (default
+"<code>other</code>"), <a>report timing</a> for <var>response</var>,
+<var>global</var>, <var>initiatorType</var>, and the result of calling <a>TAO origin check</a> with
+<var>response</var> and the result of <a lt="ASCII serialization of an origin">serializing</a>
+<var>origin</var>.
```
annevk, line 4347:
```suggestion
<var>global</var>, <var>initiatorType</var>, the result of calling <a>TAO origin check</a> with
<var>response</var>, and the result of <a lt="ASCII serialization of an origin">serializing</a>
<var>origin</var>.
```
Right?

fetch.bs
```diff
@@ -4329,6 +4329,31 @@ steps:
...
+
+<p class=note>This algorithm is used by <cite>HTML</cite> to report resource timing for documents
+nested through <code>iframe</code> elements, as iframes report both to the <code>iframe</code>'s
+loaded document as navigation timing and to the containing document as resource timing. [[HTML]]
```
annevk, line 4351: This read a little weird to me still. How about this:
```suggestion
<p class=note>This algorithm is used by <cite>HTML</cite> to report resource timing for documents
nested through <code>iframe</code> elements to their container document. (Nested documents also
report navigation timing to themselves.) [[HTML]]
```

fetch.bs
```diff
@@ -4347,7 +4372,7 @@ steps:
  <li><p>If <var>timingInfo</var> is null, then return.
 
  <li>
-  <p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then:
+  <p>If <var>taoPass</var> is false, then:</p>
```
annevk, line 4375:
```suggestion
  <p>If <var>taoPass</var> is false, then:
```

fetch.bs
```diff
@@ -6008,14 +6033,24 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
 
+ <li><p>Let <var>origin</var> be the result of <a>serializing a request origin</a> with
+ <var>request</var>.
+
+ <li><p>Return the result of calling <a>TAO origin check</a> with <var>response</var> and
+ <var>origin</var>.
+</ol>
+
+<p>To perform a <dfn id=concept-tao-origin-check>TAO origin check</dfn> for a <a for=/>response</a>
+<var>response</var> and an <a for=/>ASCII string</a> <var>origin</var>, run these steps:
```
annevk, line 6044:
```suggestion
<p>To perform a <dfn>TAO origin check</dfn> for a <a for=/>response</a> <var>response</var> and an 
<a for=/>ASCII string</a> <var>origin</var>, run these steps:
```
This does not need to be exported, right?

annevk (commented):
So you can end up with that scenario when the cross-origin isolated capability is not delegated through Permissions Policy. Is the problem here that we use the capability of the parent and not the child? It seems that's what we should address somehow, right?

noamr:
> So you can end up with that scenario when the cross-origin isolated capability is not delegated through Permissions Policy. Is the problem here that we use the capability of the parent and not the child? It seems that's what we should address somehow, right?

We use the capability of the child and not the parent I believe, based on the global object used as the fetch client. Thinking to address this inside resource timing rather than here - to make sure the global object at the time of accessing properties has a chance to coarse.

annevk:
In that case it should be fine, no? Theoretically it seems nicer to address it in the network layer as otherwise the "sensitive" value would have to be available in an agent cluster that can be attacker read. (Not the biggest problem by far, but would still be nice to have theoretically sound standards.)

noamr:
> In that case it should be fine, no? Theoretically it seems nicer to address it in the network layer as otherwise the "sensitive" value would have to be available in an agent cluster that can be attacker read. (Not the biggest problem by far, but would still be nice to have theoretically sound standards.)

Yes, I think it's fine and we can proceed. Implementations do this at the API layer

noamr:
actually, reading this again, I don't think we need to perform another TAO check. The origin of the initial navigation request is already the container's origin IIUC (it's a bit difficult to track this down in the HTML spec). Which means FETCH doesn't need to have special support for iframe reporting.

annevk:
Well, I think it depends. When you create a new `iframe` element that is the case. But subsequent navigations might come from another authority due to named targeting.

noamr:
> Well, I think it depends. When you create a new `iframe` element that is the case. But subsequent navigations might come from another authority due to named targeting.

Right, but those navs don't report resource timing (only `iframe.src`')

PR #1392, opened 2022-01-30
Add `Response.json` static method

### LINKED ISSUE #1389: Proposal: `Response.json` helper
The `Response` object currently has a `Response.redirect` helper method to create `Response` objects with preset `Location` headers. `Request` and `Response` already have a `json` prototype method that allows parsing bodies as JSON, but there is currently no trivial one-liner to **create** JSON `Response` objects directly from a POJO.

I propose the addition of a `Response.json(obj, opts)` static method that creates a `Response` object with a body string containing the JSON serialized representation of `obj`. `opts` would allow further modification of the JSON response, namely setting headers and status code. By default the status code is 200, and the headers are the default response headers + `content-type: application/json`.

```webidl
interface Response {
  [NewObject] static Response json(any object, optional ResponseInit init = {});
}
```

The motivation for this addition is that in server side runtimes using WHATWG `Request` and `Response` objects, folks often create helper methods for this behaviour that need to be duplicated across each project. This being built in would be very useful.

Thoughts?

Implementer interest:
- Chromium ✔️  
- Firefox ✔️ 
- Safari ❔ 
- Deno ✔️  
- Cloudflare Workers ✔️

lucacasonato:
This commit adds a `Response.json` static method that can be used to
create well formed JSON responses will very little effort.

The JSON response is _not_ pretty printed.

Closes #1389

---

   * [Chrome](https://github.com/whatwg/fetch/pull/1392#issuecomment-1123133441)
   * [Firefox](https://github.com/whatwg/fetch/pull/1392#pullrequestreview-970555855)
   * Deno
   * Cloudflare Workers
   * https://github.com/web-platform-tests/wpt/pull/32825
   * Chrome: https://crbug.com/1305358
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1758943
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=240375
   * Deno: https://github.com/denoland/deno/issues/13902

***
<a href="https://whatpr.org/fetch/1392.html" title="Last updated on May 13, 2022, 9:02 AM UTC (7ef2f43)">Preview</a> | <a href="https://whatpr.org/fetch/1392/3ab5102...7ef2f43.html" title="Last updated on May 13, 2022, 9:02 AM UTC (7ef2f43)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7298,6 +7298,7 @@ interface Response {
 
   [NewObject] static Response error();
   [NewObject] static Response redirect(USVString url, optional unsigned short status = 302);
+  [NewObject] static Response json(any data, optional ResponseInit init = {});
 
   readonly attribute ResponseType type;
 
@@ -7348,6 +7349,10 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
  <dt><code><var>response</var> = <a idl>Response</a> . <a method for=Response lt=redirect()>redirect</a>(<var>url</var>, <var>status</var> = 302)</code>
  <dd><p>Creates a redirect {{Response}} that redirects to <var>url</var> with status
  <var>status</var>.
+
+ <dt><code><var>response</var> = <a idl>Response</a> . <a method for=Response lt=json()>json</a>(<var>data</var> [, <var>init</var>])</code>
+ <dd><p>Creates a {{Response}} whose body is the JSON-encoded <var>data</var>, and status, status
+ message, and headers are provided by <var>init</var>.
 
  <dt><code><var>response</var> . <a attribute for=Response>type</a></code>
  <dd><p>Returns <var>response</var>'s type, e.g., "<code>cors</code>".
@@ -7476,6 +7481,48 @@ are:
 
  <li><p><a for="header list">Append</a> (`<code>Location</code>`, <var>value</var>) to
  <var>responseObject</var>'s <a for=Response>response</a>'s <a for=response>header list</a>.
+
+ <li><p>Return <var>responseObject</var>.
+</ol>
+
+
+<p>The static
+<dfn method for=Response><code>json(<var>data</var>, <var>init</var>)</code></dfn> method steps
+are:
+
+<ol>
+ <li><p>If <var>init</var>["{{ResponseInit/status}}"] is not in the range 200 to 599, inclusive,
+ then <a>throw</a> a {{RangeError}}.
+
+ <li><p>If <var>init</var>["{{ResponseInit/status}}"] is a <a>null body status</a>, then
+ <a>throw</a> a {{TypeError}}.
+
+ <li><p>If <var>init</var>["{{ResponseInit/statusText}}"] does not match the
+ <a spec=http>reason-phrase</a> token production, then <a>throw</a> a {{TypeError}}.
+
+ <li><p>Let <var>responseObject</var> be the result of <a for=Response>creating</a> a {{Response}}
+ object, given a new <a for=/>response</a>, "<code>response</code>", and <a>this</a>'s
+ <a>relevant Realm</a>.
+
+ <li><p>Set <var>responseObject</var>'s <a for=Response>response</a>'s <a for=response>status</a> to
+ <var>init</var>["{{ResponseInit/status}}"].
+
+ <li><p>Set <var>responseObject</var>'s <a for=Response>response</a>'s
+ <a for=response>status message</a> to <var>init</var>["{{ResponseInit/statusText}}"].
+
+ <li><p>If <var>init</var>["{{ResponseInit/headers}}"] <a for=map>exists</a>, then
+ <a for=Headers>fill</a> <var>responseObject</var>'s <a for=Response>headers</a> with
+ <var>init</var>["{{ResponseInit/headers}}"].
+
+ <li><p>Set <var>responseObject</var>'s <a for=Response>response</a>'s <a for=response>body</a> to
+ the <a>UTF-8 encoding</a> of the result of running
+ <a>serialize a JavaScript value to a JSON string</a> on <var>data</var>.
+
+ <li><p>If <var>responseObject</var>'s <a for=Response>response</a>'s <a for=response>header list</a>
+ <a for="header list">does not contain</a> `<code>Content-Type</code>`, then
+ <a for="header list">append</a> (`<code>Content-Type</code>`,
+ `<code>application/json;charset=utf-8</code>`) to <var>responseObject</var>'s
+ <a for=Response>response</a>'s <a for=response>header list</a>.
 
  <li><p>Return <var>responseObject</var>.
 </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -7657,7 +7657,7 @@ are:
  <a>relevant Realm</a>.
 
  <li><p>Perform <a>initialize a response</a> given <var>responseObject</var>, <var>init</var>, and
- (body, <code>"application/json;charset=UTF-8"</code>).
+ (body, <code>"application/json"</code>).
 
  <li><p>Return <var>responseObject</var>.
 </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -1259,7 +1259,7 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
 <p>A <dfn export id=concept-body-with-type>body with type</dfn> is a
 <a for=/>tuple</a> that consists of a
 <dfn export for="body with type" id=concept-body-with-type-body>body</dfn> (a <a for=/>body</a>) and
-<dfn export for="body with type" id=concept-body-with-type>type</dfn> (a <a for=/>header value</a>).
+<dfn export for="body with type" id=concept-body-with-type-type>type</dfn> (a <a for=/>header value</a>).
 
 <hr>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -7544,8 +7544,8 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
  <li><p>Return <var>responseObject</var>.
 </ol>
 
-<p>To <dfn>initialize a response</dfn>, given a {{Response}} object <var>response</var>, 
-{{ResponseInit}} <var>init</var>, and an optional <a for=/>body with type</a> <var>body</var>, run 
+<p>To <dfn>initialize a response</dfn>, given a {{Response}} object <var>response</var>,
+{{ResponseInit}} <var>init</var>, and an optional <a for=/>body with type</a> <var>body</var>, run
 these steps:
 
 <ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -1256,10 +1256,9 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
 
 <hr>
 
-<p>A <dfn export id=concept-body-with-type>body with type</dfn> is a
-<a for=/>tuple</a> that consists of a
-<dfn export for="body with type" id=concept-body-with-type-body>body</dfn> (a <a for=/>body</a>) and
-<dfn export for="body with type" id=concept-body-with-type-type>type</dfn> (a <a for=/>header value</a>).
+<p>A <dfn export>body with type</dfn> is a <a for=/>tuple</a> that consists of a
+<dfn export for="body with type">body</dfn> (a <a for=/>body</a>) and
+<dfn export for="body with type">type</dfn> (a <a for=/>header value</a>).
 
 <hr>
 
@@ -7447,7 +7446,7 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
 </ol>
 
 <p>To <dfn>initialize a response</dfn>, given a {{Response}} <var>response</var>, {{ResponseInit}}
-<var>init</var>, and an optional <a for/>body with type</a> <var>body</var>, run these steps:
+<var>init</var>, and an optional <a for=/>body with type</a> <var>body</var>, run these steps:
 
 <ol>
  <li><p>If <var>init</var>["{{ResponseInit/status}}"] is not in the range 200 to 599, inclusive,
@@ -7485,6 +7484,7 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
    `<code>Content-Type</code>`, then <a for="header list">append</a> (`<code>Content-Type</code>`,
    <var>body</var>'s <a for="body with type">type</a>) to <var>response</var>'s
    <a for=response>header list</a>.
+  </ol>
 </ol>
 
 <hr>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -7480,6 +7485,48 @@ are:
  <li><p>Return <var>responseObject</var>.
 </ol>
 
+
+<p>The static
+<dfn method for=Response><code>json(<var>data</var>, <var>init</var>)</code></dfn> method steps
```
lucacasonato, line 7490: The implementation here has a lot of duplication with `new Response`. I could rewrite this as follows if that is preferred (not real spec text):

1. Create new header list.
2. Fill header list with init.headers.
3. If header list does not contain `content-type`, set it to `application/json`.
4. Let body be `JSON.stringify(body)`.
5. Set init.headers = header list.
5. Return `new Response(body, init)`

fetch.bs
```diff
@@ -7480,6 +7485,48 @@ are:
...
+ <li><p>Set <var>responseObject</var>'s <a for=Response>response</a>'s
+ <a for=response>status message</a> to <var>init</var>["{{ResponseInit/statusText}}"].
+
+ <li><p>If <var>init</var>["{{ResponseInit/headers}}"] <a for=map>exists</a>, then
+ <a for=Headers>fill</a> <var>responseObject</var>'s <a for=Response>headers</a> with
+ <var>init</var>["{{ResponseInit/headers}}"].
+
+ <li><p>Set <var>responseObject</var>'s <a for=Response>response</a>'s <a for=response>body</a> to
+ the <a>UTF-8 encoding</a> of the result of running
+ <a>serialize a JavaScript value to a JSON string</a> on <var>data</var>.
```
andreubotella, line 7519: Infra has a ["serialize a JavaScript value to JSON bytes"](https://infra.spec.whatwg.org/#serialize-a-javascript-value-to-json-bytes) algorithm.

domenic:
Note that we'd need to fix Web IDL first to allow regular and static operations with the same name:

> The identifier of a static operation also must not be the same as the identifier of a regular operation defined on the same interface.

from https://webidl.spec.whatwg.org/#idl-operations

lucacasonato:
@domenic Do you think that is desirable to change? It may be confusing to have both a prototype and static method with the same name. Maybe this method should be called `fromJSON`.

domenic:
I think it is desirable to change and do not see anything confusing about having such methods, personally.

fetch.bs
annevk: Reusing the steps makes sense to me, but note that we cannot invoke global constructors in the way the pseudo-code suggests. It would have to be some kind of shared abstraction.

lucacasonato: Ok, rewrote this so the shared steps are extracted into a common "initialize a response" algorithm.

annevk (commented):
This looks reasonable, but I think it can be improved by making initialize a response also take a body and Content-Type header value (potentially as a tuple?). That would allow for sharing even more steps.

I also wonder what we should do here if browser interest isn't forthcoming. Perhaps there's a way in which we can still have mutual agreement, but it's not exposed in browsers for now. Perhaps this is worth bringing up in the HTML triage call (if it doesn't otherwise get resolved).

jasnell:
Just fyi... I have an implementation of this in Cloudflare Workers just waiting on this to land or at least signal that it's moving forward.

domenic:
I don't think we have any signal that this is moving forward in browsers, unfortunately, due to it just being low-priority sugar for browser teams. As such I don't think it'll land in Fetch, at least until a couple of browsers have engineers looking for starter projects.

I'd really encourage you to use this as a case study for https://github.com/whatwg/js-hosts . Figure out what sort of documentation you want to create there, which would enable you to land such things with confidence, and go for it! I'm happy to participate and give guarantees like "Chromium will never ship a `Response.json()` with different semantics than those specified here".

lucacasonato:
@domenic Would this get a :+1: from Chromium if we did the implementation work?

fetch.bs
```diff
@@ -7520,6 +7542,25 @@ are:
  <li><p>Return <var>responseObject</var>.
 </ol>
 
+
+<p>The static
+<dfn method for=Response><code>json(<var>data</var>, <var>init</var>)</code></dfn> method steps
+are:
+
+<ol>
+ <li><p>Let <var>bytes</var> the result of running <a>serialize a JavaScript value to JSON bytes</a>
```
yutakahirano, line 7551: This can fail (e.g., when a getter throws an exception), what do we want to do in that case? Do we want to throw an exception?

lucacasonato: Yes. From what I can tell the `JSON.stringify` call in `serialize a JavaScript value to a JSON string` will throw an exception in this case, which should just be propagated up.

lucacasonato: I'll add a test for this to the WPT.

lucacasonato: https://github.com/web-platform-tests/wpt/pull/32825/commits/004ac56823ea7b19160a9a97f9722bdd23bb7726

fetch.bs
```diff
@@ -1256,6 +1256,13 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
 
 <hr>
 
+<p>A <dfn export id=concept-body-with-type>body with type</dfn> is a
+<a for=/>tuple</a> that consists of a
+<dfn export for="body with type" id=concept-body-with-type-body>body</dfn> (a <a for=/>body</a>) and
+<dfn export for="body with type" id=concept-body-with-type-type>type</dfn> (a <a for=/>header value</a>).
```
annevk, line 1262: Drop the explicit IDs here. For new stuff we want whatever Bikeshed comes up with generally.

fetch.bs
```diff
@@ -7434,11 +7446,8 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
  <li><p>Return <var>responseObject</var>.
 </ol>
 
-<hr>
-
-<p>The
-<dfn constructor for=Response id=dom-response lt="Response(body, init)"><code>new Response(<var>body</var>, <var>init</var>)</code></dfn>
-constructor steps are:
+<p>To <dfn>initialize a response</dfn>, given a {{Response}} <var>response</var>, {{ResponseInit}}
+<var>init</var>, and an optional <a for/>body with type</a> <var>body</var>, run these steps:
```
annevk, line 7450:
```suggestion
<var>init</var>, and an optional <a for=/>body with type</a> <var>body</var>, run these steps:
```

fetch.bs
```diff
@@ -7447,45 +7456,58 @@ constructor steps are:
...
-   <li><p>If <var>Content-Type</var> is non-null and <a>this</a>'s <a for=Response>response</a>'s
-   <a for=response>header list</a> <a for="header list">does not contain</a>
+   <li><p>If <var>body</var>'s <a for="body with type">type</a> is non-null and
+   <var>response</var>'s <a for=response>header list</a> <a for="header list">does not contain</a>
    `<code>Content-Type</code>`, then <a for="header list">append</a> (`<code>Content-Type</code>`,
-   <var>Content-Type</var>) to <a>this</a>'s <a for=Response>response</a>'s
+   <var>body</var>'s <a for="body with type">type</a>) to <var>response</var>'s
    <a for=response>header list</a>.
-  </ol>
+</ol>
```
annevk, line 7488: I think there's a missing closing tag here as we're going from 3 spaces to 0 too quickly.

fetch.bs
```diff
@@ -7447,45 +7456,58 @@ constructor steps are:
...
+ <li><p>Set <a>this</a>'s <a for=Response>response</a> to a new <a for=/>response</a>.
+
+ <li><p>Set <a>this</a>'s <a for=Response>headers</a> to a <a for=/>new</a> {{Headers}} object with
+ <a>this</a>'s <a>relevant Realm</a>, whose <a for=Headers>header list</a> is <a>this</a>'s
+ <a for=Response>response</a>'s <a for=response>header list</a> and <a for=Headers>guard</a> is
+ "<code>response</code>".
+
+ <li><p>Let <var>bodyWithType</var> be null.
+
+ <li><p>If <var>body</var> is non-null, then let <var>bodyWithType</var> be the result of
```
annevk, line 7506:
```suggestion
 <li><p>If <var>body</var> is non-null, then set <var>bodyWithType</var> to the result of
```

fetch.bs
```diff
@@ -7447,45 +7456,58 @@ constructor steps are:
...
+ <a for=BodyInit>extracting</a> <var>body</var>.
```
annevk, line 7507: We should have a follow-up on using "body with type" throughout so this doesn't stay as shaky as it is here.

fetch.bs
```diff
@@ -7520,6 +7542,25 @@ are:
...
+
+ <li><p>Let <var>body</var> be the result of <a for=BodyInit>extracting</a> <var>bytes</var>.
+
+ <li><p>Let <var>responseObject</var> be the result of <a for=Response>creating</a> a {{Response}}
+ object, given a new <a for=/>response</a>, "<code>response</code>", and <a>this</a>'s
+ <a>relevant Realm</a>.
+
+ <li><p>Perform <a>initialize a response</a> given <var>responseObject</var>, <var>init</var>, and
+ (body, <code>"application/json;charset=utf-8"</code>).
+</ol>
```
annevk, line 7562: You need to return _responseObject_.

annevk (commented):
Mostly nits and request for a follow-up. Shared algorithm looks rather good! Thanks.

fetch.bs
lucacasonato: https://github.com/whatwg/fetch/issues/1410

lucacasonato:
@ricea I have an implementation for Chromium up on Gerrit now: https://chromium-review.googlesource.com/c/chromium/src/+/3639064. Is this enough to get an 'implementer interest' from Chromium? :)

ricea:
I am generally in favour, since most of the work is already done for us. @yutakahirano what do you think?

yutakahirano:
SGTM

lucacasonato:
I'll take that as a yes. Thanks folks!

@annevk what are Mozilla's thoughts?

Who should I ping for a vague interest position from the Safari team?

fetch.bs
```diff
@@ -7520,6 +7542,27 @@ are:
...
+ on <var>data</var>.
+
+ <li><p>Let <var>body</var> be the result of <a for=BodyInit>extracting</a> <var>bytes</var>.
+
+ <li><p>Let <var>responseObject</var> be the result of <a for=Response>creating</a> a {{Response}}
+ object, given a new <a for=/>response</a>, "<code>response</code>", and <a>this</a>'s
+ <a>relevant Realm</a>.
+
+ <li><p>Perform <a>initialize a response</a> given <var>responseObject</var>, <var>init</var>, and
+ (body, <code>"application/json;charset=UTF-8"</code>).
```
annevk, line 7561: The quotation marks need to be on the outside.

But also, I think we don't want a `charset` parameter here. None is defined for this MIME type.

annevk (commented):
Yeah this is reasonable and small enough. Mozilla supports this.

cc @KershawChang

annevk:
I suspect @youennf or @cdumez can comment for WebKit.

fetch.bs
lucacasonato: Removed charset.

fetch.bs
```diff
@@ -7552,50 +7554,61 @@ constructor steps are:
...
- <a for=Response>response</a>'s <a for=response>header list</a> and <a for=Headers>guard</a> is
- "<code>response</code>".
-
- <li><p>Set <a>this</a>'s <a for=Response>response</a>'s <a for=response>status</a> to
+ <li><p>Set <var>response</var>'s <a for=Response>response</a>'s <a for=response>status</a> to
  <var>init</var>["{{ResponseInit/status}}"].
 
- <li><p>Set <a>this</a>'s <a for=Response>response</a>'s <a for=response>status message</a> to
+ <li><p>Set <var>response</var>'s <a for=Response>response</a>'s <a for=response>status message</a> to
  <var>init</var>["{{ResponseInit/statusText}}"].
```
annevk, line 7561:
```suggestion
 <li><p>Set <var>response</var>'s <a for=Response>response</a>'s <a for=response>status message</a>
 to <var>init</var>["{{ResponseInit/statusText}}"].
```

fetch.bs
```diff
@@ -7539,11 +7544,8 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
  <li><p>Return <var>responseObject</var>.
 </ol>
 
-<hr>
-
-<p>The
-<dfn constructor for=Response id=dom-response lt="Response(body, init)"><code>new Response(<var>body</var>, <var>init</var>)</code></dfn>
-constructor steps are:
+<p>To <dfn>initialize a response</dfn>, given a {{Response}} <var>response</var>, {{ResponseInit}}
+<var>init</var>, and an optional <a for=/>body with type</a> <var>body</var>, run these steps:
```
annevk, line 7548:
```suggestion
<p>To <dfn>initialize a response</dfn>, given a {{Response}} object <var>response</var>, 
{{ResponseInit}} <var>init</var>, and an optional <a for=/>body with type</a> <var>body</var>, run 
these steps:
```

fetch.bs
```diff
@@ -7628,6 +7641,27 @@ are:
...
+ on <var>data</var>.
+
+ <li><p>Let <var>body</var> be the result of <a for=BodyInit>extracting</a> <var>bytes</var>.
+
+ <li><p>Let <var>responseObject</var> be the result of <a for=Response>creating</a> a {{Response}}
+ object, given a new <a for=/>response</a>, "<code>response</code>", and <a>this</a>'s
+ <a>relevant Realm</a>.
+
+ <li><p>Perform <a>initialize a response</a> given <var>responseObject</var>, <var>init</var>, and
+ (body, "<code>application/json</code>").
```
annevk, line 7660:
```suggestion
 (<var>body</var>, "<code>application/json</code>").
```

annevk (commented):
Thanks, this looks good to me now modulo some final nits.

(It's a bit unfortunate JSON ended up with `application/json` rather than `text/json`. Most other web text formats use `text/`. And while I believe this might be the first place we directly expose it in the web platform, going with `text/json` would be rather brazen.)

lucacasonato:
> It's a bit unfortunate JSON ended up with application/json rather than text/json. Most other web text formats use text/.

The platform acks that `appliction/json` is the commonly used JSON mime type already though: https://mimesniff.spec.whatwg.org/#json-mime-type (even if it never directly exposes this string). Some WICG work does make use of this content type too (both questionably relevant though):

- https://wicg.github.io/media-feeds/#retrieval-of-media-feeds
- https://wicg.github.io/conversion-measurement-api/#create-report-request

annevk (approved):
Yeah I know, it's just sad and mostly a consequence of some early ideas around `text/*` that ended up not mattering (at least not on the web).

Anyway, thanks for all your hard work on this. This was not an easy feature given all the work on Web IDL, refactoring of various Fetch algorithms, etc. 😊

I suggest we give folks until May 18 to voice any remaining concerns and merge it that day sometime.

foolip:
This change isn't possible to roll into WPT's idlharness.js test, see https://github.com/web-platform-tests/wpt/pull/34155. Does anyone want to give it a go to update idlharness.js to avoid this problem?

Issue #981, opened 2019-12-11
Trailer support in the API


annevk:
I still think this is something we should do as it's part of HTTP and with newer iterations of H/2 it's a feature that's a lot easier to make use of due to overall improved infrastructure.

My current thinking is that building this on top of #980 and #607 (as you can have multiple trailer header lists per request/response) is the way to go. `FetchObserver` could have a `sendTrailer(Headers trailer)` and `ontrailer` event handler or some such. Details probably best sorted once there's more firm implementer interest.

mlasak:
We are in a unsatisfying situation. Using fetch API with chunked transfer encoding there is currently no way to get the time when the chunk reception has started, see https://github.com/whatwg/streams/issues/1126 

When i stumbled upon HTTP Trailer [1] i shortly believed that we could solve our issue with it. The idea is to send a list with timestamps when each single chunk was put into "pipe" at server side. With some calculation we would be then enabled to calculate exact e2e network throughput. 

But unfortunately, fetch API doesn't support trailers. 
If you need a use case: we are in need for exact throughput calculation in low latency video streaming. 

Is there any chance for Trailer support in fetch API in future?

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer

annevk:
There's a chance, otherwise this would be closed. It does however require folks to be willing to implement it and drive the associated work that is required and that thus far has not materialized.

mnot:
Note that HTTP provides no guarantee that chunks are preserved end-to-end; furthermore, many (most?) HTTP APIs don't guarantee that chunk delineation is exposed accurately.

davidben:
I don't think that scenario is a good justification for adding trailers to the platform. It seems much better addressed by looking at timing and when bytes come out of the stream. Additionally, with the various layers in between JavaScript and HTTP-level chunking (TLS, QUIC, HTTP/2 and HTTP/3 frames, the browser's own mechanisms), guarantees about chunk boundaries wouldn't be coherent anyway.

mlasak:
> Note that HTTP provides no guarantee that chunks are preserved end-to-end; furthermore, many (most?) HTTP APIs don't guarantee that chunk delineation is exposed accurately.

[...]

PR #1406, opened 2022-02-27
Add server-timing to timing info struct

noamr:
Will be used by https://github.com/w3c/server-timing/pull/87

Decode and split `Server-Timing` headers, and append the raw headers to the `fetch timing info`. The server timing spec would later parse these headers and expose them.

   * Already implemented
  
   * Rigorously tested in https://github.com/web-platform-tests/wpt/tree/master/server-timing
   * N/a

***
<a href="https://whatpr.org/fetch/1406.html" title="Last updated on Apr 20, 2022, 11:33 AM UTC (6d47c8f)">Preview</a> | <a href="https://whatpr.org/fetch/1406/3a65729...6d47c8f.html" title="Last updated on Apr 20, 2022, 11:33 AM UTC (6d47c8f)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4314,7 +4314,8 @@ steps:
    <a for="fetch params">timing info</a>.</p></li>
   </ol>
  </li>
- <li><p>Otherwise, set <var>response</var>'s <a for=response>timing info</a>'s
+ <li><p>Otherwise, if <var>response</var>'s <a for=response>URL</a>'s <a for=url>scheme</a> is
+ "<code>https</code>", then set <var>response</var>'s <a for=response>timing info</a>'s
  <a for="fetch timing info">server timing headers</a> to the result of
  <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>` from
  <var>response</var>'s <a for=response>header list</a>.</p></li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4314,11 +4314,16 @@ steps:
    <a for="fetch params">timing info</a>.</p></li>
   </ol>
  </li>
- <li><p>Otherwise, if <var>response</var>'s <a for=response>URL</a>'s <a for=url>scheme</a> is
- "<code>https</code>", then set <var>response</var>'s <a for=response>timing info</a>'s
- <a for="fetch timing info">server timing headers</a> to the result of
- <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>` from
- <var>response</var>'s <a for=response>header list</a>.</p></li>
+ <li>
+  <p>Otherwise, if <var>response</var>'s <a for=response>URL</a>'s <a for=url>scheme</a> is
+  "<code>https</code>", then set <var>response</var>'s <a for=response>timing info</a>'s
+  <a for="fetch timing info">server timing headers</a> to the result of
+  <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>` from
+  <var>response</var>'s <a for=response>header list</a>.
+
+  <p>The user agent may decides to expose `<code>Server-Timing</code>` headers to
+  "<code>http</code>" requests as well.
+ </li>
 
  <li>
   <p>Let <var>processResponseEndOfBody</var> be the following steps:
--- a/fetch.bs
+++ b/fetch.bs
@@ -4315,14 +4315,14 @@ steps:
   </ol>
  </li>
  <li>
-  <p>Otherwise, if <var>response</var>'s <a for=response>URL</a>'s <a for=url>scheme</a> is
-  "<code>https</code>", then set <var>response</var>'s <a for=response>timing info</a>'s
-  <a for="fetch timing info">server timing headers</a> to the result of
-  <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>` from
-  <var>response</var>'s <a for=response>header list</a>.
+  <p>Otherwise, if <var>fetchParams</var>'s <a for="fetch params">request</a>'s
+  <a for=request>client</a> is a <a>secure context</a>, then set <var>response</var>'s
+  <a for=response>timing info</a>'s <a for="fetch timing info">server timing headers</a> to the
+  result of <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>`
+  from <var>response</var>'s <a for=response>header list</a>.
 
   <p>The user agent may decides to expose `<code>Server-Timing</code>` headers to
-  "<code>http</code>" requests as well.
+  non-secure contexts requests as well.
  </li>
 
  <li>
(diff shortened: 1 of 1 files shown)


annevk (commented):
Why can't the calling specification look at the response itself?

Also, I thought these headers might appear as trailer headers?

noamr:
> Why can't the calling specification look at the response itself?

We didn't want to pass the whole response to `resource timing`. We're passing only the `timing info`.

> Also, I thought these headers might appear as trailer headers?
They can, but I wanted to do this in iterations.

annevk:
So dealing with them in trailer headers doesn't have implementer agreement? It seems a little weird for Fetch to process this header as it doesn't seem to be related to information hiding, but if this is the only thing Resource Timing needs I guess it makes sense. Perhaps we should add some comments though that in case this changes we need to change the setup.

noamr:
> So dealing with them in trailer headers doesn't have implementer agreement? It seems a little weird for Fetch to process this header as it doesn't seem to be related to information hiding, but if this is the only thing Resource Timing needs I guess it makes sense. Perhaps we should add some comments though that in case this changes we need to change the setup.

Added a comment, is that sufficient? Not sure if any more clarification is needed at this point.

annevk:
"So dealing with them in trailer headers doesn't have implementer agreement?" is something I'd still like an answer to. The note also doesn't tackle this adequately I think as trailers can be received at a much later point so you probably want a separate data structure for them.

I also didn't really mean to ask for note, but rather a comment that indicates we might want to change approach if we end up having to inspect and parse other headers for Resource Timing. At that point it might become better to just pass Resource Timing the header list.

noamr:
> "So dealing with them in trailer headers doesn't have implementer agreement?" is something I'd still like an answer to. The note also doesn't tackle this adequately I think as trailers can be received at a much later point so you probably want a separate data structure for them.

I don't think that would be necessary. Resource timing is only reported after response end, at which point those would be available. But in any case, I would defer this to when there's consensus about trailers.

> 
> I also didn't really mean to ask for note, but rather a comment that indicates we might want to change approach if we end up having to inspect and parse other headers for Resource Timing. At that point it might become better to just pass Resource Timing the header list.

Got it.

noamr:
See https://github.com/whatwg/fetch/issues/981 and https://github.com/w3c/server-timing/issues/68 for conversation about trailer suopport.

noamr:
@annevk anything missing in this?

annevk (approved):
I'd like @ddragana or @valenting to comment as well given that Firefox does support trailers here if I'm not mistaken, but this looks okay to me. I pushed some minor commits that I hope are acceptable.

I'm not entirely sure about the approach to processing this header being shared across two specifications, but I also cannot really think of a better way of doing it.

noamr:
> I'd like @ddragana or @valenting to comment as well given that Firefox does support trailers here if I'm not mistaken, but this looks okay to me. I pushed some minor commits that I hope are acceptable.
> 
> 

Thanks, of course they are.

> 
> I'm not entirely sure about the approach to processing this header being shared across two specifications, but I also cannot really think of a better way of doing it.

I agree it's not ideal but it's also how it looks like in implementations, I believe it's the the lesser of evils.

annevk:
The other thing I realized looking at Firefox's implementation a bit more is that it limits this feature to secure contexts. That's probably something we want to account for here and not on the API side.

noamr:
> The other thing I realized looking at Firefox's implementation a bit more is that it limits this feature to secure contexts. That's probably something we want to account for here and not on the API side.

Done, and added a [WPT](https://github.com/web-platform-tests/wpt/pull/33173). Chromium implementation, however, currently allows `Server-Timing` in http. (Safari doesn't support `Server-Timing` at all).
@yoavweiss WDYT about limiting ST to HTTPS?

yoavweiss:
I believe that would be a breaking change from Chromium's perspective. Is there any particular reason why Server-Timing should be limited to secure contexts? Any security/privacy issues that arise from the fact that it is not?
Also, what are the benefits of enforcing this limitation on the Fetch side?

valenting:
Looking back at the [intent to ship](https://groups.google.com/g/mozilla.dev.platform/c/MSzaY7_4mvg/m/hGpUlTzxAgAJ) I think the main motivation for limiting this to secure origins was [Mozilla's policy](https://blog.mozilla.org/security/2018/01/15/secure-contexts-everywhere/) of doing so.

I also seem to recall a discussion where trailers were causing problems with middleboxes over HTTP, but I can't find any references to that at this time.

noamr:
> Looking back at the [intent to ship](https://groups.google.com/g/mozilla.dev.platform/c/MSzaY7_4mvg/m/hGpUlTzxAgAJ) I think the main motivation for limiting this to secure origins was [Mozilla's policy](https://blog.mozilla.org/security/2018/01/15/secure-contexts-everywhere/) of doing so.
> 
> I also seem to recall a discussion where trailers were causing problems with middleboxes over HTTP, but I can't find any references to that at this time.

Seems like the conversation [didn't result in a consensus](https://github.com/w3c/server-timing/issues/54), which is unfortunate since currently server-timing is not interoperable in that way.

annevk:
>  Also, what are the benefits of enforcing this limitation on the Fetch side?

Mainly principle of least privilege.

yoavweiss:
> > Also, what are the benefits of enforcing this limitation on the Fetch side?
> 
> Mainly principle of least privilege.

OK. At the same time, that doesn't seem in line with what some implementations are doing.

noamr:
I amended the PR to reflect the Gecko/Chromium status quo: the headers are always sent for `https`, and make `http` something that's up to the user-agent's discretion. It's not perfect in terms of interoperability but at least it makes the part that are in consensus specified.

fetch.bs
```diff
@@ -4296,6 +4299,16 @@ steps:
...
  </li>
+ <li>
+  <p>Otherwise, if <var>response</var>'s <a for=response>URL</a>'s <a for=url>scheme</a> is
+  "<code>https</code>", then set <var>response</var>'s <a for=response>timing info</a>'s
+  <a for="fetch timing info">server timing headers</a> to the result of
+  <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>` from
+  <var>response</var>'s <a for=response>header list</a>.
+
+  <p>The user agent may decides to expose `<code>Server-Timing</code>` headers to
+  "<code>http</code>" requests as well.
```
annevk, line 4310: This should be a secure context check of sorts as these headers are fine on `localhost`.

noamr: @annevk I could use some guidance here :)

annevk: Can't we check if request's client is a secure context? (With that referencing the definition in HTML.) That should account for `localhost` as well as nested contexts and such.

noamr: Gotcha. done

annevk (approved):
@yoavweiss want to have another look?

yoavweiss (approved):
LGTM

PR #1411, opened 2022-03-10
Perform a CSP check when consuming preloaded response

noamr:
Closes https://github.com/whatwg/html/issues/7686

   * Implemented in WebKit & Chromiun
   * https://github.com/web-platform-tests/wpt/pull/33109
   * https://github.com/web-platform-tests/wpt/pull/33205
   * Chrome: N/A
   * Firefox: [1759002](https://bugzilla.mozilla.org/show_bug.cgi?id=1759002)
   * Safari: [238019](https://bugs.webkit.org/show_bug.cgi?id=238019)

***
<a href="https://whatpr.org/fetch/1411.html" title="Last updated on Mar 17, 2022, 1:07 PM UTC (808cb65)">Preview</a> | <a href="https://whatpr.org/fetch/1411/4561d5f...808cb65.html" title="Last updated on Mar 17, 2022, 1:07 PM UTC (808cb65)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3895,19 +3895,19 @@ the request.
    <a for="environment settings object">origin</a>.
 
    <li><p>Let <var>onPreloadedResponseAvailable</var> be an algorithm that runs the following
-   step given a <a for=/>response</a> <var>response</var>: Set <var>fetchParams</var>'s
+   step given a <a for=/>response</a> <var>response</var>: set <var>fetchParams</var>'s
    <a for="fetch params">preloaded response candidate</a> to <var>response</var>.
 
    <li><p>Let <var>foundPreloadedResource</var> be the result of invoking
-   <a>consume a preloaded resource</a> for <var>req</var>'s <a for=request>window</a>,
-   given <var>req</var>'s <a for=request>URL</a>, <var>req</var>'s <a for=request>destination</a>,
-   <var>req</var>'s <a for=request>mode</a>, <var>req</var>'s <a for=request>credentials mode</a>,
-   <var>req</var>'s <a for=request>integrity metadata</a>, and
-   <var>onPreloadedResponseAvailable</var>.
+   <a>consume a preloaded resource</a> for <var>request</var>'s <a for=request>window</a>, given
+   <var>request</var>'s <a for=request>URL</a>, <var>request</var>'s <a for=request>destination</a>,
+   <var>request</var>'s <a for=request>mode</a>, <var>request</var>'s
+   <a for=request>credentials mode</a>, <var>request</var>'s <a for=request>integrity metadata</a>,
+   and <var>onPreloadedResponseAvailable</var>.
 
    <li><p>If <var>foundPreloadedResource</var> is true and <var>fetchParams</var>'s
    <a for="fetch params">preloaded response candidate</a> is null, then set <var>fetchParams</var>'s
-   <a for="fetch params">preloaded response candidate</a> to "<code>pending</code>".</p></li>
+   <a for="fetch params">preloaded response candidate</a> to "<code>pending</code>".
   </ol>
  </li>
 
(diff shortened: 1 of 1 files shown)


annevk:
From the discussion in that HTML issue this doesn't seem like the correct fix as the other policies (e.g., Mixed Content blocking) do not end up being enforced.

noamr:
> From the discussion in that HTML issue this doesn't seem like the correct fix as the other policies (e.g., Mixed Content blocking) do not end up being enforced.

Unless I am missing something, the other policies are not modifiable after the document is created, so there is no need to re-check them. That's true at least for mixed content and was discussed in the HTML issue.

annevk:
* It seems that "Run report Content Security Policy violations for request" wouldn't happen.
* Mike's scenario of scheme upgrades seems problematic. Yes, perhaps there's an HTTP cache entry, but is that guaranteed? And either way, wouldn't it result in a service worker being asked whereas if the scheme upgrade happened before that wouldn't be the case? (Scheme upgrades is also what Mixed Content Level 2 does.)

I haven't done a complete check, but this seems concerning and also suggests a potential mismatch with implementations that might cause further issues in the future.

noamr:
> * It seems that "Run report Content Security Policy violations for request" wouldn't happen.

Does that need to happen again? The request in question had already been processed.

But I agree that this is indeed concerning.  Not so much for the preload case where we can define that all the checks happen a second time, but I'm more concerned about other memory cache scenarios that are not clearly specified.

noamr:
> * It seems that "Run report Content Security Policy violations for request" wouldn't happen.
> * Mike's scenario of scheme upgrades seems problematic. Yes, perhaps there's an HTTP cache entry, but is that guaranteed? And either way, wouldn't it result in a service worker being asked whereas if the scheme upgrade happened before that wouldn't be the case? (Scheme upgrades is also what Mixed Content Level 2 does.)
> 
> I haven't done a complete check, but this seems concerning and also suggests a potential mismatch with implementations that might cause further issues in the future.

I refactored preload handling - where a preload scenario goes through the same checks as any other fetch, but uses the preloaded response instead of seeking it in network or cache. This seems to be the safest, and allows us to incorporate different memory cache scenarios later by folding them into `consume a preloaded resource`. WDYT?

fetch.bs
```diff
@@ -3855,9 +3858,12 @@ the request.
    with <var>request</var>'s <a for=request>client</a>'s
    <a for="environment settings object">origin</a>.
 
+   <li><p>Set <var>fetchParams</var>'s <a for="fetch params">preloaded response candidate</a> to
+   "<code>pending</code>".
```
annevk, line 3862: How about instead of having this step, we only set state to "pending" in the last step if _foundPreloadedResource_ is true? Although I guess we'd have to make sure it's not a response already in case it was all available synchronously.

So you'd get

> If _foundPreloadedResource_ is true and _fetchParams_'s preloaded response candidate is null, set _fetchParams_'s preloaded response candidate to "pending".

That seems a little clearer to me.

annevk (commented):
This looks really solid. This would result in reporting twice, right? Is that a case we might want to adjust to not do that or is it fine?

noamr:
> This looks really solid. This would result in reporting twice, right? Is that a case we might want to adjust to not do that or is it fine?

I think it's ok that it reports twice if both the preload and actual request are in violation but I will add a test to verify

fetch.bs
noamr: > How about instead of having this step, we only set state to "pending" in the last step if _foundPreloadedResource_ is true? Although I guess we'd have to make sure it's not a response already in case it was all available synchronously.
> 
> 
> 
> So you'd get
> 
> 
> 
> > If _foundPreloadedResource_ is true and _fetchParams_'s preloaded response candidate is null, set _fetchParams_'s preloaded response candidate to "pending".
> 
> 
> 
> That seems a little clearer to me.

Sure, I did it this way because of the synchronous case but your proposal handles it more readably

fetch.bs
```diff
@@ -4001,6 +4008,17 @@ steps:
   corresponding to the first matching statement:
 
   <dl class=switch>
+   <dt><var>fetchParams</var>'s <a for="fetch params">preloaded response candidate</a> is not null
+
+   <dd>
+    <ol>
+     <li><p>Wait until <var>fetchParams</var>'s
+     <a for="fetch params">preloaded response candidate</a> is not "<code>pending</code>".
+
```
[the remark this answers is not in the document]
noamr: It could be null though

noamr:
> This looks really solid. This would result in reporting twice, right? Is that a case we might want to adjust to not do that or is it fine?

I wrote a [WPT](https://github.com/web-platform-tests/wpt/pull/33205) for this - apparently Chrome reports twice and Firefox/Safari reports once. I wonder if it's necessary to report twice or we can leave this optional.

annevk:
@noamr I guess we could make it conditional upon preloaded response candidate, but I think what Chrome does is reasonable. You might also get slightly different failures which could be interesting to know about. Let's file a bug against Safari? Firefox's behavior is likely due to it not enforcing policies at all.

I pushed a smallish fixup commit that also changes the variable name from _req_ to _request_. I think that ended up being wrong in the original PR due to the text moving around. Could you confirm it's now correct?

noamr:
> @noamr I guess we could make it conditional upon preloaded response candidate, but I think what Chrome does is reasonable. You might also get slightly different failures which could be interesting to know about. Let's file a bug against Safari? Firefox's behavior is likely due to it not enforcing policies at all.
> 

SG, posted a [WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=238019) and we can merge the WPT alongside the PR AFAIC

PR #1417, opened 2022-03-22
Specify connection timing for HTTP/3

noamr:
- `secure connection start` is identical to `connection start`
  (TLS and connecting are part of the same flow in HTTP/3)

- `connection end` is right after the handshake

This corresponds to current Chromium implementation.

Closes https://github.com/w3c/navigation-timing/issues/165

   * …
   * …
   * Currently WPT doesn't have support for H3 (except for `WebTransport`, TBD)
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1417.html" title="Last updated on Apr 6, 2022, 10:54 AM UTC (50514eb)">Preview</a> | <a href="https://whatpr.org/fetch/1417/92b3578...50514eb.html" title="Last updated on Apr 6, 2022, 10:54 AM UTC (50514eb)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2644,6 +2644,11 @@ boolean <var>http3Only</var>, run these steps:
  <a for="connection timing info">secure connection start time</a> should be the result of calling
  <a for=/>unsafe shared current time</a> immmediately before starting the handshake process to
  secure <var>connection</var>. [[!TLS]]
+
+ <li><p>If <var>connection</var> is an HTTP/3 connection, <var>timingInfo</var>'s
+ <a for="connection timing info">connection start time</a> and <var>timingInfo</var>'s
+ <a for="connection timing info">secure connection start time</a> should be equal, as in HTTP/3
+ the handshake process is performed as part of the initial connection setup. [[!HTTP3]]
 </ul>
 
 <p class=note>The <a for=/>clamp and coarsen connection timing info</a> algorithm ensures that
--- a/fetch.bs
+++ b/fetch.bs
@@ -2647,8 +2647,9 @@ boolean <var>http3Only</var>, run these steps:
 
  <li><p>If <var>connection</var> is an HTTP/3 connection, <var>timingInfo</var>'s
  <a for="connection timing info">connection start time</a> and <var>timingInfo</var>'s
- <a for="connection timing info">secure connection start time</a> should be equal, as in HTTP/3
- the handshake process is performed as part of the initial connection setup. [[!HTTP3]]
+ <a for="connection timing info">secure connection start time</a> must be equal. (In HTTP/3
+ the secure transport handshake process is performed as part of the initial connection setup.)
+ [[!HTTP3]]
 </ul>
 
 <p class=note>The <a for=/>clamp and coarsen connection timing info</a> algorithm ensures that


fetch.bs
```diff
@@ -2643,7 +2647,10 @@ boolean <var>http3Only</var>, run these steps:
  <li><p>If a secure transport is used, <var>timingInfo</var>'s
```
ianswett, line 2647: The above paragraphs already talk about secure connections implicitly, so maybe some of this should be moved up?

For HTTP/3, I'm not sure you need to add anything, though if you wanted to say "If a secure transport is used, such as TLS over TCP or QUIC, ..." that could clarify that you're including both.

fetch.bs
```diff
@@ -2630,6 +2630,10 @@ boolean <var>http3Only</var>, run these steps:
    <li><p>If the user agent waits for full handshake completion to send the request, this interval
    includes the full TLS handshake even if other requests were sent using early data on
    <var>connection</var>.
+
+   <li><p>If the connection is an HTTP/3 <a>connection</a>,
+   <var>timingInfo</var>'s <a for="connection timing info">connection end time</a> should represent
+   the time right after the TLS handshake. [[!HTTP3]]
```
ianswett, line 2636: I think HTTP/3 is pretty much the same as other versions of HTTP over TLS 1.3 in that if you send early data, you don't wait for the ServerHello?  If you want to use a term from the RFC for when all requests can be sent, I'd use 'Handshake Complete', which is the point at which the client has the final 1-RTT keys.

I'll note that early data can be rejected, and it might be worth clarifying whether the time the early data is initially sent should be used or after rejection.  I suspect the former, but it's worth being clear I think.

Note: 'handshare' is misspelled on line 2626.

ianswett (commented):
I'm new to this text, but some comments.

I believe the text that references TLS could probably be more specific, but I'd prefer to defer that to someone else.

fetch.bs
noamr: Gotcha. The main important addition is that in QUICK `secure connection start` and `connection start` are identical. Perhaps I should just add that and forget the rest.

ianswett: Thanks, saying that explicitly makes tons of sense to me.

LPardue:
I am not an expert in fetch but making sure there is no difference in timing of secure connect start and connect start for QUIC-based HTTP makes sense to me.

yoavweiss (approved):
LGTM

fetch.bs
```diff
@@ -2644,6 +2644,11 @@ boolean <var>http3Only</var>, run these steps:
  <a for="connection timing info">secure connection start time</a> should be the result of calling
  <a for=/>unsafe shared current time</a> immmediately before starting the handshake process to
  secure <var>connection</var>. [[!TLS]]
+
+ <li><p>If <var>connection</var> is an HTTP/3 connection, <var>timingInfo</var>'s
+ <a for="connection timing info">connection start time</a> and <var>timingInfo</var>'s
+ <a for="connection timing info">secure connection start time</a> should be equal, as in HTTP/3
+ the handshake process is performed as part of the initial connection setup. [[!HTTP3]]
```
annevk, line 2651: It's not clear to me why this is a "should" and not a "must". I also separated the justification as we don't typically do those as part of a requirement.

We already expose the connection type through ALPN negotiated protocol so that's not a new concern.
```suggestion
 <li><p>If <var>connection</var> is an HTTP/3 connection, <var>timingInfo</var>'s
 <a for="connection timing info">connection start time</a> and <var>timingInfo</var>'s
 <a for="connection timing info">secure connection start time</a> must be equal. (In HTTP/3
 the secure transport handshake process is performed as part of the initial connection setup.)
 [[!HTTP3]]
```

annevk (commented):
I have some nits on the wording and whether this should be should or must, but otherwise looks okay to me.

noamr:
> I have some nits on the wording and whether this should be should or must, but otherwise looks okay to me.

Thanks, committed the suggestion.

annevk:
@ianswett could you take a final look?

ianswett (commented):
I can't approve, but LGTM.

noamr:
> Test secureConnectionStart in HTTP/3 case web-platform-tests/wpt#33549

Done https://github.com/web-platform-tests/wpt/issues/33549

PR #1418, opened 2022-03-22
Integrity-metadata should not be a preload key

noamr:
In conjunction with https://github.com/whatwg/html/pull/7738

   * …
   * …
   * Covered thoroughly by https://wpt.live/preload/subresource-integrity.html
   * Chrome: …
   * Firefox: …
   * Safari: …

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3902,8 +3902,7 @@ the request.
    <a>consume a preloaded resource</a> for <var>request</var>'s <a for=request>window</a>, given
    <var>request</var>'s <a for=request>URL</a>, <var>request</var>'s <a for=request>destination</a>,
    <var>request</var>'s <a for=request>mode</a>, <var>request</var>'s
-   <a for=request>credentials mode</a>, <var>request</var>'s <a for=request>integrity metadata</a>,
-   and <var>onPreloadedResponseAvailable</var>.
+   <a for=request>credentials mode</a>, and <var>onPreloadedResponseAvailable</var>.
 
    <li><p>If <var>foundPreloadedResource</var> is true and <var>fetchParams</var>'s
    <a for="fetch params">preloaded response candidate</a> is null, then set <var>fetchParams</var>'s


annevk:
I'm not sure I understand the tests. Maybe @domfarolino can help explain how to read them. For instance, "Same-origin script with non-matching digest does not re-use preload with matching digest." suggests to me that integrity metadata is looked at when looking at preloads to consume. Or am I missing something?

noamr:
> Same-origin script with non-matching digest

That test has a good digest for preload and a bad one for consume, and it shows that the preload is ignored.
"not CORS request, with correct hash" tests the main use-case here; that even when the consumer doesn't specify integrity metadata, the preload (which does have integrity metadata) can be used.

There is a test missing to show that if both preload and consumer have matching integrity but with a different SHA the preload should be consumed. It's an edge case that IIRC Chromium currently doesn't implement due to it being esoteric, I can add a test for it and open a bug.

noamr:
> > That test has a good digest for preload and a bad one for consume, and it shows that the preload is ignored.
> 
> Why would the preload be ignored though? How does that follow from the specification text?

Actually you're right. The test shows that there would be two requests, while in the spec text there would be one request and the image would fail.

I think a spec text that would correspond with the tests (and current implementation) is to keep fetch as is, but to make integrity meta data an "optional" part of the preload key, meaning that if it's not present then the lookup ignores integrity metadata.

This is also reflected by "Cross-origin script, not CORS request, with hash mismatch" - there is only one (failing) request that also fails the consumer even though the consumer doesn't specify a hash.

noamr:
Amended https://github.com/whatwg/html/pull/7738 to match current implementations, accounting for the edge-case mismatch

noamr:
> As the parameters of https://html.spec.whatwg.org/multipage/links.html#consume-a-preloaded-resource are not modified by https://github.com/whatwg/html/pull/7738, I suspect this change to the caller-side of #consume-a-preloaded-resource is not needed? (The discussion above is still valid/useful for https://github.com/whatwg/html/pull/7738)

Yes, that's accurate. I am leaving this open for the discussion

noamr:
Closing as there is no action item for the fetch spec, though some of the discussion is useful for reference.

Issue #1421, opened 2022-03-29
Editorial: using serialized origin in TAO check and "null"


noamr:
In https://fetch.spec.whatwg.org/#tao-check, the TAO header is checked to see if it includes request's serialized origin.
However, for redirect-tainted requests, https://fetch.spec.whatwg.org/#serializing-a-request-origin will return the string `"null"`.
This would imply that `Timing-Allow-Origin: origin-a.com, null` would succeed for redirect-tainted requests.

Suggesting not to use `serializing-a-request-origin` here.

annevk:
This doesn't seem editorial and I'm also not entirely sure why this is wrong. CORS works the same way.

noamr:
> This doesn't seem editorial and I'm also not entirely sure why this is wrong. CORS works the same way.
Not wrong as in, `Timing-Allow-Origin: origin-a.com, null` is valid for redirect-tainted requests? That's surprising to me. If that's the case it should be better WPT-covered.

annevk:
It works for CORS as well (although there was some argument we should not accept it for credentialed requests, but I think the specification still does).

noamr:
> It works for CORS as well (although there was some argument we should not accept it for credentialed requests, but I think the specification still does).

Got it. I'll amend the iframe-TAO PR to account for this and close this issue. Will add tests separately.

Issue #1221, opened 2021-04-24
TAO check and child frame navigations


zetafunction:
The steps for performing a TAO check are defined here:
https://fetch.spec.whatwg.org/#tao-check

1. If request’s timing allow failed flag is set, then return failure.
2. If request’s response tainting is `"basic"`, then return success.
3. Let values be the result of getting, decoding, and splitting `Timing-Allow-Origin` from response’s header list.
4. If values contains `"*"`, then return success.
5. If values contains the result of serializing a request origin with request, then return success.
6. Return failure.

If I interpret https://fetch.spec.whatwg.org/#ref-for-concept-request-response-tainting%E2%91%A4 correctly, response tainting is always `"basic"` for navigate requests.

Does this imply the TAO check would always pass when reporting the resource for a child frame navigation to the parent frame, even if cross-origin?

A separate question would be if the TAO check needs to separately account for the fact that the parent frame's origin may not be the same as the request's origin. Reading https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object and https://w3c.github.io/navigation-timing/#dfn-create-the-navigation-timing-entry, I don't see where the behavior of reporting navigation timing to the parent Document is defined.

zetafunction:
@npm1 @domfarolino @yoavweiss

annevk:
Yeah, the navigate algorithm needs to perform the relevant TAO checks itself. It also follows redirects manually so I don't think Fetch can meaningfully account for it, but maybe? (I haven't followed the recent timing changes to HTML, I guess I should review those.)

cc @noamr @domenic

noamr:
> Yeah, the navigate algorithm needs to perform the relevant TAO checks itself. It also follows redirects manually so I don't think Fetch can meaningfully account for it, but maybe? (I haven't followed the recent timing changes to HTML, I guess I should review those.)
> 
> cc @noamr @domenic

Navigation entries are not (currently) reported to the parent document, so they don't need a TAO check.
IFRAME navigation is reported to the parent as a regular resource timing entry with a TAO check, though that reporting is yet to be specified (it's in the TODO list for HTML element reporting).

zetafunction:
> IFRAME navigation is reported to the parent as a regular resource timing entry with a TAO check, though that reporting is yet to be specified (it's in the TODO list for HTML element reporting).

How should response tainting be calculated for this separate TAO check?

annevk:
[...]

PR #1422, opened 2022-03-29
Perform TAO check for nested navigations

noamr:
Closes https://github.com/whatwg/fetch/issues/1221
Closes https://github.com/whatwg/fetch/issues/1421

   * Already implemented, it's an implemented piece that was missing in the spec.
   * Existing tests: https://wpt.fyi/results/resource-timing/cross-origin-iframe.html?label=experimental&label=master&aligned
   * N/A

***
<a href="https://whatpr.org/fetch/1422.html" title="Last updated on Apr 28, 2022, 7:03 AM UTC (67fbcc2)">Preview</a> | <a href="https://whatpr.org/fetch/1422/14898c0...67fbcc2.html" title="Last updated on Apr 28, 2022, 7:03 AM UTC (67fbcc2)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6087,20 +6087,33 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
 <var>response</var>, run these steps:
 
 <ol>
- <li><p>If <var>request</var>'s <a for=request>timing allow failed flag</a> is set, then return
- failure.
-
- <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>basic</code>", then
- return success.
+ <li><p>If <var>request</var>'s <a for=request>timing allow state</a> is "<code>failure</code>",
+ then return "<code>failure</code>".
 
  <li><p>Let <var>values</var> be the result of
  <a for="header list">getting, decoding, and splitting</a> `<code>Timing-Allow-Origin</code>` from
  <var>response</var>'s <a for=response>header list</a>.
 
- <li><p>If <var>values</var> <a for=list>contains</a> "<code>*</code>", then return success.
+ <li><p>If <var>values</var> <a for=list>contains</a> "<code>*</code>", then return
+ "<code>success</code>".
 
  <li><p>If <var>values</var> <a for=list>contains</a> the result of
  <a>serializing a request origin</a> with <var>request</var>, then return success.
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>mode</a> is "<code>navigate</code>", and
+  <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> is not
+  <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then
+  return failure.
+
+  <p class="note">This is necessary for nested navigations - e.g. cross-origin iframes.
+  In that case, <var>request</var>'s <a for=request>origin</a> would be the container's
+  <a for=document>origin</a>, and the <a>TAO check</a> would fail. Since navigation timing never
+  validates the results of the <a>TAO check</a>, the nested document would still have access to the
+  full timing information, but the container would not.
+
+ <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>basic</code>", then
+ return success.
 
  <li><p>Return failure.
 </ol>
--- a/fetch.bs
+++ b/fetch.bs
@@ -6087,8 +6087,8 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
 <var>response</var>, run these steps:
 
 <ol>
- <li><p>If <var>request</var>'s <a for=request>timing allow state</a> is "<code>failure</code>",
- then return "<code>failure</code>".
+ <li><p>If <var>request</var>'s <a for=request>timing allow failed flag</a> is set, then return
+ failure.
 
  <li><p>Let <var>values</var> be the result of
  <a for="header list">getting, decoding, and splitting</a> `<code>Timing-Allow-Origin</code>` from
--- a/fetch.bs
+++ b/fetch.bs
@@ -6106,8 +6106,8 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
   return failure.
 
   <p class="note">This is necessary for nested navigations - e.g. cross-origin iframes.
-  In that case, <var>request</var>'s <a for=request>origin</a> would be the container's
-  <a for=document>origin</a>, and the <a>TAO check</a> would fail. Since navigation timing never
+  In that case, <var>request</var>'s <a for=request>origin</a> would be the container document's
+  <a for=Document>origin</a>, and the <a>TAO check</a> would fail. Since navigation timing never
   validates the results of the <a>TAO check</a>, the nested document would still have access to the
   full timing information, but the container would not.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -6094,8 +6094,7 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
  <a for="header list">getting, decoding, and splitting</a> `<code>Timing-Allow-Origin</code>` from
  <var>response</var>'s <a for=response>header list</a>.
 
- <li><p>If <var>values</var> <a for=list>contains</a> "<code>*</code>", then return
- "<code>success</code>".
+ <li><p>If <var>values</var> <a for=list>contains</a> "<code>*</code>", then return success.
 
  <li><p>If <var>values</var> <a for=list>contains</a> the result of
  <a>serializing a request origin</a> with <var>request</var>, then return success.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -6030,14 +6033,31 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
  <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>basic</code>", then
  return success.
 
+ <li><p>Let <var>origin</var> be null if <var>request</var> has a
```
annevk, line 6036: You cannot initialize a variable twice. You want to use "Set ... to".

fetch.bs
```diff
@@ -6030,14 +6033,31 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+ <a for=request>redirect-tainted origin</a>, <var>request</var>'s <a for=request>origin</a>
+ otherwise.
+
+ <li><p>Return the result of performing a <a>TAO origin check</a> given <var>origin</var> and
+ <var>response</var>.
+</ol>
+
+<p>To perform a <dfn>TAO origin check</dfn> for a null or <span>origin</span>
```
annevk, line 6044: You want to use `<a for=/>` here and below, not `span`.

fetch.bs
```diff
@@ -6030,14 +6033,31 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+<var>origin</var> and a <span>response</span> <var>response</var>, run these steps:
+
+<ol>
+ <li><p>If <var>origin</var> is <a>same origin</a> with <var>response</var>'s
+ <a for=response>URL</a>'s <a for=url>origin</a>, then return success.
```
annevk, line 6049: Doesn't this do the wrong thing for A1 -> B -> A2 redirects?

annevk (commented):
I left a comment in #1421 as well. Not entirely sure about this.

noamr:
> I left a comment in #1421 as well. Not entirely sure about this.

Because of the lack of origin serialization?

Serialization and the specific notes aside, any reason to be unsure, or an alternative way to go about this?
The main issue is the "basic" response tainting due to this being a navigation. This timing info is both TAO protected and not; It's TAO protected only when reporting to the iframe.

annevk:
Sorry, I hadn't really looked at that part of the PR.

So we have a nested navigation response (after redirects) and want to see if it's okay to report its timings to its parent, which might be cross-origin. Why are redirects prior to the final navigation response not important here? Shouldn't they get a say as to whether reporting happens?

noamr:
> Sorry, I hadn't really looked at that part of the PR.
> 
> So we have a nested navigation response (after redirects) and want to see if it's okay to report its timings to its parent, which might be cross-origin. Why are redirects prior to the final navigation response not important here? Shouldn't they get a say as to whether reporting happens?

You're right, there is a missing part. Revising.

noamr:
> > Sorry, I hadn't really looked at that part of the PR.
> > So we have a nested navigation response (after redirects) and want to see if it's okay to report its timings to its parent, which might be cross-origin. Why are redirects prior to the final navigation response not important here? Shouldn't they get a say as to whether reporting happens?
> 
> You're right, there is a missing part. Revising.

Revised to follow how implementations go through the redirect chain again to TAO-check it vs. the container origin.

fetch.bs
```diff
@@ -6030,14 +6046,25 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
 
+ <li><p>Let <var>serializedOrigin</var> be the result of <a>serializing a request origin</a> given
+ <var>request</var>.
+
+ <li><p>Return the result of performing a <a>TAO origin check</a> given <var>origin</var> and
+ <var>response</var>.
+</ol>
+
+<p>To perform a <dfn>TAO origin check</dfn> for an <a for=/>ASCII string</a>
+<var>serializedOrigin</var> and a <span>response</span> <var>response</var>, run these steps:
```
annevk, line 6057: Still uses `span`.

fetch.bs
```diff
@@ -4399,15 +4402,23 @@ steps:
 
  <li><p>If <var>timingInfo</var> is null, then return.
 
- <li>
-  <p>If <var>response</var>'s <a for=response>timing allow passed flag</a> is not set, then:
+ <li><p>Let <var>taoStatus</var> be <code>success</code>.
```
annevk, line 4405: Can we make this "Let taoSucceeded be true" instead? That seems preferable to introducing new types.

fetch.bs
```diff
@@ -266,6 +266,9 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
 
  <dt><dfn export for="fetch timing info">final connection timing info</dfn> (default null)
  <dd>Null or a <a for=/>connection timing info</a>.
+
+ <dt><dfn for="fetch timing info">TAO checks</dfn> (default an empty list)
+ <dd>A list of algorithms accepting an <a for=/>ASCII string</a>
```
annevk, line 271: "A list of algorithms each of which takes an ASCII string and returns ???"

fetch.bs
```diff
@@ -4678,6 +4689,11 @@ these steps:
 
    <li><p>If the <a>TAO check</a> for <var>request</var> and <var>response</var> returns failure,
    then set <var>request</var>'s <a for=request>timing allow failed flag</a>.
+
+   <li><p>Otherwise, <a for=list>append</a> the following steps given <a for=/>ASCII string</a>
```
annevk, line 4693: given an*

annevk (commented):
I don't really understand the setup I'm afraid. Could you explain it?

* Does this require to keep all the intermediate responses in memory until "finalize and report timing" runs?
* Why does "finalize and report timing" now end up doing multiple TAO checks for everything, rather than just nested documents?
* Is the global in "finalize and report timing" the correct one?

noamr:
> I don't really understand the setup I'm afraid. Could you explain it?

We have to re-run the TAO checks when we know which global is gong to expose the RT entry. In actuality, we can this TAO check and drop the one we're currently running during a response (that's sort-of what implementations do today).

> * Does this require to keep all the intermediate responses in memory until "finalize and report timing" runs?

Only the bits of the response necessary for the TAO check (TAO header, origin, etc.). I can modify the TAO check algorithm to receive the minimal argument set instead. It should not have an observable difference.

> * Why does "finalize and report timing" now end up doing multiple TAO checks for everything, rather than just nested documents?

We can perform this final TAO check only when the global to report to is not the same as the original `client`. As above, it should not have an observable difference but perhaps it would more readable and optimized in advance.

> * Is the global in "finalize and report timing" the correct one?

Yes, it's the global that's going to expose the RT entry.

annevk:
Yeah, I think trying to make this more readable plus perhaps some notes would go a long way.

noamr:
> Yeah, I think trying to make this more readable plus perhaps some notes would go a long way.

I revised this completely:

- TAO would fail for cross-origin navigations without a proper TAO header
- Navigation timing anyway doesn't check TAO, so this TAO failure would only affect the resource timing entry in the container

fetch.bs
```diff
@@ -6102,6 +6099,21 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
  <li><p>If <var>values</var> <a for=list>contains</a> the result of
  <a>serializing a request origin</a> with <var>request</var>, then return success.
 
+ <li>
+  <p>If <var>request</var>'s <a for=request>mode</a> is "<code>navigate</code>", and
+  <var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> is not
+  <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, then
+  return failure.
```
annevk, line 6106: So this succeeds when A1 embeds A2. It does not succeed when A1 embeds B. It also succeeds when A1 embeds A2 and A2 redirects to B which redirects to A3. That case seems concerning.

This should probably also use "redirect-tainted origin"?

noamr: It wouldn't succeed if `A2` redirects to `B` which redirects to `A3`, as the TAO check is done in the `B` redirect response and would set the TAO failed flag. And it uses the [serialized request origin](https://fetch.spec.whatwg.org/#serializing-a-request-origin) which takes redirect-tainting into account.

All of these concerns are baked into the TAO check already, all my patch does is ignore the `basic` [response tainting](https://fetch.spec.whatwg.org/#concept-request-response-tainting) when in a nested navigation.

annevk (approved):

annevk:
Thanks for explaining the "navigate" case in more detail. I relatively positive this works now. Once the remaining checkboxes in OP are complete this can land I think.

noamr:
> Thanks for explaining the "navigate" case in more detail. I relatively positive this works now. Once the remaining checkboxes in OP are complete this can land I think.

Great! Done.

PR #1424, opened 2022-04-15
Export main fetch

arichiv:
I need to reference this in https://github.com/WICG/client-hints-infrastructure/pull/106

***
<a href="https://whatpr.org/fetch/1424.html" title="Last updated on Apr 15, 2022, 2:27 PM UTC (53445ea)">Preview</a> | <a href="https://whatpr.org/fetch/1424/583986a...53445ea.html" title="Last updated on Apr 15, 2022, 2:27 PM UTC (53445ea)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4002,7 +4002,7 @@ the request.
 
 <h3 id=main-fetch>Main fetch</h3>
 
-<p>To <dfn id=concept-main-fetch>main fetch</dfn>, given a <a for=/>fetch params</a>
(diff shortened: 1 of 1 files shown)


annevk:
I don't think we should export terms for the sole purpose of enabling monkey patches. I recommend using an actual link or referencing the fetch algorithm and specify the details in prose.

domenic:
For monekypatches I use the `<a spec=FETCH>main fetch</a>` syntax which makes it clearer that you're not doing a supported thing.

Issue #1077, opened 2020-08-18
Handle blob URLs without a backing blob


annevk:
Presumably these should result in a network error. The current specification doesn't state anything and it's not tested either.

See the README for instructions on where to add tests. In the specification the "blob" case of https://fetch.spec.whatwg.org/#scheme-fetch needs updating to handle current URL’s blob URL entry being null.

M-ZubairAhmed:
@annevk can i look into this one, just trying to understand what needs to be done here

annevk:
Hey @M-ZubairAhmed, there are basically two separate tasks:

1. Create test cases for `blob:` URLs not backed by a `Blob` object. E.g., trying to fetch `blob:not-backed-by-a-blob` would be such an example. Presumably this should result in a network error.
2. Update the specification to check for that condition as per OP.

Thanks for looking into it! Let me know if you have further questions, I'm also on IRC: https://whatwg.org/irc.

Krishnaraj-shankar:
Hi @annevk ,Im a begginer . Would you guide me to solve this issue..

PR #1435, opened 2022-05-05
Only allow blob: URLs that have a backing blob

dlrobertson:
Document the need to return a network error on fetch if the blob URL has
no backing blob.

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/33941
   * Chrome: …
   * Firefox: …
   * Safari: …

Fixes: https://github.com/whatwg/fetch/issues/1077

***
<a href="https://whatpr.org/fetch/1435.html" title="Last updated on May 13, 2022, 1:13 PM UTC (3fccfa7)">Preview</a> | <a href="https://whatpr.org/fetch/1435/3ab5102...3fccfa7.html" title="Last updated on May 13, 2022, 1:13 PM UTC (3fccfa7)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4503,12 +4503,14 @@ steps:
     <a for="fetch params">canceled</a>:
 
     <ol>
-     <li><p>Let <var>blob</var> be <var>request</var>'s <a for=request>current URL</a>'s
-     <a for=url>blob URL entry</a>'s <a for="blob URL entry">object</a>.
+     <li><p>Let <var>blobURLEntry</var> be <var>request</var>'s <a for=request>current URL</a>'s
+     <a for=url>blob URL entry</a>.
 
      <li>
-      <p>If <var>request</var>'s <a for=request>method</a> is not `<code>GET</code>` or
-      <var>blob</var> is not a {{Blob}} object, then return a <a>network error</a>. [[!FILEAPI]]
+      <p>If <var>request</var>'s <a for=request>method</a> is not `<code>GET</code>`,
+      <var>blobURLEntry</var> is null, or <var>blobURLEntry</var>'s
+      <a for="blob URL entry">object</a> is not a {{Blob}} object, then return a
+      <a>network error</a>. [[!FILEAPI]]
 
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.


annevk (commented):
Thanks for tackling this!

Based on #1077 and my vague recollection of things you will need to rewrite the previous step so you can null check the blob URL entry. I suspect you want to instead initialize _blobURLEntry_ there and then rewrite this step to something like

> If request's method is not GET, blobURLEntry is null, or blobURLEntry's object is not a Blob object, then return ...

dlrobertson:
> Based on #1077 and my vague recollection of things you will need to rewrite the previous step so you can null check the blob URL entry. I suspect you want to instead initialize _blobURLEntry_ there and then rewrite this step to something like
> 
> > If request's method is not GET, blobURLEntry is null, or blobURLEntry's object is not a Blob object, then return ...

Awesome! Thanks for the info!

fetch.bs
```diff
@@ -4503,12 +4503,13 @@ steps:
...
-     <li><p>Let <var>blob</var> be <var>request</var>'s <a for=request>current URL</a>'s
-     <a for=url>blob URL entry</a>'s <a for="blob URL entry">object</a>.
+     <li><p>Let <var>blobURLEntry</var> be the <a for=url>blob URL entry</a> for the
+     <var>request</var>'s <a for=request>current URL</a>.
 
      <li>
-      <p>If <var>request</var>'s <a for=request>method</a> is not `<code>GET</code>` or
-      <var>blob</var> is not a {{Blob}} object, then return a <a>network error</a>. [[!FILEAPI]]
+      <p>If <var>request</var>'s <a for=request>method</a> is not `<code>GET</code>`,
+      <var>blobURLEntry</var> is null, or <var>blobURLEntry</var>'s object is not
```
annevk, line 4511: object should be `<a for="blob URL entry">object</a>` here, as it was before.

fetch.bs
```diff
@@ -4503,12 +4503,14 @@ steps:
     <a for="fetch params">canceled</a>:
 
     <ol>
-     <li><p>Let <var>blob</var> be <var>request</var>'s <a for=request>current URL</a>'s
-     <a for=url>blob URL entry</a>'s <a for="blob URL entry">object</a>.
+     <li><p>Let <var>blobURLEntry</var> be the associated <a for=url>blob URL entry</a> of
```
dlrobertson, line 4506: Much of the rewording here may not be necessary. My intent was to avoid using successive possessive nouns, which could improve readability. I'm certainly open to suggestions here.

fetch.bs
dlrobertson: Good catch!

fetch.bs
annevk: We tend to use the current style in specifications so I'd suggest we stick to that as it might actually hurt readability overall when you keep switching styles.

dlrobertson: :+1: Sounds good

annevk:
Great! You'll need to accept the invite for https://github.com/mozilla-standards and make your membership of the organization public to appease the IPR bot. After that this can be merged. \o/

dlrobertson:
Bots appeased! Thanks for the tips

PR #1439, opened 2022-05-11
Editorial: consistently use a body + type tuple

### LINKED ISSUE #1410: Use "body with type" from #1392 throughout spec
Currently `(safely) extract a body` returns a mime type and a body. #1392 introduces a new tuple type for the combo of these two types. We should switch to using it for the `extract a body` return type and other relevant fields after #1392 lands.

lucacasonato:
Previously multiple functions returned ad-hoc tuples of body + mime
type. This commit consolidates all of these into a concrete "body with
type" tuple.

This is an editorial change (does not impact behaviour).

Closes #1410

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1297,6 +1297,14 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
 
 <hr>
 
+<p>A <dfn export>body with type</dfn> is a <a for=/>tuple</a> that consists of a
+<dfn export for="body with type">body</dfn> (a <a for=/>body</a>) and a
+<dfn export for="body with type">type</dfn> (a <a for=/>header value</a>).
+
+<p class="note no-backref">Be aware that a <a for=/>body with type</a>'s <a for="body with type">type</a> can be null.
+
+<hr>
+
 <p>To <dfn export>handle content codings</dfn> given <var>codings</var> and <var>bytes</var>, run
 these steps:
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1297,6 +1297,14 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
 
 <hr>
 
+<p>A <dfn export>body with type</dfn> is a <a for=/>tuple</a> that consists of a
+<dfn export for="body with type">body</dfn> (a <a for=/>body</a>) and a
+<dfn export for="body with type">type</dfn> (a <a for=/>header value</a>).
+
+<p class="note no-backref">Be aware that a <a for=/>body with type</a>'s <a for="body with type">type</a> can be null.
```
annevk, line 1304: In that case it should say "null or a header value" above.

fetch.bs
```diff
@@ -3881,8 +3889,8 @@ the request.
  <var>crossOriginIsolatedCapability</var>.
 
  <li><p>If <var>request</var>'s <a for=request>body</a> is a <a for=/>byte sequence</a>, then set
- <var>request</var>'s <a for=request>body</a> to the first return value of
- <a for=BodyInit>safely extracting</a> <var>request</var>'s <a for=request>body</a>.
+ <var>request</var>'s <a for=request>body</a> to the <a for="body with type">body</a> in the result
```
annevk, line 3892: I think it should be "of the result" (applies several times), but not a 100%.

fetch.bs
```diff
@@ -6451,9 +6461,8 @@ typedef (Blob or BufferSource or FormData or URLSearchParams or USVString) XMLHt
 
 typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;</pre>
 
-<p>To <dfn for=BodyInit export>safely extract</dfn> a <a for=/>body</a> and a
-`<code>Content-Type</code>` <a for=/>header value</a> from a <a for=/>byte sequence</a> or
-{{BodyInit}} object <var>object</var>, run these steps:
+<p>To <dfn for=BodyInit export>safely extract</dfn> a <a for=/>body with type</a> from a
+<a for=/>byte sequence</a> or {{BodyInit}} object <var>object</var>, run these steps:
```
annevk, line 6465: In this algorithm you want to change "results" to "result" as it now returns a singular thing.

annevk (commented):
Thanks for working on this! This needs a PR for XHR too, right?

fetch.bs
lucacasonato: I switched it between "of" and "in" a couple of times before submitting the PR. I am not sure either. It's kind of a weird sentence construction.

lucacasonato:
> This needs a PR for XHR too, right?

Yup, opened now: https://github.com/whatwg/xhr/pull/349

PR #1442, opened 2022-05-17
Define opaque-response blocking 

annevk:
This is good enough for early review, but there are a number of issues that still need resolving: https://github.com/annevk/orb/labels/mvp.

There are also some inline TODO comments.

A PR against HTML is needed to ensure it passes the appropriate metadata for media element and classic script requests. We might also want to depend on HTML for parsing JavaScript.

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -33,6 +33,11 @@ urlPrefix:https://w3c.github.io/hr-time/#;spec:hr-time
         url:dfn-coarsened-shared-current-time;text:coarsened shared current time
         url:dfn-unsafe-shared-current-time;text:unsafe shared current time
     type:typedef;url:dom-domhighrestimestamp;text:DOMHighResTimeStamp
+
+urlPrefix: https://tc39.es/ecma262/;type:dfn;spec:ecma-262
+    url:sec-parsetext;text:ParseText
+    url:prod-Script;text:Script
+    url:script-record;text:Script Record
 </pre>
 
 <pre class=link-defaults>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4656,19 +4941,23 @@ these steps:
...
    <a>HTTP-network-or-cache fetch</a> given <var>fetchParams</var>.
 
-   <li>
-    <p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>cors</code>" and a
-    <a>CORS check</a> for <var>request</var> and <var>response</var> returns failure, then return a
-    <a>network error</a>.
+   <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>",
+   <var>response</var>'s <a for=response>status</a> is not a <a>redirect status</a>, and the
+   <a>opaque-response-safelist check</a> given <var>request</var> and <var>response</var> returns
+   false, then return a <a>network error</a>.
```
annevk, line 4947: Perhaps we should make the redirect status check part of the algorithm?

domenic:
So I think using ParseText here is pretty clean, especially since ParseScript seems to really prefer you giving it a Realm (even though it's used just for pass through).

However there may be a slight issue here, which is that ParseText does early-error detection. I wonder if people really plan to do early-error detection (which IIUC involves pretty detailed JavaScript engine semantics) inside the network code where ORB is implemented.

annevk:
Yes, if you look at https://whatpr.org/fetch/1442.html#http-fetch it currently handles service workers in step 5. Then in step 6, if there's still no response, it'll do a network fetch in substep 3. Then in substep 4 it'll do the opaque-response-safelist check. This is also clarified in a note at the end of step 6.

PR #1447, opened 2022-06-01
Add algorithm to extract an encoding from a MIME type

annevk:
This is useful for HTML's fetch a classic script and Fetch's upcoming opaque-response blocking. It will also be useful to more tightly define equivalent legacy setups elsewhere.

***
<a href="https://whatpr.org/fetch/1447.html" title="Last updated on Jun 1, 2022, 4:31 PM UTC (5d92f26)">Preview</a> | <a href="https://whatpr.org/fetch/1447/4128e96...5d92f26.html" title="Last updated on Jun 1, 2022, 4:31 PM UTC (5d92f26)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3492,18 +3492,18 @@ Content-Type:
 </div>
 
 <p>To <dfn export>legacy extract an encoding</dfn> given failure or a <a for=/>MIME type</a>
-<var>mimeType</var> and an <a for=/>encoding</a> <var>encoding</var>, run these steps:
+<var>mimeType</var> and an <a for=/>encoding</a> <var>fallbackEncoding</var>, run these steps:
 
 <ol>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -3491,6 +3491,30 @@ Content-Type:
  </table>
 </div>
 
+<p>To <dfn export>legacy extract an encoding</dfn> given failure or a <a for=/>MIME type</a>
+<var>mimeType</var> and an <a for=/>encoding</a> <var>encoding</var>, run these steps:
```
domenic, line 3495: encoding -> fallbackEncoding would be a big improvement IMO.

domenic (approved):
At first I thought this might belong better in mimesniff but if it's generally meant to be used with "extract a MIME type" then here makes sense.

PR #1453, opened 2022-06-14
Add `Set-Cookie` as a forbidden header name

lucacasonato:
Following a suggestion from @annevk in #973, this adds `set-cookie` as a
forbidden request header name. The rationale here is to prevent complication of
internal data structures for internal header lists that are attached to
`Request` objects, in light of #1346.

I added a UseCounter to Chromium in January to check that this is a web
compatible change. The analysis uncovered no issues, and as such I consider that
this change is web compatible. See
https://github.com/whatwg/fetch/issues/973#issuecomment-1154958094 for more
elaboration.

   * Deno
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -997,6 +997,7 @@ is a <a>byte-case-insensitive</a> match for one of
  <li>`<code>Keep-Alive</code>`
  <li>`<a http-header><code>Origin</code></a>`
  <li>`<code>Referer</code>`
+ <li>`<code>Set-Cookie</code>`
  <li>`<code>TE</code>`
  <li>`<code>Trailer</code>`
  <li>`<code>Transfer-Encoding</code>`


annevk (commented):
Please add a note at the end explaining why Set-Cookie is part of this list as it might surprise some folks.

@youennf this addresses your concern with respect to exposing `Set-Cookie` on `Headers`, right?

@yutakahirano @ddragana @KershawChang concerns?

lucacasonato:
> Please add a note at the end explaining why Set-Cookie is part of this list as it might surprise some folks.

Done.

annevk (commented):
I pushed a nit commit to combine this note with the existing one. Let me know what you think. I'm happy with this.

lucacasonato:
Thanks. Looks great.

yutakahirano:
> @yutakahirano @ddragana @KershawChang concerns?

No concerns.

KershawChang:
> 
> @yutakahirano @ddragana @KershawChang concerns?

No concerns. This makes sense to me.

lucacasonato:
Implementation bugs are now filed.

jmanico:
I struggle to see why Set-Cookie (a response header) is now a forbidden Request header.

Is this just to make sure developers do not use Set-Cookie in a request (demonstrating a misunderstanding of cookies)? HTTPOnly currently stops reading cookies from response headers and Set-Cookie never appears in requests since Set-Cookie is a response header.

annevk:
As I said on Twitter, it's explained by the note this PR added. `Set-Cookie` is the only header that cannot be combined (in the HTTP sense) and we don't want implementations to have to support that for requests as it would require a special side table in optimized implementations.

Issue #1070, opened 2020-08-06
blob: URLs and range requests


mkruisselbrink:
The current fetch spec says that all blob: URL requests should result in a 200/OK response (a separate issue, it should probably also mention how to reply if the URL's blob url entry is null). That doesn't entirely match at least chrome's implementation though. In chrome we also support range requests on blob URLs, as long as the range consists of a single range. In that case we reply with a 206 and just the requested range. If multiple ranges are passed in we fail the request with a network error.

I haven't tried what other browsers do for range requests to blob URLs.

jakearchibald:
fwiw, I'd like to make this work for the cache API too at some point.

youennf:
I haven't tested it but I would hope this is working in Safari as well.
WPT tests would be nice.

annevk:
@mkruisselbrink by design the URL's blob URL entry cannot be null. The URL parser does the lookup. I guess perhaps something could still go amiss if it has to transfer process boundaries (OOM) and we should account for that somehow. That should be a network error.

Is this specific to blob or should this also work for `data:` URLs and `about:blank`?

mkruisselbrink:
I'm not sure why URL's blob URL entry wouldn't be able to be null? Step 4 in https://url.spec.whatwg.org/#url-parsing "Set url’s blob URL entry to [...], and null otherwise." There needs to be some way to deal with resolving/fetching blob URLs that don't actually correspond to blobs after all...

I haven't looked into what chrome does for range requests data: URLs or about:blank. Those are at least very different code paths in chrome.

annevk:
Thanks, I split out null handling into #1077.

annevk:
This is a good issue for someone with some experience writing [web-platform-tests](https://github.com/web-platform-tests/wpt). There's a couple steps toward solving this:

1. Write tests for range requests to `blob:` URLs (and ideally also `data:` URLs and `about:blank`) and use those tests to tease out what browsers are doing.
2. If browsers agree, great. We probably just want to specify what they do. If they don't agree, let's discuss the results in this issue.
3. Adjust the requirements in the specification accordingly. (This might be a little involved and as such I can help out here once it gets to this stage.)

dlrobertson:
This looks like an interesting issue, so I started writing some wpt tests for this. I haven't tried `data:` or `blob:` urls yet, but I see the following for `blob:`:

| Engine | Basic Range | Multiple Ranges |
|--------|-------------|------------------|
[...]

PR #1454, opened 2022-06-16
Add support for parsing a suffix byte range spec

dlrobertson:
Add support for parsing a range header in the suffix byte range spec
form.

   * …
   * …
   * Partially [here](https://github.com/web-platform-tests/wpt/pull/34384#discussion_r898469534)
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1454.html" title="Last updated on Sep 23, 2022, 4:36 PM UTC (3ebbb33)">Preview</a> | <a href="https://whatpr.org/fetch/1454/0d95632...3ebbb33.html" title="Last updated on Sep 23, 2022, 4:36 PM UTC (3ebbb33)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1122,6 +1122,9 @@ run these steps:
  <li><p>Let <var>rangeStart</var> be the result of <a>collecting a sequence of code points</a> that
  are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
 
+ <li><p>Let <var>rangeStartValue</var> be <var>rangeStart</var>, interpreted as decimal number, if
+ <var>rangeStart</var> is not the empty string; otherwise null.
+
  <li><p>If the <a>code point</a> at <var>position</var> within <var>data</var> is not U+002D (-),
  then return failure.
 
@@ -1130,25 +1133,21 @@ run these steps:
  <li><p>Let <var>rangeEnd</var> be the result of <a>collecting a sequence of code points</a> that
  are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
 
+ <li><p>Let <var>rangeEndValue</var> be <var>rangeEnd</var>, interpreted as decimal number, if
+ <var>rangeEnd</var> is not the empty string; otherwise null.
+
  <li><p>If <var>position</var> is not past the end of <var>data</var>, then return failure.
 
- <li><p>If <var>rangeEnd</var>'s <a for=string>length</a> is 0 and <var>rangeStart</var>'s
- <a for=string>length</a> is 0, then return failure.
+ <li><p>If <var>rangeEndValue</var> and <var>rangeStartValue</var> are null, then return failure.
 
- <li><p>If <var>rangeEnd</var>'s <a for=string>length</a> is 0, then return (<var>rangeStart</var>,
- null).
+ <li><p>If <var>rangeStartValue</var> and <var>rangeEndValue</var> are numbers, and
+ <var>rangeStartValue</var> is greater than <var>rangeEndValue</var>, then return failure.
 
  <li>
-  <p>If <var>rangeStart</var>'s <a for=string>length</a> is 0, then return (null,
-  <var>rangeEnd</var>).
+  <p>Return (<var>rangeStartValue</var>, <var>rangeEndValue</var>).
 
   <p class="note">The range end or start can be omitted, e.g., `<code>bytes=0-</code>` or
   `<code>bytes=-500</code>` are valid ranges.
-
- <li><p>If <var>rangeStart</var>, interpreted as decimal number, is greater than
- <var>rangeEnd</var>, interpreted as decimal number, then return failure.
-
- <li><p>Return (<var>rangeStart</var>, <var>rangeEnd</var>).
 </ol>
 
 <p class=note><a>Parse a single range header value</a> succeeds for a subset of allowed range header
--- a/fetch.bs
+++ b/fetch.bs
@@ -887,7 +887,13 @@ fetch("https://victim.example/naïve-endpoint", {
     </div>
 
    <dt>`<code>range</code>`
-   <dd><p>If <var>value</var> is not a <a>simple range header value</a>, then return false.
+   <dd>
+    <ol>
+     <li><p>Let <var>rangeResult</var> be the result of <a>parse a single range header value</a>
+     given <var>value</var>.
+
+     <li><p>If <var>rangeResult</var> is failure, then return false.
+    </ol>
 
    <dt>Otherwise
    <dd><p>Return false.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1103,17 +1103,23 @@ run these steps:
  <li><p>If <var>position</var> is not past the end of <var>data</var>, then return false.
 
  <li>
-  <p>If <var>rangeEnd</var>'s <a for=string>length</a> is 0, then return true.
+  <p>If <var>rangeEnd</var>'s <a for=string>length</a> is 0 and <var>rangeStart</var>'s
```
dlrobertson, line 1106: This would mean `bytes=-` would be invalid.

dlrobertson (commented):
I can wait to implement this until after/with the blob range PR.

annevk:
It seems good to separate this out, but shall we turn it into a value parser at the same time so it's forward-compatible with the blob PR?

So we call this "parse a single range header value" and it returns (_start_, null), (null, _end_), (_start_, _end_), or failure. And then adjust the caller to account for that.

dlrobertson (commented):
> So we call this "parse a single range header value" and it returns (start, null), (null, end), (start, end), or failure. And then adjust the caller to account for that.

Yeah this makes sense, but after reading I wonder if it would also make sense to tackle the following comment and also change the parser to operate on a header list as part of this patch.

From https://github.com/whatwg/fetch/issues/1070#issuecomment-1141315124
> Note that this parser needs to take a header list for consistency with "extract a MIME type" and friends. That also makes it clear that if you have multiple Range headers what value ends up being operated on and why that ends up failing (the comma).

fetch.bs
```diff
@@ -1077,14 +1081,14 @@ run these steps:
  <li><p>Return <var>values</var>.
 </ol>
 
-<p>To determine if a <a>byte sequence</a> <var>value</var> is a
-<dfn>simple range header value</dfn>, perform the following steps. They return a <a>boolean</a>.
+<p>To <dfn lt="parse a single range header value|single range header value">parse a single range header value</dfn>
+from a <a>byte sequence</a> <var>value</var>, run these steps:
```
dlrobertson, line 1085: Many parsers indicate the returned value type. I couldn't find a primitive type that would encompass a range type `(start, end)`.

dlrobertson (commented):
Updated `simple range header` to `parse a single range header`.

fetch.bs
```diff
@@ -867,7 +867,11 @@ fetch("https://victim.example/naïve-endpoint", {
     </div>
 
    <dt>`<code>range</code>`
-   <dd><p>If <var>value</var> is not a <a>simple range header value</a>, then return false.
+   <dd>
+    <ol>
+     <li><p>Let <var>rangeResult</var> be the result of <a>parse a single range header value</a>.
```
annevk, line 872: We need to pass the parser something to operate on and we usually separate the list items by a newline.
```suggestion
     <li><p>Let <var>rangeResult</var> be the result of <a>parse a single range header value</a>
     given <var>value</var>.

```

fetch.bs
```diff
@@ -1077,14 +1081,14 @@ run these steps:
  <li><p>Return <var>values</var>.
 </ol>
 
-<p>To determine if a <a>byte sequence</a> <var>value</var> is a
-<dfn>simple range header value</dfn>, perform the following steps. They return a <a>boolean</a>.
+<p>To <dfn lt="parse a single range header value|single range header value">parse a single range header value</dfn>
```
annevk, line 1084: I don't think we need to make "single range header value" able to reference this? We should also keep the ID `simple-range-header-value` working I think.

fetch.bs
annevk: It returns failure or a https://infra.spec.whatwg.org/#tuples of two items each of which is null or a number. I'm not sure it's worth writing that out however.

fetch.bs
```diff
@@ -1093,27 +1097,35 @@ run these steps:
...
 
  <li><p>Advance <var>position</var> by 1.
 
  <li><p>Let <var>rangeEnd</var> be the result of <a>collecting a sequence of code points</a> that
  are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
 
- <li><p>If <var>position</var> is not past the end of <var>data</var>, then return false.
+ <li><p>If <var>position</var> is not past the end of <var>data</var>, then return failure.
 
  <li>
```
annevk, line 1109: If an `<li>` contains a single `<p>` there should not be a newline as is done above.

annevk (commented):
This looks good, modulo a couple of small nits. Note that I'll be out for at least another two weeks. Hope that's not too discouraging as I really like your contributions!

fetch.bs
dlrobertson: > We should also keep the ID simple-range-header-value working I think.

:+1: I think that is what I meant to do here, but failed at the implementation 😃 Will fix on the next update.

dlrobertson (commented):
Thanks for the review!

> Note that I'll be out for at least another two weeks.

No problem! Enjoy your time off

fetch.bs
```diff
@@ -1077,14 +1083,14 @@ run these steps:
  <li><p>Return <var>values</var>.
 </ol>
 
-<p>To determine if a <a>byte sequence</a> <var>value</var> is a
-<dfn>simple range header value</dfn>, perform the following steps. They return a <a>boolean</a>.
+<p>To <dfn lt="parse a single range header value|simple range header value">parse a single range header value</dfn>
```
annevk, line 1086:
```suggestion
<p>To <dfn id=simple-range-header-value>parse a single range header value</dfn>
```

fetch.bs
```diff
@@ -1093,29 +1099,36 @@ run these steps:
...
+ <var>rangeEnd</var>, interpreted as decimal number, then return failure.
 
- <li><p>Return true.
+ <li><p>Return (<var>rangeStart</var>, <var>rangeEnd</var>).
 </ol>
 
-<p class="note">A <a>simple range header value</a> is a subset of allowed range header values, but
-it is the most common form used by user agents when requesting media or resuming downloads. This
-format of range header value can be set using <a>add a range header</a>.
+<p class="note"><a>parse a single range header value</a> succeeds for a subset of allowed range
```
annevk, line 1129:
```suggestion
<p class="note"><a>Parse a single range header value</a> succeeds for a subset of allowed range
```

fetch.bs
```diff
@@ -867,7 +867,13 @@ fetch("https://victim.example/naïve-endpoint", {
...
 
    <dt>`<code>range</code>`
-   <dd><p>If <var>value</var> is not a <a>simple range header value</a>, then return false.
+   <dd>
+    <ol>
+     <li><p>Let <var>rangeResult</var> be the result of <a>parse a single range header value</a>
+     given <var>value</var>.
+
+     <li><p>If <var>rangeResult</var> is failure, then return false.
+    </ol>
```
annevk, line 876: Optional: yours is okay, but I think it's also simple enough still we could shorten it to:
```suggestion
   <dd><p>If the result of <a>parse a single range header value</a> given <var>value</var> is
   failure, then return false.
```

annevk (commented):
This looks great! Couple nits left. I'm happy to address those myself, but I wanted your take on them before going ahead.

dlrobertson (commented):
Thanks for the review! All of the changes make sense.

annevk:
Looking at this again, it seems we return start and end as byte sequences, not numbers. We probably want to interpret them as numbers and return those numbers in the tuple?

dlrobertson:
> Looking at this again, it seems we return start and end as byte sequences, not numbers. We probably want to interpret them as numbers and return those numbers in the tuple?

Good point! Yeah we should definitely return those as numbers.

annevk:
Attempted to fix that. Not gonna land on a Friday evening, but if this looks good I think this can land Monday as it doesn't change any normative requirements.

dlrobertson:
Yeah, looks good to me!

Issue #1438, opened 2022-05-11
Streaming upload


yutakahirano:
Past discussions: #425, #966, ...

Please also see: [other upload-streaming-blocking issues](https://github.com/whatwg/fetch/labels/upload-streaming-blocking).

Let's discuss whether we want to have the streaming upload feature, i.e., a way to attach a `ReadableStream` to a `Request`, on the Web.
Last time we discussed, all the browser vendors agreed that supporting full-duplex (c.f., #1254) would be very difficult, so I'd like to exclude the option from this discussion. Feel free discuss that at #1254 or another issue.

We, Chrome, have a partial implementation, but failed to show its effectiveness thourgh [a past experiment](https://github.com/whatwg/fetch/pull/425#issuecomment-1122620305).

We have two paths forward:

1. Get more support from other browser vendors and web developers, and ship this feature.
1. Give up this feature and remove the logic from the spec.

Thoughts?

yutakahirano:
(Feel free to restate your position here even if you've already done that in #425).

drzraf:
It's a great opportunity to make the web a bit more bidirectional.

ReadableStream means on-the-fly transformation, the first candidate being... encryption.
> `Use case`: I want to submit a form (to an openstack POST-handler) which contains attachments encrypted on-the-fly. In the current situation, **client memory is ~ twice the file size** what is unsuitable for multi-GB uploads.

The lack of ReadableStream upload incentives the use of chunking as an alternative (which has the additional benefit of being easily resumable).
Since I believe streams find their use for large data (unpredictable size), **reliability** (resumable/error-handing), if not directly involved, needs to at least be considered by implementations.

benbucksch:
I'd like to echo https://github.com/whatwg/fetch/pull/425#issuecomment-1122646961 . This feature is desperately needed for streaming media from a client to a server.

The alternatives like WebSockets etc. all need manual, custom code on both the client and server side, which makes interoperability practically impossible. That's more like FTP. The beauty of HTTP is its simplicity.

We can stream from server to client, in response bodies, so we should also be able to stream from client to server in request bodies. Like I can stream a video from server to client, I should be able to stream a video or audio from client to server. This avoids having to create custom protocols and servers and clients for streaming media.

benbucksch:
One important project of mine died exactly at this point, because this wasn't possible.

ioquatix:
[...]

Issue #1254, opened 2021-06-16
Fetch body streams are not full duplex


lucacasonato:
With request body streams and response body streams one would think it is now possible to use `fetch` for full duplex communication over HTTP, like in the example below:

```js
const intervalStream = new ReadableStream({
  start(c) {
    let count = 0;
    const timer = setInterval(() => {
      console.log("sent!");
      c.enqueue("Hello\n");
      if (count === 5) {
        clearInterval(timer);
        c.close();
      }
      count++;
    }, 1000);
  },
}).pipeThrough(new TextEncoderStream());

const resp = await fetch("https://full-duplex-server.deno.dev/", {
  method: "POST",
  body: intervalStream,
});

console.log("foo");

const reader = resp.body.pipeThrough(new TextDecoderStream()).getReader();

while (true) {
  const { value, done } = await reader.read();
  if (done) break;
  console.log(value);
}

console.log("done!");
```

In this example the server just pipes the incoming request body stream into the response body steam with no modification.

If full duplex communication was working,  you would see "sent!" and "Hello\n" being logged in close proximity to each other, live. This is however not the case, because the promise returned from `fetch` does not resolve until the entire request body stream has been sent and closed.

The solution here would be to resolve the `Promise<Response>` returned from `fetch` before the request body stream has completed. To do this, body upload streaming has to happen in parallel. The main issue with the spec in this area is that the promise returned from `fetch` is expected to reject if a chunk of the request body stream is not a Uint8Array (behaviour is tested in https://staging.wpt.fyi/results/fetch/api/basic/request-upload.any.html).

Are there any plans to resolve this shortcoming before request body streaming is sthbilized in browsers? HTTP/2 and QUIC are very capable duplex stream protocols (HTTP/1 is theoretically also capable) and it would be a shame if that power was not exposed to the web platform.

lucacasonato:
To be clear, I am not explicitly vouching for duplex streams to make it into the initial request body uploading implementation in browsers, I just want to make sure that it would be possible to add down the line. Deno implements `fetch` w/ request body streaming, and our implementation is capable of full duplex too, but it has to deviate from the spec to allow that. The backing HTTP implementations for both Node.js and Servo are also capable of full duplex body streams.

annevk:
See #229 for some history. Since then, I've learned that this would require a lot of work in browsers and it's not planned to be supported initially.

However, I do think we should allow for it in theory, so if you could elaborate on what prohibits it that would help.

[...]

PR #1457, opened 2022-06-20
Introduce RequestInit.duplex

yutakahirano:
Introduce the member, and throw if init.duplex is not set in the
Request constructor, as discussed at #1438.

   * Chrome
   * Mozilla
   * Deno
   * https://github.com/web-platform-tests/wpt/pull/34496
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1337696
   * Firefox: This is a part of the streaming upload feature.
   * Safari: This is a part of the streaming upload feature.
   * Deno (not for CORS changes): This is a part of the streaming upload feature.

***
<a href="https://whatpr.org/fetch/1457.html" title="Last updated on Jul 1, 2022, 8:09 AM UTC (557f24d)">Preview</a> | <a href="https://whatpr.org/fetch/1457/4c93f89...557f24d.html" title="Last updated on Jul 1, 2022, 8:09 AM UTC (557f24d)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6888,7 +6888,7 @@ object), initially null.
    before starting receiving the response) fetch, and "full" for initiating a full-duplex (i.e.,
    the user agent starts receiving the response before sending all the request content) fetch.
    This needs to be set when {{RequestInit/body}} is a {{ReadableStream}}. <span class=note>The
-   semantics for "full" has not been specified yet. See
+   semantics for "full" have not been specified yet. See
    <a href="https://github.com/whatwg/fetch/issues/1254">issue #1254</a>.</span>
 </dl>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -6884,14 +6884,14 @@ object), initially null.
    <dd>Can only be null. Used to disassociate <var>request</var> from any {{Window}}.
 
    <dt>{{RequestInit/duplex}}
-   <dd>"<code>half</code>" is the only valid value and it is for initiating a half-duplex (i.e.,
-   the user agent sends all the request content before starting receiving the response) fetch. We
-   reserve "<code>full</code>" for future use, for initiating a full-duplex (i.e., the user agent
-   starts receiving the response before sending all the request content) fetch. This member needs
-   to be set when {{RequestInit/body}} is a {{ReadableStream}}. <span class=note>The semantics for
-   "<code>full</code>" have not been specified yet. See
-   <a href="https://github.com/whatwg/fetch/issues/1254">issue #1254</a>.</span>
-</dl>
+   <dd>"<code>half</code>" is the only valid value and it is for initiating a half-duplex fetch
+   (i.e., the user agent sends the entire request before processing the response).
+   "<code>full</code>" is reserved for future use, for initiating a full-duplex fetch (i.e., the
+   user agent can process the response before sending the entire request). This member needs to be
+   set when {{RequestInit/body}} is a {{ReadableStream}} object. <span class=note>See
+   <a href="https://github.com/whatwg/fetch/issues/1254">issue #1254</a> for defining
+   "<code>full</code>".</span>
+  </dl>
 
  <dt><code><var>request</var> . <a attribute for=Request>method</a></code>
  <dd>Returns <var>request</var>'s HTTP method, which is "<code>GET</code>" by default.
--- a/fetch.bs
+++ b/fetch.bs
@@ -6784,7 +6784,7 @@ dictionary RequestInit {
   DOMString integrity;
   boolean keepalive;
   AbortSignal? signal;
-  RequestDuplex? duplex;
+  RequestDuplex duplex;
   any window; // can only be set to null
 };
 
@@ -6793,7 +6793,7 @@ enum RequestMode { "navigate", "same-origin", "no-cors", "cors" };
 enum RequestCredentials { "omit", "same-origin", "include" };
 enum RequestCache { "default", "no-store", "reload", "no-cache", "force-cache", "only-if-cached" };
 enum RequestRedirect { "follow", "error", "manual" };
-enum RequestDuplex { "half", "full" };
+enum RequestDuplex { "half" };
 </pre>
 
 <p class="note no-backref">"<code>serviceworker</code>" is omitted from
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -6880,7 +6882,15 @@ object), initially null.
...
    <dt>{{RequestInit/window}}
    <dd>Can only be null. Used to disassociate <var>request</var> from any {{Window}}.
-  </dl>
+
+   <dt>{{RequestInit/duplex}}
+   <dd>"half" for initiating a half-duplex (i.e., the user agent send all the request content
+   before starting receiving the response) fetch, and "full" for initiating a full-duplex (i.e.,
+   the user agent starts receiving the response before sending all the request content) fetch.
+   This needs to be set when {{RequestInit/body}} is a {{ReadableStream}}. <span class=note>The
+   semantics for "full" has not been specified yet. See
```
lucacasonato, line 6891:
```suggestion
   semantics for "full" have not been specified yet. See
```

lucacasonato (commented):
We need to add a Request#duplex attribute too, otherwise the `new Response(resp.body, resp);` pattern breaks.

yutakahirano:
> We need to add a Request#duplex attribute too, otherwise the `new Response(resp.body, resp);` pattern breaks.

I don't understand, this change is sorely for requests, not responses.

lucacasonato:
Sorry, I meant `new Request(req.url, req);`

yutakahirano:
As you see in the change, currently `duplex = 'full'` is unconditionally rejected in the constructor. I expect you to fix that in #1254. I expect you to define `Request.duplex` altogether.

Does this make sense?

lucacasonato:
> I expect you to fix that in https://github.com/whatwg/fetch/issues/1254.

Yeah, that is my understanding too. 

> I expect you to define Request.duplex altogether.

I don't quite understand. Are you saying to not introduce the `Request.duplex` attribute in this PR, and instead do that as part of #1254? I'd rather see `Request.duplex` added at the same time that `RequestInit.duplex` is added (so in this PR).

yutakahirano:
>> I expect you to define Request.duplex altogether.

> I don't quite understand. Are you saying to not introduce the Request.duplex attribute in this PR, and instead do that as part of https://github.com/whatwg/fetch/issues/1254?

Right, because I don't see a reason to specify it now, and I see a reason not to specify it now.

Specifically,

 1. Until #1254 is resolved, web developers cannot use `Request.duplex` effectively, as it always returns 'half'.
 2. If we fail to specify #1254, it will be difficult to remove the member without compatibility problems.

This is different from the situation for `RequestInit.duplex`.

 1. `RequestInit.duplex` forces web developers to specify the member (when the body is a stream), and that will allow us not to specify the default value for now.
 2. We can remove `RequestInit.duplex` if we fail to specify #1254. There will not be any compatibility problems.

yutakahirano:
> Sorry, I meant new Request(req.url, req);

You can use `new Request(req);`

yutakahirano:
Thank you! @annevk , can you take a look at this change?

fetch.bs
```diff
@@ -6784,6 +6784,7 @@ dictionary RequestInit {
   DOMString integrity;
   boolean keepalive;
   AbortSignal? signal;
+  RequestDuplex? duplex;
```
annevk, line 6787: There's no need for this to be nullable.

fetch.bs
```diff
@@ -6792,6 +6793,7 @@ enum RequestMode { "navigate", "same-origin", "no-cors", "cors" };
 enum RequestCredentials { "omit", "same-origin", "include" };
 enum RequestCache { "default", "no-store", "reload", "no-cache", "force-cache", "only-if-cached" };
 enum RequestRedirect { "follow", "error", "manual" };
+enum RequestDuplex { "half", "full" };
```
annevk, line 6796: We don't need to have "full" here at this point.

fetch.bs
```diff
@@ -6880,7 +6882,15 @@ object), initially null.
...
    <dd>Can only be null. Used to disassociate <var>request</var> from any {{Window}}.
-  </dl>
+
+   <dt>{{RequestInit/duplex}}
+   <dd>"half" for initiating a half-duplex (i.e., the user agent send all the request content
+   before starting receiving the response) fetch, and "full" for initiating a full-duplex (i.e.,
+   the user agent starts receiving the response before sending all the request content) fetch.
+   This needs to be set when {{RequestInit/body}} is a {{ReadableStream}}. <span class=note>The
+   semantics for "full" have not been specified yet. See
+   <a href="https://github.com/whatwg/fetch/issues/1254">issue #1254</a>.</span>
```
annevk, line 6892: Use `"<code>half</code>"` for formatting. As we don't specify "full" at this point this text needs to be changed a little bit.

fetch.bs
```diff
@@ -7305,11 +7315,16 @@ constructor steps are:
  <li><p>Let <var>inputOrInitBody</var> be <var>initBody</var> if it is non-null; otherwise
  <var>inputBody</var>.
 
+ <li><p>If <var>init</var>["{{RequestInit/duplex}}"] is "full", then throw a {{TypeError}}.
```
annevk, line 7318: This will happen automatically as any unknown value ends up throwing so we can remove this step.

annevk (commented):
Thanks, this looks good to me, but we can simplify it a bit further.

I like that we have a note about specifying "full" though. Let's keep that.

fetch.bs

fetch.bs

fetch.bs
yutakahirano: Done.

fetch.bs

annevk:
I'll merge this on Friday unless anything new comes up. (I assume we'll have the tests sorted by then and @lucacasonato will have had a chance to look at those as well.)

jimrandomh:
This change was breaking, wasn't communicated by downstream libraries as breaking, and caused us a brief outage. This is the second time I've been impacted by a backwards-incompatible change in this specification, and while the other instance arguably had a security justification, this one was just "we introduced a new parameter and forgot to give it a default".

annevk:
This change was introduced before any browser implemented request streams so I suspect your issue is mainly with a non-browser implementation and the way it handles incompatible changes. Pretty sure that airing your grievances about it here won't result in meaningful change.

PR #1468, opened 2022-07-09
Pass in response status to mark resource timing

abinpaul1:
These changes are to support addition of http response status code to Perfomance resource timing. Further details are available at https://github.com/w3c/resource-timing/pull/335

   * Chromium : https://chromestatus.com/feature/5163838794629120
   * Firefox : https://github.com/mozilla/standards-positions/issues/665
   * https://chromium-review.googlesource.com/c/chromium/src/+/3754923
   * https://chromium-review.googlesource.com/c/chromium/src/+/3958184
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1343293
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1796785
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=246857

***
<a href="https://whatpr.org/fetch/1468.html" title="Last updated on Oct 21, 2022, 4:24 PM UTC (08d1251)">Preview</a> | <a href="https://whatpr.org/fetch/1468/1fbc40c...08d1251.html" title="Last updated on Oct 21, 2022, 4:24 PM UTC (08d1251)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4387,12 +4387,19 @@ steps:
 
       <p class=note>This covers the case of <var>response</var> being a <a>network error</a>.
 
+     <li><p>Let <var>responseStatus</var> be <var>response</var>'s <a for=response>status</a>.
+
+     <li><p>If <var>fetchParams</var>'s <a for="fetch params">request</a>'s
+     <a for=request>mode</a> is "<code>navigate</code>" and <var>request</var>'s
+     <a for=request>current URL</a>'s <a for=url>origin</a> is not <a>same origin</a> with
+     <var>request</var>'s <a for=request>origin</a>, then set <var>responseStatus</var> to 0.
+
      <li><p>If <var>fetchParams</var>'s <a for="fetch params">request</a>'s
      <a for=request>initiator type</a> is not null, then
      <a for=/>mark resource timing</a> given <var>timingInfo</var>, <var>request</var>'s
      <a for=request>URL</a>, <a for="fetch params">request</a>'s
      <a for=request>initiator type</a>, <var>global</var>, <var>cacheState</var>,
-     <var>bodyInfo</var>, and <var>response</var>'s <a for=response>status</a>.
+     <var>bodyInfo</var>, and <var>responseStatus</var>.
     </ol>
 
    <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4391,8 +4391,8 @@ steps:
      <a for=request>initiator type</a> is not null, then
      <a for=/>mark resource timing</a> given <var>timingInfo</var>, <var>request</var>'s
      <a for=request>URL</a>, <a for="fetch params">request</a>'s
-     <a for=request>initiator type</a>, <var>global</var>, <var>cacheState</var>, and
-     <var>bodyInfo</var>.
+     <a for=request>initiator type</a>, <var>global</var>, <var>cacheState</var>,
+     <var>bodyInfo</var>, and <var>response</var>'s <a for=response>status</a>.
     </ol>
 
    <li>
--- a/fetch.bs
+++ b/fetch.bs
@@ -4387,12 +4387,11 @@ steps:
 
       <p class=note>This covers the case of <var>response</var> being a <a>network error</a>.
 
-     <li><p>Let <var>responseStatus</var> be <var>response</var>'s <a for=response>status</a>.
-
-     <li><p>If <var>fetchParams</var>'s <a for="fetch params">request</a>'s
-     <a for=request>mode</a> is "<code>navigate</code>" and <var>request</var>'s
-     <a for=request>current URL</a>'s <a for=url>origin</a> is not <a>same origin</a> with
-     <var>request</var>'s <a for=request>origin</a>, then set <var>responseStatus</var> to 0.
+     <li><p>Let <var>responseStatus</var> be 0 if <var>fetchParams</var>'s
+     <a for="fetch params">request</a>'s <a for=request>mode</a> is "<code>navigate</code>" and
+     <var>response</var>'s <a for=response>URL</a>'s <a for=url>origin</a> is not
+     <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, and
+     <var>response</var>'s <a for=response>status</a> otherwise.
 
      <li><p>If <var>fetchParams</var>'s <a for="fetch params">request</a>'s
      <a for=request>initiator type</a> is not null, then
--- a/fetch.bs
+++ b/fetch.bs
@@ -2195,6 +2195,9 @@ this is also tracked internally using the request's <a for=request>timing allow 
 <dfn export for=response>service worker timing info</dfn> (null or a
 <a for=/>service worker timing info</a>), which is initially null.
 
+<p>A <a for=/>response</a> has an associated <dfn for=response>has-cross-origin-redirects</dfn>
+(a boolean), which is initially false.
+
 <hr>
 
 <p>A <a for=/>response</a> whose
@@ -4258,6 +4261,9 @@ steps:
  <!-- If you are ever tempted to move this around, carefully consider responses from about URLs,
       blob URLs, service workers, HTTP cache, HTTP network, etc. -->
 
+ <li><p>If <var>request</var> has a <a for=request>redirect-tainted origin</a>, then set
+ <var>internalResponse</var>'s <a for=response>has-cross-origin-redirects</a> to true.
+
  <li><p>If <var>request</var>'s <a for=request>timing allow failed flag</a> is unset, then set
  <var>internalResponse</var>'s <a for=response>timing allow passed flag</a>.
 
@@ -4402,9 +4408,8 @@ steps:
 
      <li><p>Let <var>responseStatus</var> be 0 if <var>fetchParams</var>'s
      <a for="fetch params">request</a>'s <a for=request>mode</a> is "<code>navigate</code>" and
-     <var>response</var>'s <a for=response>URL</a>'s <a for=url>origin</a> is not
-     <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, and
-     <var>response</var>'s <a for=response>status</a> otherwise.
+     <var>response</var>'s <a for=response>has-cross-origin-redirects</a> is true; otherwise
+     <var>response</var>'s <a for=response>status</a>.
 
      <li><p>If <var>fetchParams</var>'s <a for="fetch params">request</a>'s
      <a for=request>initiator type</a> is not null, then


fetch.bs
```diff
@@ -4379,6 +4382,9 @@ steps:
 
      <li><p>Let <var>bodyInfo</var> be <var>response</var>'s <a for=response>body info</a>.
 
+     <li><p>Let <var>bodyInfo</var>'s <a for="response body info">response status</a> be
+     <var>response</var>'s <a for=response>status</a>.
```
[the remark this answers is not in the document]
abinpaul1: > At the same time, maybe there are other places that modify the response status that I missed.

I couldn't find any other place where this status gets modified, so I've updated such that bodyInfo response status gets set at the step you pointed out.

fetch.bs
```diff
@@ -5536,6 +5539,9 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
        <li><p>Let <var>status</var> be the HTTP response's status code.
```
yoavweiss, line 5540: @annevk - Would it make sense to also explicitly set the response's status to |status| here? (maybe as a separate PR)

yoavweiss (approved):
Non-authoritative LGTM

yoavweiss:
Also, we'd want tests and implementation issues before this lands.

fetch.bs
```diff
@@ -301,6 +301,9 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
  <dt><dfn export for="response body info" id="fetch-timing-info-decoded-body-size">decoded
  size</dfn> (default 0)
  <dd>A number.
+ <dt><dfn export for="response body info" id="fetch-timing-info-response-status">response
```
noamr, line 304: This is not needed. You have [response's status](https://fetch.spec.whatwg.org/#concept-response-status) when you set the `report timing steps` in [fetch response handover](https://fetch.spec.whatwg.org/#fetch-finale). I would add it as another parameter to `Mark resource timing`.

Also, as we talked on the Chromium CL, I don't think we should automatically make this TAO-protected rather than CORS-protected. It's not exactly timing information.

abinpaul1: Okay, just to be sure I understand. We would pass in the status as a new parameter to [Mark resource time](https://w3c.github.io/resource-timing/#dfn-mark-resource-timing) which would in turn pass it to [Setup the resource timing entry](https://w3c.github.io/resource-timing/#dfn-setup-the-resource-timing-entry) where we would use it to set the entry's `responseStatus`.

> Also, as we talked on the Chromium CL, I don't think we should automatically make this TAO-protected rather than CORS-protected. It's not exactly timing information.

Yeah, we don't want to. I am a bit unsure as to how we would would specify this in the specification . I initially thought mentioning in the Note in [Cross-origin Resources section of resource timing](https://w3c.github.io/resource-timing/#sec-cross-origin-resources) that the status would be `0` if `CORS check` fails for the cross origin resource would suffice, but it seems I'm missing something.

noamr: > Okay, just to be sure I understand. We would pass in the status as a new parameter to [Mark resource time](https://w3c.github.io/resource-timing/#dfn-mark-resource-timing) which would in turn pass it to [Setup the resource timing entry](https://w3c.github.io/resource-timing/#dfn-setup-the-resource-timing-entry) where we would use it to set the entry's `responseStatus`.

Exactly

> 
> > Also, as we talked on the Chromium CL, I don't think we should automatically make this TAO-protected rather than CORS-protected. It's not exactly timing information.
> 
> Yeah, we don't want to. I am a bit unsure as to where the exact place to specify this is. Does it make sense to add this to [fetch response handover](https://fetch.spec.whatwg.org/#fetch-finale) as well - If the CORS check fails pass in `0` as the status to `Mark resource timing`

Actually I don't think you need to do anything. If this is a no-cors request, `response` would already be an [opaque filtered response](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque) with a status of 0.

fetch.bs
annevk: It might be worth having a follow-up issue on how to better integrate with HTTP or indicate that's what's happening. We basically want to map an HTTP response to a Fetch response. There has been some discussion about this also with the HTTP WG, but I don't think we'll get anything formal out of that.

fetch.bs
```diff
@@ -301,6 +301,9 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
...
+ status</dfn> (default 0)
```
annevk, line 305: One minor thing here that I notice is incorrect for some other fields as well is that `fetch.bs` shouldn't wrap inside phrasing elements such as `<dfn>` so you can more easily search for the terms inside of them.

fetch.bs
annevk: I think there might be an existing problem here in that an opaque response wouldn't expose body info.

noamr: We set the body info to an empty body info if TAO fails.

noamr (approved):
Non-authoritative LGTM

fetch.bs
```diff
@@ -4391,8 +4391,8 @@ steps:
      <a for=request>initiator type</a> is not null, then
      <a for=/>mark resource timing</a> given <var>timingInfo</var>, <var>request</var>'s
      <a for=request>URL</a>, <a for="fetch params">request</a>'s
-     <a for=request>initiator type</a>, <var>global</var>, <var>cacheState</var>, and
-     <var>bodyInfo</var>.
+     <a for=request>initiator type</a>, <var>global</var>, <var>cacheState</var>,
+     <var>bodyInfo</var>, and <var>response</var>'s <a for=response>status</a>.
```
noamr, line 4395: I was wrong about this - in the case of iframes the status would leak here. You've reached the same problem in the Chromium implementation, need to think about this :)

abinpaul1: Updated it to also take into account `request`'s mode and also origin

fetch.bs
```diff
@@ -4387,12 +4387,18 @@ steps:
 
       <p class=note>This covers the case of <var>response</var> being a <a>network error</a>.
 
+     <li><p>Let <var>responseStatus</var> be 0 if <var>fetchParams</var>'s
+     <a for="fetch params">request</a>'s <a for=request>mode</a> is "<code>navigate</code>" and
+     <var>response</var>'s <a for=response>URL</a>'s <a for=url>origin</a> is not
+     <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, and
+     <var>response</var>'s <a for=response>status</a> otherwise.
```
annevk, line 4394: The style we try to follow (see Infra) is like this:

> origin; otherwise _response_'s status.

annevk, line 4394: I don't think the same origin check here is good. This doesn't take redirects into account.

Also, is this the internal response or not?

annevk (commented):
If you rebase this I think Build will succeed again. There was a temporary issue that is now resolved.

fetch.bs
abinpaul1: To take redirects into acount, would introducing a new flag to request be a good idea - `is navigation same origin`. As [HTTP fetch step 6.6](https://fetch.spec.whatwg.org/#ref-for-timing-allow-failed%E2%91%A2) we would check if request’s [mode](https://fetch.spec.whatwg.org/#concept-request-mode) is "navigate" and request’s [current URL](https://fetch.spec.whatwg.org/#concept-request-current-url)’s [origin](https://url.spec.whatwg.org/#concept-url-origin) is [same origin](https://html.spec.whatwg.org/multipage/origin.html#same-origin) with request’s [origin](https://fetch.spec.whatwg.org/#concept-request-origin). If at any time this is not the case we set `is navigation same origin` to false.

This flag would then be used instead of the current check. Would this work?

annevk: Can we use https://fetch.spec.whatwg.org/#concept-request-tainted-origin? I think request's URL list is still being build up by HTML's navigate algorithm so that should be fine. cc @noamr

abinpaul1: Okay, so just to be clear we let `responseStatus` be 0 if request’s [mode](https://fetch.spec.whatwg.org/#concept-request-mode) is "navigate" and request has a [redirect-tainted origin](https://fetch.spec.whatwg.org/#concept-request-tainted-origin)?

Also, I believe Noam is OOO for a few months.


> Also, is this the internal response or not?

I don't think it is the internal response. We want the filtered response's status, so that status would be 0 if its say a no-cors request as well.

annevk: I think so, but now I wonder why this is different from the other measurements. If this is specifically for navigation timing it should probably follow the current model for those. cc @noamr

yoavweiss: This is not about navigation timing but about iframe navigations being reported to their parent. Since they cannot be CORS fetches, we don't want to report their status if they are cross origin.

yoavweiss: I don't believe we do, but I could be wrong. Are you aware of such infrastructure?

yoavweiss: Wait, we have https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-has-cross-origin-redirects for NavTiming. I can see if/how it fits here.

yoavweiss:
A rough sketch of using existing infrastructure:
* [navigate an iframe](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element:navigate) defines `reportFrameTiming`. It needs to also take in a "has cross origin redirects" boolean and pass it to Fetch's [report timing steps](https://fetch.spec.whatwg.org/#finalize-and-report-timing)
  - We need to fix up the frame reporting so that it passes an algorithm that works with the current [report timing steps](https://fetch.spec.whatwg.org/#fetch-controller-report-timing-steps)
* The passed algorithm would need to nullify the status code in case of a non-same-origin iframe or one that [has cross-origin redirects](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-has-cross-origin-redirects).
* In order for that to happen, we'd need to pass "has cross origin redirects" to [process response end of body](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents:navigation-params-process-response-end-of-body-4)

@abinpaul1 @annevk - does this make sense?

annevk:
Kinda, but then "has cross-origin redirects" needs to be state that Fetch maintains, I suspect. I'm also confused by "navigate an iframe" as I suspect some of those steps also need to apply to `embed` and `object` and processing the final response doesn't just result in a report but also needs to actually render something and such.

abinpaul1:
- The information regarding if there were any cross origin redirects can be determined if needed in fetch, but currently its stored in the [has cross-origin redirects](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-has-cross-origin-redirects) boolean on navigation params struct in HTML.
- I could see this boolean is used mainly in two places ([[1]](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents:navigation-params-has-cross-origin-redirects-4),  [[2]](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents:navigation-params-has-cross-origin-redirects-5)). In both these places we also have access to the response as [navigation param's response](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-response).
- Is it possible to consider moving the [has cross-origin redirects](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-has-cross-origin-redirects) boolean as a field to the [response](https://fetch.spec.whatwg.org/#concept-response)? Its value could be determined in the Fetch spec and it could be used as needed in the above two places as well and for our use case as well. This would maybe help us to easily reuse the boolean and not have to try to pass it along.

annevk:
Yeah that seems reasonable. And since we copy request's URL list over to the eventual response we might be able to reuse part of https://fetch.spec.whatwg.org/#concept-request-tainted-origin somehow.

(Apologies for taking so long. In the future please bug me on Matrix if I'm not responsive. Or ask the design question there as others might be able to answer as well.)

fetch.bs
abinpaul1: Updated

abinpaul1:
Have added the `has-cross-origin-redirects` boolean to the response. I'll also modify html spec to reuse this in place of [has cross-origin redirects](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-has-cross-origin-redirects) boolean on navigation params.

abinpaul1:
@annevk Thanks for the pointer. I've opened the HTML PR with the changes. https://github.com/whatwg/html/pull/8405

annevk:
@abinpaul1 can you file bugs on Gecko and WebKit? That's the last thing remaining here.

abinpaul1:
Filed the bugs

fetch.bs
```diff
@@ -4400,12 +4406,17 @@ steps:
 
       <p class=note>This covers the case of <var>response</var> being a <a>network error</a>.
 
+     <li><p>Let <var>responseStatus</var> be 0 if <var>fetchParams</var>'s
+     <a for="fetch params">request</a>'s <a for=request>mode</a> is "<code>navigate</code>" and
+     <var>response</var>'s <a for=response>has-cross-origin-redirects</a> is true; otherwise
+     <var>response</var>'s <a for=response>status</a>.
```
annevk, line 4412: I was about to merge this, but I don't understand how this protects the response status when mode isn't "navigate". Shouldn't we be using the timing allow passed flag in that case?

abinpaul1: As noted in https://github.com/whatwg/fetch/pull/1468#discussion_r931266882, wouldn't the response be a filtered response here thereby protecting the response status?

PR #1469, opened 2022-07-11
Editorial: Use controller for navigation redirect

noamr:
Remove the existing navigation redirect, and allow the caller
to use the controller to continue to the next redirect.

This simplifies the navigation HTML<->FETCH flow, as now there is
a single `fetch params` struct for the course of the entire navigation.
Up until now some data from fetch params would be lost, like timing info
and task desination, because navigation fetches create a brand new
`fetch params` struct. Now that data is retained, and the only thing that
HTML needs to do is call `process the next redirect` on the fetch controller.

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fetch/1469.html" title="Last updated on Oct 13, 2022, 7:21 AM UTC (05ef536)">Preview</a> | <a href="https://whatpr.org/fetch/1469/283c2ae...05ef536.html" title="Last updated on Oct 13, 2022, 7:21 AM UTC (05ef536)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -8307,10 +8307,18 @@ in a <a for=/>parallel queue</a> if <a for=fetch><i>useParallelQueue</i></a> is 
  <cite>HTML</cite>. [[HTML]]
 
  <dt><a for=fetch><i>processResponse</i></a>
- <dd><p>Takes an algorithm that will be passed a <a for=/>response</a>. Indicates
- <a for=/>response</a>'s <a for=response>header list</a> has been received and initialized. This
- is primarily useful for standards that want to operate on <a for=/>response</a>'s
- <a for=response>body</a>'s <a for=body>stream</a> directly.
+ <dd>
+  <p>Takes an algorithm that will be passed a <a for=/>response</a>. Indicates
+  <a for=/>response</a>'s <a for=response>header list</a> has been received and initialized. This
+  is primarily useful for standards that want to operate on <a for=/>response</a>'s
+  <a for=response>body</a>'s <a for=body>stream</a> directly.
+
+  <p>If the <a for=/>request</a>'s <a for=request>redirect mode</a> was "<code>manual</code>", then
+  callers need to follow a very specific flow with this algorithm to get the intended behavior.
+  They should compute the appropriate <a for=response>location URL</a>, and if it is not null or a
+  failure, then they should call <a for="fetch controller">process the next manual redirect</a>.
+  This will result in <a for=fetch><i>processResponse</i></a> being called again, with the next
+  <a for=/>response</a> in the redirect chain.
 
  <dt><a for=fetch><i>processResponseEndOfBody</i></a>
  <dd><p>Takes an algorithm that will be passed a <a for=/>response</a>. Indicates the network is
--- a/fetch.bs
+++ b/fetch.bs
@@ -8313,12 +8313,13 @@ in a <a for=/>parallel queue</a> if <a for=fetch><i>useParallelQueue</i></a> is 
   is primarily useful for standards that want to operate on <a for=/>response</a>'s
   <a for=response>body</a>'s <a for=body>stream</a> directly.
 
-  <p>If the <a for=/>request</a>'s <a for=request>redirect mode</a> was "<code>manual</code>", then
-  callers need to follow a very specific flow with this algorithm to get the intended behavior.
-  They should compute the appropriate <a for=response>location URL</a>, and if it is not null or a
-  failure, then they should call <a for="fetch controller">process the next manual redirect</a>.
-  This will result in <a for=fetch><i>processResponse</i></a> being called again, with the next
-  <a for=/>response</a> in the redirect chain.
+  <p>If the <a for=/>request</a>'s <a for=request>mode</a> is "<code>navigate</code>" and its
+  <a for=request>redirect mode</a> is "<code>manual</code>", then callers need to follow a very
+  specific flow with this algorithm to get the intended behavior. They should compute the
+  appropriate <a for=response>location URL</a>, and if it is not null or a failure, then they
+  should call <a for="fetch controller">process the next manual redirect</a>. This will result in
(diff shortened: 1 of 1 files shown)


domenic (commented):
LGTM. @annevk do you mind if we merge this, since it unblocks some HTML work?

/cc @jeremyroman as well whose review would be appreciated if he has time.

noamr:
@annevk ping?

fetch.bs
```diff
@@ -4766,7 +4779,10 @@ these steps:
 
      <dt>"<code>manual</code>"
      <dd><p>Set <var>response</var> to an <a>opaque-redirect filtered response</a> whose
-     <a for="filtered response">internal response</a> is <var>actualResponse</var>.
+     <a for="filtered response">internal response</a> is <var>actualResponse</var>, and set
+     <var>fetchParams</var>'s <a for="fetch params">controller</a>'s <a
+     for="fetch controller">next redirect steps</a> to run <a>HTTP-redirect fetch</a> given
+     <var>fetchParams</var> and <var>response</var>.
```
annevk, line 4785: We should only do this when mode is "navigate".

annevk (commented):
How does this work in terms of callbacks and such? The prior approach was that the algorithm would return a response. I'm not entirely sure that was workable, but I'm not sure what the new approach is.

noamr:
> How does this work in terms of callbacks and such? The prior approach was that the algorithm would return a response. I'm not entirely sure that was workable, but I'm not sure what the new approach is.

The new approach is reviewable here: https://github.com/whatwg/html/pull/8097

noamr:
> > How does this work in terms of callbacks and such? The prior approach was that the algorithm would return a response. I'm not entirely sure that was workable, but I'm not sure what the new approach is.
> 
> The new approach is reviewable here: [whatwg/html#8097](https://github.com/whatwg/html/pull/8097)

Specifically, we don't need new callbacks because we don't replace `fetch params`. `process response` etc still work the same way, the only thing that changes is that instead of processing redirects automatically we have to call `next` from the outside.

fetch.bs
noamr: Done

noamr:
> > > How does this work in terms of callbacks and such? The prior approach was that the algorithm would return a response. I'm not entirely sure that was workable, but I'm not sure what the new approach is.
> 
> > 
> 
> > The new approach is reviewable here: [whatwg/html#8097](https://github.com/whatwg/html/pull/8097)
> 
> 
> 
> Specifically, we don't need new callbacks because we don't replace `fetch params`. `process response` etc still work the same way, the only thing that changes is that instead of processing redirects automatically we have to call `next` from the outside.

@annevk anything else missing here? Other changes in the HTML spec depend on this.

annevk:
But how would HTML be notified that it needs to call "next"? Would it call that from "process response"? If so we need to document that "process response" can be invoked multiple times in this mode, no?

I did look at https://whatpr.org/html/8097/browsing-the-web.html but it doesn't actually fill in "process response" (although it does talk about it in a later step).

noamr:
> But how would HTML be notified that it needs to call "next"? Would it call that from "process response"? If so we need to document that "process response" can be invoked multiple times in this mode, no?

Yes, that's the idea, I can add a comment about this when describing `process response`.

> 
> I did look at https://whatpr.org/html/8097/browsing-the-web.html but it doesn't actually fill in "process response" (although it does talk about it in a later step).

process a navigate fetch, 15.5.2: process the next redirect and then waits for process a response. What's missing?

annevk:
"process response" is an argument of the fetch algorithm, not something you can wait for.

noamr:
> "process response" is an argument of the fetch algorithm, not something you can wait for.

This was there in the HTML spec before and I didn't care to refactor it here... but I can if it makes a difference - perhaps `process the next redirect` can pass a new callback if it's more readable, but the behavior would be the same.

annevk:
The existing setup is indeed broken. I'm wondering what our long(er) term perspective is here. I'm also worried that if we keep using the same fetch, fetch controller, and fetch params, we'll end up with weird timing and state situations on them. Typically when you've seen "process response" you wouldn't expect to start uploading again, for instance.

noamr:
> When you do a `<form>` POST and you get one or more 308 redirects.
> 
> And the concern is that if we ever added state such as "final response" (which we have for timings!) it would potentially get a very different meaning here and might end up being invalidated upon "follow".

Is the concern something that exists in the spec now? I don't see any reference to 308 in fetch/HTML.
Would adding a new callback for this feel more future-proof? Other suggestions on how to unbreak this?

annevk:
308 is a redirect status and 301/302/303 change the request method to `GET`. So it exists, but it's a bit harder to find what its implications are.

I don't think reusing the callback is the problem per se. We should document it's a bit weird for "navigate" though. I think the main problem might be all the state that we keep and don't reset (and some of it we need to keep and not reset, such as the URL list). Perhaps a follow-up issue to look at that more closely would be best for now.

domenic:
I audited the state which is not reset. That is the fetch params:

- request: this is the thing we need to preserve, and that the existing spec does
- algorithms: preserving these will make things easier, in the event we ever fix HTML to use them properly
- task destination, cross-origin-isolated capability: we don't want to reset these
- controller: seems to work well. The timing info being carried over is probably a win.
- timing info: this seems like a win.
- preloaded response candidate: not applicable to navigations, will always be null.

I overall agree with the idea that, since we don't reset this state during non-"manual" redirects, this change is an improvement since it makes "manual" redirects more like non-"manual" ones.

I might tweak this PR to change "next redirect steps" to "next manual redirect steps", and "process the next redirect" to "process the next redirect manually" or even "process the next HTTP redirect manually".

@annevk, what do you think? Can we merge this, if I rebase it and maybe perform those renames?

annevk (approved):
Thanks @domenic and @noamr!

PR #1481, opened 2022-08-22
[draft] Pass in content type to resource-timing

abinpaul1:
These changes are to support addition of conten type to Perfomance resource timing. Further details are available at https://github.com/w3c/resource-timing/pull/341

   * Chromium: https://chromestatus.com/feature/5156068351541248
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …


fetch.bs
```diff
@@ -4383,6 +4383,13 @@ steps:
 
      <li><p>Let <var>bodyInfo</var> be <var>response</var>'s <a for=response>body info</a>.
 
+     <li><p>Let <var>mimeType</var> be the empty string if <var>fetchParams</var>'s
+     <a for="fetch params">request</a>'s <a for=request>mode</a> is "<code>navigate</code>" and
+     <var>response</var>'s <a for=response>URL</a>'s <a for=url>origin</a> is not
+     <a>same origin</a> with <var>request</var>'s <a for=request>origin</a>, and
+     result of <a for="header list">extracting a MIME type</a> from <var>response</var>'s
```
yoavweiss, line 4390: Can this be called with an opaque response?

abinpaul1: Yeah, when it's opaque we have to keep an empty string. 

This seems to effectively be the same condition that we are applying to `responseStatus` as well. Would it make sense to keep the result of this condition in a variable and use it to determine `responseStatus`, `contentType` and others fields if any that would have to be gated on CORS.

Something like : 
Let _shouldReport_ be `true` if response's type is  "opaque"  or  fetchParams ’s request ’s mode is "navigate " and  response’s url ’s origin is not same origin with  request’s origin, and `false` otherwise.

Let _mimeType_ be result of extracting a MIME type from response’s  header list if _shouldReport_ is true and 0 otherwise.

Let _responseStatus_ be response's status if _shouldReport_ is true and 0 otherwise.

yoavweiss: What @noamr did elsewhere (e.g. with "body info" is to only initialize it when the response is not opaque. It might be worthwhile to repeat the same pattern (or hook the info you want here into "body info" itself, as mime type kinda feels related.

@noamr - thoughts?

noamr: > What @noamr did elsewhere (e.g. with "body info" is to only initialize it when the response is not opaque. It might be worthwhile to repeat the same pattern (or hook the info you want here into "body info" itself, as mime type kinda feels related.
> 
> @noamr - thoughts?

Yes mime-type should be inside `body info`.

abinpaul1: I've moved content-type inside `body info`. But can we rely on `body info`'s initialization for `content type` as well ? 

`body info` initialization seems to depend on [timing allow passed flag](https://fetch.spec.whatwg.org/#concept-response-timing-allow-passed). In [fetch response handover](https://fetch.spec.whatwg.org/#fetch-finale) 3.3.5, `body info` is reinitialized depending on `timing allow passed flag` which is different from the conditions we need for `content type`

fetch.bs
```diff
@@ -4391,6 +4394,14 @@ steps:
 
       <p class=note>This covers the case of <var>response</var> being a <a>network error</a>.
 
+     <li><p>Let <a for=response>body info</a>'s <a for="response body info">content type</a> be the
+     empty string if <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>" or
+     <var>fetchParams</var>'s <a for="fetch params">request</a>'s <a for=request>mode</a> is
+     `<code>navigate</code>` and <var>response</var>'s <a for=response>URL</a>'s
+     <a for=url>origin</a> is not <a>same origin</a> with <var>request</var>'s
+     <a for=request>origin</a>, and result of <a for="header list">extracting a MIME type</a> from
+     <var>response</var>'s <a for=response>header list</a> otherwise.
```
annevk, line 4403: I have similar concerns here to your other PR about the same origin check and redirects, and also how to word otherwise cases.

Also, when you can only "let" a variable, you need to "set" a field ("to" something).

annevk:
It also seems like we need to have https://github.com/w3c/server-timing/issues/89 resolved before continuing here.

yoavweiss:
> It also seems like we need to have [w3c/server-timing#89](https://github.com/w3c/server-timing/issues/89) resolved before continuing here.

@annevk the equivalent concerns related to this are currently handled in the [ResourceTiming PR](https://github.com/w3c/resource-timing/pull/341/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R714), essentially enabling user agents to trim entropy off of mime types to align them with a well-known list.

We tried to go off a [well-known list of types in MIMESNIFF](https://mimesniff.spec.whatwg.org/#mime-type-groups), but there seem to be types that have inherent entropy in their essence ("+zip" and "+xml" types specifically) which made that unreasonably hard.

So, I guess the open questions are:
a) Is implementation-defined trimming good enough from your perspective? Given that the concern doesn't have consensus around it, I thought this would be good enough.
b) Do you prefer the trimming to happen in Fetch rather than in RT?

/cc @achristensen07

annevk:
I think it should happen in Fetch, if that is something that we plan on doing. I also think it should be standardized as it seems like it could be an interop hazard.

Could we perhaps normalize +zip to application/zip and +xml to application/xml and call it a day? And perhaps special case a few types such as image/svg+xml so they are preserved.

yoavweiss:
I'm wary of pushing for a complex mechanism here, given that there's no consensus that https://github.com/w3c/server-timing/issues/89 is indeed an issue. Does the fact that this would require CORS (rather than TAO) change the calculus on Apple's threat model?

yoavweiss:
> @yoavweiss I think so, but then we also don't need stripping to just the essence, right?

From my perspective, we don't need that (for neither this nor ServerTiming, which is TAO protected). I know that @achristensen07 disagrees with me on the ServerTiming/TAO aspects, but I'm not sure if he does on exposing such info to CORS enabled resources.

> Not entirely clear to me how that relates to the "navigate"-related aspects of this PR, either.

For the "navigate" parts, they require same-origin, so I don't think @achristensen07's threat model applies there.

yoavweiss:
@annevk - friendly ping! Do we need the essence stripping from your perspective?

fetch.bs
```diff
@@ -346,6 +346,9 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
  <dt><dfn export for="response body info" id="fetch-timing-info-decoded-body-size">decoded
  size</dfn> (default 0)
  <dd>A number.
+<dt><dfn export for="response body info" id="response-body-info-content-type">content
```
annevk, line 349: Nit: indentation.

fetch.bs
```diff
@@ -346,6 +346,9 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
...
+ type</dfn> (default empty string)
+ <dd>A string.
```
annevk, line 351: I don't think extracting a MIME type returns a string.

fetch.bs
```diff
@@ -4518,6 +4521,14 @@ steps:
      <var>response</var>'s <a for=response>has-cross-origin-redirects</a> is true; otherwise
      <var>response</var>'s <a for=response>status</a>.
 
+     <li><p>Set <a for=response>body info</a>'s <a for="response body info">content type</a> to the
+     empty string if <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>" or
```
annevk, line 4525: As with the other check we probably don't need to check "opaque" here as "opaque" responses won't have a `Content-Type` header anyway? (Unless you poke at its internals, which we're not doing here if I remember correctly.)

annevk (commented):
Right, I agree we can expose this as-is. `Content-Type` is a safelisted response header when it comes to CORS.

fetch.bs
annevk: I think it would be cleanest btw if we store null or a MIME type here and let callers serialize as needed.

fetch.bs
abinpaul1: Fixed

fetch.bs
abinpaul1: Updated to hold a MIME type.

fetch.bs
abinpaul1: Oh right..Thanks. I've dropped the check for `opaque`.

fetch.bs
```diff
@@ -4518,6 +4521,13 @@ steps:
      <var>response</var>'s <a for=response>has-cross-origin-redirects</a> is true; otherwise
      <var>response</var>'s <a for=response>status</a>.
 
+     <li><p>If <var>fetchParams</var>'s <a for="fetch params">request</a>'s <a for=request>mode</a> is
```
annevk, line 4524: Nit: exceeds 100 columns.

fetch.bs
```diff
@@ -4518,6 +4521,13 @@ steps:
...
+     not "<code>navigate</code>" or <var>response</var>'s
+     <a for=response>has-cross-origin-redirects</a> is false, then set
+     <a for=response>body info</a>'s <a for="response body info">content type</a> to the result of
```
annevk, line 4527: This should be _bodyInfo_ presumably?

abinpaul1: Yes...updated

abinpaul1:
@annevk Added tests for checking the parsing over at https://chromium-review.googlesource.com/c/chromium/src/+/4058549.
Please have a look

annevk:
@abinpaul1 I didn't realize that directory doesn't test a bunch of the error cases around MIME type handling. It's essentially only handling the `Content-Type` `charset` parameter defaulting.

However, it also seems like you're expecting the wrong thing. It should be the mimeType field, not documentContentType, right?

abinpaul1:
> It still seems good though, but maybe add https://github.com/web-platform-tests/wpt/blob/master/mimesniff/mime-types/resources/mime-types.json to the mix?

I've tried to include those as well now.

> However, it also seems like you're expecting the wrong thing. It should be the mimeType field, not documentContentType, right?

Yeah. Changed to `mimeType` field.

annevk:
Thanks, that looks good. For the problem you're running into with `mime-types.json`, I recommend taking a look at the README https://github.com/web-platform-tests/wpt/tree/master/mimesniff/mime-types (or one of the tests using it).

abinpaul1:
Thanks again for the pointers. Updated the tests now (extra checks similar to [charset-parameter.window.js#L41](https://github.com/web-platform-tests/wpt/blob/7d03aa883513a5ffdfc12e2938ca485f8dce31b3/mimesniff/mime-types/charset-parameter.window.js#L41) before running the test case from json.

annevk (approved):
This looks good. At this point we need either Mozilla or WebKit to support this change.

annevk:
@abinpaul1 I updated this to account for the MIME Sniffing change (doesn't pass CI yet due to indexing, probably tomorrow). Will you update the tests?

fetch.bs
```diff
@@ -4753,6 +4755,20 @@ steps:
...
+      not "<code>navigate</code>" or <var>response</var>'s
+      <a for=response>has-cross-origin-redirects</a> is false:
+
+      <ol>
+       <li><p>Let <var>mimeType</var> be the result of
+       <a for="header list">extracting a MIME type</a> from <var>response</var>'s
+       <a for=response>header list</a>.
+
+       <li><p>Set <var>bodyInfo</var>'s <a for="response body info">content type</a> to the result
+       of <a>minimizing a supported MIME type</a> given <var>mimeType</var>.
```
annevk, line 4769: This needs to be conditional upon _mimeType_ not being null.

annevk:
@abinpaul1 I think we could add a `minimizedMIMEType` member. In particular for https://github.com/web-platform-tests/wpt/blob/master/mimesniff/mime-types/resources/mime-types.json that seems good. But there need to be new tests as well, yes. It might well make sense to create a JSON resource specifically for inputs and outputs of the minimized algorithm.

For types that fall into the "is supported by" conditional, I suggest we stick with types we know all browsers implement. And keep types that are in the process of being standardized/implemented in a tentative file or some such.

Also, did we have cross-origin tests already? If not we probably should have some as well, even if only for images and scripts or some such.

abinpaul1:
@annevk

I've updated the tests here : https://chromium-review.googlesource.com/c/chromium/src/+/4509283

> And keep types that are in the process of being standardized/implemented in a tentative file or some such.

I'm not sure  if I know which types are being standardized and have to be included as tentative. Would it make sense for us to just add them to the JSON for the minimized algorithm tests later when they do become standardized ?


> Also, did we have cross-origin tests already? If not we probably should have some as well, even if only for images and scripts or some such.

I think these were covered in  [content-type.html](https://github.com/web-platform-tests/wpt/blob/master/resource-timing/content-type.html) - [1](https://github.com/web-platform-tests/wpt/blob/master/resource-timing/content-type.html#L63) and [2](https://github.com/web-platform-tests/wpt/blob/master/resource-timing/content-type.html#L30)

annevk (approved):

annevk:
That seems fine, but e.g., you could assume "`image/png`" is always supported.

Issue #1486, opened 2022-09-08
Ensure that a Request object can be used as RequestInit


davidbarratt:
When constructing a `Request` object, this works in every browser:
```js
const base = new Request('https://example.com', { headers: { 'accept': 'text/html' }});
const request = new Request('https://new.example.com', base);
console.log(request.headers.get('accept')); // text/html
console.log(request.url); // https://new.example.com
```
| Browser      | Works |
| ----------  | ------- |
| Firefox       | ✅         |
| Chrome     | ✅         |
| Safari        | ✅         |
| Deno         | ✅         |

I assumed this was the _prefered_ way to make modifications to a URL of a request object, given that this does **not** work in any browser given that it drops the `headers` from the `base` completely.
```js
const base = new Request('https://example.com', { headers: { 'accept': 'text/html' }});
const request = new Request('https://new.example.com', { ...base });
console.log(request.headers.get('accept')); // null
console.log(request.url); // https://new.example.com
```
| Browser      | Works |
| ----------  | ------- |
| Firefox       | ❌         |
| Chrome     | ❌         |
| Safari        | ❌         |
| Deno         | ❌         |

This seems straightforward enough, if you need to change the URL of a request object, pass the existing request object as the second parameter.

Things get _really_ weird though when there is a body on the request.

```js
const base = new Request('https://example.com', { method: 'POST', body: 'hello!', headers: { 'accept': 'text/html' }});
const request = new Request('https://new.example.com', base);
console.log(request.headers.get('accept')); // text/html
console.log(request.url); // https://new.example.com
console.log(await request.text()); // hello!
```
| Browser      | Works | Result |
| ----------  | ------- | ------- |
| Firefox       | ❌         | `body` is dropped
| Chrome     | ❌         | ```Uncaught TypeError: Failed to construct 'Request': The `duplex` member must be specified for a request with a streaming body``` |
| Safari        | ✅         | 
| Deno         | ✅         |

Could the spec clarify that Request should work as a RequestInit?

annevk:
Can you remind me what `...base` ends up doing? The last case looks like a result from the recent changes around upload streams.

cc @jakearchibald

davidbarratt:
> Can you remind me what ...base ends up doing? 

Surprisingly, it becomes an _empty_ object.

The only way I could convert `Request` to a "plain" object was like this:
```js
const base = new Request('https://example.com', { method: 'POST', body: 'hello!', headers: { 'accept': 'text/html' }});

[...]

PR #1493, opened 2022-09-23
Add duplex property to Request

jakearchibald:
   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

Working on tests now…

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6780,6 +6780,7 @@ interface Request {
   readonly attribute boolean isReloadNavigation;
   readonly attribute boolean isHistoryNavigation;
   readonly attribute AbortSignal signal;
+  readonly attribute RequestDuplex duplex;
 
   [NewObject] Request clone();
 };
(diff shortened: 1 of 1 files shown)


jakearchibald:
Tests added. I'll create the implementation bugs when this is reviewed.

annevk (approved):
Thanks, this looks good. I was initially thinking we might want to de-duplicate some of the domintro with the duplex dictionary member, but for now this actually seems fine as they are somewhat different features even if they share a name. And we can revise when we add "full".

annevk:
I think we can land this once there are impl bugs. This is really a bug fix for #1457 as #1486 shows it's rather broken without it.

annevk:
cc @lucacasonato

lucacasonato (approved):
I agree with the motivation for this change.

I do have a concern with this implementation though, namely that this effectively sets the default value for `duplex` to `"half"` (even if the user did not specify a value). This is incompatible with Deno, because our default value has always been `"full"` there (even though this is not explicitly exposed on a `.duplex` member on the `Request` object.

I think we'll just have to swallow the "implementation will vary from spec" pill here though, as we implemented full duplex streaming before standardization.

The user observable change here will be:

```
// Chrome
const req = new Request("Hello World");
req.duplex; // "half"

// Deno
const req = new Request("Hello World");
req.duplex; // "full"
```

This is not much worse than the status quo though, as the actual behaviour would be different between Deno and other browsers regardless of this PR. This PR just explicitly exposes this difference between the default duplex streaming behaviours in a string value that the user can see.

Also of note: this does not close the door to updating the default for `duplex` being `"full"` when a user specifies a `ReadableStream` body.

LGTM

jakearchibald:
@annevk 

> I was initially thinking we might want to de-duplicate some of the domintro with the duplex dictionary member

Yeah, I felt the same, but figured developers may go straight to either bit of documentation, so the duplication seem ok.

PR #1495, opened 2022-09-27
Add a destination type for FedCM

cbiesinger:
We think it makes sense to define a destination type specific to FedCM requests (https://fedidcg.github.io/FedCM/)

This would replace the Sec-FedCM-CSRF header that is currently in the FedCM spec.

   * Chrome is interested (we're the ones proposing this)
   * Firefox is interested in FedCM [in general](https://mozilla.github.io/standards-positions/#fedcm) and has also suggested a change along these lines (https://github.com/fedidcg/FedCM/issues/320, though that issue is kind of meandering and has morphed into general CORS discussions)
   * I will add WPT tests once I update Chrome's implementation to use this destination
   * Chrome: https://crbug.com/1368382
   * Firefox: n/a, fedcm impl still in progress
   * Safari: n/a, does not support fedcm
   * Deno (not for CORS changes): n/a

***
<a href="https://whatpr.org/fetch/1495.html" title="Last updated on Oct 4, 2022, 12:50 PM UTC (58d6ae4)">Preview</a> | <a href="https://whatpr.org/fetch/1495/1fbc40c...58d6ae4.html" title="Last updated on Oct 4, 2022, 12:50 PM UTC (58d6ae4)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1517,6 +1517,7 @@ the empty string,
 "<code>style</code>",
 "<code>track</code>",
 "<code>video</code>",
+"<code>webidentity</code>",
 "<code>worker</code>", or
 "<code>xslt</code>". Unless stated otherwise it is the empty string.
 
@@ -1548,7 +1549,7 @@ not always relevant and might require different behavior.
    <th>CSP directive
    <th>Features
   <tr>
-   <td rowspan=19>""
+   <td rowspan=20>""
    <td>"<code>report</code>"
    <td rowspan=2>&mdash;
    <td>CSP, NEL reports.
@@ -1611,6 +1612,10 @@ not always relevant and might require different behavior.
    <td>"<code>sharedworker</code>"
    <td><code>child-src</code>, <code>script-src</code>, <code>worker-src</code>
    <td><code>SharedWorker</code>
+  <tr>
+   <td>"<code>webidentity</code>"
+   <td><code>connect-src</code>
+   <td><code>Federated Credential Management requests</code>
   <tr>
    <td>"<code>worker</code>"
    <td><code>child-src</code>, <code>script-src</code>, <code>worker-src</code>


cbiesinger:
xref https://github.com/fedidcg/FedCM/issues/353

fetch.bs
```diff
@@ -1517,6 +1517,7 @@ the empty string,
 "<code>style</code>",
 "<code>track</code>",
 "<code>video</code>",
+"<code>web-identity</code>",
```
annevk, line 1520: All the other names here, such as "`sharedworker`" are without hyphen. I think I'd favor local consistency, but will defer to @domenic on this.

annevk (commented):
When are implementations planning to ship this? If that's soon, will it move out of the CG soon as well?

This also requires a PR to CSP I think, to make it use connect-src.

annevk:
The Build script discovered you need to update `<td rowspan=19>` to account for this new row.

fetch.bs
domenic: Yeah, probably best to stick with local consistency (no hyphen).

cbiesinger: Done!

cbiesinger:
To answer the questions:
- Chrome is hoping to ship in 108, but we might slip to 109
- Not sure about Firefox's plans, maybe @bvandersloot-mozilla can advise?
- We spec'd the CSP check in https://fedidcg.github.io/FedCM/#fetch-the-internal-manifest, is that not enough?
- Thanks for pointing out the rowspan thing, now fixed

annevk:
The way it works is that Fetch calls out to CSP (to ensure all fetches are governed by CSP). With what you have now we'd end up checking CSP twice and slightly different CSP policies at that.

cbiesinger:
Oh OK, I will work on that. Can you elaborate on how it is a slightly different policy...?

annevk:
Just because of the code path it goes down in CSP which doesn't currently know about this destination.

cbiesinger:
I have sent https://github.com/w3c/webappsec-csp/pull/567

Is there anything else that's blocking merging this PR?

cbiesinger:
Tests are in https://github.com/web-platform-tests/wpt/pull/36230

bvandersloot-mozilla:
>     * Not sure about Firefox's plans, maybe @bvandersloot-mozilla can advise?

We are supportive and implementing FedCM. It is harder for us to be too precise at this point. Current ballpark is early next year.

annevk (approved):

annevk:
Thanks @cbiesinger!

fetch.bs
```diff
@@ -1517,6 +1517,7 @@ the empty string,
 "<code>style</code>",
 "<code>track</code>",
 "<code>video</code>",
+"<code>webidentity</code>",
```
smaug----, line 1520: Shouldn't this have been added to the dictionary too, or how is https://fetch.spec.whatwg.org/#dom-request-destination supposed to work?

annevk: It depends a bit on whether it's exposed to service workers, but if it is then yes. I filed https://github.com/whatwg/fetch/pull/1500 on that.

cbiesinger: Thanks for pointing that out. It is not exposed to service workers but I'll let Anne handle the question of whether it should be added to the enum anyway, since they have a patch already.

Issue #1496, opened 2022-09-30
Chromium's implementation of use-URL-credentials probably does not match the spec


domenic:
Spinning off from https://github.com/whatwg/fetch/pull/465#issuecomment-1263103213 .

We suspect that Chromium implements this flag by stripping the username and password from the URL before doing the fetch, which will cause the server worker, or any redirect destinations, to observe the modified URL. Whereas in the Fetch spec, there's a separate boolean which causes the URL credentials to be not-used.

This might be just a Chromium bug, but it's worth checking at least WebKit given the shared lineage. It's possible we might want to update the spec to match Chromium's behavior instead, as arguably reducing the number of URLs with usernames/passwords in them throughout the ecosystem is nice.

First step is to write some proper web platform tests, I guess.

domenic:
Stripping the username and password also doesn't match the algorithm in [HTTP-network-or-cache fetch](https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch), which roughly uses the flag only when both URL credentials and authentication side table credentials are present. In other words, if *only* URL credentials are present, the spec still uses them, whereas Chromium does not (since they are stripped from the URL). See https://github.com/whatwg/fetch/pull/1498#discussion_r985369200 for a more detailed analysis.

So yeah, tests needed in general...

PR #1498, opened 2022-09-30
Explain the use-URL-credentials flag

domenic:
***
<a href="https://whatpr.org/fetch/1498.html" title="Last updated on Oct 5, 2022, 3:53 AM UTC (7c8c508)">Preview</a> | <a href="https://whatpr.org/fetch/1498/1fbc40c...7c8c508.html" title="Last updated on Oct 5, 2022, 3:53 AM UTC (7c8c508)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1783,7 +1783,7 @@ Unless stated otherwise, it is unset.
 
 <p class=note>When this flag is set, when a <a for=/>request</a>'s
 <a for=request>URL</a> has a <a for=url>username</a> and <a for=url>password</a>, and there is an
-available <a>authentication entry</a> for the <a for=/>request</a>, then the <a>URL</a>'s
+available <a>authentication entry</a> for the <a for=/>request</a>, then the <a for=/>URL</a>'s
 credentials are preferred over that of the <a>authentication entry</a>. Modern specifications avoid
 setting this flag, since putting credentials in <a for=/>URLs</a> is discouraged, but some older
 features set it for compatibility reasons.
--- a/fetch.bs
+++ b/fetch.bs
@@ -1781,6 +1781,12 @@ which is "<code>omit</code>", "<code>same-origin</code>", or
 <dfn export for=request id=concept-request-use-url-credentials-flag>use-URL-credentials flag</dfn>.
 Unless stated otherwise, it is unset.
 
+<p class=note>This flag controls whether the <a for=/>request</a>'s <a for=request>URL</a>'s
+<a for=url>username</a> and <a for=url>password</a> will be used to look up an
+<a>authentication entry</a> or not. Modern specifications avoid setting it, since putting
+credentials in <a for=/>URLs</a> is discouraged, but some older features set it for compatibility
+reasons.
+
 <p>A <a for=/>request</a> has an associated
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1781,6 +1781,12 @@ which is "<code>omit</code>", "<code>same-origin</code>", or
 <dfn export for=request id=concept-request-use-url-credentials-flag>use-URL-credentials flag</dfn>.
 Unless stated otherwise, it is unset.
 
+<p class=note>This flag controls whether the <a for=/>request</a>'s <a for=request>URL</a>'s
+<a for=url>username</a> and <a for=url>password</a> will be used to look up an
+<a>authentication entry</a> or not. Modern specifications avoid setting it, since putting
+credentials in <a for=/>URLs</a> is discouraged, but some older features set it for compatibility
+reasons.
```
annevk, line 1788: This is incorrect. If this flag is true and URL contains username/password, we'll use the URL's username/password instead of the authentication entry (which is stored in a side table).

domenic: Eek, OK, I think I understand the relevant conditionals now. So let me break it down to make sure.

There are three input conditions which interact with the UUC flag:

1. authentication entry exists in the side table
2. URL includes credentials
3. isAuthenticationFetch is true, i.e., are we doing a second fetch after a 401 with WWW-Authenticate

If the UUC flag is not set, then:

- (1) + (2) + (3) => use side table
- (1) + (2) => use side table
- (1) + (3) => use side table
- (2) + (3) => use credentials from the URL
- (1) => use side table
- (2) => use credentials from the URL
- (3) => no credentials

If the UUC flag is set, then:

- (1) + (2) + (3) => **use credentials from the URL**
- (1) + (2) => **no credentials**
- (1) + (3) => use side table
- (2) + (3) => use credentials from the URL
- (1) => use side table
- (2) => no credentials
- (3) => no credentials

Does this seem right? The (1) + (2) case is particularly a bit surprising of a change.

annevk: Yeah, this seems to match the spec, but I'm not sure it's what we want or whether it matches browsers.

If you mean "(1) + (2) => no credentials" from the second list, that seems somewhat reasonable to me. Presumably that would then result in a 401 after which the URL credentials would be used, no?

I think what's more surprising is that we end up using URL credentials when the UUC flag is not set. That seems like a bug of sorts. (Some user agents associate credentials with a request by serializing them in the URL, which is perhaps where some of this stems from. It's been a long time.)

domenic: > Presumably that would then result in a 401 after which the URL credentials would be used, no?

Eh, seems like it depends on whether the 401 results in a re-request with the prompted credentials in the URL, or in the side table. Which I'm not sure about.

> I think what's more surprising is that we end up using URL credentials when the UUC flag is not set. That seems like a bug of sorts.

My interpretation (and what I'm revising this PR to do) is that the current spec uses the flag to preference URL credentials over side table credentials, if both are present. If only one of the two is present, then it still gets used, regardless of the flag.

But, that's not how Chromium seems to behave per #1496 at least, so yeah, maybe a bug. I'll comment over there. We could add an XXX pointing to that issue if you think it's a good idea?

annevk (approved):
Apologies for the slow follow-up, I'd be happy either way.

PR #1500, opened 2022-10-06
Add "webidentity" to RequestDestination enum too

annevk:
***
<a href="https://whatpr.org/fetch/1500.html" title="Last updated on Oct 7, 2022, 7:06 AM UTC (3cd0aea)">Preview</a> | <a href="https://whatpr.org/fetch/1500/c9039b9...3cd0aea.html" title="Last updated on Oct 7, 2022, 7:06 AM UTC (3cd0aea)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1556,9 +1556,9 @@ the empty string,
 "<code>webidentity</code>",
 "<code>worker</code>", or
 "<code>xslt</code>". Unless stated otherwise it is the empty string.
-<!-- Do not forget to update the corresponding RequestDestination enum! Note that it's okay that it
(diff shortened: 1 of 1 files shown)


cbiesinger:
No, it's not exposed to serviceworkers. See e.g. https://fedidcg.github.io/FedCM/#fetch-the-client-metadata where service-worker mode is set to none.

cbiesinger:
In case you meant whether service workers can call the API, the answer is also no, https://w3c.github.io/webappsec-credential-management/#credentialscontainer has Exposed=Window

annevk:
Thanks, I meant the former. In that case I'll just add a note that clarifies why it's not in the enum.

Issue #1504, opened 2022-10-20
Make Blob reading use File API's "get stream"


annevk:
This needs to happen in "scheme fetch" and `BodyInit`'s "extract".

annevk:
In fact, extracting is so nice that "scheme fetch" can build entirely on top of that.

PR #1505, opened 2022-10-20
Ensure scheme fetch always returns a response with an actual body

annevk:
Helps with https://github.com/whatwg/fetch/issues/1504.

***
<a href="https://whatpr.org/fetch/1505.html" title="Last updated on Oct 20, 2022, 3:25 PM UTC (1322132)">Preview</a> | <a href="https://whatpr.org/fetch/1505/e5f025a...1322132.html" title="Last updated on Oct 20, 2022, 3:25 PM UTC (1322132)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4569,7 +4569,8 @@ steps:
   "<code>blank</code>", then return a new <a for=/>response</a> whose
   <a for=response>status message</a> is `<code>OK</code>`, <a for=response>header list</a> is «
   (`<code>Content-Type</code>`, `<code>text/html;charset=utf-8</code>`) », and
-  <a for=response>body</a> is the empty byte sequence.
+  <a for=response>body</a> is the <a for="body with type">body</a> of the result of
+  <a for=BodyInit>safely extracting</a> the empty byte sequence.
 
   <p>Otherwise, return a <a>network error</a>.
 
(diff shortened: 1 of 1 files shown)


andreubotella (approved):
LGTM

fetch.bs
```diff
@@ -4569,7 +4569,8 @@ steps:
   "<code>blank</code>", then return a new <a for=/>response</a> whose
   <a for=response>status message</a> is `<code>OK</code>`, <a for=response>header list</a> is «
   (`<code>Content-Type</code>`, `<code>text/html;charset=utf-8</code>`) », and
-  <a for=response>body</a> is the empty byte sequence.
+  <a for=response>body</a> is the <a for="body with type">body</a> of the result of
+  <a for=BodyInit>safely extracting</a> the empty byte sequence.
```
domenic, line 4573: A shortcut for this would be appreciated, for use in HTML as well. Basically body-from-byte-sequence. It looks like you could use it below too.

fetch.bs
```diff
@@ -4597,22 +4598,21 @@ steps:
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
-     <li><p>Let <var>response</var> be a new <a for=/>response</a> whose
-     <a for=response>status message</a> is `<code>OK</code>`.
+     <li><p>Let <var>bodyWithType</var> be the result of <a for=BodyInit>safely extracting</a>
+     <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>.
```
domenic, line 4602: Hmm. "extract"'s Blob processing is much vaguer than https://w3c.github.io/FileAPI/#blob-get-stream . What is the plan for when to use one versus the other? Should File API just use "safely extract" and get rid of "extract stream"? Basically, who should own the blob -> stream logic, File API or Fetch?

domenic: Oh, I guess https://github.com/whatwg/fetch/pull/1506 fixes this, nevermind.

fetch.bs
annevk: #1507

PR #1506, opened 2022-10-20
Actually define Blob-to-ReadableStream conversion

### LINKED ISSUE #1504: Make Blob reading use File API's "get stream"

annevk:
This depends on https://github.com/w3c/FileAPI/pull/182.

Together with #1505 this fixes #1504.

***
<a href="https://whatpr.org/fetch/1506.html" title="Last updated on Oct 20, 2022, 3:44 PM UTC (599ebef)">Preview</a> | <a href="https://whatpr.org/fetch/1506/e5f025a...599ebef.html" title="Last updated on Oct 20, 2022, 3:44 PM UTC (599ebef)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6543,9 +6543,18 @@ typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;</pre>
 steps:
 
 <ol>
- <li><p>Let <var>stream</var> be <var>object</var> if <var>object</var> is a {{ReadableStream}}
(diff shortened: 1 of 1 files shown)


andreubotella (approved):
With this change it doesn't make that much sense to have the stream be initialized first before the switch, but I guess this is fine for now. We might revisit this if we define `multipart/form-data` serialization in terms of streams.

domenic (approved):

PR #1508, opened 2022-10-21
Remove HTTP cache talk in HTTP-network fetch

annevk:
This was overlooked in https://github.com/whatwg/fetch/commit/cbca2c2f3a37084e336e14348de683f8ffa0fed9 and is already handled by HTTP-network-or-cache fetch.

Also remove the "abort when" wrapper as this can all be done synchronously. If the fetch is canceled it's caught immediately after.

And finally use the tracking-vector feature instead of a note.

***
<a href="https://whatpr.org/fetch/1508.html" title="Last updated on Oct 21, 2022, 12:36 PM UTC (53a367e)">Preview</a> | <a href="https://whatpr.org/fetch/1508/4a7bf35...53a367e.html" title="Last updated on Oct 21, 2022, 12:36 PM UTC (53a367e)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5511,12 +5511,6 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
  <li><p>Let <var>timingInfo</var> be <var>fetchParams</var>'s <a for="fetch params">timing info</a>.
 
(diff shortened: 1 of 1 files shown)


domenic (approved):
Abort and fingerprinting changes LGTM. I haven't verified the correctness of removing the no-store setting.

Issue #1512, opened 2022-10-24
The end of fetch response handover needs to operate on the internal response


annevk:
Steps 5-7 of https://fetch.spec.whatwg.org/#fetch-finale don't work for opaque responses. We want to actually give a signal after the entire body is read which if we're just looking at the filtered response wouldn't work because body would be null. And despite it being somewhat unsafe, I think @domenic also expressed that he would expect callers of "process response consume body" to get all the bytes of an opaque response without ceremony.

Found as part of #1509. Should maybe be fixed together with #1511.

cc @noamr

annevk:
Should also update the warning in https://fetch.spec.whatwg.org/#fetch-elsewhere-fetch as part of this.

Issue #1511, opened 2022-10-24
Does Server-Timing work when not exposed through Access-Control-Expose-Headers?


annevk:
While working on #1509 I realized something about this step in [fetch response handover](https://fetch.spec.whatwg.org/#fetch-finale):

> If response is not a network error and fetchParams’s request’s client is a secure context, then set timingInfo’s server-timing headers to the result of getting, decoding, and splitting \``Server-Timing`\` from response’s header list.

If you have a CORS filtered response here it's highly likely you don't have access to the \``Server-Timing`\` header, except if the server explicitly safelisted it. Are those the intended semantics?

cc @yoavweiss @noamr

yoavweiss:
Thanks for catching that! Those are not the intended semantics. Server-Timing + Timing-Allow-Origin is the opt-in to get Server-Timing data (and how implementations currently work AFAIK).

What's the best way to fix that? Attach the Server-Timing value to the response further up the chain (e.g. before step 13 of Main Fetch)?

annevk:
Using https://fetch.spec.whatwg.org/#concept-internal-response plus a note as to why it's safe seems good enough. And maybe we should have an explicit test that it's not exposed to `fetch()` callers (i.e., only through this side channel).

PR #1513, opened 2022-10-24
Describe filtered responses better

### LINKED ISSUE #1509: Describe filtered responses better
There's a couple things that need to be done:

1. There's a number of call sites that assume filtered responses have the same associated concepts as normal responses do. They probably should, except if this causes a problem of some kind. Needs to be double checked and then we should update the definition of a filtered response to make it clear that unless stated otherwise it has all the same associated concepts.
2. It would be good to check if response call sites in general use the filtered response correctly as I haven't been as disciplined about this as I should have been.

annevk:
Fixes #1509. Follow-ups: #1511 & #1512.

***
<a href="https://whatpr.org/fetch/1513.html" title="Last updated on Oct 25, 2022, 8:44 AM UTC (9f5e5c4)">Preview</a> | <a href="https://whatpr.org/fetch/1513/6291752...9f5e5c4.html" title="Last updated on Oct 25, 2022, 8:44 AM UTC (9f5e5c4)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2321,7 +2321,7 @@ of defining the concrete types of <a for=/>filtered responses</a>.)
  equivalent parameters exposes <a for=/>filtered responses</a> to callers to ensure they do not
  accidentally leak information. If the information needs to be revealed for legacy reasons, e.g., to
  feed image data to a decoder, the associated <a for="filtered response">internal response</a> can
- only be used by specification algorithms.
+ be used by specification algorithms.
 
  <p>New specifications ought not to build further on <a>opaque filtered responses</a> or
  <a>opaque-redirect filtered responses</a>. Those are legacy constructs and cannot always be
--- a/fetch.bs
+++ b/fetch.bs
@@ -2317,10 +2317,11 @@ that offers a limited view on an associated <a for=/>response</a>. This associat
 of defining the concrete types of <a for=/>filtered responses</a>.)
 
 <div class=note>
- <p>The <a for=/>fetch</a> algorithm returns such a view to ensure APIs do not accidentally leak
- information. If the information needs to be exposed for legacy reasons, e.g., to feed image data to
- a decoder, the associated <a for="filtered response">internal response</a> can be used, which is
- only "accessible" to internal specification algorithms.
+ <p>The <a for=/>fetch</a> algorithm by way of <a for=fetch><i>processResponse</i></a> and
+ equivalent parameters exposes <a for=/>filtered responses</a> to callers to ensure they do not
+ accidentally leak information. If the information needs to be revealed for legacy reasons, e.g., to
+ feed image data to a decoder, the associated <a for="filtered response">internal response</a> can
+ be used by internal specification algorithms.
 
  <p>New specifications ought not to build further on <a>opaque filtered responses</a> or
  <a>opaque-redirect filtered responses</a>. Those are legacy constructs and cannot always be
@@ -2379,6 +2380,24 @@ is a <a>filtered response</a> whose
 
  <p>This also means that JavaScript APIs, such as
  <a attribute for=Response lt=ok><code>response.ok</code></a>, will return rather useless results.
+</div>
+
+<div class=example id=example-filtered-responses>
+ <pre><code class=lang-javascript>
+// The <a for=response>type</a> of a <a for=/>response</a> is exposed to script.
+
+(async function filteredResponseDemo() {
+  console.log(new Response().type); // "default"
+
+  console.log((await fetch("/")).type); "basic"
+
+  console.log((await fetch("https://api.example/status")).type); // "cors"
+
+  console.log((await fetch("https://crossorigin.example/image", { mode: "no-cors" })).type); // "opaque"
+
+  console.log((await fetch("/surprise-me", { redirect: "manual"})).type); // "opaqueredirect"
+})();
+</code></pre>
 </div>
 
 <hr>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2304,18 +2304,28 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
...
+<a for=/>response</a> that is neither a <a for=/>network error</a> nor a
+<a for=/>filtered response</a>).
+
+<p>Unless stated otherwise a <a for=/>filtered response</a>'s associated concepts (such as its
+<a for=response>body</a>) refer to the associated concepts of its
+<a for="filtered response">internal response</a>. (The exceptions to this are listed below as part
+of defining the concrete types of <a for=/>filtered responses</a>.)
+
+<div class=note>
+ <p>The <a for=/>fetch</a> algorithm returns such a view to ensure APIs do not accidentally leak
```
domenic, line 2320: Fetch no longer "returns" responses, so maybe tweak this?

fetch.bs
```diff
@@ -2304,18 +2304,28 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
...
+ information. If the information needs to be exposed for legacy reasons, e.g., to feed image data to
+ a decoder, the associated <a for="filtered response">internal response</a> can be used, which is
+ only "accessible" to internal specification algorithms.
```
domenic, line 2323: What internal specification algorithms is it accessible to?

fetch.bs
```diff
@@ -2304,18 +2304,28 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
...
+
+ <p>New specifications ought not to build further on <a>opaque filtered responses</a> or
+ <a>opaque-redirect filtered responses</a>. Those are legacy constructs and cannot always be
+ adequately protected given contemporary computer architecture.
+</div>
 
 <p>A <dfn export id=concept-filtered-response-basic>basic filtered response</dfn> is a
```
domenic, line 2330: Maybe out of scope / better as a followup, but I don't know what each of these types are used for. Especially "basic filtered response", and why there are two types of responses that sound similar ("basic" vs. "default"). Examples would be helpful.

fetch.bs
annevk: Any that need access? E.g., scripts, images, style sheets.

fetch.bs
annevk: "basic" is essentially about hiding cookies. "default" is a response that's not filtered at all. I think you only get that with synthetic responses that haven't gone through fetch.

fetch.bs
domenic: I stil find this sentence confusing them... what is an example of a non-internal specification algorithm, which doesn't have access?

fetch.bs
```diff
@@ -2371,6 +2382,32 @@ is a <a>filtered response</a> whose
  <a attribute for=Response lt=ok><code>response.ok</code></a>, will return rather useless results.
 </div>
 
+<div class=example id=example-filtered-responses>
+ <p>The <a for=response>type</a> of a <a for=/>response</a> is exposed to script through the
+ {{Response/type}} getter:
+
+ <pre><code class=lang-javascript>
+(async function filteredResponseDemo() {
```
domenic, line 2390: You can assume top-level await now, IMO, so the wrapper is no longer necessary :).

fetch.bs
```diff
@@ -2304,18 +2304,29 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
...
-<dfn export id=concept-internal-response for="filtered response">internal response</dfn>.
-
-<p class="note no-backref">The <a for=/>fetch</a> algorithm returns such a view to ensure APIs do
-not accidentally leak information. If the information needs to be exposed for legacy reasons, e.g.,
-to feed image data to a decoder, the associated <a for="filtered response">internal response</a> can
-be used, which is only "accessible" to internal specification algorithms and is never a
-<a>filtered response</a> itself.
+<p>A <dfn export id=concept-filtered-response>filtered response</dfn> is a <a for=/>response</a>
+that offers a limited view on an associated <a for=/>response</a>. This associated
+<a for=/>response</a> can be accessed through <a>filtered response</a>'s associated
```
domenic, line 2309:
```suggestion
<a for=/>response</a> can be accessed through <a>filtered response</a>'s
```

to reduce the repetition of "associated"

fetch.bs
annevk: Oh, I could remove the word "internal".

domenic (approved):

fetch.bs
```diff
@@ -2304,18 +2304,29 @@ known as an <dfn export id=concept-aborted-network-error>aborted network error</
...
+<a for=response>body</a>) refer to the associated concepts of its
+<a for="filtered response">internal response</a>. (The exceptions to this are listed below as part
+of defining the concrete types of <a for=/>filtered responses</a>.)
+
+<div class=note>
+ <p>The <a for=/>fetch</a> algorithm by way of <a for=fetch><i>processResponse</i></a> and
+ equivalent parameters exposes <a for=/>filtered responses</a> to callers to ensure they do not
+ accidentally leak information. If the information needs to be revealed for legacy reasons, e.g., to
+ feed image data to a decoder, the associated <a for="filtered response">internal response</a> can
+ only be used by specification algorithms.
```
annevk, line 2324:
```suggestion
 be used by specification algorithms.
```

PR #1516, opened 2022-10-25
Editorial: WebSockets has its own standard now

annevk:
Depends on https://github.com/whatwg/websockets/pull/36.

Co-Authored-By: Yutaka Hirano <yhirano@chromium.org>

***
<a href="https://whatpr.org/fetch/1516.html" title="Last updated on Oct 26, 2022, 9:11 AM UTC (3709ea4)">Preview</a> | <a href="https://whatpr.org/fetch/1516/b56692a...3709ea4.html" title="Last updated on Oct 26, 2022, 9:11 AM UTC (3709ea4)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -62,9 +62,6 @@ urlPrefix:https://tc39.es/ecma262/#;type:dfn;spec:ecma-262
     "HSTS": {
         "aliasOf": "RFC6797"
     },
-    "WSP": {
-        "aliasOf": "RFC6455"
-    },
     "HTTPVERBSEC1": {
         "publisher": "US-CERT",
         "href": "https://www.kb.cert.org/vuls/id/867593",
--- a/fetch.bs
+++ b/fetch.bs
@@ -8171,10 +8171,17 @@ requests</a> — and do not use `<code>Vary</code>`.
 
 <h3 class=no-num id=websocket-protocol oldids=websocket-connections,websocket-opening-handshake,fail-the-websocket-connection,the-websocket-connection-is-established>WebSockets</h2>
 
+<p>As part of establishing a connection, the {{WebSocket}} object initiates a special kind of
+<a for=/>fetch</a> (using a <a for=/>request</a> whose <a for=request>mode</a> is
+"<code>websocket</code>") which allows it to share in many fetch policy decisions, such
+<cite>HTTP Strict Transport Security (HSTS)</cite>. Ultimately this results in fetch calling into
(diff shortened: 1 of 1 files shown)


annevk:
It would be even better if this contained some actual brief "Background reading" on how WebSockets layers on top of Fetch.

fetch.bs
```diff
@@ -805,8 +802,8 @@ A: 3
  <li><p>Otherwise, <a for=list>append</a> (<var>name</var>, <var>value</var>) to <var>list</var>.
 </ol>
 
-<p class="note no-backref"><a for="header list">Combine</a> is used by {{XMLHttpRequest}} and the
-<a lt="establish a WebSocket connection">WebSocket protocol handshake</a>.
+<p class=note><a for="header list">Combine</a> is used by {{XMLHttpRequest}} and the
+<a spec=websockets lt="establish a WebSocket connection">WebSocket protocol handshake</a>.
```
domenic, line 806: This `spec=` shouldn't be necessary, here and below. It pierces the `export=""` barrier so is generally bad.

I guess it is necessary for now since Bikeshed gets confused between Fetch and WebSockets as the link destination. But that's better fixed with link-defaults plus a TODO to remove them in the future, I believe.

fetch.bs
```diff
@@ -8316,6 +8169,14 @@ resource — for non-<a>CORS requests</a> as well as <a>CORS
 requests</a> — and do not use `<code>Vary</code>`.
 
 
+<h3 class=no-num id=websocket-protocol oldids=websocket-connections,websocket-opening-handshake,fail-the-websocket-connection,the-websocket-connection-is-established>WebSockets</h2>
+
+<p class=note>Fetch used to define
```
domenic, line 8174: A section with only a note looks strange. Maybe just make this a normal paragraph.

fetch.bs
annevk: In the previous PR you argued we should use `spec=` for this one since there was no reason for WebSockets to export it.

domenic: Oh, I thought you were making them export it. I guess I got confused. It's fine then if you're not.

annevk: I made WebSockets export obtain a WebSocket connection (that is used by Fetch). Establish a WebSocket connection is somewhat confusingly named but that is the handshake part the API ends up invoking.

domenic (approved):
LGTM since my only remaining valid comment is about the section with only a note, about which I don't feel strongly.

PR #1520, opened 2022-10-29
Add support for blob range requests

dlrobertson:
   * Chromium and Webkit seem to have at least partial implementations
   * https://github.com/web-platform-tests/wpt/pull/34384
   * https://github.com/web-platform-tests/wpt/pull/37569
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1784880
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1399678
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=248994

Implement support for blob range requests.

Related to: https://github.com/whatwg/fetch/issues/1070

***
<a href="https://whatpr.org/fetch/1520.html" title="Last updated on Feb 20, 2023, 9:14 AM UTC (e330519)">Preview</a> | <a href="https://whatpr.org/fetch/1520/f6e9b8d...e330519.html" title="Last updated on Feb 20, 2023, 9:14 AM UTC (e330519)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4766,21 +4766,105 @@ steps:
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
-     <li><p>Let <var>bodyWithType</var> be the result of <a for=BodyInit>safely extracting</a>
-     <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>.
+     <li><p>Let <var>response</var> be a new <a for=/>response</a>.
 
-     <li><p>Let <var>body</var> be <var>bodyWithType</var>'s <a for="body with type">body</a>.
-
-     <li><p>Let <var>length</var> be <var>body</var>'s <a for=body>length</a>,
+     <li><p>Let <var>fullLength</var> be <var>blobURLEntry</var>'s
+     <a for="blob URL entry">object</a>'s {{Blob/size}},
      <a lt="serialize an integer">serialized</a> and <a>isomorphic encoded</a>.
 
-     <li><p>Let <var>type</var> be <var>bodyWithType</var>'s <a for="body with type">type</a> if it
-     is non-null; otherwise the empty byte sequence.
+     <li><p>Let <var>type</var> be <var>blobURLEntry</var>'s
+     <a for="blob URL entry">object</a>'s {{Blob/type}}.
 
-     <li><p>Return a new <a for=/>response</a> whose <a for=response>status message</a> is
-     `<code>OK</code>`, <a for=response>header list</a> is « (`<code>Content-Length</code>`,
-     <var>length</var>), (`<code>Content-Type</code>`, <var>type</var>) », and
-     <a for=response>body</a> is <var>body</var>.
+     <li>
+      <p>If <var>request</var>'s <a for=request>header list</a>
+      <a for="header list">does not contain</a> `<code>Range</code>`:
+
+      <ol>
+       <li><p>Let <var>bodyWithType</var> be the result of <a for=BodyInit>safely extracting</a>
+       <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>.
+
+       <li><p>Set <var>response</var>'s <a for=response>status message</a> to `<code>OK</code>`.
+
+       <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>bodyWithType</var>'s
+       <a for="body with type">body</a>.
+
+       <li><p>Set <var>response</var>'s <a for=response>header list</a> to «
+       (`<code>Content-Length</code>`, <var>fullLength</var>), (`<code>Content-Type</code>`,
+       <var>type</var>) ».
+      </ol>
+
+     <li>
+      <p>Otherwise:
+
+      <ol>
+       <li><p>Set <var>response</var>'s <a for=response>range-requested flag</a>.
+
+       <li><p>Let <var>rangeHeader</var> be the result of <a for="header list">getting</a>
+       `<code>Range</code>` from <var>request</var>'s <a for=request>header list</a>.
+       <!-- Range should be added to the headers defined in HTTP extensions. See
+            https://github.com/whatwg/fetch/issues/1553 for future work -->
+
+       <li><p>Let <var>rangeValue</var> be the result of <a>parsing a single range header value</a>
+       given <var>rangeHeader</var> and true.
+
+       <li><p>If <var>rangeValue</var> is failure, then return a <a>network error</a>.
+
+       <li><p>Let <var>sliceEndRange</var> be null.
+
+       <li>
+        <p>If <var>rangeValue</var>[1] is non-null, then set <var>sliceEndRange</var> to
+        <var>rangeValue</var>[1] + 1.
+
+        <p class=note>A range header denotes an inclusive byte range, while the <a>slice blob</a>
+        algorithm input range does not. To use the <a>slice blob</a> algorithm, we have to increment
+        the parsed range header end value.
+
+       <li><p>Let <var>slicedBlob</var> be the result of invoking <a>slice blob</a> given
+       <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>, <var>rangeValue</var>[0],
+       <var>sliceEndRange</var>, and <var>type</var>.
+
+       <li><p>Let <var>slicedBodyWithType</var> be the result of
+       <a for=BodyInit>safely extracting</a> <var>slicedBlob</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>slicedBodyWithType</var>'s
+       <a for="body with type">body</a>.
+
+       <li><p>Let <var>slicedLength</var> be <var>slicedBlob</var>'s {{Blob/size}},
+       <a lt="serialize an integer">serialized</a> and <a>isomorphic encoded</a>.
+
+       <!-- The following steps for content-range should be definined in a separate algorithm.
+            See https://github.com/whatwg/fetch/issues/1552 for future work -->
+       <li><p>Let <var>contentRange</var> be `<code>bytes </code>`.
+
+       <li><p>If <var>rangeValue</var>[0] is non-null, then
+       <a lt="serialize an integer">serialize</a> and <a>isomorphic encode</a>
+       <var>rangeValue</var>[0], and append the result to <var>contentRange</var>.
+
+       <li><p>Otherwise, append 0x30 (0) to <var>contentRange</var>.
+
+       <li><p>Append 0x2D (-) to <var>contentRange</var>.
+
+       <li><p>If <var>sliceEndRange</var> is non-null, then
+       <a lt="serialize an integer">serialize</a> and <a>isomorphic encode</a>
+       <var>sliceEndRange</var>, and append the result to <var>contentRange</var>.
+
+       <li><p>Otherwise, append <var>fullLength</var> to <var>contentRange</var>.
+
+       <li><p>Append 0x2F (/) to <var>contentRange</var>.
+
+       <li><p>Append <var>fullLength</var> to <var>contentRange</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>status</a> to 206.
+
+       <li><p>Set <var>response</var>'s <a for=response>status message</a> to
+       `<code>Partial Content</code>`.
+
+       <li><p>Set <var>response</var>'s <a for=response>header list</a> to «
+       (`<code>Content-Length</code>`, <var>slicedLength</var>), (`<code>Content-Type</code>`,
+       <var>type</var>), (`<code>Content-Range</code>`, <var>contentRange</var>) ».
+      </ol>
+
+     <li><p>Return <var>response</var>.
     </ol>
 
    <dt>"<code>data</code>"


dlrobertson:
Posting this a bit early in the implementation phase... See the many `TODO`s 😄 

Any feedback would be really helpful!

fetch.bs
```diff
@@ -4618,21 +4618,93 @@ steps:
...
+
+       <li><p>If <var>rangeValue</var>[1] is null, append <var>bodyLength</var> to
+       <var>contentRange</var>
+
+       <li><p>Append 0x2F (/) to <var>contentRange</var>.
+
+       <li><p>Append <var>bodyLength</var> to <var>contentRange</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>status message</a> to
+       `<code>Partial Content</code>`
```
annevk, line 4700: You want to set status to 206 as well here. (It might be okay to not set status message at all and let it be the empty byte sequence as it would be in HTTP/2 and up, but I don't care strongly.)

fetch.bs
```diff
@@ -4618,21 +4618,93 @@ steps:
...
+       <li><p>Let <var>rangeValue</var> be the result of <a>parsing a single range header value</a>
+       given <var>rangeHeader</var>.
+
+       <li><p>If <var>rangeValue</var> is failure, then return a <a>network error</a>.
+
+       <li>
+         <p>blob.slice(<var>rangeStart</var>[0], <var>rangeValue</var>[1], <var>type</var>) here
+         on <var>body</var>.
+
+         <p class="note">TODO(dlrobertson): This requires a lot more thought.
```
annevk, line 4666: Primarily we don't want to use the public JS API here but instead call an internal algorithm.

fetch.bs
```diff
@@ -4618,21 +4618,93 @@ steps:
...
+
+         <p class="note">TODO(dlrobertson): Split this out into something like
+         <a>add a range header</a>
+
+       <li><p>If <var>rangeValue</var>[0] is not null, <a lt="serialize an integer">Serialize</a> and
+       <a>isomorphic encode</a> <var>rangeValue</var>[0], and append the result to
+       <var>contentRange</var>.
+
+       <li>
+         <p>If <var>rangeValue</var>[0] is null, <a lt="serialize an integer">Serialize</a> and
```
annevk, line 4679: Maybe use "Otherwise" here instead to make it clear it's one or the other?

fetch.bs
```diff
@@ -4618,21 +4618,93 @@ steps:
...
+     <li><p>If <var>request</var>'s <a for=response>header list</a> contains `<code>Range</code>`,
+     set <var>response</var>'s <a for=response>range-requested flag</a>.
+
      <li><p>Let <var>bodyWithType</var> be the result of <a for=BodyInit>safely extracting</a>
      <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>.
 
      <li><p>Let <var>body</var> be <var>bodyWithType</var>'s <a for="body with type">body</a>.
 
-     <li><p>Let <var>length</var> be <var>body</var>'s <a for=body>length</a>,
+     <li><p>Let <var>bodyLength</var> be <var>body</var>'s <a for=body>length</a>,
```
annevk, line 4631: Maybe `initialLength`/`originalLength`/`fullLength` is clearer?

fetch.bs
dlrobertson: Made a first attempt at this in https://github.com/w3c/FileAPI/pull/183.

dlrobertson:
Updated to use [slice blob](https://w3c.github.io/FileAPI/#slice-blob).

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+       <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>, <var>rangeValue</var>[0],
+       <var>sliceEndRange</var>, and <var>type</var>.
+
+       <li><p>Let <var>slicedBodyWithType</var> be the result of
+       <a for=BodyInit>safely extracting</a> <var>slicedBlob</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>slicedBodyWithType</var>'s
+       <a for="body with type">body</a>.
+
+       <li><p>Let <var>contentRange</var> be `<code>bytes </code>`.
```
dlrobertson, line 4714: I didn't move `contentRange` into another algorithm, but this should definitely be done. I'm currently planning on doing this in a follow up issue, but I can do it here if that is preferred.

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
+     <li><p>Set <var>response</var> to a new <a for=/>response</a>.
+
+     <li><p>If <var>request</var>'s <a for=response>header list</a> contains `<code>Range</code>`,
+     set <var>response</var>'s <a for=response>range-requested flag</a>.
```
annevk, line 4654: Nit:
```suggestion
     then set <var>response</var>'s <a for=response>range-requested flag</a>.
```

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
      <li><p>Let <var>type</var> be <var>bodyWithType</var>'s <a for="body with type">type</a> if it
      is non-null; otherwise the empty byte sequence.
 
-     <li><p>Return a new <a for=/>response</a> whose <a for=response>status message</a> is
-     `<code>OK</code>`, <a for=response>header list</a> is « (`<code>Content-Length</code>`,
-     <var>length</var>), (`<code>Content-Type</code>`, <var>type</var>) », and
-     <a for=response>body</a> is <var>body</var>.
+     <li>
+      <p>If <var>response</var>'s <a for=response>range-requested flag</a> is unset, run the
+      following steps:
```
annevk, line 4669:
```suggestion
      <p>If <var>response</var>'s <a for=response>range-requested flag</a> is unset:
```

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+       <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>body</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>header list</a> to «
+       (`<code>Content-Length</code>`, <var>fullLength</var>), (`<code>Content-Type</code>`,
+       <var>type</var>) ».
+      </ol>
+
+     <li>
+      <p>Otherwise, if <var>response</var>'s <a for=response>range-requested flag</a> is set, run the
+      following steps:
```
annevk, line 4683:
```suggestion
      <p>Otherwise:
```
(I was initially thinking we could maybe keep it without the "if" or have it as assert, but since there's only two branches I don't think that's warranted.)

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+
+      <ol>
+       <li><p>Let <var>rangeHeader</var> be the result of <a>extracting header list values</a> given
```
annevk, line 4686: You want to use "get" or "get, decode, and split". We're trying to get rid of "extracting header list values".

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+
+       <li><p>Let <var>rangeValue</var> be the result of <a>parsing a single range header value</a>
+       given <var>rangeHeader</var>.
+
+       <li><p>If <var>rangeValue</var> is failure, then return a <a>network error</a>.
+
+       <li><p>Let <var>sliceEndRange</var> be null.
+
+       <li>
+        <p>If <var>rangeValue</var>[1] is non-null, let <var>sliceEndRange</var> be
```
annevk, line 4697: then set _sliceEndRange_ to*

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+       <a for="body with type">body</a>.
+
+       <li><p>Let <var>contentRange</var> be `<code>bytes </code>`.
+
+       <li><p>If <var>rangeValue</var>[0] is not null, <a lt="serialize an integer">Serialize</a> and
+       <a>isomorphic encode</a> <var>rangeValue</var>[0], and append the result to
+       <var>contentRange</var>.
+
+       <li><p>Otherwise, if <var>rangeValue</var>[0] is null, append <code>`0`</code> to
+       <var>contentRange</var>.
```
annevk, line 4721:
```suggestion
       <li><p>Otherwise, append 0x30 (0) to <var>contentRange</var>.
```

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+
+       <li><p>Append 0x2D (-) to <var>contentRange</var>.
+
+       <li><p>If <var>sliceEndRange</var> is not null, <a lt="serialize an integer">Serialize</a>
```
annevk, line 4725:
```suggestion
       <li><p>If <var>sliceEndRange</var> is non-null, then <a lt="serialize an integer">serialize</a>
```

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+       and <a>isomorphic encode</a> <var>sliceEndRange</var>, and append the result to
+       <var>contentRange</var>.
+
+       <li><p>Otherwise, if <var>sliceEndRange</var> is null, append <var>fullLength</var> to
+       <var>contentRange</var>.
```
annevk, line 4730:
```suggestion
       <li><p>Otherwise, append <var>fullLength</var> to <var>contentRange</var>.
```

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+
+       <li><p>Append <var>fullLength</var> to <var>contentRange</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>status message</a> to
+       `<code>Partial Content</code>`.
+
+       <li><p>Set <var>response</var>'s <a for=response>status</a> to `<code>206</code>`.
+
+       <li><p>Set <var>response</var>'s <a for=response>header list</a> to «
+       (`<code>Content-Length</code>`, <var>length</var>), (`<code>Content-Type</code>`,
```
annevk, line 4742: Where is `<var>length</var>` defined?

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+       <var>type</var>), (`<code>Content-Range</code>`, <var>contentRange</var>) ».
```
annevk, line 4743: I guess the assumption here is that _type_ couldn't have changed?

fetch.bs
annevk: I think that's fine, but let's make sure it's filed and referenced from an HTML comment?

fetch.bs
```diff
@@ -4648,21 +4648,102 @@ steps:
...
+      </ol>
+
+     <li><p> return <var>response</var>.
```
annevk, line 4746:
```suggestion
     <li><p>Return <var>response</var>.
```

annevk (commented):
I spotted a lot of tiny nits, but I might have missed something bigger therefore. Probably needs another close look once these are addressed. Hope that's understandable.

fetch.bs
dlrobertson: Yeah, I don't see how `type` could change, so I figured I'd continue to use it.

fetch.bs
dlrobertson: :+1: Added a HTML comment and plan to open a issue when merged. I can open one now if that is the preferred workflow.

dlrobertson (commented):
> I spotted a lot of tiny nits, but I might have missed something bigger therefore. Probably needs another close look once these are addressed. Hope that's understandable.

Definitely. Feedback is greatly appreciated!

fetch.bs
```diff
@@ -4648,21 +4648,106 @@ steps:
...
+       <li><p>Let <var>rangeValue</var> be the result of <a>parsing a single range header value</a>
+       given <var>rangeHeader</var>.
+
+       <li><p>If <var>rangeValue</var> is failure, then return a <a>network error</a>.
+
+       <li><p>Let <var>sliceEndRange</var> be null.
+
+       <li>
+        <p>If <var>rangeValue</var>[1] is non-null, then set <var>sliceEndRange</var> to
+        <code><var>rangeValue</var>[1] + 1</code>.
```
annevk, line 4697:
```suggestion
        <var>rangeValue</var>[1] + 1.
```
Equations aren't code. At least that's not what we've gone with thus far.

fetch.bs
```diff
@@ -4648,21 +4648,106 @@ steps:
...
+
+        <p class=note>A range header denotes an inclusive byte range, while the <a>slice blob</a>
+        algorithm input range does not. To use the <a>slice blob</a> algorithm, we have to increment
+        the parsed range header end value.
+
+       <li><p>Let <var>slicedBlob</var> be the result of invoking <a>slice blob</a> given
+       <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>, <var>rangeValue</var>[0],
+       <var>sliceEndRange</var>, and <var>type</var>.
```
annevk, line 4705: So technically "slice blob" cannot be invoked with arguments that are null. Perhaps "slice blob" should require the arguments (and allow null) and `slice()` should do some appropriate defaulting before forwarding?

fetch.bs
```diff
@@ -4648,21 +4648,106 @@ steps:
...
+       <var>type</var>) ».
+      </ol>
+
+     <li>
+      <p>Otherwise:
+
+      <ol>
+       <li><p>Let <var>rangeHeader</var> be the result of <a for="header list">getting</a>
+       `<code>Range</code>` from <var>request</var>'s <a for=request>header list</a>.
+       <!-- Range should be added to the headers defined in HTTP extensions. -->
```
annevk, line 4686: Yeah, I guess that makes sense. And then move the single range header value parser there?

fetch.bs
```diff
@@ -4648,21 +4648,106 @@ steps:
...
+       <li><p>Otherwise, append <var>fullLength</var> to <var>contentRange</var>.
+
+       <li><p>Append 0x2F (/) to <var>contentRange</var>.
+
+       <li><p>Append <var>fullLength</var> to <var>contentRange</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>status message</a> to
+       `<code>Partial Content</code>`.
+
+       <li><p>Set <var>response</var>'s <a for=response>status</a> to 206.
```
annevk, line 4743: Let's flip these two steps. Status is more significant.

fetch.bs
annevk: Yeah please open it now so we can link it from the comment.

annevk (commented):
Thanks, this looks quite clean. I still have a couple inline comments and a longer thought here for a somewhat more substantial change.

So in principle we can get the type and size from a blob, although unfortunately they are not internal concepts as-of-yet.

That means we can avoid extracting twice.

It also means we can avoid setting up a new response upfront and can probably have two return statements at the end of the two branches where we create a response and set all its fields as we do now. (They're equivalent, but I kinda like the atomic look of the current setup.)

We also need only one conditional, whether request contains a `Range` header and the two branches can flow from that. (This is doable either way.)

Even if editorial, this would be some work. What do you think?

fetch.bs
dlrobertson: This is a great point! This should be fixed by https://github.com/w3c/FileAPI/pull/184.

fetch.bs
dlrobertson: Created https://github.com/whatwg/fetch/issues/1552.

dlrobertson:
> Thanks, this looks quite clean. I still have a couple inline comments and a longer thought here for a somewhat more substantial change.
> 
> So in principle we can get the type and size from a blob, although unfortunately they are not internal concepts as-of-yet.
> 
> That means we can avoid extracting twice.

Yeah, this makes sense. I didn't change this in order to attempt to keep things closer to how they were before, but we could easily remove the `bodyWithType` steps in the section of the algorithm before the branch.

> It also means we can avoid setting up a new response upfront and can probably have two return statements at the end of the two branches where we create a response and set all its fields as we do now. (They're equivalent, but I kinda like the atomic look of the current setup.)
> 
> We also need only one conditional, whether request contains a `Range` header and the two branches can flow from that. (This is doable either way.)
> 
> Even if editorial, this would be some work. What do you think?

I like the idea of the single return statement using `Blob` [size](https://w3c.github.io/FileAPI/#dfn-size) and [type](https://w3c.github.io/FileAPI/#dfn-type) definitions, but I'm happy to change this if this isn't preferred.

fetch.bs
```diff
@@ -4648,21 +4648,108 @@ steps:
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
-     <li><p>Let <var>bodyWithType</var> be the result of <a for=BodyInit>safely extracting</a>
-     <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>.
+     <li><p>Set <var>response</var> to a new <a for=/>response</a>.
 
-     <li><p>Let <var>body</var> be <var>bodyWithType</var>'s <a for="body with type">body</a>.
+     <li><p>If <var>request</var>'s <a for=response>header list</a> contains `<code>Range</code>`,
+     then set <var>response</var>'s <a for=response>range-requested flag</a>.
```
annevk, line 4654: I still think we should deduplicate this with the conditional below.

fetch.bs
```diff
@@ -4648,21 +4648,108 @@ steps:
...
 
-     <li><p>Let <var>length</var> be <var>body</var>'s <a for=body>length</a>,
+     <li><p>Let <var>fullLength</var> be <var>blobURLEntry</var>'s
+     <a for="blob URL entry">object</a>'s <a for=Blob data-link-type=attribute>size</a>,
```
annevk, line 4657: We can use `{{Blob/size}}` as we do elsewhere in this document. Same for type.

fetch.bs
```diff
@@ -4648,21 +4648,108 @@ steps:
...
+
+       <li><p>Let <var>slicedBodyWithType</var> be the result of
+       <a for=BodyInit>safely extracting</a> <var>slicedBlob</var>.
+
+       <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>slicedBodyWithType</var>'s
+       <a for="body with type">body</a>.
+
+       <li><p>Let <var>slicedLength</var> be <var>slicedBlob</var>'s
+       <a for=Blob data-link-type=attribute>size</a>, <a lt="serialize an integer">serialized</a>
+       and <a>isomorphic encoded</a>.
```
annevk, line 4716: Here we could still refer to the length of the body (that's the result of extracting), but I'm on the fence as to whether that's better. In any event we should use `{{Blob/size}}` until we can reference some internal field on `Blob`.

annevk (commented):
Specifications are generally quite happy with multiple return statements, but I'm okay sticking with one here. Couple final nits and this seems good to go.

fetch.bs
dlrobertson: No problem! Just for future reference, what is the benefit of `{{Blob/size}}` and how does it differ from `<a for=Blob data-link-type=attribute>size</a>`? Both create links to https://w3c.github.io/FileAPI/#dfn-size for me.

annevk: Does the latter include also result in `<code>` being included? Anyway, the former is a shorthand that should do the right thing and it's useful to do it consistently within a document so you can easily find all instances.

Now having said all that, typically you would never want to reference an IDL attribute as it's the public API. The problem here is that `Blob` still doesn't define its internal model.

dlrobertson: > Does the latter include also result in <code> being included?

Ah! No it does not.

> Anyway, the former is a shorthand that should do the right thing and it's useful to do it consistently within a document so you can easily find all instances.

This makes sense.

dlrobertson:
Updated to follow your suggestions. Let me know if there is anything else you'd like changed!

annevk (commented):
Modulo a couple tiny nits this looks good.

Are Chromium and WebKit passing all the tests or should there be bugs on them as well?

dlrobertson:
> Are Chromium and WebKit passing all the tests or should there be bugs on them as well?

Good question! Both fail most of the upsupported range cases tested in https://github.com/web-platform-tests/wpt/pull/34384. WebKit returns a response with a 416 status when the range is invalid and Chromium only fails 3 of the unsupported ranges, but does return a network error. We'll need to create bugs for this I guess.

dlrobertson:
Filed the bugs:
  - WebKit: https://bugs.webkit.org/show_bug.cgi?id=248994
  - Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1399678

I definitely didn't file the chromium bug right... any feedback on filing non-firefox implementation bugs would be appreciated

annevk:
I was about to merge this, but then read https://github.com/whatwg/fetch/issues/1070#issuecomment-1279076672 again. I think we should address the whitespace issue first, perhaps in a separate PR to constrain it to CORS?

dlrobertson:
No problem! I'll take a stab at it this evening

fetch.bs
```diff
@@ -4766,21 +4766,105 @@ steps:
...
+      <ol>
+       <li><p>Set <var>response</var>'s <a for=response>range-requested flag</a>.
+
+       <li><p>Let <var>rangeHeader</var> be the result of <a for="header list">getting</a>
+       `<code>Range</code>` from <var>request</var>'s <a for=request>header list</a>.
+       <!-- Range should be added to the headers defined in HTTP extensions. See
+            https://github.com/whatwg/fetch/issues/1553 for future work -->
+
+       <li><p>Let <var>rangeValue</var> be the result of <a>parsing a single range header value</a>
+       given <var>rangeHeader</var>.
```
annevk, line 4808:
```suggestion
       given <var>rangeHeader</var> and true.
```
I think this is the only change needed here. But if you could gloss it over one last time that would be appreciated.

I suspect we'll land this somewhere next week given the Bikeshed issues.

dlrobertson: :+1: I'll re-read and post an update this weekend

dlrobertson:
Thanks @annevk for all the help on this!

Issue #1319, opened 2021-10-04
Standardize Priority Hints


pmeenan:
Priority Hints is an API that lets developers signal to the browser the relative importance of resources, so that the browser might take this into consideration when prioritizing the request. The WICG repository for Priority Hints is [here](https://github.com/WICG/priority-hints) and the explainer spec is [here](https://wicg.github.io/priority-hints/). The explainer spec lays out some good use cases. Prioritization might take the form of modifying a request's HTTP/2 or HTTP/3 priority, delaying when a request is sent out, etc.

Browsers already have an internal implicit mapping for fetch prioritization, usually based on content type and combined with other factors (background tab, preload, prefetch, in-viewport, parser-blocking, sync XHR, etc). The prioritization usually impacts the order in which requests are made and, if the underlying protocol supports multiplexing and prioritization, the protocol-specific prioritization of each stream.

Priority Hints gives a developer a mechanism to indicate to the underlying browser, a hint about the relative priority of a given request, either indirectly through the ```importance``` attribute on an HTML element that requires a resource load (img, script, preload, etc) or explicitly as an attribute on a fetch API call. The browser can take the importance hint into account when evaluating priorities to assign to individual resources.

The intent is to provide an additional developer-controlled signal to help with cases where heuristics are not enough to result in optimal loading. There should be no functional changes as the underlying processing model is not altered.

The working group settled on ```high/low``` signals (in addition to the default of ```auto```) as a way to give developers some input into the prioritization without exposing the complexity of the different underlying protocol implementations and to continue to allow for browser heuristics. Chrome has documented the current prioritization heuristics [here](https://docs.google.com/document/d/1bCDuq9H1ih9iNjgzyAL0gpwNFiEP4TZS-YLRp_RuMlc/edit?usp=sharing) including the changes that Priority Hints apply.

### Boosting Image Priority
By default, Chrome gives images a relatively low priority because they do not block the parser or the DOM Content Loaded event from firing. There are some heuristics to boost the priority of in-viewport to get the visual content loaded sooner but that priority boost can only be applied after the initial layout has been performed and the images within the viewport can be identified.

Priority hints on either the image tag or on a preload for a given image would enable developers to signal the most important images for the user experience and cause them to potentially load sooner than they would otherwise. In Chrome, this could be done at the earliest stages of the HTML parsing and not rely on being able to re-prioritize requests that are already in-flight.

### Signalling Relative API Priority
[...]

PR #1523, opened 2022-10-31
Add the Priority Hints changes to the fetch spec

pmeenan:
This applies the [specified changes](https://wicg.github.io/priority-hints/#fetch-integration) from the [Priority Hints spec](https://wicg.github.io/priority-hints/) to fetch for [#1319](https://github.com/whatwg/fetch/issues/1319).

The main changes are:
* Rename the existing implementor-internal `priority` to `internal priority`.
* Add a new `priority` enum to the `Request` and `RequestInit` objects.

   * Chrome (Shipped in 103)
   * [Mozilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1797715)
- [ ] [Tests](https://github.com/web-platform-tests/wpt) for all of the interfaces are already being run as part of the Priority Hints spec:
   * https://github.com/web-platform-tests/wpt/blob/master/interfaces/priority-hints.idl
   * https://github.com/web-platform-tests/wpt/blob/master/priority-hints/fetch-api-request.tentative.any.js
   * https://github.com/web-platform-tests/wpt/blob/master/priority-hints/service-worker-fetch-request-priority.tentative.https.html
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=821464
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1797715
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=247291
   * Deno (not for CORS changes): https://github.com/denoland/deno/issues/16502
   * Node.js: https://github.com/nodejs/undici/issues/1748
- [ ] [MDN](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests):
   * Already [documented](https://developer.mozilla.org/en-US/docs/Web/API/Request/priority) as experimental from the existing spec, can be moved to non-experimental after merged.

***
<a href="https://whatpr.org/fetch/1523.html" title="Last updated on Dec 15, 2022, 9:54 AM UTC (724abaa)">Preview</a> | <a href="https://whatpr.org/fetch/1523/7d951ac...724abaa.html" title="Last updated on Dec 15, 2022, 9:54 AM UTC (724abaa)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1832,8 +1832,9 @@ not always relevant and might require different behavior.
 <dfn export for=request>priority</dfn>, which is "<code>high</code>", "<code>low</code>", or
 "<code>auto</code>". Unless stated otherwise it is "<code>auto</code>".
 
-<p>A <a for=/>request</a> has an associated <dfn export for=request>internal priority</dfn> (null
-or an <a>implementation-defined</a> object). Unless otherwise stated it is null.
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request oldids=concept-request-priority>internal priority</dfn> (null or an
+<a>implementation-defined</a> object). Unless otherwise stated it is null.
 
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-origin>origin</dfn>, which is
@@ -4248,9 +4249,9 @@ the request.
  <li>
   <p>If <var>request</var>'s <a for=request>internal priority</a> is null, then use
   <var>request</var>'s <a for=request>priority</a>, <a for=request>initiator</a>,
-  <a for=request>destination</a>, and <a for=request>render-blocking</a> appropriately in setting
-  <var>request</var>'s <a for=request>internal priority</a> to an <a>implementation-defined</a>
-  object.
+  <a for=request>destination</a>, and <a for=request>render-blocking</a> in an
+  <a>implementation-defined</a> manner to set <var>request</var>'s
+  <a for=request>internal priority</a> to an <a>implementation-defined</a> object.
 
   <p class=note>The <a>implementation-defined</a> object could encompass stream weight and
   dependency for HTTP/2, priorities used in <cite>Extensible Prioritization Scheme for HTTP</cite>
@@ -7479,7 +7480,8 @@ constructor steps are:
 
   <ol>
    <li><p>If <var>request</var>'s <a for=request>internal priority</a> is not null, then update
-   <var>request</var>'s <a for=request>internal priority</a> appropriately.
+   <var>request</var>'s <a for=request>internal priority</a> in an <a>implementation-defined</a>
+   manner.
 
    <li><p>Otherwise, set <var>request</var>'s <a for=request>priority</a> to
    <var>init</var>["{{RequestInit/priority}}"].
--- a/fetch.bs
+++ b/fetch.bs
@@ -1829,8 +1829,11 @@ not always relevant and might require different behavior.
 <hr>
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-priority>priority</dfn> (null or a
-user-agent-defined object). Unless otherwise stated it is null.
+<dfn export for=request>priority</dfn>, which is "<code>high</code>", "<code>low</code>", or
+"<code>auto</code>". Unless stated otherwise it is "<code>auto</code>".
+
+<p>A <a for=/>request</a> has an associated <dfn export for=request>internal priority</dfn> (null
+or an <a>implementation-defined</a> object). Unless otherwise stated it is null.
 
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-origin>origin</dfn>, which is
@@ -4243,13 +4246,16 @@ the request.
  <a for=/>header value</a>) to <var>request</var>'s <a for=request>header list</a>.
 
  <li>
-  <p>If <var>request</var>'s <a for=request>priority</a> is null, then use <var>request</var>'s
-  <a for=request>initiator</a>, <a for=request>destination</a>, and
-  <a for=request>render-blocking</a> appropriately in setting <var>request</var>'s
-  <a for=request>priority</a> to a user-agent-defined object.
+  <p>If <var>request</var>'s <a for=request>internal priority</a> is null, then use
+  <var>request</var>'s <a for=request>priority</a>, <a for=request>initiator</a>,
+  <a for=request>destination</a>, and <a for=request>render-blocking</a> appropriately in setting
+  <var>request</var>'s <a for=request>internal priority</a> to an <a>implementation-defined</a>
+  object.
 
-  <p class=note>The user-agent-defined object could encompass stream weight and dependency for
-  HTTP/2, and equivalent information used to prioritize dispatch and processing of HTTP/1 fetches.
+  <p class=note>The <a>implementation-defined</a> object could encompass stream weight and
+  dependency for HTTP/2, priorities used in <cite>Extensible Prioritization Scheme for HTTP</cite>
+  for transports where it applies (including HTTP/3), and equivalent information used to prioritize
+  dispatch and processing of HTTP/1 fetches. [[!RFC9218]]
 
  <li>
   <p>If <var>request</var> is a <a>subresource request</a>, then:
@@ -7009,6 +7015,7 @@ dictionary RequestInit {
   boolean keepalive;
   AbortSignal? signal;
   RequestDuplex duplex;
+  RequestPriority priority;
   any window; // can only be set to null
 };
 
@@ -7018,6 +7025,7 @@ enum RequestCredentials { "omit", "same-origin", "include" };
 enum RequestCache { "default", "no-store", "reload", "no-cache", "force-cache", "only-if-cached" };
 enum RequestRedirect { "follow", "error", "manual" };
 enum RequestDuplex { "half" };
+enum RequestPriority { "high", "low", "auto" };
 </pre>
 
 <p class="note no-backref">"<code>serviceworker</code>" is omitted from
@@ -7115,6 +7123,9 @@ object), initially null.
    set when {{RequestInit/body}} is a {{ReadableStream}} object. <span class=note>See
    <a href="https://github.com/whatwg/fetch/issues/1254">issue #1254</a> for defining
    "<code>full</code>".</span>
+
+   <dt>{{RequestInit/priority}}
+   <dd>A string to set <var>request</var>'s <a for=request>priority</a>.
   </dl>
 
  <dt><code><var>request</var> . <a attribute for=Request>method</a></code>
@@ -7299,8 +7310,8 @@ constructor steps are:
    <dt><a for=request>window</a>
    <dd><var>window</var>.
 
-   <dt><a for=request>priority</a>
-   <dd><var>request</var>'s <a for=request>priority</a>.
+   <dt><a for=request>internal priority</a>
+   <dd><var>request</var>'s <a for=request>internal priority</a>.
 
    <dt><a for=request>origin</a>
    <dd><var>request</var>'s <a for=request>origin</a>. <span class=note>The propagation of the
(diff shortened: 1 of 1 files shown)


pmeenan:
From what I understand, Mozilla will be implementing it so I'll ping them to commit to the implementation on the issue.

valenting:
> From what I understand, Mozilla will be implementing it so I'll ping them to commit to the implementation on the issue.

That is correct. We are tracking this in https://bugzilla.mozilla.org/show_bug.cgi?id=1797715

fetch.bs
```diff
@@ -1733,7 +1733,12 @@ not always relevant and might require different behavior.
...
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-priority>priority</dfn> (null or a
+<dfn export id=concept-request-priority>priority</dfn>, which is
+"<code>high</code>", "<code>low</code>" or "<code>auto</code>". Unless stated otherwise it is
+"<code>auto</code>".
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-internal-priority>internal priority</dfn> (null or a
 user-agent-defined object). Unless otherwise stated it is null.
```
annevk, line 1742: Can we make this `an <a>implementation-defined</a> object` while here? (Also applies below.)

fetch.bs
```diff
@@ -7070,6 +7081,10 @@ object), initially null.
  <span class=note>See <a href="https://github.com/whatwg/fetch/issues/1254">issue #1254</a> for
  defining "<code>full</code>".</span>
 
+<dt><code><var>request</var> . <a attribute for=Request>priority</a></code>
```
annevk, line 7084: Indentation is wrong.

fetch.bs
```diff
@@ -7070,6 +7081,10 @@ object), initially null.
...
+ <dd>Returns the <a>priority</a> of the <var>request</var>, e.g., "<code>high</code>",
+ "<code>low</code>" or "<code>auto</code>".
```
annevk, line 7086: Oxford comma.

annevk (commented):
Thank you for working on this!

I'm not seeing the outcome of https://github.com/WICG/priority-hints/issues/69 incorporated here.

We might also want to say a bit more about "internal priority".

Please also draft a commit message (as a comment) that links the appropriate HTML changes and makes it clear that all of this essentially obsoletes the Priority Hints specification. (I'm assuming that's the intent given there's no normative references.)

fetch.bs
pmeenan: Done. All 3 of the user-agent-defined references have been changed.

fetch.bs
pmeenan: Fixed

fetch.bs
pmeenan: Fixed

pmeenan:
@annevk I think I got most of the changes done and the last commit message includes the obsoleting language. I added a comment to the render-blocking issue to discuss it a bit more before we figure out what the changes mean for the spec (needless to say, this merge will wait on the outcome of that thread).

pmeenan:
I removed `priority` from the exposed `Request` attribute interface and have it as internal-only as discussed [here](https://github.com/WICG/priority-hints/issues/69#issuecomment-1323274220).

I also have `render-blocking` and `priority` both being copied along with `internal-priority` when constructed from an existing Request in step 12 of https://fetch.spec.whatwg.org/#dom-request (let me know if you'd rather not explicitly require that they be copied).

fetch.bs
```diff
@@ -1734,8 +1734,13 @@ not always relevant and might require different behavior.
 <hr>
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-priority>priority</dfn> (null or a
-user-agent-defined object). Unless otherwise stated it is null.
+<dfn export for=request id=concept-request-priority>priority</dfn>, which is
+"<code>high</code>", "<code>low</code>" or "<code>auto</code>". Unless stated otherwise it is
+"<code>auto</code>".
```
annevk, line 1739: We need to update HTML's existing references to this concept to use "internal priority" instead.

As far as I can tell from https://dontcallmedom.github.io/webdex/p.html#priority%40%40request%40dfn HTML is the only specification that has references (and they are more of an informative nature).

fetch.bs
```diff
@@ -1734,8 +1734,13 @@ not always relevant and might require different behavior.
...
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-internal-priority>internal priority</dfn> (null or a
+<a>implementation-defined</a> object). Unless otherwise stated it is null.
```
annevk, line 1743:
```suggestion
<p>A <a for=/>request</a> has an associated <dfn export for=request>internal priority</dfn> (null or
an <a>implementation-defined</a> object). Unless otherwise stated it is null.
```

fetch.bs
```diff
@@ -4134,13 +4139,16 @@ the request.
...
+  <a for=request>destination</a>, and <a for=request>render-blocking</a> appropriately in setting
+  <var>request</var>'s <a for=request>internal priority</a> to an <a>implementation-defined</a>
+  object.
 
-  <p class=note>The user-agent-defined object could encompass stream weight and dependency for
-  HTTP/2, and equivalent information used to prioritize dispatch and processing of HTTP/1 fetches.
+  <p class=note>The <a for=request>internal priority</a> <a>implementation-defined</a> object
+  could encompass stream weight and dependency for HTTP/2, priorities used in
+  <a href=https://datatracker.ietf.org/doc/rfc9218/>HTTP extensible priorities</a>, and equivalent
+  information used to prioritize dispatch and processing of HTTP/1 fetches.
```
annevk, line 4151: We probably ought to say something about H/3.

Nit: I also think the original text here was correct in not repeating the field name.

fetch.bs
```diff
@@ -6892,6 +6900,7 @@ dictionary RequestInit {
   boolean keepalive;
   AbortSignal? signal;
   RequestDuplex duplex;
+  FetchPriority priority;
```
annevk, line 6903: I'd be somewhat inclined to name this `RequestPriority` to match the others. It doesn't matter much as it's not exposed, but still.

fetch.bs
```diff
@@ -1734,8 +1734,13 @@ not always relevant and might require different behavior.
 <hr>
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-priority>priority</dfn> (null or a
-user-agent-defined object). Unless otherwise stated it is null.
+<dfn export for=request id=concept-request-priority>priority</dfn>, which is
+"<code>high</code>", "<code>low</code>" or "<code>auto</code>". Unless stated otherwise it is
```
annevk, line 1738: Oxford comma please.

fetch.bs
```diff
@@ -7185,6 +7198,12 @@ constructor steps are:
    <dt><a for=request>priority</a>
    <dd><var>request</var>'s <a for=request>priority</a>.
 
+   <dt><a for=request>render-blocking</a>
+   <dd><var>request</var>'s <a for=request>render-blocking</a>.
```
annevk, line 7202: I saw your comment about adding these, but given that internal priority is copied over and therefore non-null it's not clear to me when these would actually end up doing anything. Unless we decided to expose them to service workers, but we should probably do this at that point then.

annevk (commented):
What will happen to the Priority Hints specification? Does it essentially disappear?

fetch.bs
pmeenan: The [matching PR for html](https://github.com/whatwg/html/pull/8470) has the change for intental-priority along with the plumbing for fetchPriority.  Would you prefer to have the internal-priority change separate?

fetch.bs
pmeenan: Works for me.  Removed. Implementations can still copy them over if they are needed anyway.

fetch.bs
pmeenan: Done

fetch.bs
pmeenan: Done

fetch.bs
pmeenan: HTTP extensible priorities are used by H3 but it's not strictly limited to only H3.  Tweaked the language a bit to hopefully make it clearer.

pmeenan:
As far as the existing Priority Hints spec goes, once both HTML and fetch are updated I'm planning on removing most of the content but leaving a placeholder with some high-level details and links to the relevant parts in the HTML and fetch specs.  That way any external links in blog posts and articles will still find their way to the specs with the relevant details.

fetch.bs
```diff
@@ -4134,13 +4137,17 @@ the request.
...
+  <var>request</var>'s <a for=request>priority</a>, <a for=request>initiator</a>,
+  <a for=request>destination</a>, and <a for=request>render-blocking</a> appropriately in setting
+  <var>request</var>'s <a for=request>internal priority</a> to an <a>implementation-defined</a>
+  object.
+
+  <p class=note>The <a>implementation-defined</a> object could encompass stream weight and
+  dependency for HTTP/2, priorities used in
+  <a href=https://datatracker.ietf.org/doc/rfc9218/>HTTP extensible priorities</a> for transports
+  where it applies (including HTTP/3), and equivalent information used to prioritize dispatch and
+  processing of HTTP/1 fetches.
```
annevk, line 4150:
```suggestion
  <p class=note>The <a>implementation-defined</a> object could encompass stream weight and
  dependency for HTTP/2, priorities used in <cite>Extensible Prioritization Scheme for HTTP</cite>
  for transports where it applies (including HTTP/3), and equivalent information used to prioritize 
  dispatch and processing of HTTP/1 fetches. [[!RFC9218]]
```

annevk (commented):
The Request constructor is not correct:

* It copies priority rather than internal priority.
* When priority is given this is not taken into account. I suspect that when it's given we want to set request's priority to it and we want to set internal priority to null (so request's priority will be used upon fetching to initialize it). Or would we want to adjust internal priority instead in some manner so the existing prioritization information is taken into account?

pmeenan:
Thanks, looks like I accidentally deleted `internal priority` instead of priority from the copy step.  Should be fixed now.

I also added a step 27 to the constructor to take init's `priority` into account in setting `internal priority` (I don't think we can be more specific than that for how each browser will choose to modify or override an existing `internal priority`).

fetch.bs
```diff
@@ -7346,6 +7357,10 @@ constructor steps are:
  <li><p>If <var>init</var>["{{RequestInit/signal}}"] <a for=map>exists</a>, then set
  <var>signal</var> to it.
 
+ <li><p>If <var>init</var>["{{RequestInit/priority}}"] <a for=map>exists</a>, then use
+ it appropriately in setting <var>request</var>'s <a for=request>internal priority</a>
+ to an <a>implementation-defined</a> object.
```
annevk, line 7362: But this is a subset of what the fetch algorithm takes into account. Is that what we want? An alternative might be that the fetch algorithm also has a branch when internal priority is an implementation-defined object already and then modifies it to take into account potentially changed signals.

pmeenan: I tweaked the logic to only update the existing internal priority if one was already set.  If it is null then the existing fetching logic will take all of the inputs available to construct one.  If an existing one is updated, the additional signals aren't explicitly copied so I left it vague for implementations to decide how much state they need to maintain with the implementation-defined object to be able to modify an existing internal priority with a (possibly) new priority signal.

I think this walks the line without having to require that all of the underlying signals in an existing Request be copied over while still allowing for priority to tweak it.

annevk: I think that works, but then you do need to set it on the request if internal priority is null, right?

pmeenan: Argh, yep.  Think I got it this time. The check for when `init['priority']` is set and using an existing request in the constructor now updates internal priority if it was already set, otherwise it sets the value of priority to be what was specified in init.

annevk (approved):
Thanks! This looks good to me. I have a couple final nits I can take care of when merging.

Where is the corresponding HTML PR again?

Could you post a suggested commit message in a new comment that includes all the relevant details and pointers to issues/documents/PRs? You can use https://github.com/whatwg/meta/blob/main/COMMITTING.md for guidance.

I think this can be merged Wed or Thu next week assuming it all checks out.

pmeenan:
The HTML PR is [here (#8470)](https://github.com/whatwg/html/pull/8470)

The latest commit ([afce7fc](https://github.com/whatwg/fetch/pull/1523/commits/afce7fcb0cea8f9b1b403b639a657bf3178ee372)) has a commit message with the summary details of the change and links to the Issue and previous spec.

annevk:
Tests:

* Let's have a non-throwing test as well for `Request` and `fetch()`. (You can use `about:blank` as URL to avoid it taking up more resources than needed as I guess we cannot actually test priority that well?)
* Maybe: test that `Request` doesn't have `priority`.
* Fetch tests need to be inside `fetch/`.
* `/interfaces/priority-hints.idl` needs to be removed as it'll be obsoleted by Fetch and HTML. @foolip how can we stop Priority Hints from being imported?

annevk:
The test situation still needs a bit of work, but once that is done I think this can be merged. It seems okay that HTML will be merged later as it builds on top of this.

annevk:
I decided to merge this because of:

* @foolip's comment
* triaging https://github.com/WICG/priority-hints/issues didn't yield anything new
* tests are written up and reviewed by Domenic

I hope @pmeenan can sort out the WPT IDL situation and make progress on the HTML PR next.

Thanks @pmeenan for your work on this feature!

PR #1540, opened 2022-11-18
Add WebDriver BiDi network request logging

jgraham:
  * Changes landed in WebDriver-BiDi with agreement in Browser Testing and Tools WG.
   * https://wpt.fyi/results/webdriver/tests/bidi/network?label=experimental&label=master&aligned (there's no normative changes to fetch behaviour other than when WebDriver is enabled).
   * Chromium: …
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1790366 (and similarly for other events)
  * No WebDriver BiDi documentation on MDN yet.

***
<a href="https://whatpr.org/fetch/1540.html" title="Last updated on Jul 21, 2023, 9:24 AM UTC (e5bc282)">Preview</a> | <a href="https://whatpr.org/fetch/1540/0291bf1...e5bc282.html" title="Last updated on Jul 21, 2023, 9:24 AM UTC (e5bc282)">Diff</a>


fetch.bs
```diff
@@ -110,6 +110,12 @@ urlPrefix:https://tc39.es/ecma262/#;type:dfn;spec:ecma-262
         "href": "https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https",
         "publisher": "IETF",
         "title": "Service binding and parameter specification via the DNS (DNS SVCB and HTTPS RRs)"
+    },
+    "WEBDRIVER-BIDI": {
+        "authors": [],
+        "href": "https://w3c.github.io/webdriver-bidi/",
+        "publisher": "W3C",
+        "title": "WebDriver BiDi"
```
annevk, line 118: Any reason this isn't indexed automatically? Normally this shouldn't be needed.

fetch.bs
```diff
@@ -2164,6 +2175,15 @@ otherwise, it is unset.
...
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-id>request id</dfn> which is a unique
+string automatically set when the <a for=/>request</a> is created.
+
+Note: the [=request/request id=] is used by WebDriver-BiDi. It remains constant
+across all requests resultsing from a redirect of an initial request. When a
+request is [=request/cloned=], the created request gets a unique
+[=request/request id=]. [[!WEBDRIVER-BIDI]]
```
annevk, line 2185: These two statements are contradictory. You probably have to manage this a bit more explicitly throughout the specification.

fetch.bs
```diff
@@ -2123,6 +2129,11 @@ Unless stated otherwise, it is false.
 
 <p class=note>This is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
+<p>A <a for=/>request</a> has an associated <dfn export for=request>navigation id</dfn>.
+Unless stated otherwise, it is null.
```
annevk, line 2133: You should state the type somewhere. Null or a string?

fetch.bs
```diff
@@ -5129,6 +5154,13 @@ these steps:
      <a>filtered response</a>; otherwise to <var>response</var>'s
      <a for="filtered response">internal response</a>.
 
+     <!-- the serviceworker spec is responsible for emmitting the WebDriver BiDi
+     request events in this case. That's necessary to ensure that the events are
+     only generated if the serviceworker will handle the fetch, and to get the
+     correct event ordering in the case of network fallback -->
```
annevk, line 5160: > Service Workers is responsible

and later on

> the service worker will handle

Also please align indentation with similar multiline comments (also applies elsewhere).

annevk (commented):
Thanks for tackling this. Looks reasonable overall, but I found some issues.

fetch.bs
jgraham: Observationally:
```
FATAL ERROR: Couldn't find 'WEBDRIVER-BIDI' in bibliography data. Did you mean:
  webdriver2
  webdriver
  webdriver1
  webmidi
  entries-api
```

I don't think any of those suggestions are what we want. Maybe something has to happen on the WebDriver side to get this into the biblo data, but I'm not sure what.

dontcallmedom: filed https://github.com/tobie/specref/pull/746 to add it to specref (it would have been added automatically if it had been published as TR doc)

fetch.bs
```diff
@@ -2128,6 +2128,11 @@ Unless stated otherwise, it is false.
 
 <p class=note>This is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
+<p>A <a for=/>request</a> has an associated <dfn export for=request>navigation
+id</dfn> (a string). Unless stated otherwise, it is null.
```
annevk, line 2132: "null or a string", presumably?

Also, no wrapping in inline elements.

fetch.bs
```diff
@@ -2169,6 +2174,15 @@ otherwise, it is unset.
 <a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-id>request id</dfn> which is a unique
+string automatically set when the <a for=/>request</a> is created.
```
annevk, line 2179: I'd prefer if we call this just "id" or "WebDriver id".

fetch.bs
```diff
@@ -2245,7 +2259,10 @@ is to return the result of <a>serializing a request origin</a> with <var>request
 
 <ol>
  <li><p>Let <var>newRequest</var> be a copy of <var>request</var>, except for its
- <a for=request>body</a>.
+ <a for=request>body</a> and <a for=request>request id</a>.
+
+ <li><p>Let <var>newRequest</var>'s <a for=request>request id</a> be a new
```
annevk, line 2264: Set?

fetch.bs
```diff
@@ -5143,6 +5165,13 @@ these steps:
      <a>filtered response</a>; otherwise to <var>response</var>'s
      <a for="filtered response">internal response</a>.
 
+     <!-- Service Workers is responsible for emmitting the WebDriver BiDi
+          request events in this case. That's necessary to ensure that the events are
+          only generated if the service worker will handle the fetch, and to get the
+          correct event ordering in the case of network fallback -->
```
annevk, line 5171: Slight preference for comments after the step.

fetch.bs
```diff
@@ -2169,6 +2174,15 @@ otherwise, it is unset.
 <a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-id>request id</dfn> which is a unique
+string automatically set when the <a for=/>request</a> is created.
+
+Note: the [=request/request id=] is used by WebDriver-BiDi. It remains constant
```
annevk, line 2181: `<p class=note>`

annevk (commented):
@jgraham apologies for the delay. Please resolve existing comment threads and maybe ping me if you haven't seen an update in a while.

I was wondering if there was a high-level description available somewhere so I can verify the various calls happen at the right time. Also happy to trust you, but it might help to have some additional review?

fetch.bs
jgraham: I went with just "id".

fetch.bs
```diff
@@ -2157,6 +2157,11 @@ Unless stated otherwise, it is false.
 
 <p class=note>This is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
+<p>A <a for=/>request</a> has an associated <dfn export for=request>navigation id</dfn>
+(null or a string). Unless stated otherwise, it is null.
+
+<p class=note>This is for exclusive use by HTML's navigate algorithm. [[!HTML]]
```
annevk, line 2163: What do you think about calling these fields "WebDriver navigation id" and "WebDriver id"? If we ever make them more generic we can rename them and the way WebDriver consumes them can just be as ID, but at least that would discourage others from using them for purposes they're not intended for without additional review.

fetch.bs
```diff
@@ -2274,7 +2287,9 @@ is to return the result of <a>serializing a request origin</a> with <var>request
 
 <ol>
  <li><p>Let <var>newRequest</var> be a copy of <var>request</var>, except for its
- <a for=request>body</a>.
+ <a for=request>body</a> and <a for=request>id</a>.
+
+ <li><p>Set <var>newRequest</var>'s <a for=request>id</a> to a new unique string.
```
annevk, line 2292: Is this a UUID in practice? Do we just want to settle on that?

annevk (commented):
Question about the overview still stands. Looks good otherwise.

fetch.bs
jgraham: It doesn't seem to be at the moment. I'd happily require a UUID here, but I feel like we don't quite have consensus on UUIDs everywhere yet.

fetch.bs
jgraham: I don't really have a strong opinion so I changed them.

fetch.bs
whimboo: @annevk would you mind taking another look at this PR? Thanks!

whimboo:
@annevk gently ping for review. Thanks.

annevk:
Hey @whimboo @jgraham did my question about some kind of overview get lost?

I also wonder if @noamr perhaps wants to take a look.

I have a number of nits in mind looking at the text now, but I can push a commit for that once the higher level concern is resolved.

annevk:
Thanks, that helps a lot! I left some questions in the authentication issue. I think we should probably try to solve for the model we want first. At least it doesn't make much sense to me to land something we don't feel confident about.

fetch.bs
```diff
@@ -5699,6 +5731,10 @@ run these steps:
     <p class=note>This intentionally does not depend on <var>httpRequest</var>'s
     <a for=request>credentials mode</a>.
 
+   <!-- After this point the request is not further modified before being either
+        retrieved from the cache or sent -->
+   <li><p>Run the <a>WebDriver BiDi before request sent</a> steps with <var>request</var>.
```
sadym-chromium, line 5736: @jgraham does it intentionally use `request` instead of `httpRequest`?

fetch.bs
```diff
@@ -4826,6 +4840,11 @@ steps:
      <a for=request>URL</a>, <var>fetchParams</var>'s <a for="fetch params">request</a>'s
      <a for=request>initiator type</a>, <var>global</var>, <var>cacheState</var>,
      <var>bodyInfo</var>, and <var>responseStatus</var>.
+
+     <li><p>If <var>response</var> is a <a>network error</a>, run
+     the <a>WebDriver BiDi fetch error</a> steps with <var>request</var>. Otherwise
+     run the <a>WebDriver BiDi response completed</a> steps with
```
OrKoN, line 4846: I failed to find a conclusive answer so far so posting a question here: do we expect this to be called if the client never reads the response body (e.g., `void fetch('...')`)? As far as I see, this is used as a flush algorithm for a stream pipe and it looks like it might not get called if the body is not read by the client until the browser decides to clean up the stream. cc @jgraham @sadym-chromium @juliandescottes

juliandescottes: @orkon normally the way data collection works is that the stream is cloned as soon as it is created in fetch, which I don't think depends on clients reading the stream? Typically I think the stream cloning should happen at this step https://fetch.spec.whatwg.org/#ref-for-concept-response-body%E2%91%A1%E2%91%A4:~:text=Set%20response%E2%80%99s%20body%20to%20a%20new%20body%20whose%20stream%20is%20stream%2E

OrKoN: @juliandescottes what about the situation when we do not collect any data for WebDriver BiDi but just sending events?

OrKoN: I am not 100% sure that the linked line implies that the stream will be read resulting in the `WebDriver BiDi response completed` hook being called? or did you manage to trace it? Here it says that the body might still be enqueued https://fetch.spec.whatwg.org/#ref-for-concept-response-body%E2%91%A1%E2%91%A4:~:text=be%20a%20redirect).-,Return%20response.%20Typically%20response%E2%80%99s%20body%E2%80%99s%20stream%20is%20still%20being%20enqueued%20to%20after%20returning.,-4.8.%20CORS

juliandescottes: (urgh sorry was exhausted yesterday I somehow thought the question was about data collection)

juliandescottes: I will try to move the trigger earlier, so that it doesn't depend on consuming the body (but I will have to do it on another PR, I don't have push access here)

sadym-chromium: FWIW I believe the question is critical for optimizations, where the response is read in a lazy way only on-demand.

Issue #1526, opened 2022-11-03
Add `<div algorithm>` wrappers around algorithms


jgraham:
This will enable bikeshed features including interactive highlighting of variables in the output. It will also make it easier in general to determine where an algorithm begins and ends.

annevk:
I guess to minimize disruption we shouldn't rewrap when we do this but just put them on the same indentation level as the algorithm itself.

jakearchibald:
_twitches about wrapping_

PR #1549, opened 2022-11-24
Editorial: add algorithm wrapper to header algorithms

dlrobertson:
Add `<div algorithm>` wrappers to defined header algorithms.

Related to: https://github.com/whatwg/fetch/issues/1526

***
<a href="https://whatpr.org/fetch/1549.html" title="Last updated on Nov 24, 2022, 7:20 PM UTC (eb48da0)">Preview</a> | <a href="https://whatpr.org/fetch/1549/223ca89...eb48da0.html" title="Last updated on Nov 24, 2022, 7:20 PM UTC (eb48da0)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -548,6 +548,7 @@ for consistency.
 <p class="note no-backref">A <a for=/>header list</a> is essentially a
 specialized multimap: an ordered list of key-value pairs with potentially duplicate keys.
 
+<div algorithm>
 <p>To
 <dfn export for="header list" id=concept-header-list-get-structured-header>get a structured field value</dfn>
 given a <a for=/>header name</a> <var>name</var> and a string <var>type</var> from a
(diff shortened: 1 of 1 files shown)


annevk:
When I look at Preview I don't see the variable highlighting.

https://tabatkins.github.io/bikeshed/#algorithms suggests the `algorithm` attribute needs to have a value, but I'm not sure if that's the complete story.

It doesn't look like we explicitly disable this in https://github.com/tabatkins/bikeshed-boilerplate/blob/main/boilerplate/whatwg/defaults.include so maybe you have to repeat the algorithm name and it doesn't want to look at the first `<dfn>`?

cc @jgraham @tabatkins

dlrobertson:
> When I look at Preview I don't see the variable highlighting.

Did you click the variables for an algorithm in [HTTP Headers (2.2.2)](https://whatpr.org/fetch/1549.html#terminology-headers)? E.g. when clicking the variables for `get a structured field value`, I see the following.

Screenshot 2022-11-25 at 10 26 04 AM

I probably should have specified that I only added the tag for the HTTP Headers section in this commit.

> https://tabatkins.github.io/bikeshed/#algorithms suggests the `algorithm` attribute needs to have a value, but I'm not sure if that's the complete story.

This may be the preferred case, but [css-scroll-anchoring](https://github.com/w3c/csswg-drafts/blob/f2e3f006d09bc6ce5b2d03be0abef47c98bd1408/css-scroll-anchoring-1/Overview.bs#L135-L149) uses `algorithm` like we do here. We can definitely add descriptions if this is preferred.

annevk:
Ah, it's not working in WebKit for some reason. I'll investigate and merge this early next week. Could you file an issue against Bikeshed for updating its documentation?

PR #1559, opened 2022-12-04
Editorial: <div algorithm> wrappers for more definitions.

dlrobertson:
Related to: #1526


dlrobertson:
Just trying to add these tags as I do some reading (because I find them to be helpful)... If this process is adding too much noise or if there is something more helpful I can be doing while reading, please let me know 😄

fetch.bs
```diff
@@ -2575,6 +2580,7 @@ or a <a for=/>URL</a>.
 
  <li><p>Return <var>location</var>.
 </ol>
+</div>
 
 <p class=note>The <a for=response>location URL</a> algorithm is exclusively used for redirect
```
annevk, line 2585: I think it makes semantic sense for this note to be part of the algorithm.

annevk (commented):
No suggestions from me, I think this works.

I do have a small suggestion for this PR.

fetch.bs
dlrobertson: Yeah, that makes sense.

annevk (approved):

PR #1564, opened 2022-12-10
Add HTTP OWS handling to single range header parsing

dlrobertson:
Add HTTP optional white space handling to the single range header parser.

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/34384
   * https://github.com/web-platform-tests/wpt/pull/37569
   * Chromium: …
   * Gecko: …
   * WebKit: …

I didn't see a concrete consensus made about the CORS case in https://github.com/whatwg/fetch/issues/1070, so I went with the more conservative option here.

***
<a href="https://whatpr.org/fetch/1564.html" title="Last updated on Feb 16, 2023, 1:41 AM UTC (1be32b3)">Preview</a> | <a href="https://whatpr.org/fetch/1564/093c622...1be32b3.html" title="Last updated on Feb 16, 2023, 1:41 AM UTC (1be32b3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1020,7 +1020,7 @@ fetch("https://victim.example/naïve-endpoint", {
    <dd>
     <ol>
      <li><p>Let <var>rangeValue</var> be the result of <a>parsing a single range header value</a>
-     given <var>value</var>.
+     given <var>value</var> and false.
 
      <li><p>If <var>rangeValue</var> is failure, then return false.
 
@@ -1292,16 +1292,27 @@ run these steps:
 
 <div algorithm>
 <p>To <dfn id=simple-range-header-value>parse a single range header value</dfn> from a
-<a>byte sequence</a> <var>value</var>, run these steps:
+<a>byte sequence</a> <var>value</var> and a boolean <var>allowWhitespace</var>, run these steps:
 
 <ol>
  <li><p>Let <var>data</var> be the <a>isomorphic decoding</a> of <var>value</var>.
 
- <li><p>If <var>data</var> does not <a for=string>start with</a> "<code>bytes=</code>", then return
+ <li><p>If <var>data</var> does not <a for=string>start with</a> "<code>bytes</code>", then return
  failure.
 
  <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially
- pointing at the 6th <a>code point</a> of <var>data</var>.
+ pointing at the 5th <a>code point</a> of <var>data</var>.
+
+ <li><p>If <var>allowWhitespace</var> is true, <a>collect a sequence of code points</a> that are
+ <a>HTTP tab or space</a>, from <var>data</var> given <var>position</var>.
+
+ <li><p>If the <a>code point</a> at <var>position</var> within <var>data</var> is not U+003D (=),
+ then return failure.
+
+ <li><p>Advance <var>position</var> by 1.
+
+ <li><p>If <var>allowWhitespace</var> is true, <a>collect a sequence of code points</a> that are
+ <a>HTTP tab or space</a>, from <var>data</var> given <var>position</var>.
 
  <li><p>Let <var>rangeStart</var> be the result of <a>collecting a sequence of code points</a> that
  are <a>ASCII digits</a>, from <var>data</var> given <var>position</var>.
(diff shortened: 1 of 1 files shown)


annevk (commented):
@jakearchibald @youennf @rayankans does any of you think this boolean is worth having for CORS? I think I'd rather allow whitespace in CORS for the `Range` header so we can keep `Range` header parsing consistent. In particular this would allow an arbitrary number of 0x09 and 0x20 after `bytes=`. I don't see the security benefit.

(We also use this parser for `blob:` URL requests and there we already allow whitespace.)

dlrobertson:
> @jakearchibald @youennf @rayankans does any of you think this boolean is worth having for CORS? I think I'd rather allow whitespace in CORS for the `Range` header so we can keep `Range` header parsing consistent. In particular this would allow an arbitrary number of 0x09 and 0x20 after `bytes=`. I don't see the security benefit.
> 
> (We also use this parser for `blob:` URL requests and there we already allow whitespace.)

I've tested Blink and WebKit and both disallow the whitespace allowed in HTTP-RANGE as tested [here](https://github.com/web-platform-tests/wpt/pull/37569). Would it make sense to remove the boolean, but add a note that some implementations don't allow this optional whitespace?

fetch.bs
```diff
@@ -1243,7 +1243,7 @@ run these steps:
 
 <div algorithm>
 <p>To <dfn id=simple-range-header-value>parse a single range header value</dfn> from a
-<a>byte sequence</a> <var>value</var>, run these steps:
+<a>byte sequence</a> <var>value</var> and a boolean <var>allowWhiteSpace</var>, run these steps:
```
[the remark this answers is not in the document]
dlrobertson: Would this still be the case following the discussion in https://github.com/whatwg/fetch/issues/1070

annevk:
It seems that as per more recent discussion in #1070 I misunderstood the ABNF and no whitespace is allowed after `bytes=` in theory. Let's get to the bottom of that and then revisit this as it might well be that the status quo is correct now.

dlrobertson:
> If we keep this we should name this disallowWhitespace so it can default to false. Although as long as we don't export it we should probably keep it as a required argument.

Updated to use `disallowWhitespace`. Sorry for the delay, I thought I had already done this.

annevk (commented):
This only removes whitespace after `=` if I understand it correctly. Whereas https://github.com/whatwg/fetch/issues/1070#issuecomment-1381651996 argues for removing it in 6 places.

dlrobertson:
HTTP 2616 allowed whitespace:

|whitespace| supported by parse-single-range|
|------------|----|
| Before `=` |  ✅ |
| After `=` |  ✅ |
| Before `-` | ✅ |
| After `-` | ✅ |
| Before `,` | ❌ |
| After `,` | ❌ |

Added some extra cases to https://github.com/web-platform-tests/wpt/pull/37569

fetch.bs
```diff
@@ -1292,28 +1292,45 @@ run these steps:
 
 <div algorithm>
 <p>To <dfn id=simple-range-header-value>parse a single range header value</dfn> from a
-<a>byte sequence</a> <var>value</var>, run these steps:
+<a>byte sequence</a> <var>value</var> and a boolean <var>disallowWhitespace</var>, run these steps:
```
annevk, line 1295: Since it's a required argument, perhaps we should invert this to not have a double negative.

annevk (commented):
Thanks, that makes sense as we don't support multiple values here. One nit and with that this is good to go.

fetch.bs
dlrobertson: 👍 yeah that makes sense

annevk (approved):
This looks great, thanks!

Blocked on https://github.com/speced/bikeshed/issues/2476 unfortunately.

dlrobertson:
> Blocked on [speced/bikeshed#2476](https://github.com/speced/bikeshed/issues/2476) unfortunately.

No problem! Thanks for the review... Will I need to rebase once the fix for [speced/bikeshed#2476](https://github.com/speced/bikeshed/issues/2476) lands?

annevk:
No, this is all good. I also looked at #1520 again and apart from invoking this algorithm with true I think that's also good (and I can make that change when merging if it all looks good to you).

PR #1569, opened 2022-12-13
Editorial: Amend body reading to use no-promise version

### LINKED ISSUE #1568: Remove promise dependency from "Fully read a body"
[Fully read a body](https://fetch.spec.whatwg.org/#body-fully-read) is used in contexts that are not necessarily JS related, e.g. loading `<link>`s.
The implementation should go via the streams API without requiring promises internally.
Promises should only be used in JS-related contexts.

noamr:
Following https://github.com/whatwg/streams/pull/1250
Closes #1568

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1569.html" title="Last updated on Dec 22, 2022, 2:06 PM UTC (c6dc3a2)">Preview</a> | <a href="https://whatpr.org/fetch/1569/84ea7db...c6dc3a2.html" title="Last updated on Dec 22, 2022, 2:06 PM UTC (c6dc3a2)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6912,6 +6912,8 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step 
 <a for=/>byte sequence</a> <var>bytes</var>: Return a new {{ArrayBuffer}} whose contents are
 <var>bytes</var>.
 
+<p class="note">The above method may <a>reject</a> with a {{RangeError}}.
+
 <p>The <dfn method for=Body><code>blob()</code></dfn> method steps are to return the result
 of running <a for=Body>consume body</a> with <a>this</a> and the following step given a
 <a for=/>byte sequence</a> <var>bytes</var>: Return a {{Blob}} whose contents are <var>bytes</var>
@@ -6979,6 +6981,8 @@ running <a for=Body>consume body</a> with <a>this</a> and the following step giv
 
 <p>The <dfn method for=Body><code>json()</code></dfn> method steps are to return the result
 of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from bytes</a>.
+
+<p class="note">The above method may <a>reject</a> with a {{SyntaxError}}.
 
 <p>The <dfn method for=Body><code>text()</code></dfn> method steps are to return the result
 of running <a for=Body>consume body</a> with <a>this</a> and <a>UTF-8 decode</a>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -1465,9 +1465,6 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
  <li><p>If <var>taskDestination</var> is null, then set <var>taskDestination</var> to the result of
  <a>starting a new parallel queue</a>.
 
- <li><p>Let <var>promise</var> be the result of <a>fully reading body as promise</a> given
- <var>body</var>.
-
  <li><p>Let <var>fulfilledSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
  <a>queue a fetch task</a> to run <var>processBody</var> given <var>bytes</var>, with
  <var>taskDestination</var>.
@@ -1475,22 +1472,12 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
  <li><p>Let <var>rejectedSteps</var> be to <a>queue a fetch task</a> to run
  <var>processBodyError</var>, with <var>taskDestination</var>.
 
- <li><p><a for=promise>React</a> to <var>promise</var> with <var>fulfilledSteps</var> and
- <var>rejectedSteps</var>.
-</ol>
-</div>
-
-<div algorithm>
-<p>To <dfn export lt="fully reading body as promise">fully read body as promise</dfn>, given a
-<a for=/>body</a> <var>body</var>, run these steps:
-
-<ol>
  <li><p>Let <var>reader</var> be the result of <a for=ReadableStream>getting a reader</a> for
- <var>body</var>'s <a for=body>stream</a>. If that threw an exception, then return
- <a>a promise rejected with</a> that exception.
+ <var>body</var>'s <a for=body>stream</a>. If that threw an exception, then run
+ <var>rejectedSteps</var> with that exception.
 
- <li><p>Return the result of <a for=ReadableStreamDefaultReader>reading all bytes</a> from
- <var>reader</var>.
+ <li><p><a for=ReadableStreamDefaultReader>Read all bytes</a> from
+ <var>reader</var>, given <var>fulfilledSteps</var> and <var>rejectedSteps</var>.
 </ol>
 </div>
 
@@ -6981,8 +6968,25 @@ the associated steps:
  <li><p>Let <var>promise</var> be <a>a promise resolved with</a> an empty
  <a for=/>byte sequence</a>.
 
- <li><p>If <var>object</var>'s <a for=Body>body</a> is non-null, then set <var>promise</var> to the
- result of <a>fully reading body as promise</a> given <var>object</var>'s <a for=Body>body</a>.
+ <li>
+  <p>If <var>object</var>'s <a for=Body>body</a> is non-null, then:
+
+  <ol>
+   <li><p>Set <var>promise</var> to <a>a new promise</a>.
+
+   <li><p>Let <var>fulfilledSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
+   <a for=/>resolve</a> <var>promise</var> given <var>bytes</var>.
+
+   <li><p>Let <var>rejectedSteps</var> given <var>error</var> be to <a>reject</a> <var>promise</var>
+   given <var>error</var>.
+
+   <li><p>Let <var>reader</var> be the result of <a for=ReadableStream>getting a reader</a> for
+   <var>body</var>'s <a for=body>stream</a>. If that threw an exception, then run
+   <var>rejectedSteps</var> with that exception.
+
+   <li><p><a for=ReadableStreamDefaultReader>Read all bytes</a> from <var>reader</var>, given
+   <var>fulfilledSteps</var> and <var>rejectedSteps</var>.
+  </ol>
 
  <li><p>Let <var>steps</var> be to return the result of <a>package data</a> with the first argument
  given, <var>type</var>, and <var>object</var>'s <a for=Body>MIME type</a>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -6916,7 +6916,7 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step 
 
 <p>The <dfn method for=Body><code>blob()</code></dfn> method steps are to return the result
 of running <a for=Body>consume body</a> with <a>this</a> and the following step given a
-<a for=/>byte sequence</a> <var>bytes</var>: Return a {{Blob}} whose contents are <var>bytes</var>
+<a for=/>byte sequence</a> <var>bytes</var>: return a {{Blob}} whose contents are <var>bytes</var>
 and whose {{Blob/type}} attribute is <a>this</a>'s <a for=Body>MIME type</a>.
 
 <p>The <dfn method for=Body><code>formData()</code></dfn> method steps are to return the result of
--- a/fetch.bs
+++ b/fetch.bs
@@ -6879,85 +6879,6 @@ returns failure or a <a for=/>MIME type</a>.
 <a>this</a>'s <a for=Body>body</a> is non-null and <a>this</a>'s <a for=Body>body</a>'s
 <a for=body>stream</a> is <a for=ReadableStream>disturbed</a>; otherwise false.
 
-<p>The <dfn id=concept-body-package-data for=Body>package data</dfn> algorithm, given
-<var>bytes</var>, <var>type</var>, and a <var>mimeType</var>, switches on <var>type</var>, and runs
-the associated steps:
-
-<dl class=switch>
- <dt><i>ArrayBuffer</i>
- <dd>
-  <p>Return a new {{ArrayBuffer}} whose contents are <var>bytes</var>.
-
-  <p class=note>Allocating an {{ArrayBuffer}} can throw a {{RangeError}}.
-
- <dt><i>Blob</i>
- <dd><p>Return a {{Blob}} whose contents are <var>bytes</var> and {{Blob/type}} attribute is
- <var>mimeType</var>.
-
- <dt><i>FormData</i>
- <dd>
-  <p>If <var>mimeType</var>'s <a for="MIME type">essence</a> is "<code>multipart/form-data</code>",
-  then:
-
-  <ol>
-   <li>
-    <p>Parse <var>bytes</var>, using the value of the `<code>boundary</code>` parameter from
-    <var>mimeType</var>, per the rules set forth in
-    <cite>Returning Values from Forms: multipart/form-data</cite>. [[!RFC7578]]</p>
-
-    <p>Each part whose `<code>Content-Disposition</code>` header contains a `<code>filename</code>`
-    parameter must be parsed into an <a for="entry list">entry</a> whose value is a {{File}} object
-    whose contents are the contents of the part. The {{File/name}} attribute of the {{File}} object
-    must have the value of the `<code>filename</code>` parameter of the part. The {{Blob/type}}
-    attribute of the {{File}} object must have the value of the `<code>Content-Type</code>` header
-    of the part if the part has such header, and `<code>text/plain</code>` (the default defined by
-    [[!RFC7578]] section 4.4) otherwise.</p>
-
-    <p>Each part whose `<code>Content-Disposition</code>` header does not contain a
-    `<code>filename</code>` parameter must be parsed into an <a for="entry list">entry</a> whose
-    value is the <a lt="UTF-8 decode without BOM">UTF-8 decoded without BOM</a> content of the part.
-    <span class=note>This is done regardless of the presence or the value of a
-    `<code>Content-Type</code>` header and regardless of the presence or the value of a
-    `<code>charset</code>` parameter.</span></p>
-
-    <p class=note>A part whose `<code>Content-Disposition</code>` header contains a
-    `<code>name</code>` parameter whose value is `<code>_charset_</code>` is parsed like any other
-    part. It does not change the encoding.</p>
-   </li>
-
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1465,32 +1465,19 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
  <li><p>If <var>taskDestination</var> is null, then set <var>taskDestination</var> to the result of
  <a>starting a new parallel queue</a>.
 
- <li><p>Let <var>promise</var> be the result of <a>fully reading body as promise</a> given
- <var>body</var>.
-
  <li><p>Let <var>fulfilledSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
```
domenic, line 1468: Should be renamed to successSteps/errorSteps since there is no more promise fulfillment/rejection going on.

fetch.bs
```diff
@@ -6958,8 +6945,25 @@ the associated steps:
...
 
- <li><p>If <var>object</var>'s <a for=Body>body</a> is non-null, then set <var>promise</var> to the
- result of <a>fully reading body as promise</a> given <var>object</var>'s <a for=Body>body</a>.
+ <li>
+  <p>If <var>object</var>'s <a for=Body>body</a> is non-null, then:
+
+  <ol>
+   <li><p>Set <var>promise</var> to <a>a new promise</a>.
+
+   <li><p>Let <var>fulfilledSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
```
domenic, line 6954: Same here. (Although I guess resolveSteps/rejectSteps would be OK too.)

fetch.bs

fetch.bs

fetch.bs
```diff
@@ -6958,8 +6945,25 @@ the associated steps:
...
 
- <li><p>If <var>object</var>'s <a for=Body>body</a> is non-null, then set <var>promise</var> to the
- result of <a>fully reading body as promise</a> given <var>object</var>'s <a for=Body>body</a>.
+ <li>
+  <p>If <var>object</var>'s <a for=Body>body</a> is non-null, then:
+
+  <ol>
+   <li><p>Set <var>promise</var> to <a>a new promise</a>.
+
+   <li><p>Let <var>successStep</var> given a <a>byte sequence</a> <var>bytes</var> be to
```
domenic, line 6954:
```suggestion
   <li><p>Let <var>successSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
```

fetch.bs
```diff
@@ -6958,8 +6945,25 @@ the associated steps:
...
+
+   <li><p>Let <var>errorSteps</var> given <var>error</var> be to <a>reject</a> <var>promise</var>
+   given <var>error</var>.
+
+   <li><p>Let <var>reader</var> be the result of <a for=ReadableStream>getting a reader</a> for
+   <var>body</var>'s <a for=body>stream</a>. If that threw an exception, then run
+   <var>errorSteps</var> with that exception.
+
+   <li><p><a for=ReadableStreamDefaultReader>Read all bytes</a> from <var>reader</var>, given
+   <var>successStep</var> and <var>errorSteps</var>.
```
domenic, line 6965:
```suggestion
   <var>successSteps</var> and <var>errorSteps</var>.
```

domenic (approved):
LGTM with nits

noamr:
@annevk @domenic can we merge this please? :)

fetch.bs
```diff
@@ -1465,32 +1465,19 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
...
-<div algorithm>
-<p>To <dfn export lt="fully reading body as promise">fully read body as promise</dfn>, given a
-<a for=/>body</a> <var>body</var>, run these steps:
-
-<ol>
  <li><p>Let <var>reader</var> be the result of <a for=ReadableStream>getting a reader</a> for
- <var>body</var>'s <a for=body>stream</a>. If that threw an exception, then return
- <a>a promise rejected with</a> that exception.
+ <var>body</var>'s <a for=body>stream</a>. If that threw an exception, then run
+ <var>errorSteps</var> with that exception.
```
annevk, line 1477:
```suggestion
 <var>errorSteps</var> with that exception and return.
```
Right?

fetch.bs
```diff
@@ -1465,32 +1465,19 @@ null), run these steps. <var>processBody</var> must be an algorithm accepting a
...
+ <li><p>Let <var>errorSteps</var> be to <a>queue a fetch task</a> to run
  <var>processBodyError</var>, with <var>taskDestination</var>.
 
- <li><p><a for=promise>React</a> to <var>promise</var> with <var>fulfilledSteps</var> and
- <var>rejectedSteps</var>.
-</ol>
-</div>
-
-<div algorithm>
-<p>To <dfn export lt="fully reading body as promise">fully read body as promise</dfn>, given a
```
annevk, line 1484: It seems nobody used this so this is safe: https://dontcallmedom.github.io/webdex/f.html#fully%20reading%20body%20as%20promise%40%40fetch%25%25dfn.

annevk (commented):
Two thoughts:

1. I don't understand how "read all bytes" works as it doesn't seem to yield to the event loop.
2. If it somehow works I think we can rewrite "consume body" in a better way whereby it only uses a promise for the return value and not any internal logic.

noamr:
> Two thoughts:
> 
> 1. I don't understand how "read all bytes" works as it doesn't seem to yield to the event loop.

This is true in general for streams I believe. Not sure I have a good answer. I think this problem exists in the current wording and in all the callers of streams if I understand this correctly.

> 2. If it somehow works I think we can rewrite "consume body" in a better way whereby it only uses a promise for the return value and not any internal logic.

Gotcha, will amend.

noamr:
> > Two thoughts:
> > 
> > 1. I don't understand how "read all bytes" works as it doesn't seem to yield to the event loop.
> 
> This is true in general for streams I believe. Not sure I have a good answer. I think this problem exists in the current wording and in all the callers of streams if I understand this correctly.
> 
> > 2. If it somehow works I think we can rewrite "consume body" in a better way whereby it only uses a promise for the return value and not any internal logic.
> 
> Gotcha, will amend.

Done, I think it looks a lot clearer

fetch.bs
```diff
@@ -6978,16 +6965,30 @@ the associated steps:
...
+
+ <li><p>If <var>object</var>'s <a for=Body>body</a> is null, then return
+ the result of executing <var>packageSteps</var> given an empty <a for=/>byte sequence</a>.
+
+ <li><p>Set <var>promise</var> to <a>a new promise</a>.
 
- <li><p>If <var>object</var>'s <a for=Body>body</a> is non-null, then set <var>promise</var> to the
- result of <a>fully reading body as promise</a> given <var>object</var>'s <a for=Body>body</a>.
+ <li><p>Let <var>successSteps</var> given a <a>byte sequence</a> <var>bytes</var> be to
+ <a for=/>resolve</a> <var>promise</var> given the result of calling <var>packageSteps</var>
```
annevk, line 6978:
```suggestion
 <a for=/>resolve</a> <var>promise</var> given the result of running <var>packageSteps</var>
```

fetch.bs
```diff
@@ -6978,16 +6965,30 @@ the associated steps:
...
  a {{TypeError}}.
 
- <li><p>Let <var>promise</var> be <a>a promise resolved with</a> an empty
- <a for=/>byte sequence</a>.
+ <li><p>Let <var>packageSteps</var> given a <a for=/>byte sequence</a> <var>bytes</var> be to
+ return the result of <a lt="package data">packaging</a> <var>bytes</var> given <var>type</var> and
+ <var>object</var>'s <a for=Body>MIME type</a>.
+
+ <li><p>If <var>object</var>'s <a for=Body>body</a> is null, then return
+ the result of executing <var>packageSteps</var> given an empty <a for=/>byte sequence</a>.
```
annevk, line 6973:
```suggestion
 <li><p>If <var>object</var>'s <a for=Body>body</a> is null, then return
 <a>a promise resolved with</a> the result of running <var>packageSteps</var> given an empty
 <a for=/>byte sequence</a>.
```

annevk (commented):
Thanks! Looks good modulo nits.

annevk (commented):
Sigh, sorry, the rewrite of consume body doesn't work because package data can throw.

_packageSteps_ would have to take a promise as argument that is either resolved or rejected depending on the result of running package data. It ends up being even longer than the original version, but probably still better as at least we're not resolving promises with Infra types?

MattiasBuelens (approved):
LGTM.

> Two thoughts:
> 
> 1. I don't understand how "read all bytes" works as it doesn't seem to yield to the event loop.

If the stream has unread chunks, the read loop of "read all bytes" will immediately process all of those without yielding. Afterwards, the read request from the read loop will sit in the reader's queue until *something else* calls ["enqueue"](https://streams.spec.whatwg.org/#readablestream-enqueue). This will synchronously call the "chunk steps" of that read request (through [step 3 in ReadableStreamDefaultControllerEnqueue](https://streams.spec.whatwg.org/commit-snapshots/f894acdd417926a2121710803cef593e15127964/#readable-stream-default-controller-enqueue)) and advance the read loop.

So indeed, we don't yield to any particular event loop, instead we assume that there's some other event loop out there that's calling "enqueue", and we run the "chunk steps" as part of that same call.

noamr:
> LGTM.
> 
> > Two thoughts:
> > 
> > 1. I don't understand how "read all bytes" works as it doesn't seem to yield to the event loop.
> 
> If the stream has unread chunks, the read loop of "read all bytes" will immediately process all of those without yielding. Afterwards, the read request from the read loop will sit in the reader's queue until _something else_ calls ["enqueue"](https://streams.spec.whatwg.org/#readablestream-enqueue). This will synchronously call the "chunk steps" of that read request (through [step 3 in ReadableStreamDefaultControllerEnqueue](https://streams.spec.whatwg.org/commit-snapshots/f894acdd417926a2121710803cef593e15127964/#readable-stream-default-controller-enqueue)) and advance the read loop.
> 
> So indeed, we don't yield to any particular event loop, instead we assume that there's some other event loop out there that's calling "enqueue", and we run the "chunk steps" as part of that same call.

Right, which means we might resolve a promise here from a different event loop. I think I can fix it in this PR.

fetch.bs
```diff
@@ -6892,102 +6879,122 @@ returns failure or a <a for=/>MIME type</a>.
...
+   written. Volunteers welcome.
+
+   <p>Otherwise, if <var>mimeType</var>'s <a for="MIME type">essence</a> is
+   "<code>application/x-www-form-urlencoded</code>", then:
+
+   <ol>
+    <li><p>Let <var>entries</var> be the result of
+    <a lt="urlencoded parser">parsing</a> <var>bytes</var>.
+
+    <li><p>If <var>entries</var> is failure, then <a>throw</a> a {{TypeError}}.
```
MattiasBuelens, line 6970:
```suggestion
    <li><p>If <var>entries</var> is failure, then run <var>rejectStep</var> with a {{TypeError}}.
```

fetch.bs
```diff
@@ -6892,102 +6879,122 @@ returns failure or a <a for=/>MIME type</a>.
...
+
+    <li><p>Run <var>resolveStep</var> with a new {{FormData}} object whose
+    <a for=FormData>entry list</a> is <var>entries</var>.
+   </ol>
+
+   <p>Otherwise, run <var>rejectStep</var> with a {{TypeError}}.
+
+  <dt><i>JSON</i>
+  <dd><p>Run <var>resolveStep</var> with the result of running <a>parse JSON from bytes</a> on
+  <var>bytes</var>.
```
MattiasBuelens, line 6980: This calls into ["parse a JSON string to a JavaScript value"](https://infra.spec.whatwg.org/#parse-a-json-string-to-a-javascript-value), which can throw if the string is not valid JSON. In that case, we should call `rejectStep`.

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
...
+ <a for=/>resolve</a> <var>promise</var> with the result of running <var>convertBytesToJSValue</var>
+ with <var>data</var>. If that threw an exception, <a>reject</a> <var>promise</var> with that
+ exception.
 
- <dt><i>FormData</i>
- <dd>
-  <p>If <var>mimeType</var>'s <a for="MIME type">essence</a> is "<code>multipart/form-data</code>",
-  then:
+ <li><p>If <var>object</var>'s <a for=Body>body</a> is null, then run <var>successSteps</var>
+ with an empty <a for=/>byte sequence</a>, and return.
```
annevk, line 6898: You need to return _promise_ here.

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
...
+
+ <li>Let <var>errorSteps</var> given <var>error</var> be to <a>reject</a> <var>promise</var> with
+ <var>error</var>.
```
annevk, line 6901: If you define these before _successSteps_ you can invoke them as part of defining _successSteps_.

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
...
+ <li><p><a for=body>Fully read</a> <var>object</var>'s <a for=Body>body</a> given
+ <var>successSteps</var>, <var>errorSteps</var>, and <var>object</var>'s
+ <a>relevant settings object</a>'s <a for="environment settings object">global object</a>.
+
+ <li><p>Return <var>promise</var>.
+</ol>
+
+<p>The <dfn method for=Body><code>arrayBuffer()</code></dfn> method steps are to return the result
+of running <a for=Body>consume body</a> with <a>this</a> and the following step given a
+<a for=/>byte sequence</a> <var>bytes</var>: Return a new {{ArrayBuffer}} whose contents are
```
annevk, line 6912:
```suggestion
<a for=/>byte sequence</a> <var>bytes</var>: return a new {{ArrayBuffer}} whose contents are
```

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
...
+<var>bytes</var>.
+
+<p class="note">The above method may <a>reject</a> with a {{RangeError}}.
 
+<p>The <dfn method for=Body><code>blob()</code></dfn> method steps are to return the result
+of running <a for=Body>consume body</a> with <a>this</a> and the following step given a
+<a for=/>byte sequence</a> <var>bytes</var>: Return a {{Blob}} whose contents are <var>bytes</var>
```
annevk, line 6919:
```suggestion
<a for=/>byte sequence</a> <var>bytes</var>: return a {{Blob}} whose contents are <var>bytes</var>
```

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
...
+and whose {{Blob/type}} attribute is <a>this</a>'s <a for=Body>MIME type</a>.
+
+<p>The <dfn method for=Body><code>formData()</code></dfn> method steps are to return the result of
+running <a for=Body>consume body</a> with <a>this</a> and the following step given a
+<a for=/>byte sequence</a> <var>bytes</var>: switch on <a>this</a>'s <a for=Body>MIME type</a>'s
+<a for="MIME type">essence</a>:
```
annevk, line 6925: and run the corresponding steps*

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
...
+
+ <li><p>Return <var>promise</var>.
+</ol>
+
+<p>The <dfn method for=Body><code>arrayBuffer()</code></dfn> method steps are to return the result
+of running <a for=Body>consume body</a> with <a>this</a> and the following step given a
+<a for=/>byte sequence</a> <var>bytes</var>: Return a new {{ArrayBuffer}} whose contents are
+<var>bytes</var>.
+
+<p class="note">The above method may <a>reject</a> with a {{RangeError}}.
```
annevk, line 6915:
```suggestion
<p class="note">The above method can reject with a {{RangeError}}.
```

fetch.bs
```diff
@@ -6948,62 +6963,29 @@ the associated steps:
...
-running <a for=Body>consume body</a> with <a>this</a> and <i>Blob</i>.
-
-<p>The <dfn method for=Body><code>formData()</code></dfn> method steps are to return the result of
-running <a for=Body>consume body</a> with <a>this</a> and <i>FormData</i>.
+<p>The <dfn method for=Body><code>json()</code></dfn> method steps are to return the result
+of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from bytes</a>.
 
-<p>The <dfn method for=Body><code>json()</code></dfn> method steps are to return the result of
-running <a for=Body>consume body</a> with <a>this</a> and <i>JSON</i>.
+<p class="note">The above method may <a>reject</a> with a {{SyntaxError}}.
```
annevk, line 6985:
```suggestion
<p class="note">The above method can reject with a {{SyntaxError}}.
```

fetch.bs
```diff
@@ -6892,30 +6879,58 @@ returns failure or a <a for=/>MIME type</a>.
 <a>this</a>'s <a for=Body>body</a> is non-null and <a>this</a>'s <a for=Body>body</a>'s
 <a for=body>stream</a> is <a for=ReadableStream>disturbed</a>; otherwise false.
 
-<p>The <dfn id=concept-body-package-data for=Body>package data</dfn> algorithm, given
-<var>bytes</var>, <var>type</var>, and a <var>mimeType</var>, switches on <var>type</var>, and runs
-the associated steps:
+<p>The <dfn id=concept-body-consume-body for=Body>consume body</dfn> algorithm, given a {{Body}}
+<var>object</var> and <var>convertBytesToJSValue</var>, which is an algorithm that accepts a
+<a for=/>byte sequence</a> and returns a JavaScript value or throws an exception, runs these steps:
```
annevk, line 6884:
```suggestion
<p id=concept-body-package-data>The <dfn id=concept-body-consume-body for=Body>consume body</dfn>
algorithm, given an object that includes {{Body}} <var>object</var> and an algorithm that takes a
<a for=/>byte sequence</a> and returns a JavaScript value or throws an exception
<var>convertBytesToJSValue</var>, runs these steps:
```

annevk (commented):
I think this works, nice! Couple nits left.

fetch.bs

fetch.bs

fetch.bs

Issue #1580, opened 2022-12-27
Fix bikeshed link errors for implementation-defined and url


dlrobertson:
## Summary

There appear to be some link errors that bikeshed throws some warnings for the following, due to multiple definitions:

 - [x] `url` definitions ([concept-url from url](https://url.spec.whatwg.org/#concept-url) and [navigation-status-url from webdriver](https://w3c.github.io/webdriver-bidi/#navigation-status-url))
 - [ ] `implementation defined` definitions ([implementation-defined from infra](https://infra.spec.whatwg.org/#implementation-defined) and [implementation-defined from ecmascript](https://tc39.es/ecma262/#implementation-defined))

## Example

```
 make local
bikeshed spec fetch.bs fetch.html --md-Text-Macro="COMMIT-SHA LOCAL COPY"
LINK ERROR: Multiple possible 'url' dfn refs for '['/']'.
Arbitrarily chose https://url.spec.whatwg.org/#concept-url
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:url; type:dfn; text:url
spec:webdriver-bidi; type:dfn; text:url
<a data-link-for="/" data-link-type="dfn" data-lt="URL">URL</a>
LINK ERROR: Multiple possible 'implementation-defined' dfn refs.
Arbitrarily chose https://infra.spec.whatwg.org/#implementation-defined
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:infra; type:dfn; text:implementation-defined
spec:ecmascript; type:dfn; text:implementation-defined
<a data-link-type="dfn" data-lt="implementation-defined">implementation-defined</a>
LINK ERROR: Multiple possible 'urls' dfn refs for '['/']'.
Arbitrarily chose https://url.spec.whatwg.org/#concept-url
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:url; type:dfn; text:url
spec:webdriver-bidi; type:dfn; text:url
<a data-link-for="/" data-link-type="dfn" data-lt="URLs">URLs</a>
 ✔  Successfully generated, with 3 linking errors
```

whimboo:
The underlying issue with the WebDriver BiDi spec should be fixed now (https://github.com/w3c/webdriver-bidi/issues/347)

dlrobertson:
[...]

PR #1581, opened 2022-12-28
Editorial: Fix implementation-defined links

dlrobertson:
Related to: #1580

***
<a href="https://whatpr.org/fetch/1581.html" title="Last updated on Jan 2, 2023, 6:54 PM UTC (88c8ca2)">Preview</a> | <a href="https://whatpr.org/fetch/1581/464326e...88c8ca2.html" title="Last updated on Jan 2, 2023, 6:54 PM UTC (88c8ca2)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -113,6 +113,10 @@ urlPrefix:https://tc39.es/ecma262/#;type:dfn;spec:ecma-262
 }
 </pre>
 
+<pre class=link-defaults>
+spec:infra; type:dfn; text:implementation-defined
+</pre>
+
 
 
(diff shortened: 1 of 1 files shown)


domenic:
Why not use link-defaults like the error message suggests?

dlrobertson:
> Why not use link-defaults like the error message suggests?

I didn't see another one in this spec, but that probably makes sense

dlrobertson:
Updated to use a `link-defaults` block

dlrobertson:
Updated to only specify the default for `implementation-defined` after the webdriver spec was fixed.

PR #1584, opened 2023-01-01
Use internal response for Server-Timing

### LINKED ISSUE #1511: Does Server-Timing work when not exposed through Access-Control-Expose-Headers?

noamr:
The `Server-Timing` header is protected by TAO, so there's no need to safelist it in CORS as well.

Closes #1511

   * Already implemented
   * https://github.com/web-platform-tests/wpt/pull/37714
   * Already implemented

***
<a href="https://whatpr.org/fetch/1584.html" title="Last updated on Jan 3, 2023, 8:07 AM UTC (edff313)">Preview</a> | <a href="https://whatpr.org/fetch/1584/464326e...edff313.html" title="Last updated on Jan 3, 2023, 8:07 AM UTC (edff313)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4624,8 +4624,9 @@ steps:
   from <var>response</var>'s <a for="filtered response">internal response</a>'s
   <a for=response>header list</a>.
 
-  <p class="note">We can safely use filtered response here as the opt-in to protect
-  `<code>Server-Timing</code>` responses is `<code>Timing-Allow-Origin</code>`.
+  <p class=note>Using _response_'s <a for="filtered response">internal response</a> is safe as
+  exposing `<code>Server-Timing</code>` header data is guarded through the
+  `<code>Timing-Allow-Origin</code>` header.
 
   <p>The user agent may decide to expose `<code>Server-Timing</code>` headers to non-secure contexts
   requests as well.
--- a/fetch.bs
+++ b/fetch.bs
@@ -4621,7 +4621,11 @@ steps:
   <a for="fetch params">request</a>'s <a for=request>client</a> is a <a>secure context</a>, then set
   <var>timingInfo</var>'s <a for="fetch timing info">server-timing headers</a> to the
   result of <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>`
-  from <var>response</var>'s <a for=response>header list</a>.
+  from <var>response</var>'s <a for="filtered response">internal response</a>'s
+  <a for=response>header list</a>.
+
+  <p class="note">We can safely use filtered response here as the opt-in to protect
+  `<code>Server-Timing</code>` responses is `<code>Timing-Allow-Origin</code>`.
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4621,7 +4621,11 @@ steps:
   <a for="fetch params">request</a>'s <a for=request>client</a> is a <a>secure context</a>, then set
   <var>timingInfo</var>'s <a for="fetch timing info">server-timing headers</a> to the
   result of <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>`
-  from <var>response</var>'s <a for=response>header list</a>.
+  from <var>response</var>'s <a for="filtered response">internal response</a>'s
+  <a for=response>header list</a>.
+
+  <p class="note">We can safely use filtered response here as the opt-in to protect
+  `<code>Server-Timing</code>` responses is `<code>Timing-Allow-Origin</code>`.
```
annevk, line 4628:
```suggestion
  <p class=note>Using _response_'s <a for="filtered response">internal response</a> is safe as
  exposing `<code>Server-Timing</code>` header data is guarded through the
  `<code>Timing-Allow-Origin</code>` header.
```

fetch.bs
```diff
@@ -4621,7 +4621,11 @@ steps:
   <a for="fetch params">request</a>'s <a for=request>client</a> is a <a>secure context</a>, then set
   <var>timingInfo</var>'s <a for="fetch timing info">server-timing headers</a> to the
   result of <a for="header list">getting, decoding, and splitting</a> `<code>Server-Timing</code>`
-  from <var>response</var>'s <a for=response>header list</a>.
+  from <var>response</var>'s <a for="filtered response">internal response</a>'s
+  <a for=response>header list</a>.
```
annevk, line 4625: We're sure _response_ is always a filtered response at this point, right?

annevk (commented):
Thanks, looks good modulo nit. I guess you might have to rebase to avoid the implementation-defined issue.

fetch.bs
noamr: It could be a standard CORS/same-origin response. Would that still be a filtered response? Is there a shorthand for receiving the "internal response if it's filtered or the regular response otherwise"?

annevk: Step 14.2 of main fetch seems to ensure there's either a network error or a filtered response so I think we're good.

HTML defines "unsafe response" which we should maybe define here as well, but would be good to know where it's needed first.

PR #1585, opened 2023-01-02
Editorial: Add prose about constructing a request

noamr:
This intends to be a summary of informative notes in the #Requests section, into a concise "starter" guide to populating a request, which is usually the tricky bit of using Fetch.

Requested by authors of other specs, see for example: https://github.com/w3ctag/design-principles/issues/238 https://github.com/fedidcg/FedCM/issues/320

***
<a href="https://whatpr.org/fetch/1585.html" title="Last updated on Mar 2, 2023, 1:03 PM UTC (dd89623)">Preview</a> | <a href="https://whatpr.org/fetch/1585/a5560d2...dd89623.html" title="Last updated on Mar 2, 2023, 1:03 PM UTC (dd89623)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -8456,7 +8456,7 @@ the [[WEBIDL inline]], for which every object that implements an <a>interface</a
 <a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the element's
 <a>node document</a>'s <a>relevant settings object</a>. [[WEBIDL]]
 
-<p>Choose your <a for=/>request</a>'s <a for=request>destination</a>.
+<p>Choose your <a for=/>request</a>'s <a for=request>destination</a> using the guidance in the <a>destination table</a>.
 <a for=request>destinations</a> affect <cite>Content Security Policy</cite> and have other
 implications such as the `<code>sec-fetch-dest</code>` header, so they are much more than
 informative metadata. If a new feature requires a <a for=request>destination</a> that's not in the
--- a/fetch.bs
+++ b/fetch.bs
@@ -8452,7 +8452,7 @@ body, you can assign or stream it to <a for=/>request</a>'s <a for=request>body<
 <p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
 <a>environment settings object</a> you're operating in. Web-exposed APIs usually operate on
 <a>platform objects</a> (e.g. a <cite>Web IDL</cite> based API), which have a
-<a>relevant settings object</a>. For example, a <a for=/>request</a> associated with a DOM
+<a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a DOM
 <a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the element's
 <a>node document</a>'s <a>relevant settings object</a>. [[WEBIDL]]
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -8450,8 +8450,8 @@ as defined by HTTP. If your `<code>POST</code>` or `<code>PUT</code>` <a for=/>r
 body, you can assign or stream it to <a for=/>request</a>'s <a for=request>body</a>. [[HTTP]]
 
 <p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
-<a>environment settings object</a> you're operating in. Web-exposed APIs usually operate on
-<a>platform objects</a> (e.g. a <cite>Web IDL</cite> based API), which have a
+<a>environment settings object</a> you're operating in. Web-exposed APIs are generally defined with
+the [[WEBIDL inline]], for which every object that implements an <a>interface</a> has a
 <a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a DOM
 <a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the element's
 <a>node document</a>'s <a>relevant settings object</a>. [[WEBIDL]]
--- a/fetch.bs
+++ b/fetch.bs
@@ -1703,7 +1703,7 @@ and "<code>webidentity</code>" as fetches with those destinations skip service w
 "<code>xslt</code>" as that too can cause script execution. It is not included in the list as it is
 not always relevant and might require different behavior.
 
-<div class=note id="destination-table">
+<div class=note id=destination-table>
  <p>The following table illustrates the relationship between a <a for=/>request</a>'s
  <a for=request>initiator</a>, <a for=request>destination</a>, CSP directives, and features. It is
  not exhaustive with respect to features. Features need to have the relevant values defined in their
@@ -8440,6 +8440,7 @@ correctly. This section aims to give some advice.
 
 <p class=XXX>This is a work in progress.
 
+
 <h3 id=fetch-elsewhere-request>Setting up a request</h3>
 
 <p>The first step in <a for=/>fetching</a> is to create a <a for=/>request</a>, and populate its
@@ -8447,37 +8448,36 @@ correctly. This section aims to give some advice.
 
 <p>Start by setting the <a for=/>request</a>'s <a for=request>URL</a> and <a for=request>method</a>,
 as defined by HTTP. If your `<code>POST</code>` or `<code>PUT</code>` <a for=/>request</a> needs a
-body, you set <a for=/>request</a>'s <a for=request>body</a> to a <a for=/>byte sequence</a>
-<a for="byte sequence">as a body</a>, or to a new <a for=/>body</a> whose <a for=body>stream</a>
-is a {{ReadableStream}} you created.  [[HTTP]]
+body, you set <a for=/>request</a>'s <a for=request>body</a> to a <a for=/>byte sequence</a>, or to
+a new <a for=/>body</a> whose <a for=body>stream</a> is a {{ReadableStream}} you created. [[HTTP]]
 
 <p>Choose your <a for=/>request</a>'s <a for=request>destination</a> using the guidance in the
 <a href="#destination-table">destination table</a>. <a for=request>Destinations</a> affect
-<cite>Content Security Policy</cite> and have other implications such as the
-[:Sec-Fetch-Dest:] header, so they are much more than informative metadata. If a new
-feature requires a <a for=request>destination</a> that's not in the
-<a href="#destination-table">destination table</a>, please
+<cite>Content Security Policy</cite> and have other implications such as the [:Sec-Fetch-Dest:]
+header, so they are much more than informative metadata. If a new feature requires a
+<a for=request>destination</a> that's not in the <a href="#destination-table">destination table</a>,
+please
 <a href="https://github.com/whatwg/fetch/issues/new?title=What%20destination%20should%20my%20feature%20use">file an issue</a>
-to discuss. [[!CSP]]
+to discuss. [[CSP]]
 
 <p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
 <a>environment settings object</a> you're operating in. Web-exposed APIs are generally defined with
 Web IDL, for which every object that implements an <a>interface</a> has a
-<a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a
-DOM <a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the
-element's <a>node document</a>'s <a>relevant settings object</a>. All features that are directly
-web-exposed by JavaScript, HTML, CSS, or other {{Document}} subresources should have a
+<a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with an
+<a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the element's
+<a>node document</a>'s <a>relevant settings object</a>. All features that are directly web-exposed
+by JavaScript, HTML, CSS, or other {{Document}} subresources should have a
 <a for=request>client</a>.
 
-<p>If your <a for=/>fetching</a> is not directly web-exposed, e.g. it is sent in the background
+<p>If your <a for=/>fetching</a> is not directly web-exposed, e.g., it is sent in the background
 without relying on a current {{Window}} or {{Worker}}, leave <a for=/>request</a>'s
 <a for=request>client</a> as null and set the <a for=/>request</a>'s <a for=request>origin</a>,
 <a for=request>policy container</a>, <a for=request>service-workers mode</a>, and
-<a for=request>referrer</a> to appropriate values instead, e.g. by copying them from the
+<a for=request>referrer</a> to appropriate values instead, e.g., by copying them from the
 <a>environment settings object</a> ahead of time. In these more advanced cases, make sure the
-details of how your fetch handles <cite>Content Security Place</cite> and
-<a for=/>referrer policy</a> are fleshed out. Also make sure you handle concurrency, as callbacks (see [[#fetch-elsewhere-fetch]])
-would be posted on a <a>parallel queue</a>. [[!REFERRER]] [[!CSP]]
+details of how your fetch handles <cite>Content Security Policy</cite> and
+<a for=/>referrer policy</a> are fleshed out. Also make sure you handle concurrency, as callbacks
+(see [[#fetch-elsewhere-fetch]]) would be posted on a <a>parallel queue</a>. [[REFERRER]] [[CSP]]
 
 <p>Think through the way you intend to handle cross-origin resources. Some features may only work in
 the <a>same origin</a>, in which case set your <a for=/>request</a>'s <a for=request>mode</a> to
@@ -8489,19 +8489,19 @@ to discuss.
 
 <p>For cross-origin requests, also determines if <a for=/>credentials</a> are to be included with
 the requests, in which case set your <a for=/>request</a>'s <a for=request>credentials mode</a> to
-"<code>include</code>",
+"<code>include</code>".
 
-<p>Figure out if your fetch needs to be reported to [[RESOURCE-TIMING inline]], and with which
+<p>Figure out if your fetch needs to be reported to <cite>Resource Timing</cite>, and with which
 <a for=request>initiator type</a>. By passing an <a for=request>initiator type</a> to the
 <a for=/>request</a>, reporting to <cite>Resource Timing</cite> will be done automatically once the
-fetch is done and the <a for=/>response</a> is fully downloaded.  [[RESOURCE-TIMING]]
+fetch is done and the <a for=/>response</a> is fully downloaded. [[RESOURCE-TIMING]]
 
 <p>If your request requires additional HTTP headers, set its <a for=request>header list</a> to
-a <a for=/>header list</a> that contains those headers, e.g.
-« (`<code>My-Header-Name</code>`, `<code>My-Header-Value</code>`) ». Note that sending custom
-headers may have implications, such as requiring a <a>CORS-preflight fetch</a>, so handle with care.
+a <a for=/>header list</a> that contains those headers, e.g., « (`<code>My-Header-Name</code>`,
+`<code>My-Header-Value</code>`) ». Sending custom headers may have implications, such as requiring a
+<a>CORS-preflight fetch</a>, so handle with care.
 
-<p>If you want to override the default caching mechanism, e.g. disable caching for this
+<p>If you want to override the default caching mechanism, e.g., disable caching for this
 <a for=/>request</a>, set the request's <a for=request>cache mode</a> to a value other than
 "<code>default</code>".
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -8462,7 +8462,7 @@ to discuss. [[!CSP]]
 
 <p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
 <a>environment settings object</a> you're operating in. Web-exposed APIs are generally defined with
-the [[WEBIDL inline]], for which every object that implements an <a>interface</a> has a
+Web IDL, for which every object that implements an <a>interface</a> has a
 <a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a
 DOM <a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the
 element's <a>node document</a>'s <a>relevant settings object</a>. All features that are directly
--- a/fetch.bs
+++ b/fetch.bs
@@ -1515,6 +1515,9 @@ these steps:
 
 <h4 id=requests>Requests</h4>
 
+<p class=note>This section documents how requests work in detail. To get started with populating a
+request, see <a href="fetch-elsewhere-request">populating a request</a>.
+
 <p>The input to <a for=/>fetch</a> is a
 <dfn export id=concept-request>request</dfn>.
 
@@ -1701,10 +1704,10 @@ and "<code>webidentity</code>" as fetches with those destinations skip service w
 not always relevant and might require different behavior.
 
 <div class=note>
- <p>The following table illustrates the relationship between a <a for=/>request</a>'s
- <a for=request>initiator</a>, <a for=request>destination</a>, CSP directives, and features. It is
- not exhaustive with respect to features. Features need to have the relevant values defined in their
- respective standards.
+ <p>The following <dfn lt="destination table">table</dfn> illustrates the relationship between a
+ <a for=/>request</a>'s <a for=request>initiator</a>, <a for=request>destination</a>, CSP
+ directives, and features. It is not exhaustive with respect to features. Features need to have the
+ relevant values defined in their respective standards.
 
  <table>
   <tbody><tr>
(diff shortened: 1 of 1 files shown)


noamr:
@annevk I'm sure some things here can be made more accurate but I thought I'd take a first stab as I feel this is needed...

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
...
+
+<p>Requests accept an optional <a for=request>header list</a> parameter, custom HTTP headers to be
+sent alongside the <a for=/>request</a>. Note that sending custom headers may have implications,
+such as requiring a <a>CORS-preflight fetch</a>, so handle with care.
+
+<p>One of the most important things when constructing a <a for=/>request</a> is understanding how
+it's supposed to handle cross-origin resources. This is expressed in the <a for=/>request</a>'s
+<a for=request>mode</a>. Some features are supposed to work only in the context of the
+<a>same origin</a>, in which case the "<code>same-origin</code>" <a for=request>mode</a> is
+appropriate. For new features that require accessing cross-origin resources, "<code>cors</code>"
```
jyasskin, line 8403: It might work better to write this as a [`<dl class="switch">`](https://w3c.github.io/tr-design/src/README#switch) with the options looking like 
```html
<dt>If the request's URL should be constrained to the same origin as its client's origin
<dd>Set [=request/mode=] to [=same origin=]
```

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
...
+should be used, alongside a <a for=request>credentials mode</a>, which defines whether the
+<a for=/>request</a> is anonymous or authenticated (e.g. with cookies and client certificates). The
+"<code>no-cors</code>" <a for=request>mode</a> is used for backwards compatibility with features
+that existed before the CORS mechanism was introduced. To use "<code>no-cors</code>" in a new
+feature, please discuss on the <a href=https://github.com/whatwg/fetch>Fetch Github repository</a>.
+"<code>navigate</code>" and "<code>websocket</code>" are reserved for specific features.
```
jyasskin, line 8409: Maybe don't even mention these. We could include something like "Other request modes can only be used by the feature they're named after." to head off questions.

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
...
+
+<p>A <a for=/>request</a> accepts a few security-related parameters that are often configured, such
+as <a for=request>integrity metadata</a>, a string hash of the actual data that verifies that the
+received resource matches an expectation, a <a for=request>cryptographic nonce metadata</a> and
+<a for=request>parser metadata</a>, which allows more fine-grained security via
+<cite>Content Security Policy</cite>, and <a for=request>referrer policy</a>, which allows the
+<a for=/>request</a> to use a different <a for=/>referrer policy</a> from the one provided by the
+<a for=request>client</a>. [[!CSP]] [[!REFERRER]]
```
jyasskin, line 8417: I'd give each of these a paragraph or an item in a [`<dl class="switch">`](https://w3c.github.io/tr-design/src/README#switch). Like 
> If you need to verify the received data hashes to a particular value, set [=request/integrity metadata=].

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
 
 <p class=XXX>This is a work in progress.
 
+<h3 id=fetch-elsewhere-request>Constructing a request</h3>
```
jyasskin, line 8383: A high-level comment for the whole section, which I've elaborated on in several of the specific comments:

I think we want this section to be instructions to the authors of calling specifications -- a how-to guide. It might make sense to also improve the descriptive text in https://fetch.spec.whatwg.org/#requests, but that's not the primary hole this section needs to fill.

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
...
+
+<p>The first step in <a for=/>fetching</a> is to create a <a for=/>request</a>, and populate its
+multiple associated parameters.
+
+<p>A <a for=/>request</a> oftenmost has a <a for=request>client</a>, which is the
+<a>environment settings object</a> that initiatated the request, and the receiver of the different
+<a for=/>response</a> callbacks. It then expects a <a for=request>URL</a>, and an HTTP
+<a for=request>method</a>, `<code>GET</code>` by default. `<code>POST</code>` and
+`<code>PUT</code>` requests also accept a <a for=request>body</a>, which is sent together with the
+request.
```
jyasskin, line 8393: These probably belong first:

> Start by setting the HTTP [=request/method=] and target [=request/URL=] as described in [[HTTP]]. If your `POST` or `PUT` request needs a body, you can assign or stream it to [=request/body=].

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
...
+<p>A <a for=/>request</a> accepts a few security-related parameters that are often configured, such
+as <a for=request>integrity metadata</a>, a string hash of the actual data that verifies that the
+received resource matches an expectation, a <a for=request>cryptographic nonce metadata</a> and
+<a for=request>parser metadata</a>, which allows more fine-grained security via
+<cite>Content Security Policy</cite>, and <a for=request>referrer policy</a>, which allows the
+<a for=/>request</a> to use a different <a for=/>referrer policy</a> from the one provided by the
+<a for=request>client</a>. [[!CSP]] [[!REFERRER]]
+
+<p>A <a for=/>request</a> also accepts a few metadata parameters, the important one being
+<a for=request>destination</a>. <a for=request>destinations</a> affect
```
jyasskin, line 8420: Perhaps:

> Set your request's [=request/destination=] according to how your specification is going to use the response. If you need a new destination, ...

This paragraph should link to the note that's under https://fetch.spec.whatwg.org/#request-destination-script-like that describes what each of the destinations and initiator types are for.

I think people are required to set their destination, in which case, this paragraph should appear before the optional parameters.

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
...
+<cite>Content Security Policy</cite> and have other implications such as the
+`<code>sec-fetch-dest</code>` header, so they are much more than informative metadata. If a new
+feature requires a <a for=request>destination</a> that's not in the list of
+<a for=request>destinations</a>, please file an issue in the
+<a href=https://github.com/whatwg/fetch>Fetch Github repository</a>. An aditional metadata
+paramater is <a for=request>initiator type</a>, which specifies that the <a for=/>fetching</a>
```
jyasskin, line 8426: Split this paragraph so people notice that they also need to set the initiator type. Do they always need to set it, or just if they want their resource to participate in [[resource-timing]]? Is the note under https://w3c.github.io/resource-timing/#dom-performanceresourcetiming-initiatortype the best description of how to set this?

jyasskin (commented):
Here are my thoughts about how this should work, but you should prioritize opinions from people who've had trouble calling into fetch, like @tabatkins and @npm1.

jyasskin:
Also, thanks a bunch for starting this section!

fetch.bs
```diff
@@ -8380,6 +8380,56 @@ correctly. This section aims to give some advice.
 
 <p class=XXX>This is a work in progress.
 
+<h3 id=fetch-elsewhere-request>Constructing a request</h3>
+
+<p>The first step in <a for=/>fetching</a> is to create a <a for=/>request</a>, and populate its
+multiple associated parameters.
+
+<p>A <a for=/>request</a> oftenmost has a <a for=request>client</a>, which is the
+<a>environment settings object</a> that initiatated the request, and the receiver of the different
```
npm1, line 8389: nit: initiated (typo)

fetch.bs
npm1: Agree that it would be good to format this more as a how-to, and in particular have more bullet lists (and mandatory parameters first and common optional parameters after). A hard part about using fetch is how to invoke it and continue processing the response after it's done. Based on the title of the section it would be out of scope but wanted to point out the gap there as well. There are not many good examples of using fetch in a nontrivial way where the result is 'awaited' by the algorithm, which then does something with the response.

noamr: I focused on requests first because the callback part is already kinda covered [in the subsequent bit](https://fetch.spec.whatwg.org/#fetch-elsewhere-fetch). Is something missing in that section?

jyasskin: I think that subsequent section could use an example of how to use it. I think a lot of people expect to be able to say: 
1. In parallel:
   1. fetch the request.
   2. Wait for the fetch to finish.
   3. Use the response.

And with the setup of the *processResponse* algorithms, that just doesn't work. You have to do something more like

1. Fetch the request, with the following algorithms as arguments:
   <dl><dt>processResponse<dd>Steps
   <dt>processResponseConsumeBody<dd>Other Steps

It might also be helpful to rewrite that section to map from a spec's goals about how it'll handle the response, to the set of algorithm arguments it ought to pass.

But +100 for doing all that in a separate PR. Writing this section is a great first step.

noamr: Gotcha, this makes sense. I will get to improving the callback section once we finish with the request section.
I also intend to add a section about the different cross-origin protections (CORS, TAO, CORB, CSP) with a short guide on which to use for what purpose.

fetch.bs
noamr: @jyasskin @npm1 revised based on your comments, I think it's much better now. WDYT?

npm1 (commented):
Looks good to me!

fetch.bs
```diff
@@ -8385,6 +8388,69 @@ correctly. This section aims to give some advice.
...
+"<code>same-origin</code>". As a general rule, modern web-exposed features that fetch cross-origin
+resources should use "<code>cors</code>" and decide on the <a for=request>credentials mode</a> on a
+case by case basis. For example, font loading uses "<code>same-origin</code>" (credentials are only
+sent in same-origin requests), video posters use "<code>include</code>" (credentials are always
+sent), and many features make this configurable in the API. If your feature is not web-exposed, or
+you think there is another reason for it to fetch cross-origin resources without CORS, discuss this
+with your security team and ping contributors of the
+<a href=https://github.com/whatwg/fetch>Fetch Standard</a>.
+
+<p>Figure out if your fetch needs to be reported to <cite>Resource Timing</cite>, and with which
```
jyasskin, line 8425: I wanted to suggest
```suggestion
<p>Figure out if your fetch needs to be reported to [[RESOURCE-TIMING inline]], and with which
```
to get a nice link to the spec, but it gives some ugly extra "Level 2" text as part of the full spec title. (https://github.com/tabatkins/bikeshed/issues/2441) Up to you whether you think it works better anyway.

fetch.bs
```diff
@@ -8385,6 +8388,69 @@ correctly. This section aims to give some advice.
...
+ <dt><a for=request>integrity metadata</a>
+ <dd><p>Verifies that the content of the resource matches a pre-calculated string hash.
+
+ <dt><a for=request>referrer policy</a>
+ <dd><p>Allows using a different <a for=/>referrer policy</a> from the one provided by the
+ <a for=/>environment settings object</a>. [[!REFERRER]]
+
+ <dt><a for=request>cryptographic nonce metadata</a>
+ <dd><p>Together with <cite>Content Security Policy</cite>, ensures that only elements that provide
+ this predetermined string can fetch this resource.
```
jyasskin, line 8447: I think this is
```suggestion
 <dd><p>If the site's <cite>Content Security Policy</cite> specifies a <a grammar>nonce-source</a>
 for this <a for=/>request</a>'s <a for=request>destination</a>, any request without a
 [[CSP#match-nonce-to-source-list|matching]] value in this field will be blocked.
```

fetch.bs
```diff
@@ -8385,6 +8388,69 @@ correctly. This section aims to give some advice.
...
+resources should use "<code>cors</code>" and decide on the <a for=request>credentials mode</a> on a
+case by case basis. For example, font loading uses "<code>same-origin</code>" (credentials are only
+sent in same-origin requests), video posters use "<code>include</code>" (credentials are always
+sent), and many features make this configurable in the API. If your feature is not web-exposed, or
+you think there is another reason for it to fetch cross-origin resources without CORS, discuss this
+with your security team and ping contributors of the
+<a href=https://github.com/whatwg/fetch>Fetch Standard</a>.
+
+<p>Figure out if your fetch needs to be reported to <cite>Resource Timing</cite>, and with which
+<a for=request>initiator type</a>. By passing an <a for=request>initiator type</a> to the
```
jyasskin, line 8426:
```suggestion
<p>Figure out if your fetch needs to be reported to <cite>Resource Timing</cite>. If so, set
the <a for=/>request</a>'s <a for=request>initiator type</a> to the value you want to appear
in <code>{{PerformanceResourceTiming}}.{{PerformanceResourceTiming/initiatorType}}</code>.
By passing an <a for=request>initiator type</a> to the
```
(The choice of `{{}}` or `<a idl>` actually affects the styling here.)

jyasskin (approved):
Thanks! I think @annevk and @domenic should decide how they want this section to direct feature authors to start discussions with them. Do they want to enable Github Discussions here, use Issues, or something else? Beyond that, all of my comments here are optional, and I don't expect you to wait for me once the editors approve the PR.

fetch.bs
```diff
@@ -8385,6 +8388,69 @@ correctly. This section aims to give some advice.
...
+<p>The first step in <a for=/>fetching</a> is to create a <a for=/>request</a>, and populate its
+multiple associated parameters.
+
+<p>Start by setting the <a for=/>request</a>'s <a for=request>URL</a> and <a for=request>method</a>,
+as defined by HTTP. If your `<code>POST</code>` or `<code>PUT</code>` <a for=/>request</a> needs a
+body, you can assign or stream it to <a for=/>request</a>'s <a for=request>body</a>. [[HTTP]]
+
+<p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
+<a>environment settings object</a> you're operating in. Web-exposed APIs usually operate on
+<a>platform objects</a> (e.g. a <cite>Web IDL</cite> based API), which have a
```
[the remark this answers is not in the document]
noamr: The idea is to link "platform object" with a term that people who write other APIs are familiar with

fetch.bs
```diff
@@ -8385,6 +8388,69 @@ correctly. This section aims to give some advice.
...
+<a>relevant settings object</a>. For example, a <a for=/>request</a> associated with a DOM
+<a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the element's
+<a>node document</a>'s <a>relevant settings object</a>. [[WEBIDL]]
```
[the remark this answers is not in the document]
noamr: Can you give examples outside of FedCM? Is this mostly about UA-initiated fetches that are not web-exposed in themselves?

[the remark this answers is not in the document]
jyasskin: I believe these are all web-exposed:
* https://wicg.github.io/attribution-reporting-api/#create-a-report-request
* https://wicg.github.io/background-fetch/#background-fetch-manager-fetch
* https://w3c.github.io/reporting/#try-delivery

fetch.bs
```diff
@@ -8385,6 +8388,72 @@ correctly. This section aims to give some advice.
...
+<a for=struct>items</a>.
+
+<p>Start by setting the <a for=/>request</a>'s <a for=request>URL</a> and <a for=request>method</a>,
+as defined by HTTP. If your `<code>POST</code>` or `<code>PUT</code>` <a for=/>request</a> needs a
+body, you can assign or stream it to <a for=/>request</a>'s <a for=request>body</a>. [[HTTP]]
+
+<p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
+<a>environment settings object</a> you're operating in. Web-exposed APIs usually operate on
+<a>platform objects</a> (e.g. a <cite>Web IDL</cite> based API), which have a
+<a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a DOM
```
jyasskin, line 8403: Given https://github.com/whatwg/fetch/pull/1585/files#r1062124820, I'd suggest:

```suggestion
<a>environment settings object</a> you're operating in. Web-exposed APIs are generally defined with
the [[WEBIDL inline]], for which every object that implements an <a>interface</a> has a
<a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a DOM
```

fetch.bs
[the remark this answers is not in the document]
noamr: > I believe these are all web-exposed:
> 
> * https://wicg.github.io/attribution-reporting-api/#create-a-report-request

Seems like the fetch itself is not web exposed?

> 
> * https://wicg.github.io/background-fetch/#background-fetch-manager-fetch

This is a JS API with an IDL interface, which has a relevant settings object.

> 
> * https://w3c.github.io/reporting/#try-delivery

Seems like this ends up also doing something with a JS observer, which has a relevant settings object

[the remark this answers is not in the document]
jyasskin: For the first, I might be understanding something different by "web-exposed": the result doesn't get reported to Javascript running in a web page, but developers building websites will have their designs broken if it changes.

For the second, the relevant settings object's window may have closed by the time the fetch is sent, which means it can't use that as its *client*. Maybe the right answer is to start a Service Worker to exist while the fetch is running, so that can be the *client*, but the feature's designers need someone to tell them that, and help work through the tradeoffs.

For the third, the JS observer isn't really important, but there is a global available in the [algorithm's caller](https://w3c.github.io/reporting/#send-reports), so the use of `null` is just a bug. But the developers *thought* they didn't have one, maybe because the algorithm originally ran after the originating client closed. They should have a way to ask what to do, so someone knowledgeable can find them the right client.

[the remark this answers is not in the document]
noamr: Totally agree that documentation for this is lacking. Let's fix it step by step.
btw client is not mandatory. Instead of a client you can populate the request's origin and policy container, and respond to callbacks on a parallel queue (some anonymous new thread). For some of these features maybe that's the right approach and we should address this in this documentation.

noamr:
@annevk want to have a go at reviewing this?

fetch.bs
```diff
@@ -1694,10 +1697,10 @@ and "<code>webidentity</code>" as fetches with those destinations skip service w
 not always relevant and might require different behavior.
 
 <div class=note>
- <p>The following table illustrates the relationship between a <a for=/>request</a>'s
- <a for=request>initiator</a>, <a for=request>destination</a>, CSP directives, and features. It is
- not exhaustive with respect to features. Features need to have the relevant values defined in their
- respective standards.
+ <p>The following <dfn lt="destination table">table</dfn> illustrates the relationship between a
```
domenic, line 1700: I would either make this `<dfn>destination table</dfn>`, or not make it a `<dfn>` and instead just add an `id=""` to the note. The latter seems a bit cleaner, as having a definition inside a non-normative section is weird.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
 <p class=XXX>This is a work in progress.
 
+<h3 id=fetch-elsewhere-request>Populating a request</h3>
+
+<p>The first step in <a for=/>fetching</a> is to create a <a for=/>request</a>, and populate its
+<a for=struct>items</a>.
+
+<p>Start by setting the <a for=/>request</a>'s <a for=request>URL</a> and <a for=request>method</a>,
+as defined by HTTP. If your `<code>POST</code>` or `<code>PUT</code>` <a for=/>request</a> needs a
+body, you can assign or stream it to <a for=/>request</a>'s <a for=request>body</a>. [[HTTP]]
```
domenic, line 8398: "you can assign or stream it" is confusing. How would you stream it?

I think the most common pattern here will be to use a byte sequence as the body, using the helper we recently defined: https://whatpr.org/fetch/1585.html#byte-sequence-as-a-body . I'd suggest referencing that as the primary method, and mentioning you can construct a new [=body=] for advanced cases.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+
+<p>Choose your <a for=/>request</a>'s <a for=request>destination</a> using the guidance in the
+<a>destination table</a>. <a for=request>destinations</a> affect
```
domenic, line 8401:
```suggestion
<a>destination table</a>. <a for=request>Destinations</a> affect
```

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+<cite>Content Security Policy</cite> and have other implications such as the
+`<code>sec-fetch-dest</code>` header, so they are much more than informative metadata. If a new
```
domenic, line 8403:
```suggestion
[:Sec-Fetch-Dest:] header, so they are much more than informative metadata. If a new
```

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+feature requires a <a for=request>destination</a> that's not in the <a>destination table</a>, please
+file an issue in the <a href=https://github.com/whatwg/fetch/issues>Fetch Github repository</a> to
+discuss. [[!CSP]]
```
domenic, line 8406: Usually we do these as `<a href="https://github.com/whatwg/fetch/issues/new?title=What%20destination%20should%20my%20feature%20use">file an issue</a> to discuss`.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+
+<p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
+<a>environment settings object</a> you're operating in. Web-exposed APIs are generally defined with
+the [[WEBIDL inline]], for which every object that implements an <a>interface</a> has a
+<a>relevant settings object</a> you can use. For example, a <a for=/>request</a> associated with a
+DOM <a for=/>element</a> would set the <a for=/>request</a>'s <a for=request>client</a> to the
+element's <a>node document</a>'s <a>relevant settings object</a>. All features that are directly
+web-exposed by JavaScript, HTML, CSS, or other <code>Document</code> subresources should have A
```
domenic, line 8414:
```suggestion
web-exposed by JavaScript, HTML, CSS, or other {{Document}} subresources should have a
```

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+<a for=request>client</a>.
+
+<p>If your <a for=/>fetching</a> is not directly web-exposed, e.g. it is sent in the background
+without relying on a current {{Window}} or {{Worker}}, leave <a for=/>request</a>'s
+<a for=request>client</a> as null and set the <a for=/>request</a>'s <a for=request>origin</a>,
+<a for=request>policy container</a>, and <a for=request>referrer</a> to appropriate values instead,
+e.g. by copying them from the <a>environment settings object</a> ahead of time. In these more
+advanced cases, make sure the details of how your fetch handles <cite>Content Security Place</cote>
```
domenic, line 8422:
```suggestion
advanced cases, make sure the details of how your fetch handles <cite>Content Security Policy</cite>
```

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+and <a for=/>referrer policy</a> are fleshed out. Also make sure you handle concurrency, as
+callbacks would be posted on a <a>parallel queue</a>. [[!REFERRER]] [[!CSP]]
+
+<p>Think through the way you intend to handle cross-origin resources. Some features may only work in
+the <a>same origin</a>, in which case set your <a for=/>request</a> <a for=request>mode</a> to
+"<code>same-origin</code>". Otherwise, new web-exposed features should almost always set their
+<a for=request>mode</a> to "<code>cors</code>". If your feature is not web-exposed, or you think
+there is another reason for it to fetch cross-origin resources without CORS, discuss this with a
+security team and ping contributors of the
+<a href=https://github.com/whatwg/fetch>Fetch Standard</a>.
```
domenic, line 8432: Similar suggestion for issue-filing markup.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+
+<p>Figure out if your fetch needs to be reported to [[RESOURCE-TIMING inline]], and with which
+<a for=request>initiator type</a>. By passing an <a for=request>initiator type</a> to the
+<a for=/>request</a>, reporting to <cite>Resource Timing</cite> will be done automatically once the
+fetch is done and the <a for=/>response</a> is fully downloaded.  [[RESOURCE-TIMING]]
+
+<p>If your request requires additional HTTP headers, <a for="header list">append</a> them to your
```
domenic, line 8439: Don't you actually just set the header list? Since header list is, as you recently explained to me, more like "extra header list"?

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+<p>Check if your request needs one or more of the additional security-related parameters. The
+following ones are often configurable, e.g. by an HTML attribute:
+
+<dl>
+ <dt><a for=request>credentials mode</a>
+ <dd><p>Set to "<code>include</code>" if <a for=/>credentials</a> are to be sent with cross-origin
+ requests.
+
+ <dt><a for=request>integrity metadata</a>
+ <dd><p>Verifies that the content of the resource matches a pre-calculated string hash.
```
domenic, line 8452: I think very few fetches will need this. (Only ones initiated by API surfaces which have an explicit place for web developers to put the integrity, basically.) Not sure if it's worth mentioning.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+
+ <dt><a for=request>referrer policy</a>
+ <dd><p>Allows using a different <a for=/>referrer policy</a> from the one provided by the
+ <a for=/>policy container</a>. [[!REFERRER]]
+
+ <dt><a for=request>cryptographic nonce metadata</a>
+ <dd><p>If the site's <cite>Content Security Policy</cite> specifies a <a grammar>nonce-source</a>
+ for this <a for=/>request</a>'s <a for=request>destination</a>, any request without a
+ [[CSP#match-nonce-to-source-list|matching]] value in this field will be blocked.
```
domenic, line 8461: Ditto.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+</dl>
+
+<p>Browse through the rest of the parameters for <a for=/>request</a> to see if something else is
+relevant to you. The rest of the parameters are used less frequently, often for special purposes,
+and they are documented in detail in the <a href="#requests">Requests</a> section of this standard.
```
domenic, line 8466: Use bikeshed auto-links for the section here, if possible.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
+
```
domenic, line 8467: I think service-workers mode often needs configuring for the type of "exotic" fetches we've seen recently. Maybe mention it in the section on null-client fetches.

Similarly such fetches will probably need to set the referrer.

fetch.bs
```diff
@@ -8385,6 +8388,83 @@ correctly. This section aims to give some advice.
...
```
domenic, line 8468: Cache mode and redirect mode are other parameters I sometimes see changed.

noamr: Does someone use redirect-mode other than navigation & service-workers? Seems pretty exotic

noamr:
Thanks @domenic, I addressed your notes but seems like builds are failing (CSSWG API is down?)

fetch.bs
domenic: I see a few places using it to prohibit redirects: https://dontcallmedom.github.io/webdex/r.html#redirect%20mode%40%40request%40dfn

noamr: OK, added a short note about prohibiting redirects ("manual" is the exotic one I guess)

noamr:
> Thanks @domenic, I addressed your notes but seems like builds are failing (CSSWG API is down?)

Any other comments?
Thanks 🙏🏻

fetch.bs
```diff
@@ -1693,11 +1696,11 @@ and "<code>webidentity</code>" as fetches with those destinations skip service w
 "<code>xslt</code>" as that too can cause script execution. It is not included in the list as it is
 not always relevant and might require different behavior.
 
-<div class=note>
- <p>The following table illustrates the relationship between a <a for=/>request</a>'s
- <a for=request>initiator</a>, <a for=request>destination</a>, CSP directives, and features. It is
- not exhaustive with respect to features. Features need to have the relevant values defined in their
- respective standards.
+<div class=note id="destination-table">
+ <p>The following table llustrates the relationship between a
```
domenic, line 1700: This entire paragraph's diff seems unnecessary, and mostly just inserts a typo.

fetch.bs
```diff
@@ -8385,6 +8388,78 @@ correctly. This section aims to give some advice.
...
+<cite>Content Security Policy</cite> and have other implications such as the
+[:Sec-Fetch-Dest:] header, so they are much more than informative metadata. If a new
+feature requires a <a for=request>destination</a> that's not in the
+<a href="#destination-table">destination table</a>, please
+<a href="https://github.com/whatwg/fetch/issues/new?title=What%20destination%20should%20my%20feature%20use">file an issue</a>
+to discuss. [[!CSP]]
+
+<p>Set your <a for=/>request</a>'s <a for=request>client</a> to the
+<a>environment settings object</a> you're operating in. Web-exposed APIs are generally defined with
+the [[WEBIDL inline]], for which every object that implements an <a>interface</a> has a
```
domenic, line 8413:
```suggestion
Web IDL, for which every object that implements an <a>interface</a> has a
```

fetch.bs
domenic: Still missing this

fetch.bs
```diff
@@ -8385,6 +8388,78 @@ correctly. This section aims to give some advice.
...
+web-exposed by JavaScript, HTML, CSS, or other <code>Document</code> subresources should have A
+<a for=request>client</a>.
+
+<p>If your <a for=/>fetching</a> is not directly web-exposed, e.g. it is sent in the background
+without relying on a current {{Window}} or {{Worker}}, leave <a for=/>request</a>'s
+<a for=request>client</a> as null and set the <a for=/>request</a>'s <a for=request>origin</a>,
+<a for=request>policy container</a>, <a for=request>service-workers mode</a>, and
+<a for=request>referrer</a> to appropriate values instead, e.g. by copying them from the
+<a>environment settings object</a> ahead of time. In these more advanced cases, make sure the
+details of how your fetch handles <cite>Content Security Place</cote> and
```
domenic, line 8426:
```suggestion
details of how your fetch handles <cite>Content Security Place</cite> and
```

fetch.bs
```diff
@@ -8385,6 +8388,78 @@ correctly. This section aims to give some advice.
...
+<a for=/>referrer policy</a> are fleshed out. Also make sure you handle concurrency, as callbacks
```
domenic, line 8427:
```suggestion
<a for=/>referrer policy</a> are fleshed out. Also make sure you handle concurrency, as callbacks (see [[#fetch-elsewhere-fetch]])
```

fetch.bs
```diff
@@ -8385,6 +8388,78 @@ correctly. This section aims to give some advice.
...
+<p>If your request requires additional HTTP headers, set its <a for=request>header list</a> to
+a <a for=/>header list</a> that contains those headers, e.g.
+« (`<code>My-Header-Name</code>`, `<code>My-Header-Value</code>`) ». Note that sending custom
+headers may have implications, such as requiring a <a>CORS-preflight fetch</a>, so handle with care.
+
+<p>If you want to override the default caching mechanism, e.g. disable caching for this
+<a for=/>request</a>, set the request's <a for=request>cache mode</a> to a value other than
+"<code>default</code>".
+
+<p>Determine if your request should support redirects. If it doesn't, set its
```
domenic, line 8456:
```suggestion
<p>Determine whether you want your request to support redirects. If you don't, set its
```

domenic (commented):
Still some nits.

fetch.bs

noamr:
> Still some nits.

All fix... Another review pass?

noamr:
Ping? @domenic

domenic (approved):
LGTM, sorry for missing the previous ping!

annevk (approved):

PR #1593, opened 2023-01-13
Set body with byte reading support

### LINKED ISSUE #267: Use a ReadableStream with byte source (formerly called ReadableByteStream) for .body

saschanaz:
Fixes #267, paired with https://github.com/w3c/FileAPI/pull/188

The switch is quite simple it seems 👀

   * Mozilla (already shipped)
   * Google @ricea
   * Apple? @annevk
   * https://github.com/web-platform-tests/wpt/pull/37910
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1243329
   * Gecko: (Shipped)
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=250549
   * Deno (not for CORS changes): https://github.com/denoland/deno/issues/17386

***
<a href="https://whatpr.org/fetch-1/1593.html" title="Last updated on Feb 17, 2023, 7:53 PM UTC (5473365)">Preview</a> | <a href="https://whatpr.org/fetch/1593/093c622...5473365.html" title="Last updated on Feb 17, 2023, 7:53 PM UTC (5473365)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5922,6 +5922,10 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
  <li><p>Let |pullAlgorithm| be the followings steps:
 
   <ol>
+   <li>Let |promise| be a new promise.
+
+   <li>Return |promise| and run the remaining steps [=in parallel=].
+
    <li>If |buffer| is empty and the ongoing fetch is [=fetch/suspend|suspended=],
    [=fetch/resumed|resume=] the fetch.
 
@@ -5950,6 +5954,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
    <li><p>If |stream| is [=ReadableStream/errored=], then [=fetch controller/terminate=]
    |fetchParams|'s [=fetch params/controller=].
+
+   <li>Resolve |promise| with undefined.
   </ol>
 
  <li><p>Let <var>cancelAlgorithm</var> be an algorithm that <a for="fetch controller">aborts</a>
--- a/fetch.bs
+++ b/fetch.bs
@@ -5923,16 +5923,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
  <li><p>Let <var>highWaterMark</var> be a non-negative, non-NaN number, chosen by the user agent.
 
- <li><p>Let <var>sizeAlgorithm</var> be an algorithm that accepts a <a>chunk</a> object and returns
- a non-negative, non-NaN, non-infinite number, chosen by the user agent.
-
  <li><p>Let <var>stream</var> be a <a>new</a> {{ReadableStream}}.
 
- <li><p><a for=ReadableStream>Set up</a> <var>stream</var> with
- <a for="ReadableStream/set up"><var>pullAlgorithm</var></a> set to <var>pullAlgorithm</var>,
- <a for="ReadableStream/set up"><var>cancelAlgorithm</var></a> set to <var>cancelAlgorithm</var>,
- <a for="ReadableStream/set up"><var>highWaterMark</var></a> set to <var>highWaterMark</var>, and
- <a for="ReadableStream/set up"><var>sizeAlgorithm</var></a> set to <var>sizeAlgorithm</var>.
+ <li><p>[=ReadableStream/set up with byte reading support|Set up=] <var>stream</var> with byte
+ reading support with <var>[=ReadableStream/set up/pullAlgorithm=]</var> set to
+ <var>pullAlgorithm</var>, <var>[=ReadableStream/set up/cancelAlgorithm=]</var> set to
+ <var>cancelAlgorithm</var>, <var>[=ReadableStream/set up/highWaterMark=]</var> set to
+ <var>highWaterMark</var>.
 
  <li><p>Set <var>response</var>'s <a for=response>body</a> to a new <a for=/>body</a> whose
  <a for=body>stream</a> is <var>stream</var>.
@@ -6723,7 +6720,8 @@ steps:
  running <var>object</var>'s <a for=Blob>get stream</a>.
 
  <li><p>Otherwise, set <var>stream</var> to a <a>new</a> {{ReadableStream}} object, and
- <a for=ReadableStream>set up</a> <var>stream</var>.
+ [=ReadableStream/set up with byte reading support|set up=] <var>stream</var> with byte reading
+ support.
 
  <li><p><a for=/>Assert</a>: <var>stream</var> is a {{ReadableStream}} object.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -5922,7 +5922,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
  <li><p>Let |pullAlgorithm| be the followings steps:
 
   <ol>
-   <li>[=fetch/resumed|Resume=] the ongoing fetch if it is [=fetch/suspend|suspended=].
+   <li>If |buffer| is empty and the ongoing fetch is [=fetch/suspend|suspended=],
+   [=fetch/resumed|resume=] the fetch.
 
    <li>Wait until |buffer| is not empty.
 
@@ -5949,9 +5950,6 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
    <li><p>If |stream| is [=ReadableStream/errored=], then [=fetch controller/terminate=]
    |fetchParams|'s [=fetch params/controller=].
-
-   <li><p>If |stream| doesn't [=ReadableStream/need more data=], ask the user agent to
-   [=fetch/suspend=] the ongoing fetch.
   </ol>
 
  <li><p>Let <var>cancelAlgorithm</var> be an algorithm that <a for="fetch controller">aborts</a>
@@ -6017,6 +6015,9 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <var>fetchParams</var>'s <a for="fetch params">controller</a>.
 
          <li><p>Append |bytes| to |buffer|.
+
+         <li><p>If the size of |buffer| is larger than a limit chosen by the user agent, ask the
+         user agent to [=fetch/suspend=] the ongoing fetch.
         </ol>
 
        <li><p>Otherwise, if the bytes transmission for <var>response</var>'s message body is done
--- a/fetch.bs
+++ b/fetch.bs
@@ -5922,7 +5922,7 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
  <li><p>Let |pullAlgorithm| be the followings steps:
 
   <ol>
-   <li>Let |promise| be a new promise.
+   <li>Let |promise| be [=a new promise=].
 
    <li>Return |promise| and run the remaining steps [=in parallel=].
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -5922,40 +5922,46 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
  <li><p>Let |pullAlgorithm| be the followings steps:
 
   <ol>
-   <li>Let |promise| be [=a new promise=].
+   <li><p>Let |promise| be [=a new promise=].
 
-   <li>Return |promise| and run the remaining steps [=in parallel=].
+   <li><p>Return |promise| and run the following steps [=in parallel=].
 
-   <li>If |buffer| is empty and the ongoing fetch is [=fetch/suspend|suspended=],
-   [=fetch/resumed|resume=] the fetch.
+    <ol>
+     <li><p>If the size of |buffer| is smaller than a lower limit chosen by the user agent and the
+     ongoing fetch is [=fetch/suspend|suspended=], [=fetch/resumed|resume=] the fetch.
 
-   <li>Wait until |buffer| is not empty.
+     <li><p>Wait until |buffer| is not empty.
 
-   <li>Let |available| be the size of |buffer|.
+     <li><p>[=Queue a fetch task=] to run the following steps, with |fetchParams|'s [=task destination=].
 
-   <li>Let |desiredSize| be |available|.
+      <ol>
+       <li><p>Let |available| be the size of |buffer|.
 
-   <li>If |stream|'s [=ReadableStream/current BYOB request view=] is non-null, then set
-   |desiredSize| to |stream|'s [=ReadableStream/current BYOB request view=]'s [=BufferSource/byte
-   length=].
+       <li><p>Let |desiredSize| be |available|.
 
-   <li>Let |extractSize| be the smaller value of |available| and |desiredSize|.
+       <li><p>If |stream|'s [=ReadableStream/current BYOB request view=] is non-null, then set
+       |desiredSize| to |stream|'s [=ReadableStream/current BYOB request view=]'s [=BufferSource/byte
+       length=].
 
-   <li>Let <var>bytes</var> be the result of extracting |extractSize| of bytes from
-   <var>buffer</var>.
+       <li><p>Let |extractSize| be the smaller value of |available| and |desiredSize|.
 
-   <li>If |stream|s [=ReadableStream/current BYOB request view=] is non-null, then
-   [=ArrayBufferView/write=] |bytes| into |stream|'s [=ReadableStream/current BYOB request view=],
-   and set |view| to |stream|'s [=ReadableStream/current BYOB request view=]. <li>Otherwise, set
-   |view| to the result of [=ArrayBufferView/create|creating=] a {{Uint8Array}} from |bytes| in
-   |stream|'s [=relevant Realm=].
+       <li><p>Let <var>bytes</var> be the result of extracting |extractSize| of bytes from
+       <var>buffer</var>.
 
-   <li>[=ReadableStream/Enqueue=] |view| into |stream|.
+       <li>If |stream|s [=ReadableStream/current BYOB request view=] is non-null, then
+       [=ArrayBufferView/write=] |bytes| into |stream|'s [=ReadableStream/current BYOB request
+       view=], and set |view| to |stream|'s [=ReadableStream/current BYOB request view=].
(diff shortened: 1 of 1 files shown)


domenic:
You should probably have something similar to https://wicg.github.io/serial/#readable-attribute , which uses https://streams.spec.whatwg.org/#readablestream-current-byob-request-view to require that the implementation only pull the requested number of bytes from the network, and then write into the current BYOB request view when applicable.

saschanaz:
Hmm, so currently Fetch has "If stream doesn’t [need more data](https://streams.spec.whatwg.org/#readablestream-need-more-data) ask the user agent to [suspend](https://fetch.spec.whatwg.org/#concept-fetch-suspend) the ongoing fetch.", while Web Serial has:

>1. Let desiredSize be the [desired size to fill up to the high water mark](https://streams.spec.whatwg.org/#readablestream-desired-size-to-fill-up-to-the-high-water-mark) for [this](https://webidl.spec.whatwg.org/#this).[[[readable]]](https://wicg.github.io/serial/#dfn-readable).
>2. If [this](https://webidl.spec.whatwg.org/#this).[[[readable]]](https://wicg.github.io/serial/#dfn-readable)'s [current BYOB request view](https://streams.spec.whatwg.org/#readablestream-current-byob-request-view) is non-null, then set desiredSize to [this](https://webidl.spec.whatwg.org/#this).[[[readable]]](https://wicg.github.io/serial/#dfn-readable)'s [current BYOB request view](https://streams.spec.whatwg.org/#readablestream-current-byob-request-view)'s [byte length](https://webidl.spec.whatwg.org/#buffersource-byte-length).

I guess I should revise the sentence to also check whether the current byob request view is null and suspend the fetch if it is. Am I understanding correct?

And if it's correct, should `need more data` be revised for byte streams?

domenic:
> I guess I should revise the sentence to also check whether the current byob request view is null and suspend the fetch if it is. Am I understanding correct?

Not really.

I guess there are two incompatible models in play here, and we have to pick a path.

1. Rely on buffering somewhere else in the stack. In this model, we'd:
   - Set highWaterMark = 0
   - Assume that as data comes in from the network, it's getting stored in a buffer/queue somewhere else in the stack, e.g. kernel, networking library, etc.
   - Wait for pulls. When we get a pull, either:
     - BYOB request is null, and so we create a new Uint8Array, grab whatever's in the other-part-of-the-stack buffer, put it into that Uint8Array, and enqueue it. (If nothing's there now, wait until something is.)
     - BYOB request is not null, and so we grab `byobRequestView.byteLength` bytes from the other-part-of-the-stack buffer, and write it into `byobRequestView`.
   - Write out some vague spec text about how we're assuming that stuff accumulating in the elsewhere-in-the-stack buffer causes backpressure, and thus will automatically suspend/resume bytes accumulating in that buffer, probably in a gradual fashion due to how TCP works.

2. Rely on the stream's internal queue. In this model, we'd basically do what your current draft of the PR does, or the current spec does. We only use enqueue, ignoring the current BYOB request. We'd use "suspend" and "resume" primitives to interface with the network layer, based on when the stream's internal queue is too full (i.e., when "need more data" is false).

There is no real observable difference between these for JavaScript developers. So maybe (2) is fine, perhaps supplemented with a note. But in terms of how they communicate strategies to implementers, (1) seems better. With (2), if you implemented it verbatim, you'd be doing (unobservable) copies all over the place, from the Uint8Arrays that have accumulated in the stream's internal queue via enqueue, into the BYOB request buffers the web developer supplies.

This is a bit more complicated because from what I understand, (1) is not really implementable directly on top of low-level OS socket APIs. This is because in that framing, the "buffer somewhere elsewhere in the stack" is the kernel buffer, and that has a limit, so you could eventually lose data if backpressure isn't respected fast enough. So really there's a hidden middle layer which wraps the OS socket APIs with an unlimited buffer, I think. But it's been many years since I investigated this and I'm not sure I ever fully understood it, so, I might not have that part correct...

((2) is definitely not implementable directly on top of low-level APIs, because there is no "suspend" or "resume" socket API. In some sense that is bad, but in some sense it's at least honest about how high-level it is.)

Anyway, I think thoughts from implementers on what would be the most helpful would be appreciated. /cc @ricea

saschanaz:
cc @evilpie and @mgaudet

ricea:
@domenic (1) suits me, because it matches the way we'll actually implement it.

If you're reading directly from a kernel buffer and we're using TCP then the kernel will ensure that you don't lose data, applying backpressure if you stop reading for a while.

However, in practice we're almost always using TLS, so there's a TLS library also buffering between us and the kernel. It will also be careful not to drop data and will apply backpressure when its finite buffer is full, so no problem there.

If we're using UDP, then since this is Fetch I assume we're using QUIC. QUIC provides in-order reliable streams, so you won't lose data. I'm pretty sure backpressure works similarly to TCP within a stream.

In Chrome we have a bunch of extra abstraction layers, but I assume you're thinking about server software that runs closer to the OS.

saschanaz:
Okay, so I asked around, checked the implementation, and I'm now confident that option 1 matches our behavior. I tried to spec that, except the backpressure thing. I think that one is more of the description for suspend/resume.

fetch.bs
```diff
@@ -5914,25 +5914,55 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+   <li>Let <var>bytes</var> be the result of extracting |extractSize| of bytes from
+   <var>buffer</var>.
+
+   <li>If |stream|s [=ReadableStream/current BYOB request view=] is non-null, then
+   [=ArrayBufferView/write=] |bytes| into |stream|'s [=ReadableStream/current BYOB request view=],
+   and set |view| to |stream|'s [=ReadableStream/current BYOB request view=]. <li>Otherwise, set
+   |view| to the result of [=ArrayBufferView/create|creating=] a {{Uint8Array}} from |bytes| in
+   |stream|'s [=relevant Realm=].
+
+   <li>[=ReadableStream/Enqueue=] |view| into |stream|.
```
domenic, line 5948: Don't enqueue the current BYOB request view. Only enqueue if you needed to create a new view.

fetch.bs
```diff
@@ -5914,25 +5914,55 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+   <li><p>If |stream| is [=ReadableStream/errored=], then [=fetch controller/terminate=]
+   |fetchParams|'s [=fetch params/controller=].
```
domenic, line 5951: I don't think this can happen inside pull. I think it needs to be separate, something like, if the network blows up, terminate the controller and error the stream.

I guess below there is already a line that does that, but it doesn't error the stream? Is there some way in which terminating the controller errors the stream, which I haven't seen?

fetch.bs
```diff
@@ -5914,25 +5914,55 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+   <li><p>If |stream| doesn't [=ReadableStream/need more data=], ask the user agent to
+   [=fetch/suspend=] the ongoing fetch.
```
domenic, line 5954: In this model, we don't consult "need more data" at all. We should instead have something vague near the definition of _buffer_ about how we expect that the buffer getting too full / too empty will suspend/resume.

fetch.bs
saschanaz: Isn't it something recommended by Streams? Below https://streams.spec.whatwg.org/#readablestream-current-byob-request-view:

>Specifications should, when possible, [write](https://webidl.spec.whatwg.org/#arraybufferview-write) into the [current BYOB request view](https://streams.spec.whatwg.org/#readablestream-current-byob-request-view) when it is non-null, and then call [enqueue](https://streams.spec.whatwg.org/#readablestream-enqueue) with that view.

It doesn't have a separate algorithm for `respond`, so I think enqueuing is the only way to ping the stream...

domenic: You're right, I totally forgot that we wrapped up the JS `enqueue()` + `respond()` into a single spec-level call. Sorry!

fetch.bs
saschanaz: I'm not quite sure either and honestly just copied this from the previous steps which was also right after the enqueueing step.

Looking at the [Gecko call diagram](https://searchfox.org/mozilla-central/query/default?q=calls-to%3A%27mozilla%3A%3Adom%3A%3AReadableByteStreamControllerError%27%20depth%3A4), it seems `ReadableByteStreamControllerRespond` can lead to `ReadableByteStreamControllerEnqueueClonedChunkToQueue` which can error the stream on a buffer clone failure. It's complex enough, I'd be happy if the spec can list the possible error reasons.

>I think it needs to be separate, something like, if the network blows up, terminate the controller and error the stream.

Yes, but I think that's a lil bit out of scope here as I see no relevant existing step, right?

fetch.bs
saschanaz: Hmm, I was lazy and hoping I could just reuse the existing things... but you're right, since the HWM is now zero the desired size cannot be a positive number. I'll try adding some notes below `buffer`.

saschanaz: Actually I instead moved this step back to the fetching algorithm and made it to sleep if the buffer becomes larger than a user-agent defined limit. I think that works too?

(It's me who just don't want to describe about the network layer which is not exactly in my area)

fetch.bs
```diff
@@ -5914,25 +5914,59 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+  <p class="note">This represents an internal buffer inside the network layer of the user agent.
+
+ <li><p>Let |pullAlgorithm| be the followings steps:
+
+  <ol>
+   <li>Let |promise| be [=a new promise=].
+
+   <li>Return |promise| and run the remaining steps [=in parallel=].
+
+   <li>If |buffer| is empty and the ongoing fetch is [=fetch/suspend|suspended=],
```
domenic, line 5929: It shouldn't be required to be empty before you resume. So you should add a similar user-agent-defined lower limit at which it can resume.

fetch.bs
```diff
@@ -5914,25 +5914,59 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+  <p>Let |buffer| be an empty buffer that can have bytes appended to it.
+
+  <p class="note">This represents an internal buffer inside the network layer of the user agent.
+
+ <li><p>Let |pullAlgorithm| be the followings steps:
+
+  <ol>
+   <li>Let |promise| be [=a new promise=].
+
+   <li>Return |promise| and run the remaining steps [=in parallel=].
```
domenic, line 5927: We've been avoiding this pattern these days, and instead using proper nesting.

fetch.bs
```diff
@@ -5914,25 +5914,59 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+   <li>If |buffer| is empty and the ongoing fetch is [=fetch/suspend|suspended=],
+   [=fetch/resumed|resume=] the fetch.
+
+   <li>Wait until |buffer| is not empty.
+
+   <li>Let |available| be the size of |buffer|.
+
+   <li>Let |desiredSize| be |available|.
+
+   <li>If |stream|'s [=ReadableStream/current BYOB request view=] is non-null, then set
```
domenic, line 5938: The current BYOB request view should not be accessed in parallel; nor should you create Uint8Arrays. You could probably fix this by posting a fetch task back after the waiting is done.

However, I'm not sure this is worth fixing, given the general problems with Streams being very JSey but being used even for no-JS-involved fetches (e.g. those who use a parallel queue). Thoughts from @annevk appreciated.

fetch.bs
domenic: I guess if the lack of proper erroring step is a preexisting problem then we don't need to fix it here.

I do think this step makes very little sense, so I'd prefer to remove it.

domenic (commented):
I'm happy with this approach to the network-stack buffer, modulo my comment about a nonzero lower limit

fetch.bs
saschanaz: For now I fixed it to queue a task as it was simple enough, I can revert it if Anne disagrees.

fetch.bs
saschanaz: Copypasting my question from Matrix: https://matrix.to/#/!AGetWbsMpFPdSgUrbs:matrix.org/$1y0cHr7B913RupwXNW93uLA06GT5Uls6nV8Ds8khB80?via=matrix.org&via=mozilla.org&via=igalia.com

```js
r = new ReadableStream({
  async pull(c) {
    await new Promise(r => setTimeout(r, 100));
    c.byobRequest.respond(512);
  },
  type: "bytes"
});
reader = r.getReader({mode: "byob"});
reader.read(new Uint16Array(1024));
setTimeout(() => reader.releaseLock(), 5);
```

This eventually hits https://streams.spec.whatwg.org/#abstract-opdef-readablebytestreamcontrollerenqueueclonedchunktoqueue which theoretically can error the stream. Can Fetch really ignore this?

fetch.bs
annevk: In particular you want "Run these steps in parallel:", then the nested steps, and then "Return promise.".

fetch.bs
annevk: I think it makes sense to do the right thing where we can, but I also don't mind if we take shortcuts when it cannot be observed. Although at some point we'll have to clean it all up.

fetch.bs
```diff
@@ -5961,25 +5961,67 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+       <li><p>Let |extractSize| be the smaller value of |available| and |desiredSize|.
+
+       <li><p>Let <var>bytes</var> be the result of extracting |extractSize| of bytes from
+       <var>buffer</var>.
+
+       <li>If |stream|s [=ReadableStream/current BYOB request view=] is non-null, then
+       [=ArrayBufferView/write=] |bytes| into |stream|'s [=ReadableStream/current BYOB request
+       view=], and set |view| to |stream|'s [=ReadableStream/current BYOB request view=].
+       <li>Otherwise, set |view| to the result of [=ArrayBufferView/create|creating=] a
```
domenic, line 6001: Missing newline here

domenic (approved):
LGTM with nit

fetch.bs
saschanaz:
```suggestion

       <li>Otherwise, set |view| to the result of [=ArrayBufferView/create|creating=] a
```

fetch.bs
```diff
@@ -5961,25 +5961,68 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
   </ol>
 
- <li><p>Let <var>pullAlgorithm</var> be an algorithm that <a lt=resumed for=fetch>resumes</a> the
- ongoing fetch if it is <a lt=suspend for=fetch>suspended</a>.
+ <li>
+  <p>Let |buffer| be an empty buffer that can have bytes appended to it.
+
+  <p class="note">This represents an internal buffer inside the network layer of the user agent.
+
+ <li><p>Let |pullAlgorithm| be the followings steps:
```
annevk, line 5969: If the `<li>` has multiple children, the `<p>` needs to be on its own line. (Applies several times.)

fetch.bs
```diff
@@ -5961,25 +5961,68 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+     <li><p>[=Queue a fetch task=] to run the following steps, with |fetchParams|'s [=task destination=].
+
+      <ol>
+       <li><p>Let |available| be the size of |buffer|.
+
+       <li><p>Let |desiredSize| be |available|.
+
+       <li><p>If |stream|'s [=ReadableStream/current BYOB request view=] is non-null, then set
+       |desiredSize| to |stream|'s [=ReadableStream/current BYOB request view=]'s [=BufferSource/byte
+       length=].
```
annevk, line 5991: No newlines inside terms.

fetch.bs
```diff
@@ -5961,25 +5961,68 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+       <li>Otherwise, set |view| to the result of [=ArrayBufferView/create|creating=] a
+       {{Uint8Array}} from |bytes| in |stream|'s [=relevant Realm=].
+
+       <li><p>[=ReadableStream/Enqueue=] |view| into |stream|.
+
+       <li><p>If |stream| is [=ReadableStream/errored=], then [=fetch controller/terminate=]
+       |fetchParams|'s [=fetch params/controller=].
+
+       <li><p>Resolve |promise| with undefined.
```
annevk, line 6010: xref resolve

annevk (commented):
I'll trust @domenic got the logic right. I have a few editorial comments and then it seems this can go in.

Will you file an MDN issue as well if needed?

fetch.bs
saschanaz: Hmm, that's not what the Great Rewrapper does, but okay.

saschanaz:
The build failure doesn't seem relevant 🤔

saschanaz:
Filed https://github.com/mdn/content/issues/24453

annevk (approved):

annevk:
This is blocked on https://github.com/speced/bikeshed/issues/2471, but otherwise ready to go.

fetch.bs
annevk: If that's Dominic's tool that's a known issue I think. HTML has a different convention that cares less about inline search. I want inline search to be as straightforward as possible.

annevk:
@saschanaz can you file an MDN issue? Also, thanks for making this happen!

saschanaz:
Did it already! https://github.com/whatwg/fetch/pull/1593#issuecomment-1431145522

annevk:
Ah! So the reason I asked again is because I go by the checklist in OP as I easily forget. I updated that now to include that reference.

PR #1596, opened 2023-01-25
Editorial: add <div algorithm> wrappers to Header class

dlrobertson:
Related to: #1526

***
<a href="https://whatpr.org/fetch/1596.html" title="Last updated on Jan 26, 2023, 2:20 PM UTC (8424848)">Preview</a> | <a href="https://whatpr.org/fetch/1596/a5560d2...8424848.html" title="Last updated on Jan 26, 2023, 2:20 PM UTC (8424848)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6567,6 +6567,7 @@ new Headers(meta2);
 
 <hr>
 
+<div algorithm>
 <p>To <dfn for=Headers>validate</dfn> a  <a for=/>header</a> (<var>name</var>, <var>value</var>) for
 a {{Headers}} object <var>headers</var>:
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -6667,11 +6676,16 @@ constructor steps are:
...
 
+<div algorithm>
 <p>The <dfn export for=Headers method><code>append(<var>name</var>, <var>value</var>)</code></dfn>
 method steps are to <a for=Headers>append</a> (<var>name</var>, <var>value</var>) to <a>this</a>.
+</div>
 
-<p>The <dfn export for=Headers method><code>delete(<var>name</var>)</code></dfn> method steps are:
+<div algorithm>
+<p>The <dfn export for=Headers method><code>delete(<var>name</var>)</code></dfn> method
+from a {{Headers}} object (<var>headers</var>), run these steps:
```
annevk, line 6688: You found a bug, nice! The fix is replacing _headers_ below with `<a>this</a>`. This can stay as-is. I suppose the same applies to `set()` below.

dlrobertson: Ah! Good point. Originally I thought this would be like [remove privileged no-CORS request-headers](https://github.com/whatwg/fetch/pull/1596/files#diff-b4827ee42aceaf9dc3e7216f72262d81c32217e9996fb70a657f259e3ac92604L6647), but it makes sense that we'd use `this` for these methods.

dlrobertson (commented):
Thanks for the review! I'll try to update this evening.

dlrobertson:
Updated to use `this` in the methods.

annevk (approved):
\o/

PR #1638, opened 2023-04-23
Editorial: Pass recursive=false to manual navigation redirect

### LINKED ISSUE #1629: HTTP-redirect does not call processResponse and breaks navigation expectations
["create navigation params by fetching"](https://html.spec.whatwg.org/multipage/browsing-the-web.html#create-navigation-params-by-fetching) in HTML navigation spec does following:
> "6. Otherwise, process the next manual redirect for fetchController."

and has a note:
> "This will result in calling the processResponse we supplied above, during our first iteration through the loop, and thus setting response."

but if we will look into fetch spec that what will happen while processing next manual redirect:

1. Process manual redirect is set to [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#concept-http-redirect-fetch) (https://fetch.spec.whatwg.org/#concept-http-fetch 7.2, "manual" case).
2. Step 20 of [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#concept-http-redirect-fetch) calls [main fetch](https://fetch.spec.whatwg.org/#concept-main-fetch) with fetchParams and true (redirect param).
3. [main fetch](https://fetch.spec.whatwg.org/#concept-main-fetch) does early return of response on step 13:
> If recursive is true, then return response.

because "HTTP-redirect fetch" passed recursive=true earlier.

This way [main fetch](https://fetch.spec.whatwg.org/#concept-main-fetch) never reaches [fetch response handover](https://fetch.spec.whatwg.org/#fetch-finale) where processResponse is called which means navigation code will stuck waiting for processResponse to be called.

I checked spec commit history and in the past instead of
> Return the result of running main fetch given fetchParams and true.

there was in the [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#concept-http-redirect-fetch)
> Return the result of performing a main fetch using request with recursive flag set if request's redirect mode is not manual.

note that recursive should be false for manual redirect mode was introduced here https://github.com/whatwg/fetch/pull/499/commits/c958914d772b6b1089b3f361958cbad66b546b9c

and then got removed https://github.com/whatwg/fetch/commit/12dd6fa8ca76bd3bdff0c65a0c5a84b3ca870c3d

but for me it looks like this note should not have been removed because with redirect=false in manual redirect mode processResponse will be called which is what navigation expects.

noamr:
Closes #1629

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5352,8 +5352,20 @@ run these steps:
  <li><p>Invoke <a>set <var>request</var>'s referrer policy on redirect</a> on <var>request</var> and
  <var>actualResponse</var>. [[!REFERRER]]
 
+ <li><p>Let <var>recursive</var> be true.
+
  <li>
-  <p>Return the result of running <a>main fetch</a> given <var>fetchParams</var> and true.
+  <p>If <var>request</var>'s <a for=request>redirect mode</a> is "<code>manual</code>", then:
+
+  <ol>
+   <li><p>Assert: <var>request</var>'a <a for=request>mode</a> is "<code>navigate</code>".
+
+   <li><p>Set <var>recursive</var> to false.
+  </ol>
+
+ <li>
+  <p>Return the result of running <a>main fetch</a> given <var>fetchParams</var> and
+  <var>recursive</var>.
 
   <p class=note>This has to invoke <a>main fetch</a> to get <var>request</var>'s
   <a for=request>response tainting</a> correct.


fetch.bs
```diff
@@ -5352,8 +5352,20 @@ run these steps:
  <li><p>Invoke <a>set <var>request</var>'s referrer policy on redirect</a> on <var>request</var> and
  <var>actualResponse</var>. [[!REFERRER]]
 
+ <li><p>Let <var>recursive</var> be true.
+
+ <li>
+  <p>If <var>request</var>'s <a for=request>redirect mode</a> is "<code>manual</code>", then:
+
+  <ol>
+   <li><p>Assert: <var>request</var>'a <a for=request>mode</a> is "<code>navigate</code>".
```
annevk, line 5361: This is incorrect. You can always set it to "manual" and get an opaque response at the moment. Mainly useful in service workers with navigations though.

I suggest we do something like

> Let recursive be true if ....; otherwise false.

instead.

noamr: Yea but if it's manual non-navigation you wouldn't get to this bit as fetch would never process your redirect. Anyway, revised.

annevk: Wait, that's good pushback and solid rationale. Why revise?

noamr: Mainly because that assert is potentially more confusing than helpful and I wasn't sure about it. Should we keep it or throw it? I could go either way.

noamr: Done

annevk (approved):

Issue #1567, opened 2022-12-11
Rename "extract-value flag" to "extract-value" and define it as a boolean


dlrobertson:
We should rename the "extract-value flag" for [collect-an-http-quoted-string](https://fetch.spec.whatwg.org/#collect-an-http-quoted-string) to "extract-value" and define it as a boolean. This would likely require changes to MIME Sniffing and Server Timing as well.

_Originally posted by @annevk in https://github.com/whatwg/fetch/issues/1566#issuecomment-1345477692_

annevk:
I labeled this "good first issue", but note that it is a bit involved if you want to take this on. You'll start by creating a PR for this standard (no need to ask for assignment) and then you'll need to create two downstream PRs for the standards listed above so they use the corrected term.

SatyamKharote:
Hey! I want to work on this issue can you please assign me this issue?

annevk:
Hey @SatyamKharote a good first step to taking on an issue is to read it and the comments. See the comment above yours for instructions.

SatyamKharote:
Hello @annevk, First of all, thanks for replying. I have never done any contribution to open source. This is my first step towards the open-source contribution. So can you please elaborate on what should i have to do?

annevk:
No, sorry. Sufficient knowledge of GitHub and git is required.

redone9211:
i would like to contribute to this issue , can you let me know how i can solve it.

Deepak0320Singhal:
Hey I want to work on this issue, can you assign me this?

varunseth5788:
i would like to contribute to this issue.

hiiAnup:
Hey I want to work on this issue this will be my first open source contribution

PR #1641, opened 2023-04-25
Editorial: update args for collect an HTTP quoted string

dlrobertson:
Rename the "extract-value flag" argument to the algorithm "collect an HTTP quoted string" to a boolean "extract-value" that will by default be false.

Fixes: #1567

***
<a href="https://whatpr.org/fetch/1641.html" title="Last updated on May 5, 2023, 9:40 PM UTC (ba2d76c)">Preview</a> | <a href="https://whatpr.org/fetch/1641/ca10f49...ba2d76c.html" title="Last updated on May 5, 2023, 9:40 PM UTC (ba2d76c)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -446,7 +446,7 @@ preferred. Unlike <a>ASCII whitespace</a> this excludes U+000C FF.
 <p>To
 <dfn export lt="collect an HTTP quoted string|collecting an HTTP quoted string">collect an HTTP quoted string</dfn>
 from a <a for=/>string</a> <var>input</var>, given a <a>position variable</a> <var>position</var>
-and optionally an <var>extract-value flag</var>, run these steps:
+and an optional boolean <var>extract-value</var> (default false):
 
 <ol>
  <li><p>Let <var>positionStart</var> be <var>position</var>.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -446,7 +446,7 @@ preferred. Unlike <a>ASCII whitespace</a> this excludes U+000C FF.
 <p>To
 <dfn export lt="collect an HTTP quoted string|collecting an HTTP quoted string">collect an HTTP quoted string</dfn>
 from a <a for=/>string</a> <var>input</var>, given a <a>position variable</a> <var>position</var>
-and optionally an <var>extract-value flag</var>, run these steps:
+and a boolean <var>extract-value</var> (default false), run these steps:
```
annevk, line 449:
```suggestion
and an optional boolean <var>extract-value</var> (default false):
```

fetch.bs
```diff
@@ -495,7 +495,7 @@ and optionally an <var>extract-value flag</var>, run these steps:
     </ol>
   </ol>
 
- <li><p>If the <var>extract-value flag</var> is set, then return <var>value</var>.
+ <li><p>If the <var>extract-value</var> is true, then return <var>value</var>.
```
annevk, line 498:
```suggestion
 <li><p>If <var>extract-value</var> is true, then return <var>value</var>.
```

fetch.bs
```diff
@@ -506,7 +506,7 @@ and optionally an <var>extract-value flag</var>, run these steps:
   <tr>
    <th>Input
    <th>Output
-   <th>Output with the <var>extract-value flag</var> set
+   <th>Output with the <var>extract-value</var> set to true
```
annevk, line 509:
```suggestion
   <th>Output with <var>extract-value</var> set to true
```

annevk (commented):
Thanks for tackling this. Just a couple nits. And maybe we should link the downstream PRs from the commit message?

dlrobertson:
> Thanks for tackling this. Just a couple nits. And maybe we should link the downstream PRs from the commit message?

Good point! Updated.

PR #1614, opened 2023-03-05
Editorial: Reword how-to section to explain how to use callbacks & controller

noamr:
Instead of describing the callbacks one by one, putting emphasis on how/when the caller expects responses (upon completion, chunk-by-chunk, fire-and-forget).

This (hopefully) gives other specs who need to call into Fetch a quick reference between what they're trying to do and the API.

- [ ] At least two implementers are interested (and none opposed):
   * N/A
- [ ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * N/A
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * N/A
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …
   * N/A

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***
<a href="https://whatpr.org/fetch/1614.html" title="Last updated on May 11, 2023, 2:30 PM UTC (7e104e6)">Preview</a> | <a href="https://whatpr.org/fetch/1614/aaada1f...7e104e6.html" title="Last updated on May 11, 2023, 2:30 PM UTC (7e104e6)">Diff</a>

PR #1645, opened 2023-04-30
Use internalResponse at end of fetch handover

### LINKED ISSUE #1512: The end of fetch response handover needs to operate on the internal response

noamr:
This ensures that callers of `processResponseEndOfBody` and `processResponseConsumeBody` would receive the correct body and response without additional steps.

Closes #1512

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1645.html" title="Last updated on May 22, 2023, 9:28 AM UTC (788a467)">Preview</a> | <a href="https://whatpr.org/fetch/1645/625897d...788a467.html" title="Last updated on May 22, 2023, 9:28 AM UTC (788a467)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4810,7 +4810,11 @@ steps:
  <a for="fetch params">process response</a> given <var>response</var>, with <var>fetchParams</var>'s
  <a for="fetch params">task destination</a>.
 
- <li><p>If <var>response</var>'s <a for=response>body</a> is null, then run
+ <li><p>Let <var>actualResponse</var> be <var>response</var>'s
+ <a for="filtered response">internal response</a> if <var>response</var> is a
+ <a>filtered response</a>; Otherwise <var>response</var>.
+
+ <li><p>If <var>actualResponse</var>'s <a for=response>body</a> is null, then run
  <var>processResponseEndOfBody</var>.
 
  <li>
@@ -4828,8 +4832,8 @@ steps:
    <a for="TransformStream/set up"><i>flushAlgorithm</i></a> set to
    <var>processResponseEndOfBody</var>.
 
-   <li><p>Set <var>response</var>'s <a for=response>body</a>'s <a for=body>stream</a> to the result
-   of <var>response</var>'s <a for=response>body</a>'s <a for=body>stream</a>
+   <li><p>Set <var>actualResponse</var>'s <a for=response>body</a>'s <a for=body>stream</a> to the
+   result of <var>actualResponse</var>'s <a for=response>body</a>'s <a for=body>stream</a>
    <a for=ReadableStream>piped through</a> <var>transformStream</var>.
   </ol>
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4792,7 +4792,11 @@ steps:
  <a for="fetch params">process response</a> given <var>response</var>, with <var>fetchParams</var>'s
  <a for="fetch params">task destination</a>.
 
- <li><p>If <var>response</var>'s <a for=response>body</a> is null, then run
+ <li><p>Let <var>actualResponse</var> be <var>response</var>'s
+ <a for="filtered response">internal response</a> if <var>response</var> is a
+ <a>filtered response</a>; Otherwise <var>response</var>.
```
annevk, line 4797: 1. Since this is after main fetch it's always going to be a filtered response or a network error. Hence we should have a step equivalent to step 15 of main fetch. (Might want to update the wording though as this does if/otherwise better, except that otherwise needs to be lowercase.)
2. We should use the same variable name throughout the specification. We use both actualResponse and internalResponse. _internalResponse_ seems clearer to me.

annevk (commented):
This looks mostly okay, except for some editorial work. Also, we should update the warning (as per https://github.com/whatwg/fetch/issues/1512#issuecomment-1289065530) to indicate you might get bytes back from an opaque response and thus you have to tread carefully when not using CORS.

fetch.bs

fetch.bs
```diff
@@ -4792,7 +4792,11 @@ steps:
  <a for="fetch params">process response</a> given <var>response</var>, with <var>fetchParams</var>'s
  <a for="fetch params">task destination</a>.
 
- <li><p>If <var>response</var>'s <a for=response>body</a> is null, then run
+ <li><p>Let <var>internalResponse</var> be <var>response</var>, if <var>response</var> is a
+ <a>network error</a>; otherwise <var>response</var>'s
+ <a for="filtered response">internal response</a> otherwise.
```
annevk, line 4797: Drop "otherwise" at the end here.

Also not entirely sure we need the comma before "if". I think we usually leave it out if the sentences are relatively short.

noamr:
> This looks mostly okay, except for some editorial work. Also, we should update the warning (as per [#1512 (comment)](https://github.com/whatwg/fetch/issues/1512#issuecomment-1289065530)) to indicate you might get bytes back from an opaque response and thus you have to tread carefully when not using CORS.

Regarding the comment, I want to see where we're going with https://github.com/whatwg/fetch/pull/1614 before I create a merge conflict in the how-to-fetch section... WDYT?

fetch.bs
noamr: Done (oops)

annevk:
Fair, I've taken a pass at that PR and will try to do a pass at least daily until it's done.

annevk:
It has landed. This should now be safe to rebase and then you can add the relevant text.

noamr:
> It has landed. This should now be safe to rebase and then you can add the relevant text.

Thanks! On it

noamr:
> It has landed. This should now be safe to rebase and then you can add the relevant text.

Done, please take a look if this warning is what you had in mind.

PR #1646, opened 2023-05-02
Editorial: account for removal of AbortSignal's follow

annevk:
For https://github.com/whatwg/dom/pull/1152.

---

This is incomplete. https://fetch.spec.whatwg.org/#dom-request-clone needs to be updated as well. While we could probably leave https://fetch.spec.whatwg.org/#request-create (also called by Service Workers) alone, that doesn't seem entirely right as we'd be creating a redundant `AbortSignal` object.

Would the best solution here be that "create" takes a signal(s?) argument and then uses that to create a new composite signal for itself?

It doesn't have to do anything for service workers it seems, but if service workers were to pass an empty list it'd do the same thing it does today.

cc @jakearchibald @shaseley

***
<a href="https://whatpr.org/fetch/1646.html" title="Last updated on May 17, 2023, 12:54 PM UTC (a11ffe2)">Preview</a> | <a href="https://whatpr.org/fetch/1646/018ac19...a11ffe2.html" title="Last updated on May 17, 2023, 12:54 PM UTC (a11ffe2)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7542,8 +7542,8 @@ object), initially null.
 
 <div algorithm>
 <p>To <dfn export for=Request lt=create|creating>create</dfn> a {{Request}} object, given a
-<a for=/>request</a> <var>request</var>, <a for=/>headers guard</a> <var>guard</var>, and
-<a>realm</a> <var>realm</var>, run these steps:
+<a for=/>request</a> <var>request</var>, <a for=/>headers guard</a> <var>guard</var>,
+{{AbortSignal}} object <var>signal</var>, and <a>realm</a> <var>realm</var>:
 
 <ol>
  <li><p>Let <var>requestObject</var> be a <a for=/>new</a> {{Request}} object with <var>realm</var>.
@@ -7554,8 +7554,7 @@ object), initially null.
  object with <var>realm</var>, whose <a for=Headers>headers list</a> is <var>request</var>'s
  <a for=request>headers list</a> and <a for=Headers>guard</a> is <var>guard</var>.
 
- <li><p>Set <var>requestObject</var>'s <a for=Request>signal</a> to a <a for=/>new</a>
- {{AbortSignal}} object with <var>realm</var>.
+ <li><p>Set <var>requestObject</var>'s <a for=Request>signal</a> to <var>signal</var>.
 
  <li><p>Return <var>requestObject</var>.
 </ol>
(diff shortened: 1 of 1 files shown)


shaseley:
Thanks for drafting this! I started hacking on this yesterday too, and was leaning towards having "create" take a signal and setting the Request's signal to that, with SW passing a new signal and fetch passing a composite signal.

The reason for doing that vs. always creating a composite signal is because SW invokes "signal abort" directly on the signal, acting as its own (abort) controller. Having a composite signal only be aborted by its sources is a useful property (and IMO expected), e.g. it can be GCed when all its sources are gone; having the signal itself be a source complicates that.

Aside: It might be nice if "signal abort" was private and always done through a controller, separating the control/signal functionality like we do in userland. Specs that need to trigger an abort (SW, streams, maybe navigation API) could hold onto a controller, although it adds a bit of complexity. (Implementation-wise, I've annotated these signals as "internal" and allow/expect them to act like a controller/signal pair).

annevk:
@shaseley this should be good for review now. Can I ask you to create the corresponding SW PR? See also the discussion in the Streams PR. I'm not opposed to making "signal abort" private (as the current DOM PR seems to be doing), but we'll need even more downstream PRs to pull that off.

fetch.bs
```diff
@@ -7803,11 +7802,12 @@ constructor steps are:
...
 
- <li><p>Set <a>this</a>'s <a for=Request>signal</a> to a <a for=/>new</a> {{AbortSignal}} object
- with <a>this</a>'s <a>relevant Realm</a>.
+ <li><p>Let <var>signals</var> be « <var>signal</var> » if <var>signal</var> is non-null; otherwise
+ « ».
 
- <li><p>If <var>signal</var> is non-null, then make <a>this</a>'s <a for=Request>signal</a>
- <a for=AbortSignal>follow</a> <var>signal</var>.
+ <li><p>Set <a>this</a>'s <a for=Request>signal</a> to the result of
+ <a>creating a new composite abort signal</a> from <var>signals</var>, using {{AbortSignal}} and
```
shaseley, line 7809: "creating a composite abort signal" (or we can rename the algo to add "new"). I left off "new" since I thought it was implied by "create". Same below.

shaseley:
Looks great % linking and possible rename, depending on the outcome of https://github.com/whatwg/dom/pull/1152#discussion_r1184180284.

> @shaseley this should be good for review now. Can I ask you to create the corresponding SW PR? See also the discussion in the Streams PR. I'm not opposed to making "signal abort" private (as the current DOM PR seems to be doing), but we'll need even more downstream PRs to pull that off.

Yep, happy to update SW. And I flipped "signal abort" back to exported -- I'll work on that as a follow-up.

shaseley:
I was implementing this and noticed the change is observable via event dispatch order ([wpt failure](https://github.com/web-platform-tests/wpt/blob/master/fetch/api/abort/general.any.js#L569)).  See https://github.com/whatwg/dom/pull/1152#issuecomment-1537292981.

PR #1650, opened 2023-05-04
Fully read's processBodyError argument needs to take an exception

### LINKED ISSUE #1636: "consume body" step disagrees with "fully read" on whether the error callback accepts an argument
I've noticed what I think is an inconsistency while implementing part of Fetch.

First, the definition for ["fully read a body"](https://fetch.spec.whatwg.org/commit-snapshots/8f109835dcff90d19caed4b551a0da32d9d0f57e/#ref-for-byte-sequence⑨) says:

> processBodyError must be an algorithm accepting no arguments\. 

However, ["consume a body"](https://fetch.spec.whatwg.org/commit-snapshots/8f109835dcff90d19caed4b551a0da32d9d0f57e/#concept-body-package-data
) then passes an algorithm that accepts an error argument as `processBodyError`, here named `errorSteps`:

> 3. Let errorSteps given error be to reject promise with error\.

> 6. Otherwise, fully read object’s body given successSteps, errorSteps, and object’s relevant global object\. 

Other places that call "fully read a body" that I've come across all pass it an algorithm that does not accept arguments.

We had previously implemented `processBodyError` as taking a single argument, and then the algorithms for it just ignore that argument, but this is problematic when that code also needs to be run separately when no argument is provided. (eg, ["main fetch" step 22](https://fetch.spec.whatwg.org/commit-snapshots/8f109835dcff90d19caed4b551a0da32d9d0f57e/#ref-for-concept-request-integrity-metadata①) both calls `processBodyError` without arguments in step 22.2 and passes it to "fully read a body" in step 22.4.)

annevk:
Fixes #1636.

***
<a href="https://whatpr.org/fetch/1650.html" title="Last updated on May 4, 2023, 8:08 AM UTC (47df1a3)">Preview</a> | <a href="https://whatpr.org/fetch/1650/4502481...47df1a3.html" title="Last updated on May 4, 2023, 8:08 AM UTC (47df1a3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1510,7 +1510,8 @@ must be an algorithm accepting an exception.
 <var>processBody</var>, an algorithm <var>processBodyError</var>, and an optional null,
 <a for=/>parallel queue</a>, or <a for=/>global object</a> <var>taskDestination</var> (default
 null), run these steps. <var>processBody</var> must be an algorithm accepting a
-<a for=/>byte sequence</a>. <var>processBodyError</var> must be an algorithm accepting no arguments.
+<a for=/>byte sequence</a>. <var>processBodyError</var> must be an algorithm optionally accepting an
+<a for=/>exception</a>.
 
 <ol>
  <li><p>If <var>taskDestination</var> is null, then set <var>taskDestination</var> to the result of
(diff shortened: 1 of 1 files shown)


AtkinsSJ (commented):
Looks good to me! We still have some gaps in our Fetch implementation, but as far as I can tell this solves the issue.

Issue #1630, opened 2023-04-12
Clarify priority between blob.type and Content-Type in headers for request/response.blob()


CYBAI:
Currently, the spec of [`blob()`](https://fetch.spec.whatwg.org/#dom-body-blob) says

> return a [Blob](https://w3c.github.io/FileAPI/#dfn-Blob) whose contents are bytes and whose [type](https://w3c.github.io/FileAPI/#dfn-type) attribute is [this](https://webidl.spec.whatwg.org/#this)’s [MIME type](https://fetch.spec.whatwg.org/#concept-body-mime-type).

however, it doesn't mention when the body is a Blob with type, browsers should respect the blob's type or the Content-Type in headers.

there's a test in [Chromium](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/http/tests/fetch/script-tests/request.js;l=1021-1032;drc=05c5ed4e58b4cc7d50d0018f7b4d2cdc17211389) and [WebKit](https://github.com/WebKit/WebKit/blob/512db9a6900b400f9ce93e90e6e178fa74fef03e/LayoutTests/http/wpt/fetch/fetch-as-blob.js#L62-L73).

```js
 promise_test(function(t) { 
     var req = new Request('http://localhost/', 
                           {method: 'POST', 
                            body: new Blob([''], {type: 'Text/Plain'})}); 
     req.headers.set('Content-Type', 'Text/Html'); 
     return req.blob() 
       .then(function(blob) { 
           assert_equals(blob.type, 'text/plain'); 
           assert_equals(req.headers.get('Content-Type'), 'Text/Html'); 
         }); 
   }, 
   'MIME type unchanged if headers are modified after Request() constructor'); 
```

Chromium and WebKit are respecting blob's type so they could pass the test now. For Gecko, it fails to run the test because it would respect the headers' content-type.

With confirming with @annevk in https://github.com/WebKit/WebKit/pull/12376#issuecomment-1503175231, it seems the `blob()` should always respect the header's content-type as "this’s MIME type" wins. So, I wonder maybe we would need some notes (or algorithms in the `blob()`) to clarify more about blob's type when it exists in the spec?

Finally, if we should respect headers' Content-Type, I will fix the test to

```diff
-           assert_equals(blob.type, 'text/plain'); 
+           assert_equals(blob.type, 'text/html'); 
```

[...]

PR #1657, opened 2023-05-18
Editorial: clarify how to get MIME type for request and response

CYBAI:
Fix https://github.com/whatwg/fetch/issues/1630

---

When the `Request` or `Response` has a `Blob` body with `type` and MIME type exists in its headers, browsers should always return the MIME type in headers instead of the body's type.

However, previously, the spec didn't mention this clearly so this PR will help to clarify by introducing an algorithm of `get the MIME type`.

   * https://github.com/web-platform-tests/wpt/pull/39489

***
<a href="https://whatpr.org/fetch/1657.html" title="Last updated on Jun 2, 2023, 9:41 AM UTC (23a1f62)">Preview</a> | <a href="https://whatpr.org/fetch/1657/b5a587b...23a1f62.html" title="Last updated on Jun 2, 2023, 9:41 AM UTC (23a1f62)">Diff</a>


fetch.bs
```diff
@@ -7130,8 +7130,8 @@ due course.
 <!-- https://lists.w3.org/Archives/Public/public-whatwg-archive/2014Jun/thread.html#msg72 -->
 
 <p>Objects including the {{Body}} interface mixin need to define an associated
-<dfn id=concept-body-mime-type for=Body>MIME type</dfn> algorithm which takes no arguments and
-returns failure or a <a for=/>MIME type</a>.
+<dfn id=concept-body-mime-type for=Body>MIME type</dfn> algorithm which returns the result of
+<a for=Body>get the MIME type</a> with <a>this</a>.
```
annevk, line 7134: I think we should remove this, no? Callers can use "get the MIME type" directly.

fetch.bs
```diff
@@ -7168,6 +7168,24 @@ returns failure or a <a for=/>MIME type</a>.
...
+ <li><p>If <var>requestOrResponse</var>'s has an associated <a for=Headers>header list</a>,
+ <var>headers</var>, and it's non-null, then:
+ <ol>
+   <li><p>Let <var>mimeTypeInHeaders</var> be the result of <a for="header list">extracting a
+   MIME type</a> from <var>headers</var>.
+   <li><p>If <var>mimeTypeInHeaders</var> is non-null, then return <var>mimeTypeInHeaders</var>.
+ </ol>
+ <li><p>Let <var>bodyWithType</var> be the result of <a for=BodyInit>extracting</a> <a>this</a>'s
+ <a for=Body>body</a>.
+ <li><p>Return <var>bodyWithType</var>'s <a for="body with type">type</a>.
```
annevk, line 7185: This shouldn't be needed. This already happens as part of constructing a Request or Response, no?

fetch.bs
```diff
@@ -7224,22 +7242,27 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step
...
 running <a for=Body>consume body</a> with <a>this</a> and the following step given a
-<a for=/>byte sequence</a> <var>bytes</var>: switch on <a>this</a>'s <a for=Body>MIME type</a>'s
-<a for="MIME type">essence</a> and run the corresponding steps:
+<a for=/>byte sequence</a> <var>bytes</var>:
+
+<ol>
+<li><p>Let <var>mimeType</var> be the result of <a for=Body>get the MIME type</a> with <a>this</a>.
+<li>
+  <p>Switch on <var>mimeType</var>'s <a for="MIME type">essence</a> and run the corresponding
+  steps:
```
annevk, line 7257: It looks like the text here was wrong before as MIME type can be null. So we probably need to pull the "Otherwise" branch forward and deal with that here.

I guess it's fair if you want to leave that as a follow-up as you didn't introduce it.

fetch.bs
```diff
@@ -7168,6 +7168,24 @@ returns failure or a <a for=/>MIME type</a>.
 
 <hr>
 
+<div algorithm>
+<p>To <dfn id=concept-body-get-mime-type for=Body>get the MIME type</dfn>, given an object, <var>requestOrResponse</var>, including the
+{{Body}} interface mixin, run these steps:
+
+<ol>
+ <li><p>If <var>requestOrResponse</var>'s has an associated <a for=Headers>header list</a>,
```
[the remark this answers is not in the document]
CYBAI: Thanks for helping me clarify the algorithm of getting headers here.

fetch.bs
CYBAI: Indeed! You mean, while given a Request or Response object, it must already run through the "extract" algorithm (e.g. Step 37-1 in https://fetch.spec.whatwg.org/#dom-request) and set body's type back to its headers so we don't need to "extract" again. Is my understanding correct?

annevk: Yes, `Content-Type` will already be in the header list (or not).

fetch.bs
CYBAI: As this patch aims to clarify around MIME type here, it sounds good to me to handle this along with this PR! Thanks for catching it!

> 1. If _mimeType_ is null, [Throw](https://webidl.spec.whatwg.org/#dfn-throw) a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).
> 2. Otherwise, switch on [....]

Does this ☝️ sound good to you?

fetch.bs
```diff
@@ -7178,6 +7174,22 @@ returns failure or a <a for=/>MIME type</a>.
 
 <hr>
 
+<div algorithm>
+<p>To <dfn id=concept-body-get-mime-type for=Body>get the MIME type</dfn>, given a {{Request}}
+or {{Response}} object, <var>requestOrResponse</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>headers</var> be null
```
annevk, line 7182:
```suggestion
 <li><p>Let <var>headers</var> be null.
```
Also, typically `<li>`s are separated by a newline.

fetch.bs
```diff
@@ -7178,6 +7174,22 @@ returns failure or a <a for=/>MIME type</a>.
...
+ <li><p>If <var>requestOrResponse</var> is a {{Request}} object, then set <var>headers</var> to
+ <var>requestOrResponse</var>'s <a for=Request>request</a>'s {{Request/headers}}.
```
annevk, line 7184: The last one here needs to be `<a for=request>header list</a>`. Similar error for response below.

fetch.bs
```diff
@@ -7178,6 +7174,22 @@ returns failure or a <a for=/>MIME type</a>.
...
+ <li><p>Otherwise, set <var>headers</var> to <var>requestOrResponse</var>'s <a for=Response>
+ response</a>'s {{Response/headers}}.
+ <li><p>Let <var>mimeTypeInHeaders</var> be the result of <a for="header list">extracting a
+ MIME type</a> from <var>headers</var>.
```
annevk, line 7188: No newlines in phrasing-level elements to make inline search not require a regex.

fetch.bs
```diff
@@ -7234,22 +7246,28 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step
 <p>The <dfn method for=Body><code>blob()</code></dfn> method steps are to return the result
 of running <a for=Body>consume body</a> with <a>this</a> and the following step given a
 <a for=/>byte sequence</a> <var>bytes</var>: return a {{Blob}} whose contents are <var>bytes</var>
-and whose {{Blob/type}} attribute is <a>this</a>'s <a for=Body>MIME type</a>.
+and whose {{Blob/type}} attribute is to <a for=Body>get the MIME type</a> with <a>this</a>.
```
annevk, line 7249: Please add something like:
`<!-- This is wrong. It should really set an underlying concept to this. -->`
It's an existing issue and would be quite involved to fix, but we should acknowledge it.

fetch.bs
```diff
@@ -7234,22 +7246,28 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step
...
-<a for="MIME type">essence</a> and run the corresponding steps:
+<a for=/>byte sequence</a> <var>bytes</var>:
 
-<dl class=switch>
+<ol>
+<li><p>Let <var>mimeType</var> be the result of <a for=Body>get the MIME type</a> with <a>this</a>.
+<li><p>If <var>mimeType</var> is null, then <a>Throw</a> a {{TypeError}}.
+<li>
+  <p>Otherwise, switch on <var>mimeType</var>'s <a for="MIME type">essence</a> and run the
+  corresponding steps:
```
annevk, line 7262: Here's what I think I would do

| If mimeType is non-null, then switch on _mimeType_'s essence and run the ...

and then you remove the "Otherwise" from the list and instead end with a "Throw" step.

The one unfortunate thing about all this is that you'll have to adjust indentation, which might require rewrapping due to 100 columns. If that seems like too much of a challenge I'm happy to do that for you though near the end.

annevk (commented):
Pretty close now!

fetch.bs
CYBAI: Understood! Thanks for catching it. Is following style acceptable?

```html
<li><p>Let <var>mimeTypeInHeaders</var> be the result of <a for="header list">
 extracting a MIME type</a> from <var>headers</var>.
```

annevk: No, the whole element needs to go on on the new line (otherwise you introduce whitespace between the start tag and its contents).

CYBAI: ah, I see. Thank you! I will move the whole element to next line 🙏

CYBAI:
I've tried to address those review comments and rewrap the switch cases for `formData()` in the fixup commit. If the PR is okay, I can squash the fixup commit before merge 🙏

fetch.bs
```diff
@@ -7178,6 +7174,26 @@ returns failure or a <a for=/>MIME type</a>.
 
 <hr>
 
+<div algorithm>
+<p>To <dfn id=concept-body-get-mime-type for=Body>get the MIME type</dfn>, given a {{Request}}
+or {{Response}} object, <var>requestOrResponse</var>, run these steps:
```
annevk, line 7179:
```suggestion
or {{Response}} object <var>requestOrResponse</var>:
```
(no need for a comma before the variable name and no need for "run these steps" when we start with "To")

fetch.bs
```diff
@@ -7234,69 +7250,75 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step
...
-  </ol>
+<ol>
+ <li><p>Let <var>mimeType</var> be the result of <a for=Body>get the MIME type</a>
+ with <a>this</a>.
 
-  <p class=XXX>The above is a rough approximation of what is needed for
-  `<code>multipart/form-data</code>`, a more detailed parsing specification is to be
-  written. Volunteers welcome.
+ <li><p>If <var>mimeType</var> is non-null, then switch on <var>mimeType</var>'s
+ <a for="MIME type">essence</a> and run the corresponding steps:
```
annevk, line 7267: Since this `<li>` contains both the `<p>` and the `<dl>`, the `<p>` needs to be on a newline and indented once. The `<dl>` needs to be indented one more as well therefore.

Wording-wise this looks great though!

annevk (commented):
Great, only nits remain. \o/

You don't have to squash as I'll do that through the GitHub UI. If you could work out a tentative commit message that would help though.

And perhaps we should no longer mark this as Editorial as you've fixed a null-pointer dereference.

annevk:
Finally, I suspect you're not in the Acknowledgments section yet. Feel free to add yourself!

fetch.bs
```diff
@@ -7234,69 +7250,75 @@ of running <a for=Body>consume body</a> with <a>this</a> and the following step
...
- <dt>Otherwise
- <dd><p><a>Throw</a> a {{TypeError}}.
-</dl>
+     <li><p>If <var>entries</var> is failure, then <a>throw</a> a {{TypeError}}.
+
+     <li><p>Return a new {{FormData}} object whose <a for=FormData>entry list</a> is
+     <var>entries</var>.
+    </ol>
+ </dl>
+ <li><p>Otherwise, <a>Throw</a> a {{TypeError}}.
```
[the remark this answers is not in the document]
CYBAI: Do you mean we can just state "Throw a TypeError" in this case?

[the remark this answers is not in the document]
CYBAI: I see. Thanks! I will update the PR :)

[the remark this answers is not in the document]
CYBAI: ah, yes, it's much clearer to me why the "Otherwise," is not needed here now 🙏 thank you!

CYBAI:
I've tried to reword the commit message and add more description in it. If it looks good, I can rename the PR title as well 🙏

annevk:
Thanks @CYBAI for tackling this! I noticed a few more things at the end that I cleaned up before merging. Hope that's okay.

CYBAI:
Sure! Thank you so much for the cleanup! I just learned more about how to write the spec :)

PR #1660, opened 2023-05-22
Editorial: define return types of most header list algorithms

### LINKED ISSUE #1659: Header list "get" return type is unclear
https://fetch.spec.whatwg.org/commit-snapshots/086e35f3d2b2fd1961d648f2df9b4a350fff707f/#terminology-headers

> Return the values of all headers in list whose name is a byte\-case\-insensitive match for name, separated from each other by 0x2C 0x20, in order\. 

I think this is attempting to concatenate everything into a single byte sequence? But at first I read it as returning a list. Maybe using the return-value-stating form of https://infra.spec.whatwg.org/#algorithm-declaration would help.

annevk:
Fixes #1659.

***
<a href="https://whatpr.org/fetch/1660.html" title="Last updated on May 22, 2023, 7:16 AM UTC (230c053)">Preview</a> | <a href="https://whatpr.org/fetch/1660/086e35f...230c053.html" title="Last updated on May 22, 2023, 7:16 AM UTC (230c053)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -616,8 +616,7 @@ given a <a for=/>header name</a> <var>name</var> and a string <var>type</var> fr
 <p>To
 <dfn export for="header list" id=concept-header-list-set-structured-header>set a structured field value</dfn>
 given a <a for=/>tuple</a> (<a for=/>header name</a> <var>name</var>, <a>structured field value</a>
-<var>structuredValue</var>), in a <a for=/>header list</a> <var>list</var>, run these steps. They
-return nothing.
+<var>structuredValue</var>), in a <a for=/>header list</a> <var>list</var>:
 
 <ol>
  <li><p>Let <var>serializedValue</var> be the result of executing the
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -615,7 +616,8 @@ given a <a for=/>header name</a> <var>name</var> and a string <var>type</var> fr
 <p>To
 <dfn export for="header list" id=concept-header-list-set-structured-header>set a structured field value</dfn>
 given a <a for=/>tuple</a> (<a for=/>header name</a> <var>name</var>, <a>structured field value</a>
-<var>structuredValue</var>), in a <a for=/>header list</a> <var>list</var>, run these steps:
+<var>structuredValue</var>), in a <a for=/>header list</a> <var>list</var>, run these steps. They
+return nothing.
```
domenic, line 620: Not necessary for the setter-type algorithms IMO.

domenic (approved):
LGTM but I think "return nothing" can be assumed in some cases.

fetch.bs
annevk: Fair enough.

PR #1661, opened 2023-05-25
Be more selective with handing out the internal response

annevk:
Upon reviewing #1645 again it struck me that the first argument to processResponseConsumeBody should not be the internal response. The caller would still have all the same capabilities when it's not and it's safer if it's not.

***
<a href="https://whatpr.org/fetch/1661.html" title="Last updated on May 25, 2023, 4:44 PM UTC (4189234)">Preview</a> | <a href="https://whatpr.org/fetch/1661/9003266...4189234.html" title="Last updated on May 25, 2023, 4:44 PM UTC (4189234)">Diff</a>


noamr (approved):
LGTM

noamr (approved):
LGTM

Issue #1571, opened 2022-12-14
Missing tags in create a connection


dlrobertson:
[create a connection](https://fetch.spec.whatwg.org/#create-a-connection) might be missing some tags for some steps. The steps

> If http3Only is true, then establish an HTTP/3 connection
> If credentials is false, then do not send a TLS client certificate. 
> ...

Seem to be missing tags.

dlrobertson:
> It kinda is meant to read as one big step, all impacting the connection establishment.

Would it make sense to use an unordered list for these like we do for the next step? If not, I think this issue can be closed.

annevk:
I like that idea!

PR #1678, opened 2023-06-23
Editorial: Add tags to create a connection alg

dlrobertson:
Add unordered list tags to substeps of the create a connection algorithm to improve readability.

Fixes: #1571


annevk:
I think we need a little more. Something to tie the list to the paragraph preceding it. Rewording the last sentence to make it clear the list is also to be taken into account would go a long way I think.

dlrobertson:
Sorry for the delay in updating this!

> Something to tie the list to the paragraph preceding it.

Added what the following step uses. If you feel like a more substantial rewording is needed, let me know and I'll make another attempt.

PR #1736, opened 2024-02-12
Editorial: Export fetch controller/serialized abort reason

mkruisselbrink:
This is referenced [1] by the Service Worker spec, so should either be exported or something else to achieve the same thing would need to exist (forwarding the "abort reason" from a fetch controller to an AbortController; note that the service worker algorithm is currently extra confusing because it uses the same variable name for both "controller" objects).

[1]: https://w3c.github.io/ServiceWorker/#clients-openwindow:~:text=If%20controller%E2%80%99s%20serialized%20abort%20reason%20is%20non%2Dnull

***
<a href="https://whatpr.org/fetch/1736.html" title="Last updated on Feb 12, 2024, 6:54 PM UTC (d5affe7)">Preview</a> | <a href="https://whatpr.org/fetch/1736/8dd73db...d5affe7.html" title="Last updated on Feb 12, 2024, 6:54 PM UTC (d5affe7)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -236,7 +236,7 @@ lt="authentication entry">authentication entries</a> (for HTTP authentication). 
  <dt><dfn for="fetch controller">report timing steps</dfn> (default null)
  <dd>Null or an algorithm accepting a <a for=/>global object</a>.
 
- <dt><dfn for="fetch controller">serialized abort reason</dfn> (default null)
+ <dt><dfn for="fetch controller" export>serialized abort reason</dfn> (default null)
  <dd>Null or a <a>Record</a> (result of [$StructuredSerialize$]).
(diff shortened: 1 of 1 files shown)


annevk:
@noamr @shaseley could you both look at this perhaps?

shaseley:
Exporting that looks reasonable. (Aside: it looks like "deserialize a serialized abort reason" could be used there too for consistency w/ fetch? It's already used a bit earlier in handle fetch.)

> note that the service worker algorithm is currently extra confusing because it uses the same variable name for both "controller" objects).

Ugh I think I added that second one :(. Happy to fix that if you're already doing so.

noamr (approved):

PR #1737, opened 2024-02-16
Support ServiceWorker Static Routing API

sisidovski:
We propose [ServiceWorker static routing API](https://github.com/WICG/service-worker-static-routing-api), which allows developers to configure routing, and offload things ServiceWorkers do. The main spec update is in https://github.com/w3c/ServiceWorker/pull/1701, but some behavior involves the fetch spec change (specifically, "race-network-and-fetch-handler" option).

   * Chrome is proposing this change https://groups.google.com/a/chromium.org/g/blink-dev/c/-SEMMttEqz4
   * Mozilla https://github.com/mozilla/standards-positions/pull/894
   * https://github.com/web-platform-tests/wpt/tree/master/service-workers/service-worker/tentative/static-router
   * Chromium: [crbug.com/40241479](https://crbug.com/40241479)
   * Gecko: …
   * WebKit: …

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1724,10 +1724,6 @@ is "<code>all</code>" or "<code>none</code>". Unless stated otherwise it is "<co
  </dl>
 </div>
 
-<p>A <a for=/>request</a> has an associated <dfn for=request export>service-workers race token</dfn> (a string). Unless stated otherwise it is null.
-
-<p class=note>This is only set by <a for=/>handle fetch</a> in service workers.
-
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-initiator>initiator</dfn>, which is
 the empty string,
@@ -4510,6 +4506,9 @@ steps:
 
  <li><p>If <var>recursive</var> is false, then run the remaining steps <a>in parallel</a>.
 
+ <li><p>Let |raceReesponse| be the result of [=lookup-race-response=].
+ <li><p>If |raceReesponse| is <var>response</var>, return |raceReesponse|.
+
  <li>
   <p>If <var>response</var> is null, then set <var>response</var> to the result of running the steps
   corresponding to the first matching statement:
@@ -4525,22 +4524,6 @@ steps:
      <a for="fetch params">preloaded response candidate</a> is a <a for=/>response</a>.
 
      <li><p>Return <var>fetchParams</var>'s <a for="fetch params">preloaded response candidate</a>.
-    </ol>
-
-   <dt><var>request</var>'s <a for=request>service-workers race token</a> is not null
-   <dd>
-    <ol>
-     <li><p>Let <var>raceReesponse</var> to the result of [=lookup-race-response=] given <var>request</var>'s <a for=request>service-workers race token</a>, <var>request</var>'s <a for=request>reserved client</a>, <var>request</var>'s <a for=request>URL</a>.
-     <li><p>If <var>raceReesponse</var> is non-null, then:
-      <ol>
-       <li><p>Wait until <var>raceReesponse</var> settles.
-       <li><p>If <var>raceReesponse</var> resolves with <var>response</var>, then:
-        <ol>
-         <li><p><a for=/>Assert</a>: <var>response</var> is a <a for=/>response</a>.
-         <li><p>Return <var>response</var>.
-        </ol>
-       <li><p>If <var>raceReesponse</var> rejects, return a <a>network error</a>.
-      </ol>
     </ol>
 
    <dt><var>request</var>'s <a for=request>current URL</a>'s <a for=url>origin</a> is
--- a/fetch.bs
+++ b/fetch.bs
@@ -1724,6 +1724,10 @@ is "<code>all</code>" or "<code>none</code>". Unless stated otherwise it is "<co
  </dl>
 </div>
 
+<p>A <a for=/>request</a> has an associated <dfn for=request export>service-workers race token</dfn> (a string). Unless stated otherwise it is null.
+
+<p class=note>This is only set by <a for=/>handle fetch</a> in service workers.
+
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-initiator>initiator</dfn>, which is
 the empty string,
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -4506,6 +4506,9 @@ steps:
 
  <li><p>If <var>recursive</var> is false, then run the remaining steps <a>in parallel</a>.
 
+ <li><p>Let |raceResponse| be the result of <a for=/>lookup race response</a> given <var>request</var>.
+ <li><p>If |raceResponse| is <var>response</var>, return |raceResponse|.
```
domenic, line 4510: _response_ is null at this point, so I doubt this check is what you mean.

sisidovski: I meant |response| is a [response](https://fetch.spec.whatwg.org/#concept-response). Updated.

fetch.bs
```diff
@@ -4506,6 +4506,9 @@ steps:
 
  <li><p>If <var>recursive</var> is false, then run the remaining steps <a>in parallel</a>.
 
+ <li><p>Let |raceResponse| be the result of <a for=/>lookup race response</a> given <var>request</var>.
+ <li><p>If |raceResponse| is a <a for=/>response</a>, return |raceResponse|.
```
domenic, line 4510: Nit: blank line between `<li>`s

domenic (approved):
Non-editor LGTM

fetch.bs
sisidovski: Done.

fetch.bs
```diff
@@ -4506,6 +4506,10 @@ steps:
 
  <li><p>If <var>recursive</var> is false, then run the remaining steps <a>in parallel</a>.
 
+ <li><p>Let |raceResponse| be the result of <a for=/>lookup race response</a> given <var>request</var>.
```
annevk, line 4509: As far as I can tell https://w3c.github.io/ServiceWorker/#lookup-race-response-algorithm ends up waiting, which would delay this algorithm. Is that really the intent?

annevk (commented):
Wouldn't this return a response without any filtering getting applied? I guess the idea is that these are already filtered responses? Can that be asserted?

fetch.bs
sisidovski: Yes, but actually this wouldn't delay the algorithm. When |raceResponse| is not null, the corresponding request is at least in-flight status. The map in "Lookup Race Response" is a map only contains in-flight requests/responses. At this point in 4.1. Main fetch, the actual fetch is not performed yet, we expect waiting for the in-flight request is faster in most cases.

sisidovski:
>  Wouldn't this return a response without any filtering getting applied? I guess the idea is that these are already filtered responses? Can that be asserted?

@annevk Sorry to be the slow response. This response is the result of "fetch" in [this step](https://w3c.github.io/ServiceWorker/#lookup-race-response-algorithm:~:text=Set%20raceFetchController%20to%20the%20result%20of%20calling%20fetch%20given%20request%2C%20with%20processResponse%20set%20to%20the%20following%20steps%20given%20a%20response%20raceNetworkRequestResponse%3A) in the service worker spec. So these should be filtered appropriately.

annevk:
Okay, I think that makes sense. Still have a question about the Service Worker side of things though:

> If raceNetworkRequestResponse’s [status](https://fetch.spec.whatwg.org/#concept-response-status) is [ok status](https://fetch.spec.whatwg.org/#ok-status), then:

How does this work for "no-cors" fetches?

cc @asutherland

sisidovski:
So, ok status will be filtered for "no-cors" fetches?

> If raceNetworkRequestResponse’s [status](https://fetch.spec.whatwg.org/#concept-response-status) is [ok status](https://fetch.spec.whatwg.org/#ok-status), then:

What about updating this paragraph to refer [internal response](https://fetch.spec.whatwg.org/#concept-internal-response)?  Not sure referring the ok status in internal response is concerned for information leaks, but there're a couple of lines referring internal response in the ServiceWorker spec.

annevk:
Yes, as per https://fetch.spec.whatwg.org/#concept-filtered-response-opaque status will be 0. It does seem very concerning this would tell you for every subresource whether it has an ok status or not. That's not information we reveal for images today, for instance.

sisidovski:
@annevk Sorry to be late, I had a time to think about this.

Yes, we should respect the given request mode and filtered response, and we shouldn't look at [internal response](https://fetch.spec.whatwg.org/#concept-internal-response). As the ServiceWorker sides wants to tell if the response status is OK or not so that it can return the response from the cache if the network request failed, I think it makes sense filter out the case when the request mode is "no-cors" in the ServiceWorker side. Specifically, we'll add the condition "request mode is not no-cors" to the step 13.4 in [handle fetch](https://w3c.github.io/ServiceWorker/#handle-fetch). So "race-network-and-fetch-handler" doesn't work for "no-cors" any more, so we don't store the opaque filtered response to the [race response map](https://w3c.github.io/ServiceWorker/#serviceworkerglobalscope-race-response-map).

Does this address your concern?

sisidovski:
Hmm, maybe we don't have to filter out no-cors cases? Instead, we'd check ok-status only if the response is not opaque filtered response. If the response is a opaque filtered response, we may want to check response [type](https://fetch.spec.whatwg.org/#concept-response-type) to check whether there is a network error or not instead.

domenic:
Stumbled upon this... @annevk do you have thoughts on @sisidovski's latest suggested fix?

sisidovski:
@annevk 

> Yes, as per https://fetch.spec.whatwg.org/#concept-filtered-response-opaque status will be 0. It does seem very concerning this would tell you for every subresource whether it has an ok status or not. That's not information we reveal for images today, for instance.

Service workers can already look at the OK status inside the fetch handler, so it doesn't exceed what the current service worker can do. No information is newly revealed in this change.

My previous comment looks incorrect. Maybe we can just check if the [status](https://fetch.spec.whatwg.org/#concept-response-status) is [ok status](https://fetch.spec.whatwg.org/#ok-status) or not. If the response is [opaque filtered response](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque), the status would be 0 regardless of the actual response. So the ok status check will fail anyway for the opaque response, and we think that behavior is correct.

annevk:
In general this looks good, but we should verify with tests when policies are applied:

- If you first request `https://example.com/foo` and then the service worker requests `http://example.com/foo`, will it be de-duplicated because of Mixed Content upgrading the scheme?
- If the document and service worker have different CSP policies, does the potential de-duplication happen after the security check as the patch implies?
- Do we only want to de-duplicate once as the current text in Service Workers implies (by removing the entry from the map)?

sisidovski:
Thanks! Let me address the last question first.

> Do we only want to de-duplicate once as the current text in Service Workers implies (by removing the entry from the map)?

We don't need to limit de-duplicating once. https://github.com/w3c/ServiceWorker/pull/1807 will fix it in the SW spec side.

Issue #716, opened 2018-04-23
Content codings (Content-Encoding)


annevk:
I looked into https://github.com/whatwg/fetch/pull/710#issuecomment-383000148 a bit. Browser interoperability here is far from good. Interesting scenarios:

* gzip
* unknown
* gzip,unknown
* unknown,gzip
* deflate
* gzip,deflate
* gzip,gzip
* gzip listed a large number of times with the content encoded with gzip as many times (see https://bugzilla.mozilla.org/show_bug.cgi?id=1248577)
* having a response body of which only the first half is encoded (see https://bugs.chromium.org/p/chromium/issues/detail?id=654889)

Somewhat related:
* https://bugs.chromium.org/p/chromium/issues/detail?id=567163
* https://github.com/httpwg/http-core/issues/58

It does not seem worth blocking #710 on this as it's an improvement over the status quo, but this could use more work.

cc @MoritzKn @TimothyGu @MattMenke2 @mnot

annevk:
The other thing that's wrong about the current setup is that Brotli requires HTTPS so that should also be passed through somehow. Cannot just have a codings list and bytes.

annevk:
Changes to "handling content codings":

1. Iterate over codings and do decoding for each if them, if supported, and retain input as-is otherwise. If decoding fails, return failure.
2. Pass in the response or some such so you can determine the HTTPS bit. (Might as well move parsing of `Content-Encoding` there.)

Questions:

1. Should we define a limit? Maybe at least a lower bound.
2. Can we make unsupported values an error? Probably not...
3. ...

annevk:
From @MattMenke2 at https://github.com/whatwg/fetch/pull/710#issuecomment-383614044:

> That there's also another, related issue here:  What if we support an encoding, but didn't advertise it?
> 
> Chrome used to let such requests through, applying decoding, but that led to interop issues with brotli (servers were sniffing the user agent instead of looking at the Accept-Encoding header, and sending Brotli responses over HTTP), so we changed, and now reject unadvertised encodings, at least if we recognize them.  That change led to issues with other servers that unconditionally send gzip - Chrome doesn't advertise gzip support when requesting media files, so it can use range requests.

annevk:
Another thing that's wrong here is that the handlers need to keep around some internal state for compression tables, offsets, etc. So it's effectively a sequence of specialized transform streams.

annevk:
[...]

PR #1742, opened 2024-03-25
Pass in Content-Encoding to resource-timing

Jxck:
These changes are to support addition of content type attribute to Perfomance resource timing. Further details are available at 
https://github.com/w3c/resource-timing/issues/381

   * Discussed and agreed at Feb 29, 2024 [W3C WebPerf call](https://docs.google.com/document/d/1qPPCtpg1MyVw3GGmd6VKZCdCyqoDub6Xgc8ANnq8SRI/edit#heading=h.wkdzwqaypyq6):
   * https://chromium-review.googlesource.com/c/chromium/src/+/5369775
   * Chromium: https://issues.chromium.org/issues/327941462
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1886107
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=271632
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed:
  -  https://github.com/mdn/content/issues/32823

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -369,6 +369,8 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
  <dd>A number.
  <dt><dfn export for="response body info">content type</dfn> (default the empty string)
  <dd>An <a for=/>ASCII string</a>.
+ <dt><dfn export for="response body info">content encoding</dfn> (default the empty string)
+ <dd>An <a for=/>ASCII string</a>.
 </dl>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


annevk (commented):
I think we need to actually define how the value is obtained. "extracting header list values" shouldn't be used for new features.

And similarly to `Content-Type` we probably want to minimize the information that is exchanged through some sort of safelisting.

Jxck:
@annevk Thanks for reviewing. I'm new to this area so I wanna ask you something.

> we need to actually define how the value is obtained.

Do you means that this change require "extract Content Encoding" algorithm like "extract a MIME type" ?
https://fetch.spec.whatwg.org/#concept-header-extract-mime-type

> we probably want to minimize the information
I think it should be seems like restricted to well-known codings like

- identity ?
- deflate
- gzip
- br
- zstd

is that right ? or is there some lists which I should refer to ?

annevk:
Yeah, extract a MIME type is an example of modern-style header parsing. I think all things written in terms of https://fetch.spec.whatwg.org/#concept-header-list-get or https://fetch.spec.whatwg.org/#concept-header-list-get-decode-split are good examples.

There is a registry for content codings here: https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding. I'm not sure if we have to acknowledge all of them however. `exi` and `pack200-gzip` don't need to be supported for instance and `x-gzip` can be normalized to `gzip`.

annevk:
See also #716.

yoavweiss:
We should also include [`br-d` and `zstd-d`](https://httpwg.org/http-extensions/draft-ietf-httpbis-compression-dictionary.html#name-content-encoding-2)

Jxck:
@annevk  Thanks, I'll be digging them.

@yoavweiss in this case, should new encodings be added to the IANA list first ?

yoavweiss:
@Jxck - I believe this happened as part of https://www.ietf.org/archive/id/draft-ietf-httpbis-compression-dictionary-19.html#name-iana-considerations

(In the process of which, it was changed to "dcb" and "dcz")

See https://www.iana.org/assignments/http-parameters/http-parameters.xhtml

annevk:
https://github.com/whatwg/fetch/commit/fd9beaa4c1288493328cffdd15e14bcaa6e3c1c2 was merged. So this can probably be closed?

https://github.com/w3c/resource-timing/issues/381 is still open however.

PR #1745, opened 2024-04-05
Rename http3Only to requireUnreliable

### LINKED ISSUE #1744: Consider renaming or replacing http3only?

In the context of WebTransport, the use of the term `http3only` in [obtain a connection](https://fetch.spec.whatwg.org/#concept-connection-obtain) came up in discussion of https://github.com/w3c/webtransport/issues/561.

We've been trying to adhere to more transport agnostic principles, allowing developers to express their needs based on properties of the transport, rather than hardcoding a particular protocol or a version of a protocol. 

In that context, the reason for requiring `http3only` was determined to instead represent the desire to obtain unreliable transport that can eliminate head-of-line blocking, perhaps we should use some spelling of the term `requireUnreliableTransport`.

Filing an issue to discuss if we want to transition to something more transport agnostic in fetch as well when obtaining connections. What happens when we have HTTP/4? HTTP/3.1? What if HTTP/3.1 doesn't provide the same underlying transport properties? It seems like enshrining the protocol version itself as the name of the field leaves us in an undesirable position.

annevk:
This should be more future proof.

Fixes #1744.

***
<a href="https://whatpr.org/fetch/1745.html" title="Last updated on Apr 29, 2024, 2:16 PM UTC (840df72)">Preview</a> | <a href="https://whatpr.org/fetch/1745/4c3750d...840df72.html" title="Last updated on Apr 29, 2024, 2:16 PM UTC (840df72)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2972,13 +2972,14 @@ boolean <var>requireUnreliable</var>, run these steps:
   [[!HTTP1]] [[!TLS]]
 
   <ul>
-   <li>
-    <p>If <var>requireUnreliable</var> is true, then establish a connection capable of unreliable
-    transport, e.g., HTTP/3. [[!HTTP3]]
+   <li><p>If <var>requireUnreliable</var> is true, then establish a connection capable of unreliable
+   transport, e.g., an HTTP/3 connection. [[!HTTP3]]
 
-    <p>When establishing such a connection, enable options that are necessary for WebTransport. For
-    HTTP/3, this means including SETTINGS_ENABLE_WEBTRANSPORT with a value of 1 and H3_DATAGRAM with
-    a value of 1 in the initial SETTINGS frame. [[!WEBTRANSPORT-HTTP3]] [[!HTTP3-DATAGRAM]]
+   <li><p>When establishing a connection capable of unreliable transport, enable options that are
+   necessary for WebTransport. For HTTP/3, this means including
+   <code>SETTINGS_ENABLE_WEBTRANSPORT</code> with a value of <code>1</code> and
+   <code>H3_DATAGRAM</code> with a value of <code>1</code> in the initial <code>SETTINGS</code>
+   frame. [[!WEBTRANSPORT-HTTP3]] [[!HTTP3-DATAGRAM]]
 
    <li><p>If <var>credentials</var> is false, then do not send a TLS client certificate.
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -2865,8 +2865,9 @@ steps:
 <a>network partition key</a> <var>key</var>, <a for=/>URL</a> <var>url</var>, boolean
 <var>credentials</var>, an optional <a>new connection setting</a> <var>new</var> (default
 "<code>no</code>"), and an optional boolean
-<dfn export for="obtain a connection"><var>http3Only</var></dfn> (default false), run these steps:
-<!-- new's "yes-and-dedicated" and http3Only have been added for WebTransport -->
+<dfn export for="obtain a connection"><var>requireUnreliable</var></dfn> (default false), run these
+steps:
+<!-- new's "yes-and-dedicated" and requireUnreliable have been added for WebTransport -->
 
 <ol>
  <li>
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2971,16 +2972,18 @@ boolean <var>http3Only</var>, run these steps:
...
+   <li>
+    <p>If <var>requireUnreliable</var> is true, then establish a connection capable of unreliable
+    transport, e.g., HTTP/3. [[!HTTP3]]
 
-   <li><p>When establishing an HTTP/3 connection, include SETTINGS_ENABLE_WEBTRANSPORT with a value
-   of 1 and H3_DATAGRAM with a value of 1 in the initial SETTINGS frame. [[!WEBTRANSPORT-HTTP3]]
-   [[!HTTP3-DATAGRAM]]
+    <p>When establishing such a connection, include SETTINGS_ENABLE_WEBTRANSPORT with a value of 1
+    and H3_DATAGRAM with a value of 1 in the initial SETTINGS frame. [[!WEBTRANSPORT-HTTP3]]
+    [[!HTTP3-DATAGRAM]]
```
annevk, line 2981: I would like some feedback on this part @ekinnear. Is this reasonable to desire for all unreliable transport connections? Presumably the details of this might shift in future protocol versions.

Perhaps this second paragraph should remain "When establishing an HTTP/3 connection"?

martinthomson, line 2981: Perhaps
```suggestion
    <p>When establishing such a connection, enable options that are necessary for WebTransport. 
    For HTTP/3, this means SETTINGS_ENABLE_WEBTRANSPORT with a value of 1
    and H3_DATAGRAM with a value of 1 in the initial SETTINGS frame. [[!WEBTRANSPORT-HTTP3]]
    [[!HTTP3-DATAGRAM]]
```

jasnell:
Given this work-in-progress "Quic for Streams" draft https://www.ietf.org/archive/id/draft-kazuho-quic-quic-on-streams-00.html ... which would allow implementing a functional subset of QUIC on top of TLS, I'm not sure "requireUnreliable" makes sense

annevk:
@jasnell presumably a WebTransport user that sets `requireUnreliable` to true would not want such a connection though?

ronag:
What does `requireUnreliable` even mean in a practical sense? Why would I ever want to require that something is "unreliable"?

annevk:
That's the term https://w3c.github.io/webtransport/ uses as public API and it's for them, so might as well stick with it. What it means is described in this PR.

fetch.bs
ekinnear: +1, this seems like a good improvement, and as we have additional concrete options we can include the specific configuration to use for them.

ekinnear (commented):
Thank you! This seems like it strikes a good balance between describing the needs at a high level and being clear and precise enough for implementation at a lower level.

annevk:
Thanks! I've pushed mt's suggestion. I will merge this on Monday barring any other comments.

fetch.bs
```diff
@@ -2971,16 +2972,18 @@ boolean <var>http3Only</var>, run these steps:
...
 
   <ul>
-   <li><p>If <var>http3Only</var> is true, then establish an HTTP/3 connection. [[!HTTP3]]
+   <li>
+    <p>If <var>requireUnreliable</var> is true, then establish a connection capable of unreliable
+    transport, e.g., HTTP/3. [[!HTTP3]]
 
-   <li><p>When establishing an HTTP/3 connection, include SETTINGS_ENABLE_WEBTRANSPORT with a value
-   of 1 and H3_DATAGRAM with a value of 1 in the initial SETTINGS frame. [[!WEBTRANSPORT-HTTP3]]
-   [[!HTTP3-DATAGRAM]]
```
jan-ivar, line 2978: See https://github.com/whatwg/fetch/pull/1745#discussion_r1578222835 on a resolved thread.

annevk: @jan-ivar I see, I guess then it should say "When establishing a connection capable of unreliable transport, enable options ..." I wonder if we have any cases where it's false and this is still relevant?

jan-ivar: @annevk It's likely false almost always. The WebTransport constructor's [requireUnreliable](https://w3c.github.io/webtransport/#dom-webtransportoptions-requireunreliable) option merely exists to disable HTTP/2 fallback, but likely nobody sets it since browsers still need to implement HTTP/2 fallback.

There's also [pooling](https://w3c.github.io/webtransport/#dom-webtransportoptions-allowpooling) to consider aka `!dedicated` (also not implemented yet AFAIK) where the WebTransport constructor is returned an existing connection from this algorithm's connection pool. 

Together this means both SETTINGS_ENABLE_WEBTRANSPORT and H3_DATAGRAM need to be enabled on _all_ HTTP/3 connections by default, even when not called from WebTransport. At least that's how I read it.

If this is surprising to others (or if I read this wrong), we might want to make this clearer, as I could see someone reading it the other way right now.

annevk:
Thanks, given there's been feedback that resulted in further changes I will merge it this week Thursday now to give everyone a chance to look again.

Issue #1675, opened 2023-06-21
“Realmless” ArrayBuffer creation in Body’s arrayBuffer() steps


bathos:
The `Body` mixin’s `arrayBuffer` operation steps are defined as:

> [returning] the result of running consume body with _this_ and the following step given a byte sequence bytes: return a new ArrayBuffer whose contents are bytes.

The “new ArrayBuffer” portion is unlinked. It should presumably be linking to [“create an ArrayBuffer”](https://webidl.spec.whatwg.org/#arraybuffer-create) in Web IDL and its missing argument (a realm) should be supplied. I would expect the realm here to be the [relevant realm](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-realm) of [this](https://webidl.spec.whatwg.org/#this).

(Issue discovered by @twiss when [correcting ArrayBuffer usage in WebCrypto algorithms](https://github.com/w3c/webcrypto/pull/347#discussion_r1236440541).)

annevk:
Might want to also fix

> a [Uint8Array](https://webidl.spec.whatwg.org/#idl-Uint8Array) wrapping an [ArrayBuffer](https://webidl.spec.whatwg.org/#idl-ArrayBuffer) containing the available bytes

while here.

Issue #1732, opened 2024-01-29
add a method to get a Uint8Array to Body


bakkot:
### What problem are you trying to solve?

Body has an `.arrayBuffer()` method, but that's [almost never what you actually want](https://github.com/w3ctag/design-principles/issues/463), since you can't read out of a buffer directly - you need a view. Uint8Array is the canonical "sequence of byte values" view.

### What solutions exist today?

`new Uint8Array(await body.arrayBuffer())`

### How would you solve it?

Add a `.bytes()` method which gives a new `Uint8Array`.

### Anything else?

Some brief discussion [here](https://matrixlogs.bakkot.com/WHATWG/2024-01-18#L1-L3).

PushMessageData and Blob (which live elsewhere) would need this too, since they just copied Body.

(Also, watch out for https://github.com/whatwg/fetch/issues/1675 while specifying.)

annevk:
FWIW, WebKit is okay with adding this.

cc @valenting @ricea

saschanaz:
Sounds okay to me.

ricea:
Chromium supports.

PR #1751, opened 2024-05-04
Editorial: Fix realmless ArrayBuffer creation

bakkot:
Fixes https://github.com/whatwg/fetch/issues/1675, maybe.

@annevk in that issue also points out that the [extract a body [...]](https://fetch.spec.whatwg.org/#concept-bodyinit-extract) algorithm also creates a Uint8Array in an ad-hoc way, which I've also tried to fix. Unfortunately I don't know what realm to use there - unlike the original issue, these aren't method steps. I am guessing that when the realm is unspecified (which it very often isn't - for example, earlier in the same algorithm there is "set stream to a new ReadableStream object" without a realm argument), it's implicitly the current realm? But maybe we need to thread a realm argument into this algorithm (and quite a few others), or possibly just explicitly pass "the current realm".

(If this right I'll follow up with a PR for https://github.com/whatwg/fetch/issues/1732.)

***
<a href="https://whatpr.org/fetch/1751.html" title="Last updated on May 7, 2024, 11:17 AM UTC (b9ef20a)">Preview</a> | <a href="https://whatpr.org/fetch/1751/14bad1d...b9ef20a.html" title="Last updated on May 7, 2024, 11:17 AM UTC (b9ef20a)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -7139,8 +7139,7 @@ steps:
     <p>Run <var>action</var>.
 
     <p>Whenever one or more bytes are available and <var>stream</var> is not
-    <a for=ReadableStream>errored</a>, <a for=ReadableStream>enqueue</a> a {{Uint8Array}} wrapping
-    an {{ArrayBuffer}} containing the available bytes into <var>stream</var>.
+    <a for=ReadableStream>errored</a>, <a for=ReadableStream>enqueue</a> the result of [=ArrayBufferView/create|creating=] a {{Uint8Array}} from the available bytes into <var>stream</var>.
 
     <p>When running <var>action</var> is done, <a for=ReadableStream>close</a> <var>stream</var>.
   </ol>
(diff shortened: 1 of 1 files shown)


annevk:
This looks correct, but I wonder if we have test coverage for this.

It also needs some wrapping adjustment, but happy to push a fix for that myself before landing.

The longer term plan is to fix this more generally through Web IDL by making an ambient realm available to be used for creating (most) new objects.

bakkot:
> This looks correct, but I wonder if we have test coverage for this.

There are no tests for the realm of created objects at all, that I can find. I could add some but I'd be inventing the conventions.

> The longer term plan is to fix this more generally through Web IDL by making an ambient realm available to be used for creating (most) new objects.

Presumably that would apply to the first hunk (the one in `extract a body with type`), but not the second (the one in `arrayBuffer()`), right? Since the ambient realm is not the one used in the second case. (Specifically, if you do `await firstRealm.Response.prototype.arrayBuffer.call(new secondRealm.Response('asdf'))`, the resulting `ArrayBuffer` comes from the _second_ realm, not the first, i.e. it is derived from the realm of `this` rather than the realm of the `arrayBuffer` function.)

annevk:
Well, "ambient" was doing a lot of work there. The idea would be for Web IDL to handle these kind of cases correctly so nobody has to think about them again.

If you could add a minimal `window.js` test I think inventing the conventions is fine?

bakkot:
https://github.com/web-platform-tests/wpt/pull/46104

annevk (approved):

PR #1757, opened 2024-06-03
Add process model processing for the "Priority" header.

### LINKED ISSUE #1718: Define behavior for `Priority` request header

The `Priority` request header is part of [RFC 9218](https://datatracker.ietf.org/doc/html/rfc9218#name-the-priority-http-header-fi) (Extensible HTTP Priorities) and sent by the networking layer when appropriate (for most browsers this is when using HTTP/3, Chrome may start sending it for HTTP/2 as well).

The header is not currently on the list of [forbidden request headers](https://fetch.spec.whatwg.org/#forbidden-request-header) and the behavior is undefined for how it interacts with a user-provided `Priority` header in fetch.

For actual prioritization, fetch provides [RequestPriority](https://fetch.spec.whatwg.org/#enumdef-requestpriority) (though it is not as granular) but applications may have their own use for the header name if they are already sending it.

It would be helpful to specify the behavior either by adding it to the forbidden header list or defining how the extensible priorities header should be treated if an application provides an explicit `Priority` header.

The current behavior in Firefox is to send both headers. In Chrome (behind a flag) the header will only be set by the networking stack if the application didn't include a `Priority` header as part of the request.

pmeenan:
   * Chrome
   * Firefox
   * See discussion in #1718
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

This adds the processing for the `Priority` HTTP header from HTTP Extensible Priorities ([RFC 9218](https://datatracker.ietf.org/doc/rfc9218/)).

The actual priority values are implementation-specific but this change makes it clear that an existing value should not be overwritten if the header is already present.

Fix #1718

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5654,6 +5654,12 @@ run these steps:
     default), and `<code>Accept-Charset</code>` is a waste of bytes. See
     <a>HTTP header layer division</a> for more details.
 
+   <li><p>If <var>httpRequest</var>'s <a for=request>header list</a>
+    <a for="header list">does not contain</a> `<code>Priority</code>`, then user agents may
+    <a for="header list">append</a> a `<code>Priority</code>` <a>header</a> with an
+    <a>implementation-defined</a> value derived from <var>httpRequest</var>'s
+    <a for=request>internal priority</a> in accordance with RFC 9218. [[!RFC9218]]
+
    <li>
     <p>If <var>includeCredentials</var> is true, then:
 


pmeenan:
Tests and implementation bugs coming soon.

annevk (approved):
Thanks, this looks reasonable to me modulo nits. This means that service workers do not get to see the header ever unless it was set by a web developer. Do we have test coverage for that as well?

PR #1769, opened 2024-08-19
Make get, decode, and split handle edge cases correctly

### LINKED ISSUE #1768: "get, decode, and split" doesn't return example output

https://fetch.spec.whatwg.org/commit-snapshots/4cb3cf21946113c0684f04122dd95315fd10c567/#header-value-get-decode-and-split

The algorithm doesn't seem to return the same output of the first [example](https://fetch.spec.whatwg.org/commit-snapshots/4cb3cf21946113c0684f04122dd95315fd10c567/#example-header-list-get-decode-split).

The example says that giving `nosniff,` as input we should get a list of two strings: `"nosniff"` and an empty string `""`.

If I try to follow the steps however I get a list of a single `"nosniff"` string and no empty string.

Here's how I'm interpreting it:

> 1. Let input be the result of [isomorphic decoding](https://infra.spec.whatwg.org/#isomorphic-decode) value.
> 2. Let position be a [position variable](https://infra.spec.whatwg.org/#string-position-variable) for input, initially pointing at the start of input.
> 3. Let values be a [list](https://infra.spec.whatwg.org/#list) of [strings](https://infra.spec.whatwg.org/#string), initially empty.
> 4. Let temporaryValue be the empty string.
```
input:   "nosniff,"
position: ^
values: []
temporaryValue: ""
```
>  5. While position is not past the end of input:
>     1. Append the result of [collecting a sequence of code points](https://infra.spec.whatwg.org/#collect-a-sequence-of-code-points) that are not U+0022 (") or U+002C (,) from input, given position, to temporaryValue.
```
input:   "nosniff,"
position:        ^
values: []
temporaryValue: "nosniff"
```
> 5.
>     2. If position is not past the end of input, then:
>        1. ~If the [code point](https://infra.spec.whatwg.org/#code-point) at position within input is U+0022 ("), then:~
>           1. ~Append the result of [collecting an HTTP quoted string](https://fetch.spec.whatwg.org/commit-snapshots/4cb3cf21946113c0684f04122dd95315fd10c567/#collect-an-http-quoted-string) from input, given position, to temporaryValue.~
>           2. ~If position is not past the end of input, then [continue](https://infra.spec.whatwg.org/#iteration-continue).~
>        2. Otherwise:
>           1. Assert: the [code point](https://infra.spec.whatwg.org/#code-point) at position within input is U+002C (,).
>           2. Advance position by 1.
```
input:   "nosniff,"
position:         ^
values: []
temporaryValue: "nosniff"
```
> 5.
>     2.
>        3. Remove all [HTTP tab or space](https://fetch.spec.whatwg.org/commit-snapshots/4cb3cf21946113c0684f04122dd95315fd10c567/#http-tab-or-space) from the start and end of temporaryValue.
>        4. [Append](https://infra.spec.whatwg.org/#list-append) temporaryValue to values.
```
input:   "nosniff,"
position:         ^
values: ["nosniff"]
temporaryValue: ""
```
> 5.
>     2.
>        5. Set temporaryValue to the empty string.

Here position is past the end of input, the while condition is not satisfied and this returns a list with just a single element: `["nosniff"]`.
> 6. Return values.

The examples seem right to me, I think the algorithm is wrong.

I found some WPT that indirectly test this algorithm ([via `Content-Length`](https://fetch.spec.whatwg.org/#content-length-header)), Chrome and Firefox seem to follow the behavior of the example.

https://github.com/web-platform-tests/wpt/blob/merge_pr_47660/fetch/content-length/resources/content-lengths.json#L66-L69
https://wpt.fyi/results/fetch/content-length/parsing.window.html

I.e.: a header value with a comma as the last character is split into two values, as in the example. Since the header is `Content-Length` and the spec mandates to abort if the split values don't match, the test passes if [the request](https://wpt.live/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2030%2C) results in a network error.

This should be the relevant Chrome implementation, it looks like it internally stores every header splitted
https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_response_headers.cc;l=967-970;drc=82dff63dbf9db05e9274e11d9128af7b9f51ceaa;bpv=0;bpt=1
https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_util.cc;l=968;drc=82dff63dbf9db05e9274e11d9128af7b9f51ceaa

Firefox on the other hand doesn't seem to do that, the `Content-Length` check appears to be implemented using a simpler split by just commas, ignoring quotes. It makes sense: a quote char inside `Content-Length` already makes it invalid so there's no need to care about quotes while splitting.
https://searchfox.org/mozilla-central/rev/53e68046298557fae0c922230b595bb6689bf587/netwerk/protocol/http/nsHttpHeaderArray.cpp#173-189
https://searchfox.org/mozilla-central/rev/53e68046298557fae0c922230b595bb6689bf587/netwerk/protocol/http/nsHttpHeaderArray.h#308

annevk:
Correct how it handles when a header is present but has no value (should be a list solely containing an empty string value) and when there is a trailing comma (should be a list ending with an empty string value).

Fixes #1768.

---

@domenic @MattMenke2 this was an oversight in #831 it appears. If you'd like to review in addition to @CarloCannas I'd appreciate it. Thanks!

***
<a href="https://whatpr.org/fetch/1769.html" title="Last updated on Aug 19, 2024, 1:43 PM UTC (0334504)">Preview</a> | <a href="https://whatpr.org/fetch/1769/4cb3cf2...0334504.html" title="Last updated on Aug 19, 2024, 1:43 PM UTC (0334504)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -778,30 +778,15 @@ a <a for=/>header value</a> <var>value</var>, run these steps. They return a <a 
     <p class=note>The result might be the empty string.
 
    <li>
-    <p>If <var>position</var> is not past the end of <var>input</var>, then:
+    <p>If <var>position</var> is not past the end of <var>input</var> and the
+    <a for=/>code point</a> at <var>position</var> within <var>input</var> is U+0022 ("):
 
     <ol>
-     <li>
-      <p>If the <a for=/>code point</a> at <var>position</var> within <var>input</var> is
-      U+0022 ("), then:
+     <li><p>Append the result of <a>collecting an HTTP quoted string</a> from <var>input</var>,
+     given <var>position</var>, to <var>temporaryValue</var>.
 
-      <ol>
-       <li><p>Append the result of <a>collecting an HTTP quoted string</a> from <var>input</var>,
-       given <var>position</var>, to <var>temporaryValue</var>.
-
-       <li>If <var>position</var> is not past the end of <var>input</var>, then
-       <a for=iteration>continue</a>.
-      </ol>
-
-     <li>
-      <p>Otherwise:
-
-      <ol>
-       <li><p>Assert: the <a for=/>code point</a> at <var>position</var> within <var>input</var> is
-       U+002C (,).
-
-       <li><p>Advance <var>position</var> by 1.
-      </ol>
+     <li>If <var>position</var> is not past the end of <var>input</var>, then
+     <a for=iteration>continue</a>.
     </ol>
 
    <li><p>Remove all <a>HTTP tab or space</a> from the start and end of <var>temporaryValue</var>.
@@ -809,6 +794,19 @@ a <a for=/>header value</a> <var>value</var>, run these steps. They return a <a 
    <li><p><a for=list>Append</a> <var>temporaryValue</var> to <var>values</var>.
 
    <li><p>Set <var>temporaryValue</var> to the empty string.
+
+   <li>
+    <p>If <var>position</var> is not past the end of <var>input</var>:
+
+    <ol>
+     <li><p><a for=/>Assert</a>: the <a for=/>code point</a> at <var>position</var> within
+     <var>input</var> is U+002C (,).
+
+     <li><p>Advance <var>position</var> by 1.
+
+     <li><p>If <var>position</var> is past the end of <var>input</var>, then <a for=list>append</a>
+     the empty string to <var>values</var>.
+    </ol>
   </ol>
 
  <li><p>Return <var>values</var>.
@@ -9040,6 +9038,7 @@ Brad Porter,
 Bryan Smith,
 Caitlin Potter,
 Cameron McCormack,
+Carlo Cannas,
 白丞祐 (Cheng-You Bai)<!-- CYBAI; GitHub -->,
 Chirag S Kumar<!-- fictionistique; GitHub -->,
 Chris Needham,
--- a/fetch.bs
+++ b/fetch.bs
@@ -700,6 +700,19 @@ A:
 </code></pre>
   <tr>
    <td>
+    <pre><code class=lang-http>
+A:
+B: sniff
+</code></pre>
+   <td>« "" »
+  <tr>
+   <td>
+    <pre><code class=lang-http>
+B: sniff
+</code></pre>
+   <td>null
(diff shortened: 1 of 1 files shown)


CarloCannas:
Hmm, what about an header with empty value?

If they are allowed (which seems the case to me, reading both [RFC 9110](https://httpwg.org/specs/rfc9110.html#fields.values) and [header value definition](https://fetch.spec.whatwg.org/#header-value)) I think this algorithm should return a list of a single empty string, not an empty list.

What do you think?

Maybe something like this? (sorry for the bad formatting)

```
    Let input be the result of isomorphic decoding value.

    Let position be a position variable for input, initially pointing at the start of input.

    Let values be a list of strings, initially empty.

    Let temporaryValue be the empty string.

    While true:

        Append the result of collecting a sequence of code points that are not U+0022 (") or U+002C (,) from input, given position, to temporaryValue.

        The result might be the empty string.

        If position is not past the end of input and the code point at position within input is U+0022 ("), then:

                Append the result of collecting an HTTP quoted string from input, given position, to temporaryValue.

                If position is not past the end of input, then continue.

        Remove all HTTP tab or space from the start and end of temporaryValue.

        Append temporaryValue to values.

        If position is past the end of input

        	return values

        Set temporaryValue to the empty string.

        Assert: the code point at position within input is U+002C (,).

        Advance position by 1.
```

annevk:
@CarloCannas yeah, that seems reasonable and it looks like that should be compatible with existing callers. I guess with that in theory we could make the "parent" algorithm return the empty list instead of null when the header is missing, but null is clearer. I'll see about making that change.

domenic (approved):
I can't really run this algorithm in my head enough to verify it's correct, but editorially this seems reasonable. Let's count on @CarloCannas for the normative review :).

A reference implementation and tests file with 100% code coverage would be the best solution of course.

annevk:
@CarloCannas any further thoughts on this?

I'll merge this Monday if there's no more feedback.

CarloCannas:
Hi Anne,
I reviewed the changes again, it looks fine to me. Go ahead and merge!

PR #1776, opened 2024-09-24
Clarify that request.origin cannot be "client" at certain points

### LINKED ISSUE #1773: Unclear how to serialize request origin "client"

When [`serializing a request origin`](https://fetch.spec.whatwg.org/#serializing-a-request-origin), Step 2. tells us to
`Return request's origin, serialized`. A [request's origin](https://fetch.spec.whatwg.org/#concept-request-origin) is either `"client"` or an [Origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin). However, the 

 [`ASCII serialisation of an origin`](https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin) algorithm referenced in Step 2. does not handle the case of a `"client"` origin. (Because it is about serializing an origin as described in the HTML spec, not a request's origin as described in the fetch spec.)

How should an origin of `"client"` be serialized?

simonwuelker:
Previously, the code implicitly guaranteed that a requests origin could not be "client" after Step 10 of
https://fetch.spec.whatwg.org/#concept-fetch.

Other algorithms then simply relied on this fact. This is confusing. Instead, we now explicitly assert that the origin is not client whereever we need it.

This closes #1773

---
- [X] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/fetch/1776.html" title="Last updated on Oct 1, 2024, 7:43 AM UTC (6835582)">Preview</a> | <a href="https://whatpr.org/fetch/1776/edfa8d1...6835582.html" title="Last updated on Oct 1, 2024, 7:43 AM UTC (6835582)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -458,7 +458,8 @@ and an optional boolean <var>extract-value</var> (default false):
 
  <li><p>Let <var>value</var> be the empty string.
 
- <li><p>Assert: the <a>code point</a> at <var>position</var> within <var>input</var> is U+0022 (").
+ <li><p><a for=/>Assert</a>: the <a>code point</a> at <var>position</var> within
+ <var>input</var> is U+0022 (").
 
  <li><p>Advance <var>position</var> by 1.
 
@@ -494,7 +495,7 @@ and an optional boolean <var>extract-value</var> (default false):
     <p>Otherwise:
 
     <ol>
-     <li><p>Assert: <var>quoteOrBackslash</var> is U+0022 (").
+     <li><p><a for=/>Assert</a>: <var>quoteOrBackslash</var> is U+0022 (").
 
      <li><p><a for=iteration>Break</a>.
     </ol>
@@ -591,8 +592,8 @@ given a <a for=/>header name</a> <var>name</var> and a string <var>type</var> fr
 <a>structured field value</a>.
 
 <ol>
- <li><p>Assert: <var>type</var> is one of "<code>dictionary</code>", "<code>list</code>", or
- "<code>item</code>".
+ <li><p><a for=/>Assert</a>: <var>type</var> is one of "<code>dictionary</code>",
+ "<code>list</code>", or "<code>item</code>".
 
  <li><p>Let <var>value</var> be the result of <a for="header list">getting</a> <var>name</var> from
  <var>list</var>.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2230,6 +2230,8 @@ or "<code>object</code>".
 return true:
 
 <ol>
+ <li> Assert: <var>request</var>'s <a for=request>origin</a> is not "<code>client</code>".
```
annevk, line 2233:
```suggestion
 <li><p><a for=/>Assert</a>: <var>request</var>'s <a for=request>origin</a> is not
 "<code>client</code>".
```

annevk: In particular, no space after `<li>` but it also needs a nested `<p>` like the other `<li>`s, we want to link Assert as people can use the additional guidance that provides, and then we need to wrap at a 100 columns unfortunately due to the additional text.

If you apply this throughout I'd be happy to merge this. Thanks for turning it around so quickly!

simonwuelker: Done - I've also added links for `Assert:` in all places that did not previously have them

fetch.bs
```diff
@@ -458,7 +458,7 @@ and an optional boolean <var>extract-value</var> (default false):
 
  <li><p>Let <var>value</var> be the empty string.
 
- <li><p>Assert: the <a>code point</a> at <var>position</var> within <var>input</var> is U+0022 (").
+ <li><p><a for=/>Assert</a>: the <a>code point</a> at <var>position</var> within <var>input</var> is U+0022 (").
```
annevk, line 461: 100 column wrapping please.

fetch.bs
```diff
@@ -2230,6 +2230,9 @@ or "<code>object</code>".
 return true:
 
 <ol>
+ <li><p><a for=/>Assert</a>: <var>request</var>'s <a for=request>origin</a> is not
+ "<code>client</code>"..
```
annevk, line 2234: Perhaps my suggestion was wrong, but I notice this ends in two dots now. That's also the case in a few other places.

fetch.bs
```diff
@@ -2295,7 +2301,7 @@ is to return the result of <a>serializing a request origin</a> with <var>request
 <var>last</var>, run these steps:
 
 <ol>
- <li><p>Assert: <var>last</var> is not given, or <var>first</var> is less than or equal to
+ <li><p><a for=/>Assert</a>: <var>last</var> is not given, or <var>first</var> is less than or equal to
  <var>last</var>.
```
annevk, line 2305: I suspect the wrapping isn't correct here anymore either.

PR #1783, opened 2024-10-25
Partition Blob URL fetches by Storage Key

recvfrom:
Partition Blob URL fetches by Storage Key

This change updates the spec to partition Blob URL fetches by Storage Key. This change is part of broader changes discussed in https://github.com/w3c/FileAPI/issues/153#issuecomment-2330085478. Specifically, we will also:
 - Partition Blob URL revocation by storage key - https://github.com/w3c/FileAPI/pull/201
 - Enforce noopener for cross-top-level-site Blob URLs - (Spec changes for this will follow)

I considered incorporating the Storage Key checks into the "resolve a blob URL" algorithm instead, but it seemed that this would require an environment settings object to be available as part of https://url.spec.whatwg.org/#url-parsing, and I'm not sure whether that is the case / a change we want.

   * Firefox has already implemented this - https://github.com/w3c/FileAPI/issues/153#issuecomment-2332288047
   * Safari has implemented partitioning Blob URL fetches by top-level origin and is considering partitioning them by a site-based Storage Key (Storage Key specifics TBD) - https://github.com/w3c/FileAPI/issues/153#issuecomment-2332086739
   * https://chromium-review.googlesource.com/c/chromium/src/+/5967596
   * Chromium: https://crbug.com/40057646
   * Gecko: N/A
   * WebKit: N/A
   * Deno (not for CORS changes): N/A (it seems Deno does not partition storage or more generally implement the same-origin policy: https://docs.deno.com/runtime/reference/web_platform_apis/#spec-deviations)
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/36542

***
<a href="https://whatpr.org/fetch/1783.html" title="Last updated on Dec 6, 2024, 11:17 PM UTC (ddae2dc)">Preview</a> | <a href="https://whatpr.org/fetch/1783/1dc1b03...ddae2dc.html" title="Last updated on Dec 6, 2024, 11:17 PM UTC (ddae2dc)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -3136,9 +3136,10 @@ or an <a>implementation-defined</a> value.
 <h3 id=storage-keys-for-requests>Storage keys for requests</h3>
 
 <p class=note>Requests made to Blob URLs (other than those corresponding to navigations) are
-prevented from succeeding if the <a>storage key</a> for the <a>environment</a> making the request
-is different than the <a>storage key</a> of the <environment> where the Blob URL was created.
-[[STORAGE]]
+prevented from succeeding if the <a spec=storage>storage key</a> of the
+<a>environment settings object</a> making the request is different than the
+<a spec=storage>storage key</a> of the <a>environment settings object</a> corresponding to where
+the Blob URL was created.
 
 <div algorithm>
 <p>To <dfn for=request>determine the storage key</dfn>, given a <a for=/>request</a>
@@ -3147,11 +3148,11 @@ is different than the <a>storage key</a> of the <environment> where the Blob URL
 <ol>
  <li><p>If <var>request</var>'s <a for=request>reserved client</a> is non-null, then return the
  result of running <a for=/>obtain a storage key for non-storage purposes</a> given
- <var>request</var>'s <a for=request>reserved client</a>. [[!STORAGE]]
+ <var>request</var>'s <a for=request>reserved client</a>.
 
  <li><p>If <var>request</var>'s <a for=request>client</a> is non-null, then return the result of
- running <a for=/>obtain a storage key for non-storage purposes/a> given <var>request</var>'s
- <a for=request>client</a>. [[!STORAGE]]
+ running <a for=/>obtain a storage key for non-storage purposes</a> given <var>request</var>'s
+ <a for=request>client</a>.
 
  <li><p>Return null.
 </ol>
@@ -5006,7 +5007,7 @@ steps:
      <li>
       <p>Let <var>blobStorageKey</var> be the result of running
       <a>obtain a storage key for non-storage purposes</a> with <var>blobURLEntry</var>'s
-      <a>environment settings object</a>. [[!STORAGE]]
+      <a>environment settings object</a>.
 
      <li>
       <p>Let <var>requestStorageKey</var> be the result of
@@ -5020,7 +5021,8 @@ steps:
 
        <li><p><var>requestStorageKey</var> is non-null
 
-       <li><p><var>requestStorageKey</var> is does not <a>storage key equal</a> <var>blobStorageKey</var> [[!STORAGE]]
+       <li><p><var>requestStorageKey</var> does not <a for="storage key">equal</a>
+       <var>blobStorageKey</var>
       </ul>
 
       <p>then return a <a>network error</a>.
--- a/fetch.bs
+++ b/fetch.bs
@@ -3133,32 +3133,6 @@ or an <a>implementation-defined</a> value.
 </div>
 
 
-<h3 id=storage-keys-for-requests>Storage keys for requests</h3>
-
-<p class=note>Requests made to Blob URLs (other than those corresponding to navigations) are
-prevented from succeeding if the <a spec=storage>storage key</a> of the
-<a>environment settings object</a> making the request is different than the
-<a spec=storage>storage key</a> of the <a>environment settings object</a> corresponding to where
-the Blob URL was created.
-
-<div algorithm>
-<p>To <dfn for=request>determine the storage key</dfn>, given a <a for=/>request</a>
-<var>request</var>:
-
-<ol>
- <li><p>If <var>request</var>'s <a for=request>reserved client</a> is non-null, then return the
- result of running <a for=/>obtain a storage key for non-storage purposes</a> given
- <var>request</var>'s <a for=request>reserved client</a>.
-
- <li><p>If <var>request</var>'s <a for=request>client</a> is non-null, then return the result of
- running <a for=/>obtain a storage key for non-storage purposes</a> given <var>request</var>'s
- <a for=request>client</a>.
-
- <li><p>Return null.
-</ol>
-</div>
-
-
 <h3 id=http-cache-partitions>HTTP cache partitions</h3>
 
 <div algorithm>
@@ -5007,7 +4981,7 @@ steps:
      <li>
       <p>Let <var>blobStorageKey</var> be the result of running
       <a>obtain a storage key for non-storage purposes</a> with <var>blobURLEntry</var>'s
-      <a>environment settings object</a>.
+      <a for="blob URL entry">environment</a>.
 
      <li>
       <p>Let <var>requestStorageKey</var> be the result of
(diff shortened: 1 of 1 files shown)


recvfrom:
@annevk would you be able to look at this PR as well? Or could you recommend another reviewer? Thank you!

fetch.bs
```diff
@@ -3133,6 +3133,32 @@ or an <a>implementation-defined</a> value.
 </div>
 
 
+<h3 id=storage-keys-for-requests>Storage keys for requests</h3>
+
+<p class=note>Requests made to Blob URLs (other than those corresponding to navigations) are
+prevented from succeeding if the <a spec=storage>storage key</a> of the
+<a>environment settings object</a> making the request is different than the
+<a spec=storage>storage key</a> of the <a>environment settings object</a> corresponding to where
+the Blob URL was created.
```
domenic, line 3142: This feels like it would work better after line 5028, instead of near this algorithm.

fetch.bs
```diff
@@ -4978,6 +5004,29 @@ steps:
...
+
+     <li>
+      <p>Let <var>requestStorageKey</var> be the result of
+      <a for=request>determining the storage key</a> given <var>request</var>.
+
+     <li>
+      <p>If all of the following conditions are true:
+
+      <ul class=brief>
+       <li><p><var>request</var>'s <a for=request>mode</a> is not "<code>navigate</code>"
```
domenic, line 5020:
```suggestion
       <li><p><var>request</var>'s <a for=request>mode</a> is not "<code>navigate</code>";
```

fetch.bs
```diff
@@ -4978,6 +5004,29 @@ steps:
...
+
+       <li><p><var>requestStorageKey</var> is non-null
```
domenic, line 5022:
```suggestion
       <li><p><var>requestStorageKey</var> is non-null; and
```

fetch.bs
```diff
@@ -4978,6 +5004,29 @@ steps:
...
+
+       <li><p><var>requestStorageKey</var> does not <a for="storage key">equal</a>
+       <var>blobStorageKey</var>
```
domenic, line 5025:
```suggestion
       <var>blobStorageKey</var>,
```

fetch.bs
```diff
@@ -4978,6 +5004,29 @@ steps:
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
+     <li>
+      <p>Let <var>blobStorageKey</var> be the result of running
+      <a>obtain a storage key for non-storage purposes</a> with <var>blobURLEntry</var>'s
+      <a>environment settings object</a>.
```
domenic, line 5010: This should not link to the type definition of "environment settings object", but instead link to https://w3c.github.io/FileAPI/#blob-url-entry-environment .

fetch.bs
recvfrom: I was thinking this would make sense in the Infrastructure section like how the `HTTP cache partitioning` section and `determine the HTTP cache partition` algorithm are, but I've moved the algorithm under the `Scheme fetch` section now (and removed the note since it seems kinda out of place now. WDYT?

fetch.bs
recvfrom: ah yes thanks for pointing this out. Changed

domenic (approved):
Non-editor LGTM

domenic:
It looks like there aren't _too_ many references to the blob URL entry's object... https://dontcallmedom.github.io/webdex/o.html#object%40%40blob%20URL%20entry%40dfn

annevk:
Nice, the HTML find is legit and would also need to be updated I suspect. It would benefit from having the obtaining defined in File API so there's less need for duplication. Reporting API seems bogus: https://github.com/w3c/reporting/issues/273. Web Authentication API seems similarly bogus: https://github.com/w3c/webauthn/issues/2212.

fetch.bs
```diff
@@ -4978,6 +4978,29 @@ steps:
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
+     <li>
```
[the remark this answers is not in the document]
recvfrom: Thanks, this should now be addressed

recvfrom:
Thanks for the feedback, I've updated my FileAPI PR [1] to expose an obtain algorithm as recommended here, and I've updated this PR to use it. One question I thought more about was what to do when the request doesn't have an associated environment... It seems like for Blob URLs we should allow navigations to bypass the partitioning checks, but otherwise we should just fail (this seems in line with the Chromium implementation at least). WDYT?

[1] https://github.com/w3c/FileAPI/pull/201/files

fetch.bs
```diff
@@ -4970,15 +4970,26 @@ steps:
...
+     <a for=request>determining the environment</a> given <var>request</var>.
+
+     <li><p>Let <var>isNavigation</var> be true if <var>request</var>'s <a for=request>mode</a> is
+     not "<code>navigate</code>"; otherwise, false.
+
+     <li><p>If <var>isNavigation</var> is false and <var>requestEnvironment</var> is null, then
+     return a <a>network error</a>.
+
+     <li><p>Let <var>blob</var> be the result of <a>obtaining a blob object</a> given
+     <var>blobURLEntry</var>, <var>requestEnvironment</var>, and <var>isNavigation</var>.
```
annevk, line 4989: So when isNavigation is true we don't even care about environment, right? I would merge those parameters into one.

So you have

> given an environment or the string "`navigation`" _environment_

on the other side or some such forcing callers to confront that question. And it should probably be top-level navigation?

fetch.bs
```diff
@@ -4970,15 +4970,26 @@ steps:
...
+
+     <li><p>If <var>blob</var> is failure, or <var>blob</var> is not a {{Blob}} object, then
```
annevk, line 4991: You only need the second conditional to catch both conditions.

fetch.bs
```diff
@@ -4970,15 +4970,26 @@ steps:
...
 
       <p class=note>The `<code>GET</code>` <a for=/>method</a> restriction serves no useful purpose
       other than being interoperable.
 
-     <li><p>Let <var>blob</var> be <var>blobURLEntry</var>'s <a for="blob URL entry">object</a>.
+     <li><p>Let <var>requestEnvironment</var> be the result of
+     <a for=request>determining the environment</a> given <var>request</var>.
+
+     <li><p>Let <var>isNavigation</var> be true if <var>request</var>'s <a for=request>mode</a> is
+     not "<code>navigate</code>"; otherwise, false.
```
annevk, line 4983: Perhaps instead of mode we should be checking destination for "document" which is only true for top-level navigations.

annevk (commented):
Thank you for reworking this so quickly!

fetch.bs
recvfrom: Thanks for calling this out, I had overlooked that we didn't want to exempt all navigations from the partitioning checks and instead only want to exempt top-level navigations. I confirmed that Safari and Firefox both enforce partitioning checks on subframe navigations, so we'll implement it this way in Chromium as well, write a new WPT for this, and I'll update the spec here to look for destination equaling documenting as you suggested.

recvfrom: Done!

fetch.bs
recvfrom: Thanks for the feedback - done!

fetch.bs
```diff
@@ -4970,15 +4970,32 @@ steps:
...
+     <li><p>Let <var>isTopLevelNavigation</var> be true if <var>request</var>'s <a
+     for=request>destination</a> is "<code>document</code>"; otherwise, false.
+
+     <li><p>If <var>isTopLevelNavigation</var> is false and <var>requestEnvironment</var> is null,
+     then return a <a>network error</a>.
+
+     <li><p>Let <var>blob</var> be null.
+
+     <li><p>If <var>isTopLevelNavigation</var> is true, then set <var>blob</var> to the result of
+     <a href="https://w3c.github.io/FileAPI/#blob-url-obtain-object">obtaining a blob object</a>
```
annevk, line 4991: Given that the File API PR landed we should be able to do this without `href`. Same below.

fetch.bs
```diff
@@ -4970,15 +4970,32 @@ steps:
...
+     given <var>blobURLEntry</var> and the string "<code>navigation</code>".
```
annevk, line 4992: I think it would be clearer to first initialize a variable called _navigationOrEnvironment_ and then invoke "obtaining a blob object" once. At least I find code written that way easier to follow.

annevk (commented):
Thanks again for improving this aspect of `blob:` URLs!

fetch.bs
recvfrom: Weird, at the time I couldn't get this to work without using href even after the File API PR spec had landed, but it does seem to work just fine now 🤔 Thanks for calling this out - changed

fetch.bs
recvfrom: Thanks for the feedback, I wasn't sure if it was OK to assign values of different types into a variable. It's indeed simpler with only the one all to "obtaining a blob object". Thanks, and done!

recvfrom:
> Thanks again for improving this aspect of `blob:` URLs!

NP, thanks for all of your helpful review feedback and guidance!

annevk (approved):

PR #1795, opened 2024-12-12
Correct Response's statusText check

### LINKED ISSUE #1794: Empty statusText does not match the reason-phrase token production

In [initialize a response](https://fetch.spec.whatwg.org/#initialize-a-response) algorithm:
> 2. If init["statusText"] does not match the reason-phrase token production, then throw a TypeError.

but [reason-phrase](https://httpwg.org/specs/rfc9112.html#status.line) token production requires at least one characters:

> reason-phrase  = 1*( HTAB / SP / VCHAR / obs-text )

so if `init["statusText"]` is a empty string "", this step would fail.

annevk:
This has been wrong for a decade (see 38647562f94d396b77c62f77b4024ee21a1bb400) and nobody noticed until now.

Fixes #1794.

---

Omitting template as this has been implemented "correctly" for a decade as well.

***
<a href="https://whatpr.org/fetch/1795.html" title="Last updated on Dec 12, 2024, 8:08 AM UTC (75893d9)">Preview</a> | <a href="https://whatpr.org/fetch/1795/0ce45ae...75893d9.html" title="Last updated on Dec 12, 2024, 8:08 AM UTC (75893d9)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -8273,8 +8273,8 @@ enum ResponseType { "basic", "cors", "default", "error", "opaque", "opaqueredire
  <li><p>If <var>init</var>["{{ResponseInit/status}}"] is not in the range 200 to 599, inclusive,
  then <a>throw</a> a {{RangeError}}.
 
- <li><p>If <var>init</var>["{{ResponseInit/statusText}}"] does not match the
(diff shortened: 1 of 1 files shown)


F3n67u:
Looks good to me, thank you for the fix!

PR #1796, opened 2024-12-13
Return a `content-encoding` header for resource timing and more

guohuideng2024:
The major change is to add `content-encoding` to response header list. This PR also adds description on how `content-encoding` is determined. (content negotiation) 

The purpose is to pass such value to resource timing. Further details are available at
https://github.com/w3c/resource-timing/issues/381.

Note: Per discussion at 12/05/2024 webPerWG call (https://docs.google.com/document/d/1mpFDrAWuV6IgvJ1KiL9sgIlcboC5uArtF8r_oqS1Sco/edit?tab=t.0#heading=h.af6v74wysf4m), we decided to allow arbitrary "content-encoding" value at "fetch". We only filter such value at client side, before passing the value to resource timing.

Related PR to modify resource timing specification:
https://github.com/w3c/resource-timing/pull/411

Likely,  Discussed and agreed at Feb 29, 2024 [W3C WebPerf call](https://docs.google.com/document/d/1qPPCtpg1MyVw3GGmd6VKZCdCyqoDub6Xgc8ANnq8SRI/edit#heading=h.wkdzwqaypyq6), Chromium already receiving content-encoding through fetch. But not sure about other browsers.

[to be updated with a new link] https://chromium-review.googlesource.com/c/chromium/src/+/5958411

Chromium: https://issues.chromium.org/issues/327941462
Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1886107
WebKit: https://bugs.webkit.org/show_bug.cgi?id=271632
Deno (not for CORS changes): …
[MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed:
https://github.com/mdn/content/issues/32823

Bug: https://github.com/w3c/resource-timing/issues/381

***
<a href="https://whatpr.org/fetch/1796.html" title="Last updated on May 22, 2025, 4:49 PM UTC (8b79c02)">Preview</a> | <a href="https://whatpr.org/fetch/1796/07662d3...8b79c02.html" title="Last updated on May 22, 2025, 4:49 PM UTC (8b79c02)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -6329,8 +6329,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>Otherwise, if <var>codings</var>'s <a for=list>size</a> is greater than 1, then set
          <var>filteredCoding</var> to "<code>multiple</code>".
 
-         <li><p>Otherwise, if <var>codings</var>[0] is the empty string, or it is supported by the user agent,
-         and is a <a>byte-case-insensitive</a> match for an entry listed in the
+         <li><p>Otherwise, if <var>codings</var>[0] is the empty string, or it is supported by the
+         user agent, and is a <a>byte-case-insensitive</a> match for an entry listed in the
          <cite>HTTP Content Coding Registry</cite>, then set <var>filteredCoding</var> to the result
          of <a lt=byte-lowercased>byte-lowercasing</a> <var>codings</var>[0]. [[!IANA-HTTP-PARAMS]]
 


fetch.bs
```diff
@@ -5004,6 +5041,13 @@ steps:
 
      <li><p>Let <var>type</var> be <var>blob</var>'s {{Blob/type}}.
 
+     <li><p>Let <var>accept-coding</var> be the result of <a for="header list">getting</a>
+       `<code>Accept-Encoding</code>` from <var>request</var>'s <a for=request>header list</a>.
```
noamr, line 5045: Wrong indentation

fetch.bs
```diff
@@ -5105,10 +5152,19 @@ steps:
      <li><p>Let <var>mimeType</var> be <var>dataURLStruct</var>'s
      <a for="data: URL struct">MIME type</a>, <a lt="serialize a MIME type to bytes">serialized</a>.
 
+     <li><p>Let <var>accept-coding</var> be the result of <a for="header list">getting</a>
+       `<code>Accept-Encoding</code>` from <var>request</var>'s <a for=request>header list</a>.
```
noamr, line 5156: Wrong indentation

fetch.bs
```diff
@@ -5004,6 +5041,13 @@ steps:
 
      <li><p>Let <var>type</var> be <var>blob</var>'s {{Blob/type}}.
 
+     <li><p>Let <var>accept-coding</var> be the result of <a for="header list">getting</a>
+       `<code>Accept-Encoding</code>` from <var>request</var>'s <a for=request>header list</a>.
+        If <var>accept-encoding</var> is not null and the server selects one of the encoding options,
+        let <var>coding</var> be the selected encoding option; otherwise, i.e., no encoding is used,
```
noamr, line 5047: You can't have "let" in the middle of a phrase, see style in the rest of the document.

annevk:
Thanks for taking the time to pick this up. However, it doesn't seem like this addresses all the issues with #1742? I recommend studying the feedback on that PR.

fetch.bs
guohuideng2024: fixed.

fetch.bs
guohuideng2024: fxied.

guohuideng2024:
> Thanks for taking the time to pick this up. However, it doesn't seem like this addresses all the issues with #1742? I recommend studying the feedback on that PR.

Hi Anne! I think I should have put up some background information here.

1. You mentioned in #1742 that the spec must define how the value is determined. This PR is trying to do that. The value is a result of the "content negotiation" (determine what encoding should be used) so I tried to add that into the existing text.
Note that #1742 is a change similar to one for a previously added field `contentType`. But `contentEncoding` is very different, it's not an extracted MIME type, but a result of "content negotiation". So, this PR should be very different from #1742 

2. We originally thought that the filtering should happen at the "fetch" stage. But in the last web perf meeting Patrick brought up that the returned `contentEncoding` can be a proprietary value and that value is needed by service worker. So, the unfiltered value must be kept by the browser and the filtering should happen right before reported to resourceTiming.

https://docs.google.com/document/d/1mpFDrAWuV6IgvJ1KiL9sgIlcboC5uArtF8r_oqS1Sco/edit?tab=t.0#heading=h.af6v74wysf4m 

Therefore, in this `fetch` doc I didn't mention filtering. I mentioned "filtering" in the resourceTiming spec:
https://github.com/w3c/resource-timing/pull/411
And I am going to add more details about the filtering there.

Does this sound right to you? I am new to `fetch` and I may have missed a lot of things here. Thanks for your patience and guidance.
Guohui

fetch.bs
```diff
@@ -3946,6 +3949,40 @@ Content-Type:
...
+
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>values</var>:
+
+  <ol>
+   <li><p>If <var>candidateValue</var> is null, then set <var>candidateValue</var> to
+   <var>value</var>.
+
+   <li><p>Otherwise, if <var>value</var> is not <var>candidateValue</var>, return failure.
+  </ol>
```
annevk, line 3978: So the idea here is that duplicate values are okay as long as they are fully identical? So `gzip` and `GZIP` is not okay?

fetch.bs
```diff
@@ -3946,6 +3949,40 @@ Content-Type:
...
+
+ <li><p>If <var>candidateValue</var> is the empty string or has a <a for=/>code point</a> that is
+ not an <a for=/>ASCII digit</a>, then return null.
```
annevk, line 3981: Why would it only contain digits? Typical values for `Content-Encoding` are `gzip` or `deflate` as I understand it.

fetch.bs
```diff
@@ -5004,6 +5041,15 @@ steps:
...
      <li><p>Let <var>type</var> be <var>blob</var>'s {{Blob/type}}.
 
+     <li><p>Let <var>coding</var> be the empty string.
+
+     <li><p>Let <var>accept-coding</var> be the result of <a for="header list">getting</a>
+        `<code>Accept-Encoding</code>` from <var>request</var>'s <a for=request>header list</a>.
+        If <var>accept-encoding</var> is not null and the server selects one of the encoding options,
+        set <var>coding</var> to the selected encoding option; otherwise, i.e., no encoding is used,
+        set <var>coding</var> to <a href=https://httpwg.org/specs/rfc9110.html#field.accept-encoding>
+        <code>"identity"</code></a>.
```
annevk, line 5051: The indentation here doesn't follow the style guide.

How would we know what the server selects here? This doesn't make much sense to me.

fetch.bs
```diff
@@ -5105,10 +5154,21 @@ steps:
...
+
      <li><p>Return a new <a for=/>response</a> whose <a for=response>status message</a> is
      `<code>OK</code>`, <a for=response>header list</a> is « (`<code>Content-Type</code>`,
-     <var>mimeType</var>) », and <a for=response>body</a> is <var>dataURLStruct</var>'s
-     <a for="data: URL struct">body</a> <a for="byte sequence">as a body</a>.
+     <var>mimeType</var>), (`<code>Content-Encoding</code>`, <var>coding</var>) », and
+     <a for=response>body</a> is <var>dataURLStruct</var>'s <a for="data: URL struct">
+     body</a> <a for="byte sequence">as a body</a>.
```
annevk, line 5170: This also doesn't follow the formatting guidelines. Also, do we really want to return `Content-Encoding` for `data:` URLs? Why?

fetch.bs
noamr: I think you mean `ASCII code point`? Is unicode not allowed in `Content-Encoding`?

fetch.bs
guohuideng2024: I see! I changed the text. Now something like {gzip, GZIP} (case-insensitive match) would be ok. What do you think?

fetch.bs
guohuideng2024: sorry my bad. I removed this.

fetch.bs
guohuideng2024: Thanks guys for pointing out this is on client side. I thought I was on server side. :( 
Also Noam and I had discussion offline. All the filtering happens after fetch. So we don't filter the values here.
I changed the text accordingly.

fetch.bs
guohuideng2024: I now understand what data url is and `contentEncoding` doesn't apply to it.
So I removed the text.

fetch.bs
```diff
@@ -5113,6 +5160,7 @@ steps:
      `<code>OK</code>`, <a for=response>header list</a> is « (`<code>Content-Type</code>`,
      <var>mimeType</var>) », and <a for=response>body</a> is <var>dataURLStruct</var>'s
      <a for="data: URL struct">body</a> <a for="byte sequence">as a body</a>.
+
```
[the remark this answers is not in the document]
guohuideng2024: Done

fetch.bs
```diff
@@ -5092,7 +5138,8 @@ steps:
        <li><p>Set <var>response</var>'s <a for=response>header list</a> to «
        (`<code>Content-Length</code>`, <var>serializedSlicedLength</var>),
        (`<code>Content-Type</code>`, <var>type</var>), (`<code>Content-Range</code>`,
-       <var>contentRange</var>) ».
+       <var>contentRange</var>), (`<code>Content-Encoding</code>`, <var>coding</var>) ».
+
```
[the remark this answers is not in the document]
guohuideng2024: Done

fetch.bs
```diff
@@ -5008,6 +5043,15 @@ steps:
 
      <li><p>Let <var>type</var> be <var>blob</var>'s {{Blob/type}}.
 
+     <li><p>Let <var>coding</var> be the empty string.
```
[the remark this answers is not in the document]
guohuideng2024: Done.

fetch.bs
```diff
@@ -3945,6 +3948,38 @@ Content-Type:
...
+ <li><p>Let <var>values</var> be the result of
+ <a for="header list">getting, decoding, and splitting</a> `<code>Content-Encoding</code>` from
+ <var>headers</var>.
+
+ <li><p>If <var>values</var> is null, then return null.
+
+ <li><p>Let <var>candidateValue</var> be null.
+
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>values</var>:
```
noamr, line 3970: This is basically a no-duplicates algorithm?

fetch.bs
```diff
@@ -5008,6 +5043,15 @@ steps:
 
      <li><p>Let <var>type</var> be <var>blob</var>'s {{Blob/type}}.
 
+     <li><p>Let <var>coding</var> be the empty string.
+
+     <li><p>Let <var>accept-encoding</var> be the result of <a for="header list">getting</a>
+     `<code>Accept-Encoding</code>` from <var>request</var>'s <a for=request>header list</a>.
+     If <var>accept-encoding</var> is not null and the server selects one of the encoding options,
+     set <var>coding</var> to the selected encoding option; otherwise, i.e., no encoding is used,
+     set <var>coding</var> to <a href=https://httpwg.org/specs/rfc9110.html#field.accept-encoding>
```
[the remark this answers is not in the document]
guohuideng2024: Thanks for clarifying it for me. I wrongly thought I was on server side.
Modifed the text, leaving all filtering to resourceTiming spec.

fetch.bs
guohuideng2024: I know changed the text to allow duplicates. (as long as they case-insensitive match), I am not sure if this is the best option. what do you think?

noamr: I don't know. What do browsers currently do with duplicates and multiple values? There should be WPTs for this

guohuideng2024:
Updated the patch, I just added the content encoding to the body info struct, and add the clause that updates it.

fetch.bs
```diff
@@ -1176,6 +1178,7 @@ is a <a>byte-case-insensitive</a> match for one of
    <li>`<a http-header><code>Access-Control-Request-Headers</code></a>`
    <li>`<a http-header><code>Access-Control-Request-Method</code></a>`
    <li>`<code>Connection</code>`
+   <li>`<code>Content-Encoding</code>`
```
noamr, line 1181: Is this needed?

fetch.bs
```diff
@@ -6319,6 +6321,9 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
          `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
 
+         <li><p>Set <var>response</var>'s <a for=response>body info</a>'s
+         <a for="response body info">content encoding</a> to <var>codings</var>.
```
noamr, line 6325: Do UAs today do any observable processing to this "codings" value? What happens when there are multiple values? Resource timing would make this more observable so it needs to be specified

fetch.bs
```diff
@@ -3945,7 +3948,6 @@ Content-Type:
 </div>
 </div>
 
-
```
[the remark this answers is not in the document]
guohuideng2024: Done.

fetch.bs
guohuideng2024: I removed it. After looking it up I see `Content-Encoding` is not a prohibited header.

fetch.bs
guohuideng2024: My bad. I remember at some point in the past I mentioned I observed that known encoding method was checked with actually content, and mismatch would result a fetch failure. I had that kind of observation from my local tests. However I am not able to reproduce that today. And I also tried on gerrit. https://chromium-review.googlesource.com/c/chromium/src/+/6192941?checksPatchset=4&tab=checks (see the tests have passed)

I also looked and didn't find the code that does the check in "fetch". I tried a multiple encoding test and I see the value got into the HttpHeader.

So I believe I was wrong (and very sorry about that), and the current Chromium just keeps the value in the HttpHeader.

guohuideng2024:
very sorry for so many mistakes folks. Thanks for you guys' patence.

noamr:
> very sorry for so many mistakes folks. Thanks for you guys' patence.

No worries, we've all been there! (Or at least I have...)

fetch.bs
```diff
@@ -370,6 +370,8 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
  <dd>A number.
  <dt><dfn export for="response body info">content type</dfn> (default the empty string)
  <dd>An <a for=/>ASCII string</a>.
+ <dt><dfn export for="response body info">content encoding</dfn> (default the empty string)
```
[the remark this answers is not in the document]
guohuideng2024: Yes! Changed it to "a list of strings".

noamr:
I think this is OK with me, passing on to @annevk.

annevk:
I think there's a couple of problems here still:

1. "extracting header list values" is not a well-defined algorithm and I doubt it's interoperable. So just taking its return value and exposing it without it being fixed seems problematic.
2. The return values are not filtered. This creates a new communication channel. That's why we added minimized the set of MIME types exposed: https://mimesniff.spec.whatwg.org/#minimize-a-supported-mime-type
3. The tests don't match what is specified, e.g., 'Unrecognized content encoding value should be transformed to "unknown".' is not specified here.

noamr:
> I think there's a couple of problems here still:
> 
> 1. "extracting header list values" is not a well-defined algorithm and I doubt it's interoperable. So just taking its return value and exposing it without it being fixed seems problematic.
> 2. The return values are not filtered. This creates a new communication channel. That's why we added minimized the set of MIME types exposed: https://mimesniff.spec.whatwg.org/#minimize-a-supported-mime-type
> 3. The tests don't match what is specified, e.g., 'Unrecognized content encoding value should be transformed to "unknown".' is not specified here.

(2) and (3) are specified in resource-timing, I think it's OK to move them here but I don't have a strong opinion about that.

annevk:
I think it should be specified here as that matches how we do MIME types and that reduces the chances of someone inadvertently exposing the information. In other words: the guarantee should come from Fetch, not from the caller.

guohuideng2024:
> I think it should be specified here as that matches how we do MIME types and that reduces the chances of someone inadvertently exposing the information. In other words: the guarantee should come from Fetch, not from the caller.

The "raw" contentEncoding value can be arbitrary proprietary compression the app uses, and it's leaked as a response header.
So it's indeed a new communication channel that's created :(.

Meanwhile I think moving the filtering here guarantees that the only place where the raw `contentEncoding` is leaked is the fetch response header. I would say something here that contentEncoding is filtered when accessed anywhere else.

If there is any concern pls let me know. Thanks.

annevk:
To be clear, the header is not exposed to the website passively embedding the resource, but this getter is. I don't think I understand your suggestion, could you rephrase?

noamr:
> > I think it should be specified here as that matches how we do MIME types and that reduces the chances of someone inadvertently exposing the information. In other words: the guarantee should come from Fetch, not from the caller.
> 
> The "raw" contentEncoding value can be arbitrary proprietary compression the app uses, and it's leaked as a response header.
> So it's indeed a new communication channel that's created :(.
> 
> Meanwhile I think moving the filtering here guarantees that the only place where the raw `contentEncoding` is leaked is the fetch response header. I would say something here that contentEncoding is filtered when accessed anywhere else.
> 
> If there is any concern pls let me know. Thanks.
>  

Specifically, it needs to be explicitly filtered when assigned to the response body into struct.

guohuideng2024:
> > > I think it should be specified here as that matches how we do MIME types and that reduces the chances of someone inadvertently exposing the information. In other words: the guarantee should come from Fetch, not from the caller.
> > 
> > 
> > The "raw" contentEncoding value can be arbitrary proprietary compression the app uses, and it's leaked as a response header.
> > So it's indeed a new communication channel that's created :(.
> > Meanwhile I think moving the filtering here guarantees that the only place where the raw `contentEncoding` is leaked is the fetch response header. I would say something here that contentEncoding is filtered when accessed anywhere else.
> > If there is any concern pls let me know. Thanks.
> 
> Specifically, it needs to be explicitly filtered when assigned to the response body into struct.

Got it, Thanks! I updated the PR accordingly.

guohuideng2024:
> To be clear, the header is not exposed to the website passively embedding the resource, but this getter is. I don't think I understand your suggestion, could you rephrase?

I think the website can get the arbitrary value like this: (I am new to this area so please correct me if I am wrong)
```
let myCoding = myHeaders.get("Content-Encoding");  //  |myCoding| can be a proprietary compression, i.e., an arbitrary value.
```
And the reason for that is some use cases involving service workers. See
https://github.com/w3c/resource-timing/issues/381

but the `contentEncoding` field in `resourceTiming` is filtered, where only a few pre-determined values are permitted.

noamr:
> > To be clear, the header is not exposed to the website passively embedding the resource, but this getter is. I don't think I understand your suggestion, could you rephrase?
> 
> I think the website can get the arbitrary value like this: (I am new to this area so please correct me if I am wrong)
> 
> ```
> let myCoding = myHeaders.get("Content-Encoding");  //  |myCoding| can be a proprietary compression, i.e., an arbitrary value.
> ```

You would only get access to `myHeaders` if this is an actual `fetch` or via a service worker; Those channels are not always available.

> 
> And the reason for that is some use cases involving service workers. See [w3c/resource-timing#381](https://github.com/w3c/resource-timing/issues/381)
> 
> but the `contentEncoding` field in `resourceTiming` is filtered, where only a few pre-determined values are permitted.

Yea, so filtering them when assigning to the struct wouldn't change anything observable, but any future user of that struct would get the filtered value.

guohuideng2024:
Thank you Noam!

@annevk : Would you take one more look?  (I also left a response at https://github.com/WebKit/standards-positions/issues/467 )

fetch.bs
```diff
@@ -6319,6 +6321,24 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
          `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
 
+         <li><p>Let <var>filteredCoding</var> be "<code>unknown</code>".
```
annevk, line 6324: Why is this a distinct value from the empty string? It also seems to squat on the value space of the registry, which doesn't seem ideal?

fetch.bs
```diff
@@ -6319,6 +6321,24 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+         <li><p>If <var>codings</var> is null, set <var>filteredCoding</var> to the empty string.
```
annevk, line 6326: then set*

fetch.bs
```diff
@@ -6319,6 +6321,24 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+         <li><p>Otherwise, if <var>codings</var> contains two strings or more, set <var>filteredCoding</var> to
+         "<code>multiple</code>".
```
annevk, line 6329: then set*

And this should probably compare with Infra's size concept.

fetch.bs
```diff
@@ -6319,6 +6321,24 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+         <li><p>Otherwise, if <var>codings[0]</var> is the empty string, or it is supported by the user agent,
```
annevk, line 6331: `<var>codings</var>[0]`*

fetch.bs
```diff
@@ -6319,6 +6321,24 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+         and is listed in the <a href="https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding">
+         content encoding registry on IANA</a>, set <var>filteredCoding</var> to <var>codings[0]</var>.
```
annevk, line 6333: You want `<cite>HTTP Content Coding Registry</cite>` inline and then use a reference for the actual URL.

(And also apply the earlier comments.)

fetch.bs
```diff
@@ -6319,6 +6321,24 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+         <p class='note'>
+         "<code>identity</code>" is never supported as a coding. See details
```
annevk, line 6336: This seems bogus? I'm pretty sure we have tests around this value. Also, if we want to make requirements about this value they shouldn't be in a note.

annevk (commented):
I noticed you're participating on behalf of Microsoft. That means you cannot sign the contributor's agreement as an individual. Microsoft has already signed up for the Fetch Workstream so you have to join the relevant GitHub organization (MicrosoftWHATWGContributors) and make your membership thereof public.

annevk:
Now with the filtering in place it's probably slightly more reasonable to leave the existing parsing issue unsolved for now, assuming there's adequate test coverage.

fetch.bs
guohuideng2024: `unknown` means there is a compression that's not recognized by the browser
`empty string` means there is no compression.
We would like to distinguish the two. In this discussion thread: https://github.com/w3c/resource-timing/issues/381, [nhelfman](https://github.com/nhelfman) points out that the two situations need to be distinguishable.

fetch.bs
guohuideng2024: done.

fetch.bs
guohuideng2024: Done.

fetch.bs
guohuideng2024: Done.

fetch.bs
guohuideng2024: added "then". But I am not sure how to "compare with infra's size concept". I looked at the "infra" section but I didn't see  a title related to "size". Would you please help me understand what else needs to be done with this paragraph? Thanks.

fetch.bs
guohuideng2024: I removed this paragraph.

guohuideng2024:
Thanks @annevk!  I am working on MicrosoftWHATWGContributors membership right now.

fetch.bs
annevk: Okay, what about the value space concern?

fetch.bs
```diff
@@ -6319,6 +6321,20 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+         <li><p>Let <var>filteredCoding</var> be "<code>unknown</code>".
+
+         <li><p>If <var>codings</var> is null, then set <var>filteredCoding</var> to the empty string.
+
+         <li><p>Otherwise, if <var>codings</var> contains two strings or more, then set <var>filteredCoding</var> to
+         "<code>multiple</code>".
+
+         <li><p>Otherwise, if <var>codings</var>[0] is the empty string, or it is supported by the user agent,
+         and is listed in the <a href="https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding">
+         <cite>content encoding registry on IANA</cite></a>, then set <var>filteredCoding</var> to <var>codings</var>[0].
```
annevk, line 6333: What if the server specifies "`GZIP`"? Does it get lowercased?

fetch.bs
annevk: https://infra.spec.whatwg.org/#list-size

fetch.bs
annevk: This isn't quite done. The casing is incorrect and you didn't move the URL into the references section.

fetch.bs
guohuideng2024: 1. Could you help me understand the "value space concern" problem? 
2. Yes, I made two test cases where the `contentEncoding` value is filtered to `unknown` value.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/resource-timing/content-encoding.https.html;l=36

fetch.bs
guohuideng2024: I see! I made changes accordingly.

fetch.bs
guohuideng2024: Oh, I see another example of reference, so I did the same: with an inline link reference in place, I also made an entry in the "reference section".

The reference is not in `Specref` yet, so I submitted this PR:
https://github.com/tobie/specref/pull/860

I cannot verify this yet because the `specref` PR is not merged.
Do this look correct to you?

fetch.bs
guohuideng2024: 1. Yes all output should be lowercased and I made changes in text accordingly. Thanks for pointing out.
2. In the current text, a value is "allowed" as long as it "case insensitive matches" a registered value. And it will be lowercased before exposed.  Is it O.K.?

Thanks.

fetch.bs
annevk: Well, in principle someone could register `unknown` as a content coding and user agents could implement it, but this API would not be able to distinguish it from the `unknown` case.

fetch.bs
```diff
@@ -6319,6 +6321,23 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
          `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
 
+         <li><p>Let <var>filteredCoding</var> be "<code>unknown</code>".
+
+         <li><p>If <var>codings</var> is null, then set <var>filteredCoding</var> to the empty string.
+
+         <li><p>Otherwise, if <var>codings</var>'s <a for=list>size</a> is 2 or more, then set
```
annevk, line 6328:
```suggestion
         <li><p>Otherwise, if <var>codings</var>'s <a for=list>size</a> is greater than 1, then set
```

fetch.bs
```diff
@@ -6319,6 +6321,23 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+         <var>filteredCoding</var> to "<code>multiple</code>".
+
+         <li><p>Otherwise, if <var>codings</var>[0] is the empty string, or it is supported by the user agent,
+         and is a <a>byte-case-insensitive</a> match for an entry listed in the
+         <a href="https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding">
+         <cite>HTTP Content Coding Registry</cite></a> of [[!IANA-HTTP-PARAMS]], then set
+         <var>filteredCoding</var> to <var>codings</var>[0].
```
annevk, line 6335: I think it would be clearer if you did the lowercasing here. There's no reason to lowercase the other branches.

fetch.bs
guohuideng2024: Done.

fetch.bs
guohuideng2024: Good idea. Done.

fetch.bs
```diff
@@ -6319,6 +6321,23 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+
+         <li><p>Set <var>response</var>'s <a for=response>body info</a>'s
+         <a for="response body info">content encoding</a> to the result of
+         <a lt=byte-lowercased>byte-lowercasing</a>  <var>filteredCoding</var>.
```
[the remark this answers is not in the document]
guohuideng2024: Done. This sentence is modified, and the superfluous space is gone

fetch.bs
guohuideng2024: Done.

fetch.bs
guohuideng2024: Done. Do you think I should withdraw the PR to the specref? Thanks.

fetch.bs
guohuideng2024: Ah I see!  I think a solution is to make `unknown` a reserved word and it cannot be used as `contentEncoding` in the response header. There is already one reserved word `identity`.

Do you think I should try to make that happen? I couldn't find how to propose changes to that iana "http parameters". There is a "contact" section on that page but the emails are obviously out of date. (they are @sun.com :) )

annevk: Can you email Mark Nottingham <mnot@mnot.net> and copy annevk@annevk.nl (that's me) with the information stated at https://httpwg.org/specs/rfc9110.html#content.coding.extensibility?

guohuideng2024: I sent email to Mark (and copied Anne).

guohuideng2024:
update:
1. Since I couldn't register a new reserved "unknown" value for "content coding", per the discussion with Roy and Yoav, I am replacing "unknown" with "_". "_" cannot be used as a "content coding" value so it won't cause value space problem.
2.  I removed the internal reference because the reference has been registered into "specref".

fetch.bs
```diff
@@ -6319,6 +6321,23 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
...
+         <li><p>Let <var>filteredCoding</var> be "<code>_</code>".
+
+         <li><p>If <var>codings</var> is null, then set <var>filteredCoding</var> to the empty string.
+
+         <li><p>Otherwise, if <var>codings</var>'s <a for=list>size</a> is greater than 1, then set
+         <var>filteredCoding</var> to "<code>multiple</code>".
+
+         <li><p>Otherwise, if <var>codings</var>[0] is the empty string, or it is supported by the user agent,
+         and is a <a>byte-case-insensitive</a> match for an entry listed in the
+         <a href="https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding">
```
annevk, line 6333: There's no need for the `<a>` here as we already have the [[ ]] reference. Though the [[ ]] reference should come at the end of the paragraph after the final dot.

fetch.bs
```diff
@@ -6319,6 +6321,23 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
          `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
 
+         <li><p>Let <var>filteredCoding</var> be "<code>_</code>".
+
+         <li><p>If <var>codings</var> is null, then set <var>filteredCoding</var> to the empty string.
+
+         <li><p>Otherwise, if <var>codings</var>'s <a for=list>size</a> is greater than 1, then set
```
annevk, line 6328: This doesn't work when _codings_ is failure.

fetch.bs
```diff
@@ -6319,6 +6321,23 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
          <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
          `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
 
+         <li><p>Let <var>filteredCoding</var> be "<code>_</code>".
```
annevk, line 6324: Are you updating the tests to account for this change? Can we make this `_unknown` you think? And then maybe instead of the empty string below we could make that `_missing`?

fetch.bs
guohuideng2024: I see! I removed the <a> and moved the [[]] reference to the end of the paragraph. Is the current patch good now?

fetch.bs
guohuideng2024: Fixed. "failure" and "null" would result in "the empty string".

fetch.bs
guohuideng2024: Yes. The CL is being reviewed right now:
https://chromium-review.googlesource.com/c/chromium/src/+/6506187

The following doc should be the one specifying the syntax restrictions on the values: https://www.rfc-editor.org/rfc/rfc8941.html and there is no restriction saying the value cannot start with "_". So, `_unkown` is better than `_` but I think it still requires reserving it with ietf.  Let me email Roy to get some clarification on this.

On the other hand I think "the empty string" is more conventional than `_missing` so I fee strongly for the `the empty string`.

annevk: All of this is somewhat unprecedented so I'm not sure why the empty string would clearly convey it's missing (and not that its value was the empty string for instance).

It seems that `tchar` cannot contains `@` so we could use `@unknown` and `@missing` and be fine.

fetch.bs
annevk: Why would failure be the same as missing? It seems much more like unknown to me.

guohuideng2024: "unknown" means we do receive a value (or values) that we don't want to expose, i.e., we received a value but we filter it out. This is a very particular use case requested by app. They will want to distinguish using a proprietary encoding.

fetch.bs
guohuideng2024: 1) 
As far as I know the convention is that "missing" ==> "empty string", in the performance API. (not http header)

An example would be "content type". In the fetch standard, there is this sentence:
"If mimeType is not failure, then set bodyInfo’s content type to the result of minimizing a supported MIME type given mimeType."

And for the mime type extraction, around the phrase "To extract a MIME type from a header list headers... ", the mimeType is a failure if the value is null.

Therefore, a "failure" or "null" mimeType value results in the "content type" value in bodyInfo being an empty string.
Then, in the resource timing specification, it's this "content type value in bodyinfo" to be exposed in resource timing.

This PR makes a similar change and the "content encoding" is following the same path: it is filtered here before gets into the same "bodyInfo"(and later be exposed in resource timing). And just like "content type", a "null" and "failure" value results an empty string.

2)
Yes I think `@unknown` is a good idea :)  I don't expect it there would be objects. However, how about let me update in https://github.com/w3c/resource-timing/issues/381 and wait for two days before we finalize this PR?( I will replace `_` with `@unknown`) 

Do you have any (other) concerns at this point? Would you let me know if you have any?

Cheers,
Guohui

guohuideng2024:
In the http spec I found the syntax restriction of the "content coding", which gives us a complete list of possibilities.
(see the end of this comment for more details)

 `@unknown` looks good to me but I would like to confirm we don't want to use `?unknown` nor `[unknown]` instead? @annevk 

Hopefully nobody objects and I will update this PR with `@unknown` this Wednesday.

Cheers,
Guohui

+++++++++ details from http spec +++++++++++++++++++++++++++
http spec actually defines the syntax restriction of "content-coding" parameter.
Link to the http spec: https://www.rfc-editor.org/rfc/rfc2616
The "content-coding" is a "token", and the "token" is defined as the following:

   token          = 1*<any CHAR except CTLs or separators>
   separators     = "(" | ")" | "<" | ">" | "@"
                  | "," | ";" | ":" | "\" | <">
                  | "/" | "[" | "]" | "?" | "="
                  | "{" | "}" | SP | HT
// where:
CTL = <any US-ASCII control character
(octets 0 - 31) and DEL (127)>
SP = <US-ASCII SP, space (32)>
HT = <US-ASCII HT, horizontal-tab (9)>

guohuideng2024:
I updated the PR with `@unknown` since there is no objects after asking for potential objects.
The CL to chromium and WPT tests is also updated accordingly.

annevk (commented):
The change looks good, but wrapping at 100 columns doesn't seem to be done properly. I could address that, but you're welcome to it as well.

Also, https://github.com/whatwg/fetch/pull/1796#pullrequestreview-2753138273 still needs to be resolved to satisfy the IPR bot.

guohuideng2024:
> The change looks good, but wrapping at 100 columns doesn't seem to be done properly. I could address that, but you're welcome to it as well.
> 
> Also, [#1796 (review)](https://github.com/whatwg/fetch/pull/1796#pullrequestreview-2753138273) still needs to be resolved to satisfy the IPR bot.

I will fix the 100-column wrapping and the agreement problem.

annevk (approved):
Great, this looks good. When do you expect the tests to be ready? We'll want to land those and this change around the same time.

guohuideng2024:
> Great, this looks good. When do you expect the tests to be ready? We'll want to land those and this change around the same time.

I am going to get the changes to the tests landed next week. I will update here as soon as it's done.

guohuideng2024:
@annevk  The tests are updated :)

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/resource-timing/content-encoding.https.html;l=36

I greatly appreciate the patience and guidance I received here!

PR #1821, opened 2025-04-30
Handle request's window when from browser UI

domenic:
In https://github.com/whatwg/html/pull/11250 we are working to specify browser UI-initiated navigations better. Without this change, it is unclear what to set their request's window to. They will have null clients, so "client" is not appropriate. They will show prompts, so "no-window" is not appropriate. And setting it to whatever's shown currently in the being-navigated window is not correct.

This introduces a new value, "from-browser-ui", which indicates that browser UI is initiating the request, and so any prompts should still be shown, but on a neutral backdrop, not associated with a specific Window object.

   * This matches Chrome's behavior
   * This mostly matches Firefox's behavior, although for Firefox, if you go to http://labs.w0s.jp/http/authorization/ and then enter http://labs.w0s.jp/http/authorization/basic/ in the address bar, it does not blank out the page. (Whereas if you go https://example.com/ and then enter http://labs.w0s.jp/http/authorization/basic/ in the address bar, it *does* blank out the page.) I think it's OK to say that Firefox is violating this minor part of a "should" requirement.
   * I haven't tested Safari yet.
   * This seems basically impossible to test.
   * Chromium: N/A
   * Gecko: Doesn't feel worth it for this minor divergence
   * WebKit: … TODO after testing, in the event they do something very strange
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A, too detailed

***
<a href="https://whatpr.org/fetch/1821.html" title="Last updated on Apr 30, 2025, 7:33 AM UTC (7070b90)">Preview</a> | <a href="https://whatpr.org/fetch/1821/5a96806...7070b90.html" title="Last updated on Apr 30, 2025, 7:33 AM UTC (7070b90)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1666,17 +1666,48 @@ of the <a for="environment">target browsing context</a>'s <a>active document</a>
 <a>environment settings object</a>.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-window>window</dfn>
-("<code>no-window</code>", "<code>client</code>", or an
-<a>environment settings object</a> whose
-<a for="environment settings object">global object</a> is a
-{{Window}} object). Unless stated otherwise it is
-"<code>client</code>".
+<dfn export for=request id=concept-request-window>window</dfn>, that is "<code>no-window</code>",
+"<code>from-browser-ui</code>", "<code>client</code>", or an <a>environment settings object</a>
+whose <a for="environment settings object">global object</a> is a {{Window}} object. Unless stated
+otherwise it is "<code>client</code>".
 
-<p class=note>The "<code>client</code>" value is changed to "<code>no-window</code>" or
-<a for=/>request</a>'s <a for=request>client</a> during <a lt=fetch for=/>fetching</a>. It provides
-a convenient way for standards to not have to explicitly set <a for=/>request</a>'s
-<a for=request>window</a>.
+<div class=note>
+ <p>This is used to determine whether and where to show necessary UI for the request, such as
+ authentication prompts or client certificate dialogs.
+
+ <dl>
+  <dt>"<code>no-window</code>"
+  <dd>No UI is shown; usually the request fails with a <a>network error</a>.
+
+  <dt>"<code>from-browser-ui</code>"
+  <dd>This request was initiated by browser UI, and so any UI shown will not be associated to a
+  specific window.
+
+  <dt>"<code>client</code>"
+  <dd>This value will automatically be changed to either "<code>no-window</code>" or the request's
+  <a for=request>client</a> during <a lt=fetch for=/>fetching</a>. This provides a convenient way
+  for standards to not have to explicitly set a request's <a for=request>window</a>.
+
+  <dt>an <a>environment settings object</a>
+  <dd>The UI shown will be associated with the specified {{Window}} object.
+ </dl>
+</div>
+
+<p>The <dfn for=request>appropriate user prompt context</dfn> for a <a for=/>request</a>
+<var>request</var> is determined as follows:
+
+<ol>
+ <li><p><a>Assert</a>: <var>request</var>'s <a for=request>window</a> is not "<code>client</code>".
+
+ <li><p>If the request's <a for=request>window</a> is an <a>environment settings object</a>, then
+ the prompt should occur in a way attributable to <var>request</var>'s <a for=request>window</a>.
+
+ <li><p>Otherwise, if <var>request</var>'s <a for=request>window</a> is
+ "<code>from-browser-ui</code>", then the prompt should occur in a neutral context, e.g., on top of
+ a blank page.
+
+ <li><p>Otherwise, there is no appropriate user prompt context.
+</ol>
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1666,17 +1666,48 @@ of the <a for="environment">target browsing context</a>'s <a>active document</a>
...
+
+<ol>
+ <li><p><a>Assert</a>: <var>request</var>'s <a for=request>window</a> is not "<code>client</code>".
+
+ <li><p>If the request's <a for=request>window</a> is an <a>environment settings object</a>, then
+ the prompt should occur in a way attributable to <var>request</var>'s <a for=request>window</a>.
+
+ <li><p>Otherwise, if <var>request</var>'s <a for=request>window</a> is
+ "<code>from-browser-ui</code>", then the prompt should occur in a neutral context, e.g., on top of
+ a blank page.
```
annevk, line 1707: Should we note here that the address bar should probably contain some variation on the request's current URL?

fetch.bs
```diff
@@ -6156,10 +6186,8 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
     <p>If the HTTP request results in a TLS client certificate dialog, then:
 
     <ol>
-     <li><p>If <var>request</var>'s <a for=request>window</a>
-     is an <a>environment settings object</a>, make the dialog
-     available in <var>request</var>'s
-     <a for=request>window</a>.
+     <li><p>If <var>request</var> has an <a>appropriate user prompt context</a>, then make the
+     dialog available in <var>request</var>'s <a>appropriate user prompt context</a>.
```
annevk, line 6190:
```suggestion
     dialog available therein.
```
Might be a bit nicer.

annevk (commented):
I was about to approve this, but now I'm doubting it again. Is "from-browser-ui" really correct here?

If I'm on a.com and click a link to b.com and b.com requires an end-user dialog, in that case we want the blank window as well, right? Perhaps the window cannot really be inferred from the client at all?

domenic:
> If I'm on a.com and click a link to b.com and b.com requires an end-user dialog, in that case we want the blank window as well, right?

You're right.

Let me document my findings:

|                        |Chrome|Firefox|Safari|
|------------------------|------|-------|------|
|Same-origin link-click  |Blank, URL bar updates|Old page backdrop, no URL bar update|???|
|Cross-origin link click |Blank, URL bar updates|Blank, URL bar updates|???|
|Same-origin URL bar nav |Blank, URL bar updates|Old page backdrop, URL bar updates|???|
|Cross-origin URL bar nav|Blank, URL bar updates|Blank, URL bar updates|???|

Here:

* Same-origin is http://labs.w0s.jp/http/authorization/ to https://labs.w0s.jp/http/authorization/basic/, via either link click or entering the second URL in the URL bar.
* Cross-origin is https://boom-bath.glitch.me/link-to-basic-auth.html to https://labs.w0s.jp/http/authorization/basic/, via either link click or entering the second URL in the URL bar.

and

* "blank" = the backdrop is blanked out
* "old page backdrop" = the referrer page is left in place as a backdrop, although usually tinted darker
* "no URL bar update" = the URL bar contains the referrer page URL during the prompt display
* "URL bar updates" = the URL bar updates to the new page URL during the prompt display

Honestly, Chrome's approach seems best here. It's really the response URL that is prompting you; the referrer (i.e., the request's window) is not really the party we should be attributing the prompt to.

We could still track the top-level traversable ("browser tab") if we thought that was useful, as a hint to implementations about where to display the prompt. That is, implementations should display the prompt in the place where the response is eventually going to show up, if the navigation succeeds.

But I think tying it to the referrer `Window`, as the current spec does, doesn't make sense.

Maybe if testing Safari reveals that it does something more like Firefox, we could consider speccing that sort of behavior... but I still don't like it. (I'm sick and working from home this week so I don't have access to my testing Mac. But maybe you could fill in the results?)

domenic:
I guess part of the issue is that TLS client certs are caused by the request, but WWW-Authenticate or proxy-authentication entries are caused by the response. I'm only testing WWW-Authenticate above.

valenting:
Firefox's behaviour was changed in [791594 - 401 password prompt spoofing thing](https://bugzilla.mozilla.org/show_bug.cgi?id=791594)
@Trikolon or @dveditz  might have some thoughts about this.

annevk:
Based on https://jpassing.com/2021/09/27/do-browsers-use-client-certificates-to-authenticate-the-user-the-device-or-both/ and some other search results it seems that client certificate dialogs can also be driven by the response.

I agree that would be a much better way forward here. We probably do want to track the navigable to make sure we know where to present the dialog. And that navigable then needs to figure out whether a blank interstitial is required.

Trikolon:
For Firefox we only changed the behavior (blank page, url bar updated) for cross origin sites to avoid spoofing. Demo: https://eviltrap.site/trap/http-auth-prompt-spoof/. For same-origin that's not really a concern.

I'm in favor of aligning with Chrome and inserting a blank interstitial for both cross and same origin cases to make the behavior more consistent.

PR #1823, opened 2025-05-09
Replace request's "window" with "traversable for user prompts"

domenic:
After much discussion in #1820 and #1821, this seems to be more correct. User prompts are not tied to specific `Window`s, and certainly not to the client's `Window`. They are tied to a traversable navigable, i.e. "browser tab", and that browser tab should ideally be blanked out and updated to reflect that it's the destination that's causing the prompt, not the request client.

Closes #1820. Closes #1821. Helps https://github.com/whatwg/html/pull/11250 specify browser-initiated navigations better.

   * This seems to match all browser behavior.
   * Not really testable in WPT, due to https://github.com/web-platform-tests/wpt/issues/16019 I believe.
   * Chromium: N/A
   * Gecko: N/A, although maybe a bug to encourage the blanking-out behavior would be appreciated?
   * WebKit: N/A, although maybe a bug to encourage the blanking-out behavior would be appreciated?

---

Places to update, from https://dontcallmedom.github.io/webdex/w.html#window%40%40request%40dfn:

- [x] Private Aggregation API: https://github.com/patcg-individual-drafts/private-aggregation-api/pull/182
- [ ] Topics API https://github.com/patcg-individual-drafts/topics/pull/390
- [x] Attribution Reporting https://github.com/WICG/attribution-reporting-api/pull/1519
- [ ] Shared Storage API https://github.com/WICG/shared-storage/pull/235
- [x] Content Security Policy 3 https://github.com/w3c/webappsec-csp/pull/724
- [x] Federated Credential Management API https://github.com/w3c-fedid/FedCM/pull/723
- [ ] Permissions Policy https://github.com/w3c/webappsec-permissions-policy/pull/565
- [x] WebDriver BiDi https://github.com/w3c/webdriver-bidi/pull/918

***
<a href="https://whatpr.org/fetch/1823.html" title="Last updated on May 14, 2025, 9:57 AM UTC (e485e3e)">Preview</a> | <a href="https://whatpr.org/fetch/1823/5a96806...e485e3e.html" title="Last updated on May 14, 2025, 9:57 AM UTC (e485e3e)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5566,9 +5566,6 @@ run these steps:
     "<code>error</code>", then set <var>httpFetchParams</var> to <var>fetchParams</var> and
     <var>httpRequest</var> to <var>request</var>.
 
-    <p class=note>If user prompts are possible, then we need to clone <var>request</var> because
-    the user agent might need to re-send <var>request</var>.
-
    <li>
     <p>Otherwise:
 
@@ -5587,6 +5584,12 @@ run these steps:
      <li><p>Set <var>httpFetchParams</var>'s <a for="fetch params">request</a> to
      <var>httpRequest</var>.
     </ol>
+
+    <p class=note>If user prompts or redirects are possible, then the user agent might need to
+    re-send the request with a new set of headers after the user answers the prompt or the redirect
+    location is determined. At that time, the original request body might have been partially sent
+    already, so we need to clone the request (including the body) beforehand so that we have a
+    spare copy available.
 
    <li>
     <p>Let <var>includeCredentials</var> be true if one of
--- a/fetch.bs
+++ b/fetch.bs
@@ -4469,16 +4469,16 @@ the response. [[!HTTP-CACHING]]
    "<code>no-traversable</code>".
 
    <li>
-    <p>If <var>request</var>'s <a for=request>client</a> is non-null, then:
+    <p>If <var>request</var>'s <a for=request>client</a> is non-null:
 
     <ol>
      <li><p>Let <var>global</var> be <var>request</var>'s <a for=request>client</a>'s
      <a for="environment settings object">global object</a>.
 
-     <li><p>If <var>global</var> is a {{Window}} object, and <var>global</var>'s
+     <li><p>If <var>global</var> is a {{Window}} object and <var>global</var>'s
      <a for=Window>navigable</a> is not null, then set <var>request</var>'s
-     <a for=request>traversable for user prompts</a> to <var>global</var>'s <a for=Window>navigable</a>'s
-     <a for=navigable>traversable navigable</a>.
+     <a for=request>traversable for user prompts</a> to <var>global</var>'s
+     <a for=Window>navigable</a>'s <a for=navigable>traversable navigable</a>.
     </ol>
   </ol>
 
@@ -5561,12 +5561,13 @@ run these steps:
 
   <ol>
    <li>
-    <p>If <var>request</var>'s <a for=request>traversable for user prompts</a> is "<code>no-traversable</code>" and
-    <var>request</var>'s <a for=request>redirect mode</a> is "<code>error</code>", then set
-    <var>httpFetchParams</var> to <var>fetchParams</var> and <var>httpRequest</var> to
-    <var>request</var>.
+    <p>If <var>request</var>'s <a for=request>traversable for user prompts</a> is
+    "<code>no-traversable</code>" and <var>request</var>'s <a for=request>redirect mode</a> is
+    "<code>error</code>", then set <var>httpFetchParams</var> to <var>fetchParams</var> and
+    <var>httpRequest</var> to <var>request</var>.
 
-    <p class=note>If user prompts are possible, then we need to clone <var>request</var> because ???
+    <p class=note>If user prompts are possible, then we need to clone <var>request</var> because
+    the user agent might need to re-send <var>request</var>.
 
    <li>
     <p>Otherwise:
--- a/fetch.bs
+++ b/fetch.bs
@@ -4355,8 +4355,9 @@ the response. [[!HTTP-CACHING]]
    <li><p><var>request</var>'s <a for=request>mode</a> is "<code>same-origin</code>",
    "<code>cors</code>", or "<code>no-cors</code>"
 
-   <li><p><var>request</var>'s <a for=request>client</a>'s
-   <a for="environment settings object">global object</a> is a {{Window}} object
+   <li><p><var>request</var>'s <a for=request>client</a> is not null, and <var>request</var>'s
+   <a for=request>client</a>'s <a for="environment settings object">global object</a> is a
+   {{Window}} object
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -1666,17 +1666,38 @@ of the <a for="environment">target browsing context</a>'s <a>active document</a>
 <a>environment settings object</a>.
 
 <p>A <a for=/>request</a> has an associated
-<dfn export for=request id=concept-request-window>window</dfn>
-("<code>no-window</code>", "<code>client</code>", or an
-<a>environment settings object</a> whose
-<a for="environment settings object">global object</a> is a
-{{Window}} object). Unless stated otherwise it is
-"<code>client</code>".
+<dfn export for=request id=concept-request-window>traversable for user prompts</dfn>, that is
```
annevk, line 1669: I'm trying to write out "end user" where I can myself. I'm not actually sure where we are consistency-wise, but is that something we can strife for? I'm happy to document it as the term to use.

fetch.bs
```diff
@@ -1666,17 +1666,38 @@ of the <a for="environment">target browsing context</a>'s <a>active document</a>
...
-a convenient way for standards to not have to explicitly set <a for=/>request</a>'s
-<a for=request>window</a>.
+<p>When displaying a user interface associated with a request in that request's
+<a for=request>traversable for user prompts</a>, the user agent should update the address bar to
+display something derived from the request's <a for=request>current URL</a> (and not, e.g., leave
+it at its previous value, derived from the URL of the request's initiator). Additionally, the user
+agent should avoid displaying content from the request's initiator in the
+<a for=request>traversable for user prompts</a>, especially in the case of cross-origin requests.
+Displaying a blank page behind such prompts is a good way to fulfill these requirements. Failing to
+follow these guidelines can confuse users as to which origin is responsible for the prompt.
```
annevk, line 1700:
```suggestion
follow these guidelines can confuse end users as to which origin is responsible for the prompt.
```

fetch.bs
```diff
@@ -4439,15 +4461,36 @@ the response. [[!HTTP-CACHING]]
...
- <a for="environment settings object">origin</a>.
+ <li>
+  <p>If <var>request</var>'s <a for=request>traversable for user prompts</a> is "<code>client</code>":
+
+  <ol>
+   <li><p>Set <var>request</var>'s <a for=request>traversable for user prompts</a> to
+   "<code>no-traversable</code>".
+
+   <li>
+    <p>If <var>request</var>'s <a for=request>client</a> is non-null, then:
```
annevk, line 4472:
```suggestion
    <p>If <var>request</var>'s <a for=request>client</a> is non-null:
```

fetch.bs
```diff
@@ -4439,15 +4461,36 @@ the response. [[!HTTP-CACHING]]
...
+
+    <ol>
+     <li><p>Let <var>global</var> be <var>request</var>'s <a for=request>client</a>'s
+     <a for="environment settings object">global object</a>.
+
+     <li><p>If <var>global</var> is a {{Window}} object, and <var>global</var>'s
```
annevk, line 4478:
```suggestion
     <li><p>If <var>global</var> is a {{Window}} object and <var>global</var>'s
```

fetch.bs
```diff
@@ -5517,10 +5560,13 @@ run these steps:
   <a for="fetch params">canceled</a>:
 
   <ol>
-   <li><p>If <var>request</var>'s <a for=request>window</a> is "<code>no-window</code>" and
-   <var>request</var>'s <a for=request>redirect mode</a> is "<code>error</code>", then set
-   <var>httpFetchParams</var> to <var>fetchParams</var> and <var>httpRequest</var> to
-   <var>request</var>.
+   <li>
+    <p>If <var>request</var>'s <a for=request>traversable for user prompts</a> is "<code>no-traversable</code>" and
```
annevk, line 5564: Exceeds 100 columns jingle

fetch.bs
```diff
@@ -5517,10 +5560,13 @@ run these steps:
...
+    <var>request</var>'s <a for=request>redirect mode</a> is "<code>error</code>", then set
+    <var>httpFetchParams</var> to <var>fetchParams</var> and <var>httpRequest</var> to
+    <var>request</var>.
+
+    <p class=note>If user prompts are possible, then we need to clone <var>request</var> because ???
```
annevk, line 5569: Because it's one of the scenarios where we might have to repeat request.

annevk (approved):
This looks good to me modulo some comments. @ricea and @noamr are maybe willing to review this too.

fetch.bs
domenic: Hmm, why? It seems excessively verbose...

fetch.bs
domenic: I added that, although I'm not 100% sure I understand when that would happen. Who gets the request the first time, and who gets it the second time? Does this occur for all 3 possible prompt causes (WWW-Authenticate, PACs, and TLS client certs) or just some of them?

annevk: It happens when you do a request and the server returns a challenge. Then you have to repeat the request addressing the challenge. This happens with HTTP authentication at least. I'm not sure about client certificates as they happen at the TLS level, so maybe not?

fetch.bs
domenic: I dunno, it seems a lot of churn across the ecosystem, and causes me to wonder "what's the difference between an end user and a user"? (Or is it "end-user"?)

I don't really understand the more respectful bit either.

domenic: In case it helps, after all the PRs linked in the OP land, only one spec (Permissions Policy) will actually use "traversable for user prompts", so it'd be pretty easy to change later if we do decide on a larger ecosystem movement from user -> end user.

fetch.bs
```diff
@@ -4334,7 +4355,8 @@ the response. [[!HTTP-CACHING]]
    <li><p><var>request</var>'s <a for=request>mode</a> is "<code>same-origin</code>",
    "<code>cors</code>", or "<code>no-cors</code>"
 
-   <li><p><var>request</var>'s <a for=request>window</a> is an <a>environment settings object</a>
+   <li><p><var>request</var>'s <a for=request>client</a>'s
```
noamr, line 4358: I believe `client` can be null, no? Perhaps there can be a helper algorithm to return request's document or window and performs the appropriate null checks?

noamr (commented):
All in all the direction looks good, one comment so far.

fetch.bs
domenic: Yeah... not sure about the helper algorithm as I see people reaching for it a bit too eagerly, but maybe as a followup if we take a look at existing patterns in a bit more detail. Will fix the null dereference, and thanks for catching it.

fetch.bs
annevk: I think https://datatracker.ietf.org/doc/html/rfc8890 captures best why I'd like to make this change, though arguably you could try to define "user" in the same way.

Anyway, happy to drop this for now.

annevk (approved):
r+ again for clarity. Feel free to land this as you see fit.

PR #1832, opened 2025-05-30
Update HTTPFetch algorithm to support new SW Handle Fetch returns

quasi-mod:
This will update the HttpFetch algorithm, particularly the handling of the response for Service Worker fetch (`handle fetch` step). Currently, the response of handle fetch step is assumed to return the `response` type (or null, if the ServiceWorker couldn't handle the fetch and need to fallback to the network request). However, we have changed the `handle fetch` step to also return `service worker timing info` when the ServiceWorker static routing API used, so that the corresponding timing information are correctly exposed when the ServiceWorker could not handle the fetch. To support this new return type, we need to update the handling of the response of `handle fetch`.

To expose the Service Worker Timing Info to the resource timing API, we also associate them to the Fetch Timing Info so that it could be later referenced.

   * Chromium is proposing the change
   * Waiting for reponse from other vendors (will update once we get support).
   * https://wpt.fyi/results/service-workers/service-worker/tentative/static-router?label=experimental&label=master&aligned
   * Chromium: https://g-issues.chromium.org/issues/41496865?pli=1&authuser=0
   * Gecko: https://github.com/mozilla/standards-positions/issues/1031
   * WebKit: https://github.com/WebKit/standards-positions/issues/354

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -352,6 +352,9 @@ following <a for=struct>items</a>: [[RESOURCE-TIMING]] [[NAVIGATION-TIMING]]
  <dt><dfn export for="fetch timing info">final connection timing info</dfn> (default null)
  <dd>Null or a <a for=/>connection timing info</a>.
 
+ <dt><dfn export for="fetch timing info">service worker timing info</dfn> (default null)
+ <dd>Null or a <a for=/>service worker timing info</a>.
+
  <dt><dfn export for="fetch timing info">server-timing headers</dfn> (default « »)
  <dd>A <a for=/>list</a> of strings.
  <dt><dfn export for="fetch timing info">render-blocking</dfn> (default false)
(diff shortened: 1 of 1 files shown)


quasi-mod:
Let me make this a PR and ask for opinions about this.

quasi-mod:
@yoshisatoyanagisawa Could you check if this make sense? Thanks!

fetch.bs
```diff
@@ -5248,48 +5251,61 @@ these steps:
...
+     <ol>
+     <li><p>If<var>fetchResponse</var> is <a for=/>response</a>, then:
+      <ol>
+       <li><p>Set <var>response</var> to <var>fetchResponse</var>.
+       <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
+       <a for="fetch timing info">final service worker start time</a> to
+       <var>serviceWorkerStartTime</var>.
+       <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
+       <a for="fetch timing info">service worker timing info</a> to
+       <var>response</var>'s <a for="response">service worker timing info</a>.
```
yoshisatoyanagisawa: "2.2.6. Responses" should be updated to have associated `service worker timing info`?

fetch.bs
```diff
@@ -5248,48 +5251,61 @@ these steps:
...
-     <var>serviceWorkerStartTime</var>.
+    <p>If <var>fetchResponse</var> is non-null, then:
+     <ol>
+     <li><p>If<var>fetchResponse</var> is <a for=/>response</a>, then:
+      <ol>
+       <li><p>Set <var>response</var> to <var>fetchResponse</var>.
+       <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
+       <a for="fetch timing info">final service worker start time</a> to
+       <var>serviceWorkerStartTime</var>.
+       <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
```
yoshisatoyanagisawa: It is not a blocker but resource-timing may not not have a trivial way to know `response`, right?  I just wondered why we need to copy the timing info.

fetch.bs
quasi-mod: Thanks. Actually, we already have an associated service worker timing info in the response, and was thinking that we did not need another associated one?

https://github.com/whatwg/fetch/blob/main/fetch.bs#L2508

fetch.bs
quasi-mod: I have to double check, but I think you are right. We already set the timing info directly to fetch timing info, so maybe we do not need to set them to the response.

quasi-mod: Let me double check and update the draft

fetch.bs
yoshisatoyanagisawa: Thank you.  I overlooked that.  Then the sentence sounds reasonable.

yoshisatoyanagisawa: Ah, there seems not be the way to mark this resolved in my side.

fetch.bs
```diff
@@ -5577,48 +5580,61 @@ these steps:
...
    <a for="fetch params">cross-origin isolated capability</a>. [[!HTML]] [[!SW]]
 
    <li>
-    <p>If <var>response</var> is non-null, then:
 
-    <ol>
-     <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
-     <a for="fetch timing info">final service worker start time</a> to
-     <var>serviceWorkerStartTime</var>.
+    <p>If <var>fetchResponse</var> is non-null, then:
```
```suggestion
    <p>If <var>fetchResponse</var> is non-null:
```

fetch.bs
```diff
@@ -5577,48 +5580,61 @@ these steps:
...
+     <ol>
+     <li><p>If<var>fetchResponse</var> is <a for=/>response</a>, then:
```
```suggestion
     <li>
      <p>If<var>fetchResponse</var> is <a for=/>response</a>:
```

fetch.bs
```diff
@@ -5577,48 +5580,61 @@ these steps:
...
+      <ol>
+       <li><p>Set <var>response</var> to <var>fetchResponse</var>.
+       <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
```
annevk: Newline before the `<li>`s please.

fetch.bs
```diff
@@ -5577,48 +5580,61 @@ these steps:
...
 
-      <p>then return a <a>network error</a>.
+        <p>then return a <a>network error</a>.
+      </ol>
+     <li><p>If<var>fetchResponse</var> is <a for=/>service worker timing info</a>, then:
+      <ol>
+       <li><p>Set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
+       <a for="fetch timing info">service worker timing info</a> to
+       <var>fetchResponse</var>.
+      </ol>
```
```suggestion
     <li><p>Otherwise, if <var>fetchResponse</var> is a <a for=/>service worker timing info</a>,
     then set <var>fetchParams</var>'s <a for="fetch params">timing info</a>'s
     <a for="fetch timing info">service worker timing info</a> to <var>fetchResponse</var>.
```

fetch.bs
```diff
@@ -5577,48 +5580,61 @@ these steps:
...
     </ol>
-  </ol>
+   </ol>
```
annevk: This seems suspect given the `<li>` indentation below. We use single-space indentation.

quasi-mod:
@annevk Thanks for taking a look offline! 
I have rebased on top of recent change.

Could you publish the comments you made during our offline chat at TPAC last week? Thanks!

annevk (commented):
Oh, I thought I had. Fortunately GitHub did preserve them. Sorry about that!

fetch.bs
quasi-mod: Ah, thanks. Reverted the spacing

fetch.bs
quasi-mod: Just to clarify, I think this is copying the response's timing_info's service worker timing info into the fetchParams timing info's service worker timing info, which does not expose response, so I feel like this is right.

annevk:
@quasi-mod I decided to take a pass at the formatting myself since it's your first contribution. While doing that I ended up fixing a couple of minor errors and also realized that the non-null check is redundant with the type checks. The result is a much cleaner diff. Could you please verify that all my changes were indeed editorial? (Note that I forced pushed to your branch to also rebase it so if you want to make further changes please be aware of that.)

I'd also like to hear back from @yoshisatoyanagisawa about the unresolved thread above.

fetch.bs
yoshisatoyanagisawa: Will it be used for the redirect?
I revisited this and see how fetchParams are used, and it looks used "If internalResponse’s [status](https://fetch.spec.whatwg.org/#concept-response-status) is a [redirect status](https://fetch.spec.whatwg.org/#redirect-status):".

If so, lgtm.

quasi-mod: Thanks for checking. 
On any case, we use the fetchParam's timing info in the [fetch response handover](https://whatpr.org/fetch/1832/60e9ff5...c989ab9.html#fetch-finale), and build the resource timing info in step 4-8, so I think it is OK to keep the draft as is.

yoshisatoyanagisawa: Thanks for the link.
And yes, [fetch response handover](https://fetch.spec.whatwg.org/#fetch-finale), which is called inside [Main fetch](https://fetch.spec.whatwg.org/#main-fetch), seems to set `timingInfo` which is set to be `fetchParam`'s `timing info` in Step 1.  The `timingInfo` gets eventually be used in Step 4.2.8.
Then, I understand we cannot avoid updating `timing info` in `fetchParam`.

Then, stepping back and think why we made a `response` to have `timing info`.  My understanding is that Output of an algorithm can usually be one or none.  Therefore, we need to choose either returning a tuple of response and `timing info`, or piggyback the current `response`.  We chose latter.  I think either to choose is up to the fetch editor and an author

Then, this PR sounds good to me.

quasi-mod: Thanks! Will mark this as resolved

quasi-mod:
@annevk Thank you for updating! I have reviewed the changes and seems to be OK.

annevk (approved):

Issue #808, opened 2018-09-11
After update to 3.0.0 Webpack build is failing on line 72


belcherj:
Updating from 2.0.4 to 3.0.0 webpack is failing, see error below. I assume I need to use a different loader but wanted to raise the issue since a quick web search and perusing the issues here did not result in any others having the issue.  This method of including fetch via webpack is shared a lot in examples so there will probably be others.

In my webpack config I include fetch globally like this:

```
new webpack.ProvidePlugin( {
	fetch: 'imports-loader?this=>global!exports-loader?global.fetch!whatwg-fetch',
} ),
```

```
    ERROR in ./node_modules/whatwg-fetch/fetch.js (./node_modules/imports-loader?this=>global!./node_modules/exports-loader?global.fetch!./node_modules/whatwg-fetch/fetch.js) 72:0
    Module parse failed: 'import' and 'export' may only appear at the top level (72:0)
    You may need an appropriate loader to handle this file type.
    | }
    |
    > export function Headers(headers) {
    |   this.map = {};
    |
     @ ./plugins/wccom-plugins/geo-page-replace/assets/setup.js 1:0-79
```

annevk:
This is not the repository you're looking for. Perthaps github/fetch? This repository defines the Fetch Standard.

jakearchibald:
https://github.com/github/fetch is what you're looking for.

belcherj:
Ohhh noooo, my bad.  Thanks!

PR #1841, opened 2025-07-10
Add the "speculationrules" destination

domenic:
Part of https://github.com/whatwg/html/pull/11426, for requests initiated by the `Speculation-Rules` HTTP header.

---

This should be merged a bit after https://github.com/whatwg/html/pull/11426. Otherwise it will reference the WICG draft.

   * See https://github.com/whatwg/html/pull/11426
   * This is tested by all tests of the header: https://github.com/web-platform-tests/wpt/blob/c669bcf970fe50b56e0ee53de9c2097a012dafc7/speculation-rules/prefetch/resources/ruleset.py#L40-L43
   * See https://github.com/whatwg/html/pull/11426
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/40289

***
<a href="https://whatpr.org/fetch/1841.html" title="Last updated on Jul 10, 2025, 5:24 AM UTC (c649784)">Preview</a> | <a href="https://whatpr.org/fetch/1841/5261b51...c649784.html" title="Last updated on Jul 10, 2025, 5:24 AM UTC (c649784)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1780,6 +1780,7 @@ the empty string,
 "<code>script</code>",
 "<code>serviceworker</code>",
 "<code>sharedworker</code>",
+"<code>speculationrules</code>",
 "<code>style</code>",
 "<code>track</code>",
 "<code>video</code>",
@@ -1821,13 +1822,16 @@ not always relevant and might require different behavior.
    <th>CSP directive
    <th>Features
   <tr>
-   <td rowspan=21>""
+   <td rowspan=22>""
    <td>"<code>report</code>"
-   <td rowspan=2>&mdash;
+   <td rowspan=3>&mdash;
    <td>CSP, NEL reports.
   <tr>
    <td>"<code>document</code>"
    <td>HTML's navigate algorithm (top-level only).
+  <tr>
+   <td>"<code>speculationrules</code>"
+   <td>The [:Speculation-Rules:] header
   <tr>
    <td>"<code>frame</code>"
    <td><code>child-src</code>
@@ -1887,7 +1891,7 @@ not always relevant and might require different behavior.
   <tr>
    <td>"<code>webidentity</code>"
    <td><code>connect-src</code>
-   <td><code>Federated Credential Management requests</code>
+   <td>Federated Credential Management requests
   <tr>
    <td>"<code>worker</code>"
    <td><code>child-src</code>, <code>script-src</code>, <code>worker-src</code>
@@ -7560,7 +7564,7 @@ dictionary RequestInit {
   any window; // can only be set to null
 };
 
-enum RequestDestination { "", "audio", "audioworklet", "document", "embed", "font", "frame", "iframe", "image", "json", "manifest", "object", "paintworklet", "report", "script", "sharedworker", "style",  "track", "video", "worker", "xslt" };
+enum RequestDestination { "", "audio", "audioworklet", "document", "embed", "font", "frame", "iframe", "image", "json", "manifest", "object", "paintworklet", "report", "script", "sharedworker", "speculationrules", "style",  "track", "video", "worker", "xslt" };
 enum RequestMode { "navigate", "same-origin", "no-cors", "cors" };
 enum RequestCredentials { "omit", "same-origin", "include" };
 enum RequestCache { "default", "no-store", "reload", "no-cache", "force-cache", "only-if-cached" };


domenic:
This is ready to merge!

annevk (commented):
I'm a bit torn on the spelling consistency with existing destinations as I believe Dictionary Compression and FedCM are breaking it (not sure if intentionally, but probably too late to change).

fetch.bs
```diff
@@ -1821,13 +1822,16 @@ not always relevant and might require different behavior.
    <th>CSP directive
    <th>Features
   <tr>
-   <td rowspan=21>""
+   <td rowspan=22>""
    <td>"<code>report</code>"
-   <td rowspan=2>&mdash;
+   <td rowspan=3>&mdash;
```
annevk, line 1827: Are you sure this is accurate? Wouldn't this be governed by connect-src at least?

domenic: No, per discussion with CSP folks, header-initiated fetches are not generally governed by CSP at all. The reasoning is that CSP is a mechanism for protection against HTML injection defense, not header-injection defense; if an attacker can inject headers, then they can mess up your CSP and do all sorts of bad things.

I guess `Link` might be governed by CSP, but that's more of a historical accident...

annevk: I'm not sure I agree with that.

I think it also depends on the order the headers are processed in. If the CSP header is processed first (and I hope that it is), any fetch following it would automatically be subject to CSP.

domenic: Yeah, not sure, probably something to discuss with the CSP spec editors if you want to make changes to that policy...

annevk: @antosart mind chiming in?

I think in the HTML standard we create a policy container from navigation response first thing and I wouldn't expect us to fetch anything else before that is complete. (This happens deep into https://html.spec.whatwg.org/#create-navigation-params-by-fetching if I'm reading things correctly.)

Are you also suggesting for instance that we would not take Referrer Policy into account here?

antosart: If I understand correctly, fetches do indeed happen after the policy container is created from the navigation response. However, the current proposal is to have Speculation-Rules headers explicitly bypass CSP by allowlisting the "speculationrules" destination in CSP (https://github.com/w3c/webappsec-csp/pull/776). Digging a bit, that seem to be the result of experiments and discussion (see https://chromestatus.com/feature/5123809745829888?gate=5093909324365824).

I expect the reason for this to be rooted in the attempt of CSP to do (at least) two things at the same time, i.e. control script execution and gate outgoing requests (which in the case of `script-src` end up colliding even if they are conceptually distinct - fetching a script content is not the same as executing it). Gating the Speculation-Rules header behind script-src would make it impossible to use the feature with [strict CSP](https://csp.withgoogle.com/docs/strict-csp.html). Gating it behind something else (e.g. connect-src) could be somehow confusing given that the other way to specify speculationrules is in a script tag (for which script-src does apply). It's not totally clear to me what the best option is here, but I think it would make sense to have some record/explanation of why the choice was made.

annevk: I don't think that the usage of `<script>` necessarily implies `script-src`. `script-src` makes sense when the result is script execution, but if we were to use `<script>` for JSON modules or some such it shouldn't use `script-src` but the same thing that the "`json`" destination devolves into. I think the same kind of rationale applies here, which I suspect means `connect-src` in the end.

I know some at Google don't care about the exfiltration angle of CSP, but I don't think we have ever reached agreement on giving up on it, so until we get there I think new fetch contexts should play by the same rules as existing ones.

vickiez: Hi all, I’d like to reopen this discussion as we’re working to support external speculation rules to make them easier for developers to deploy (https://github.com/whatwg/html/pull/11697 and https://github.com/w3c/webappsec-csp/pull/808).

In [CSP #808](https://github.com/w3c/webappsec-csp/pull/808), we propose introducing `speculation-rules-src` to control inline and external speculation rule sets. This again raises the question of whether the directive should also apply to rule sets fetched through the `Speculation-Rules` header.

https://github.com/w3c/webappsec-csp/pull/808#issuecomment-5398653996 highlights that a major use case for the header is platforms and CDNs such as Shopify and Cloudflare automatically enabling speculation for many customers. Requiring those customers to add `speculation-rules-src` could introduce compat risk and deployment friction, potentially reducing the benefit.

Is anyone here in contact with the affected platforms or customers so we can better understand the impact? Also, can we land the `speculationrules` destination change separately here to unblock the HTML change while we discuss the appropriate CSP behavior for header fetches?

PR #1843, opened 2025-07-25
WIP: Add a mechanism to fetch to track progress

lukewarlow:
This is not ready for review this is WIP.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1843.html" title="Last updated on Oct 9, 2025, 4:18 PM UTC (cd935ac)">Preview</a> | <a href="https://whatpr.org/fetch/1843/0a041eb...cd935ac.html" title="Last updated on Oct 9, 2025, 4:18 PM UTC (cd935ac)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -115,6 +115,7 @@ urlPrefix:https://tc39.es/ecma262/#;type:dfn;spec:ecma-262
 <pre class=link-defaults>
 spec:dom; type:dfn; text:element
 spec:dom; type:dfn; text:event;
+spec:xhr; type:event; text:progress;
 spec:infra; type:dfn; text:implementation-defined
 </pre>
 
@@ -8524,6 +8525,11 @@ callback FetchMonitorCallback = undefined (FetchMonitor monitor);
 
 [Exposed=(Window,Worker)]
 interface FetchMonitor : EventTarget {
+  readonly attribute double requestLoaded;
+  readonly attribute double responseLoaded;
+  readonly attribute double requestTotal;
+  readonly attribute double responseTotal;
+
   attribute EventHandler onrequestprogress;
   attribute EventHandler onresponseprogress;
 };
@@ -8607,6 +8613,10 @@ method steps are:
 
    <li><p>Set <var>monitor</var> to a {{FetchMonitor}}.
 
+   <li><p>Set <var>monitor</var>'s requestTotal to <var>request</var>'s
+   <a for=request>body</a>'s <a for=body>length</a>, if <var>request</var>'s <a for=request>body</a>
+   is non-null; otherwise 0.
+
    <li><p>Let <var>args</var> be « <var>monitor</var> ».
 
    <li><p>[=invoke|Invoke=] <var>monitorCallback</var> with <var>args</var>
@@ -8614,25 +8624,19 @@ method steps are:
    and return <var>p</var>.
   </ol>
 
- <li><p>Let <var>requestBodyTransmitted</var> be 0.
-
- <li><p>Let <var>requestBodyLength</var> be <var>request</var>'s <a for=request>body</a>'s
- <a for=body>length</a>, if <var>request</var>'s <a for=request>body</a> is non-null;
- otherwise 0.
-
- <li><p>Assert: <var>requestBodyLength</var> is an integer.
-
  <li>
   <p>Let <var>processRequestBodyChunkLength</var>, given a <var>bytesLength</var>, be these steps:
 
   <ol>
-   <li><p>Increase <var>requestBodyTransmitted</var> by <var>bytesLength</var>.
+   <li><p>If <var>monitor</var> is null, then return.
+
+   <li><p>Increase <var>monitor</var>'s requestLoaded by <var>bytesLength</var>.
 
    <li><p>If not roughly 50ms has passed since these steps were last invoked, then return.
 
-   <li><p>If <var>monitor</var> is not null, then <a>fire a progress event</a> named
-   <a event><code>requestprogress</code></a> at <var>monitor</var> with <var>requestBodyTransmitted</var>
-   and <var>requestBodyLength</var>.
+   <li><p><a>Fire a progress event</a> named
+   <a event><code>requestprogress</code></a> at <var>monitor</var> with
+   <var>monitor</var>'s requestLoaded and <var>monitor</var>'s requestTotal.
   </ol>
 
  <li>
@@ -8641,8 +8645,9 @@ method steps are:
   <ol>
    <li><p>If <var>monitor</var> is null, then return.
 
-   <li><p><a>Fire a progress event</a> named <a event><code>requestprogress</code></a> at <var>monitor</var>
-   with <var>requestBodyTransmitted</var> and <var>requestBodyLength</var>.
+   <li><p><a>Fire a progress event</a> named <a event><code>requestprogress</code></a> at
+   <var>monitor</var> with <var>monitor</var>'s requestLoaded and
+   <var>monitor</var>'s requestTotal.
   </ol>
 
  <li>
@@ -9260,6 +9265,12 @@ WHATWG community to ensure input from people familiar with the subject.
    <td>Progress has begun.
    <td>Once.
    <td>First.
+  <tr>
+   <th><a event><code>progress</code></a>
+   <td>In progress.
+   <td>Once or more.
+   <td>After <code>loadstart</code> has been
+    <a>dispatched</a>.
   <tr>
    <th><a event><code>requestprogress</code></a>
    <td>In progress request.


fetch.bs
```diff
@@ -7558,6 +7559,14 @@ dictionary RequestInit {
   RequestDuplex duplex;
   RequestPriority priority;
   any window; // can only be set to null
+  FetchObserverCallback observer;
+};
+
+callback FetchObserverCallback = undefined (FetchObserver requestObserver, FetchObserver responseObserver);
+
+[Exposed=(Window,Worker)]
+interface FetchObserver : EventTarget {
```
lukewarlow: Observer probably isn't the best name here?

fetch.bs
```diff
@@ -8615,10 +8681,17 @@ method steps are:
...
+ <li><p>Set <var>controller</var> to the result of calling <a for=/>fetch</a> given
+ <var>request</var> with <a for=fetch><i>processResponse</i></a> set to <var>processResponse</var>,
+ <a for=fetch><i>processRequestBodyChunkLength</i></a> set to <var>processRequestBodyChunkLength</var>,
+ and <a for=fetch><i>processRequestEndOfBody</i></a> set to <var>processRequestEndOfBody</var>.
+
  <li><p>Return <var>p</var>.
 </ol>
 </div>
 
+TEMPORARY <dfn id=event-fetchobserver-progress event for=FetchObserver><code>progress</code></dfn>
```
lukewarlow: Need to define the event somewhere properly. (potentially taking the event definitions from XHR?)

fetch.bs
```diff
@@ -9132,6 +9205,119 @@ done only by navigations). The <a>fetch controller</a> is also used to
 <a for="fetch controller">process the next manual redirect</a> for <a for=/>requests</a> with
 <a for=request>redirect mode</a> set to "<code>manual</code>".
 
+<h2 id=interface-progressevent>Interface {{ProgressEvent}}</h2>
```
lukewarlow, line 9206: This is taken from XHR and should be removed from XHR in a corresponding PR.

fetch.bs
```diff
@@ -9132,6 +9205,119 @@ done only by navigations). The <a>fetch controller</a> is also used to
...
+<p>The
+<dfn attribute for=ProgressEvent><code>lengthComputable</code></dfn>,
+<dfn attribute for=ProgressEvent><code>loaded</code></dfn>, and
+<dfn attribute for=ProgressEvent><code>total</code></dfn>
+getter steps are to return the value they were initialized to.
+
+
+<h3 id=firing-events-using-the-progressevent-interface>Firing events using the {{ProgressEvent}} interface</h3>
+
+<p>To <dfn id=concept-event-fire-progress>fire a progress event</dfn> named <var>e</var> at
```
lukewarlow, line 9236: Should fetch really fire sync events? It would match XHR but I wonder if these should be queued instead?

fetch.bs
```diff
@@ -9132,6 +9205,119 @@ done only by navigations). The <a>fetch controller</a> is also used to
...
+attribute initialized to <var>transmitted</var>, and if <var>length</var> is not 0, with the
+{{ProgressEvent/lengthComputable}} attribute initialized to true and the {{ProgressEvent/total}}
+attribute initialized to <var>length</var>.
+
+
+<h3 id=suggested-names-for-events-using-the-progressevent-interface>Suggested names for events using the {{ProgressEvent}} interface</h3>
+
+<p><em>This section is non-normative.</em>
+
+<p>The suggested {{Event/type}}
```
lukewarlow, line 9248: Most of these events aren't used in Fetch so this table should probably be kept in the XHR spec.

fetch.bs
```diff
@@ -8584,10 +8593,67 @@ method steps are:
...
+   <li><p>If <var>hasUploadListeners</var> is false, then return.
+
+   <li><p>Increase <var>requestBodyTransmitted</var> by <var>bytesLength</var>.
+
+   <li><p><a>Fire a progress event</a> named <a event><code>progress</code></a> at <var>requestObserver</var>
+   with <var>requestBodyTransmitted</var> and <var>requestBodyLength</var>.
+  </ol>
+
+ <li>
+  <p>Let <var>processResponse</var> given a <var>response</var> be these steps:
```
lukewarlow, line 8654: If we want to do progress tracking for response we'll need to do it differently to XHR, XHR consumes the body inside of this algorithm but fetch obviously can't.

One idea is to wrap response's body with a wrapper readable stream that allows use to fire progress events when the body actually gets read. This way we don't consume anything to report progress, (we can make it so we only wrapper if there's a responseObserver with a progress event listeners).

fetch.bs
```diff
@@ -7558,6 +7559,14 @@ dictionary RequestInit {
   RequestDuplex duplex;
   RequestPriority priority;
   any window; // can only be set to null
+  FetchObserverCallback observer;
+};
+
+callback FetchObserverCallback = undefined (FetchObserver requestObserver, FetchObserver responseObserver);
```
lukewarlow: Firefox has the below IDL. I split it so the observers still fire events named progress but it's maybe fine to keep it as a single eventtarget and just change the event names?

```webidl
[Exposed=Window]
callback interface ObserverCallback {
  undefined handleEvent(FetchObserver observer);
};

enum FetchState {
  // Pending states
  "requesting", "responding",
  // Final states
  "aborted", "errored", "complete"
};

[Exposed=(Window,Worker),
 Pref="dom.fetchObserver.enabled"]
interface FetchObserver : EventTarget {
  readonly attribute FetchState state;

  // Events
  attribute EventHandler onstatechange;
  attribute EventHandler onrequestprogress;
  attribute EventHandler onresponseprogress;
};
```

fetch.bs
```diff
@@ -7558,6 +7559,14 @@ dictionary RequestInit {
   RequestDuplex duplex;
   RequestPriority priority;
   any window; // can only be set to null
+  FetchObserverCallback observer;
```
lukewarlow: Should this actually be in RequestInit? I'm thinking it should probably be in a FetchInit subtype that inherits RequestInit? That way it isn't also passed to the Request constructor?

lukewarlow: If we leave it in requestinit we'll have to add a way to track it across from Request into the fetch call if you do

```js
const req = new Request(...);
fetch(req);
```

fetch.bs
```diff
@@ -8584,10 +8593,67 @@ method steps are:
         https://github.com/whatwg/dom/issues/1031#issuecomment-1233206400 -->
   </ol>
 
+ <li><p>Let <var>hasUploadListeners</var> be false.
```
lukewarlow: Should probably be `hasRequestListeners`

lukewarlow:
An alternative design that could be considered is to just make Request and event target and fire progress events at the request object directly. Would require authors to construct a request attach a listener and then pass to fetch but might be okay?

jakearchibald:
@lukewarlow I wonder if that could get confusing in cases where request/response isn't used in a fetch. Eg, if we made these things transferrable/clonable, would you expect progress events when getting a `Response` from indexeddb?

lukewarlow:
Yeah that's a good point. Perhaps it's just okay that they only fire progress within fetch? Or perhaps it would be doable to fire them elsewhere (if it would even make sense)?

My thinking is if the vast majority of cases support progress events then the nicer API is probably a worthy trade off?

Response is actually where things are less clear in general though.

annevk:
`Request` and `Response` are very much designed to be data classes. Note also how `fetch()` and `fetchLater()` essentially create a copy as one of their first steps. I don't think they should become more stateful than what is needed for the `Body` mixin.

lukewarlow:
Okay in that case I think the addition should be moved to a fetchInit rather than being part of requestInit, easy enough to do. That way it keeps it separate from request entirely.

mindplay-dk:
> Okay in that case I think the addition should be moved to a fetchInit rather than being part of requestInit, easy enough to do. That way it keeps it separate from request entirely.

what does that look like? 

a third argument to fetch?

or like `interface FetchInit extends RequestInit` with extra options?

lukewarlow:
Yeah the FetchInit dictionary would extend RequestInit

fetch.bs
```diff
@@ -8584,10 +8593,67 @@ method steps are:
...
+   <li><p>Set <var>responseObserver</var> to a {{FetchObserver}}.
+
+   <li><p>Let <var>args</var> be « <var>requestObserver</var>, <var>responseObserver</var> ».
+
+   <li><p>[=invoke|Invoke=] <var>observerCallback</var> with <var>args</var>
+   and <code>"rethrow"</code>. If this throws an exception, <a for=/>reject</a> <var>p</var> with it
+   and return <var>p</var>.
+
+   <li><p>If one or more <a event><code>progress</code></a> event listeners were added to
+   <var>requestObserver</var>, then set <var>hasUploadListeners</var> to true.
```
jakearchibald: Does this mean there's a specific time-window where listeners can be added? Isn't that… bad?

fetch.bs
```diff
@@ -8584,10 +8593,67 @@ method steps are:
...
  <li>
-  <p><p>Set <var>controller</var> to the result of calling <a for=/>fetch</a> given
-  <var>request</var> and <a for=fetch><i>processResponse</i></a> given <var>response</var> being
-  these steps:
+  <p>Let <var>processRequestEndOfBody</var> be these steps:
+
+  <ol>
+   <li><p>If <var>hasUploadListeners</var> is false, then return.
+
+   <li><p>Increase <var>requestBodyTransmitted</var> by <var>bytesLength</var>.
```
jakearchibald: `bytesLength` isn't defined.

fetch.bs
jakearchibald: Why couldn't the event be fired here https://whatpr.org/fetch/1843.html#http-network-fetch:~:text=Increase%20response%E2%80%99s%20body%20info%E2%80%99s%20encoded%20size%20by%20bytes%E2%80%99s%20length%2E ?

jakearchibald:
I wonder if this should be a single observer object, making it easier to pass around. This also makes more sense if we want somewhere to put an abort event.

Also, this would be a good opportunity to move to a more modern pattern of having the details on the target object, rather than on the event object.

```ts
interface FetchObserver extends EventTarget {
  requestLengthComputable: boolean;
  responseLengthComputable: boolean;

  requestLoaded: number;
  responseLoaded: number;

  requestTotal: number;
  responseTotal: number;

  onrequestprogress: (event: Event) => void;
  onresponseprogress: (event: Event) => void;
}
```

We could include `requeststart` and `responsestart` events which indicate when the `LengthComputable` properties have settled.

Terreii:
Awsome!
But should/could this work be synced with the [AI/Translation people](https://github.com/webmachinelearning/translation-api)? They are working on a [progress method](https://developer.chrome.com/docs/ai/translator-api#language-support), too.
It would be too bad if we had two (three if you count XMLHttpRequest) methods for progress observing on the web.

```js
const translator = await Translator.create({
  sourceLanguage: 'es',
  targetLanguage: 'fr',
  monitor(m) {
    m.addEventListener('downloadprogress', (e) => {
      console.log(`Downloaded ${e.loaded * 100}%`);
    });
  },
});
```

I think @domenic is responsible for it.

jakearchibald:
@Terreii thanks for raising this. `monitor` is a good name actually.

jakearchibald:
https://github.com/webmachinelearning/translation-api/issues/61

fetch.bs
lukewarlow: Essentially because there's no access to it from within the JS fetch() definition.

fetch.bs
lukewarlow: Renamed to FetchMonitor

fetch.bs
```diff
@@ -8512,7 +8513,19 @@ otherwise false.
...
+  FetchMonitorCallback monitor;
+};
+
+callback FetchMonitorCallback = undefined (FetchMonitor monitor);
+
+[Exposed=(Window,Worker)]
+interface FetchMonitor : EventTarget {
+  attribute EventHandler onrequestprogress;
+  attribute EventHandler onresponseprogress;
 };
```
noamr, line 8535: Perhas we can have `requestFinished` and `responseFinished` promises as well?

fetch.bs
lukewarlow: Yeah, it was based on XHR's design but I'm not really sure of the reason for it. I've removed for now.

lukewarlow: I think I understand why XHR is done that way, we can avoid that because of how this API is designed, if someone has set the monitor option that's enough of an opt-in to do to the firing of events etc.

lukewarlow:
Just coming back to this to say it's still on my radar and will try to make progress on this. I think from discussions this overall direction seems okay though?

PR #1851, opened 2025-08-19
WebDriver BiDi emulated user-agent

sadym-chromium:
Support for user-agent emulations. WebDriver BiDi specification: https://github.com/w3c/webdriver-bidi/pull/947

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -5889,6 +5889,21 @@ run these steps:
    the <a data-lt="structured field token">token</a> <code>prefetch</code>) in
    <var>httpRequest</var>'s <a for=request>header list</a>.
 
+   <li>
+    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">does not
+    contain</a> `<code>User-Agent</code>`, then:
+
+    <ol>
+     <li><p>Let <var>userAgent</var> be the result of running the [=WebDriver BiDi emulated
+     User-Agent=] with <var>httpRequest</var>'s <a for=request>client</a>.
+
+     <li><p>If <var>userAgent</var> is null, set <var>userAgent</var> to <a>default
+     `<code>User-Agent</code>` value</a>.
+
+     <li><p><a for="header list">Append</a> (`<code>User-Agent</code>`, <var>userAgent</var>) to
+     <var>httpRequest</var>'s <a for=request>header list</a>.
+    </ol>
+
    <li><p>If <var>httpRequest</var>'s <a for=request>header list</a>
    <a for="header list">does not contain</a> `<code>User-Agent</code>`, then user agents should
    <a for="header list">append</a> (`<code>User-Agent</code>`,


sadym-chromium:
@noamr PTAL

fetch.bs
```diff
@@ -5889,11 +5889,20 @@ run these steps:
...
    <var>httpRequest</var>'s <a for=request>header list</a>.
 
-   <li><p>If <var>httpRequest</var>'s <a for=request>header list</a>
-   <a for="header list">does not contain</a> `<code>User-Agent</code>`, then user agents should
-   <a for="header list">append</a> (`<code>User-Agent</code>`,
-   <a>default `<code>User-Agent</code>` value</a>) to <var>httpRequest</var>'s
-   <a for=request>header list</a>.
+   <li>
+    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">does not
+    contain</a> `<code>User-Agent</code>`, then:
```
```suggestion
    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a for="header list">does not
    contain</a> `<code>User-Agent</code>`:
```

fetch.bs
```diff
@@ -5889,11 +5889,20 @@ run these steps:
...
+
+    <ol>
+     <li><p>Let <var>userAgent</var> be the result of running the <a>WebDriver BiDi emulated
+     User-Agent</a> with <var>httpRequest</var>'s <a for=request>client</a>.
```
annevk: No newlines inside `<a>`.

fetch.bs
```diff
@@ -5889,11 +5889,20 @@ run these steps:
...
+
+     <li><p>If <var>userAgent</var> is null, set <var>userAgent</var> to <a>default
+     `<code>User-Agent</code>` value</a>.
```

fetch.bs
annevk: There's also a problem in that WebDriver hands us a string but we need a byte sequence.

sadym-chromium: I assume instead it should be:
> [UTF-8 encoding](https://encoding.spec.whatwg.org/#utf-8-encode) of <var>userAgent</var>?

sadym-chromium: Oh, apparently it is "[isomorphic encoded](https://infra.spec.whatwg.org/#isomorphic-encode)"

fetch.bs
sadym-chromium: done

fetch.bs
```diff
@@ -1420,6 +1420,21 @@ downloads. This format of range header value can be set using <a>add a range hea
...
+<ol>
+ <li><p>Let <var>userAgent</var> be the <a>WebDriver BiDi emulated User-Agent</a> for
+ <var>environment</var>.
+
+ <li><p>If <var>userAgent</var> is not null, then return <a>isomorphic encoded</a>
+ <var>userAgent</var>.
+
+ <li><p>Return <a>default `<code>User-Agent</code>` value</a>.
+</ol>
+</div>
```
```suggestion
<div algorithm>
<p>To get the <a>environment settings object</a> <var>environment</var>'s
<dfn export>environment default `<code>User-Agent</code>` value</dfn>:

<ol>
 <li><p>Let <var>userAgent</var> be the <a>WebDriver BiDi emulated User-Agent</a> for
 <var>environment</var>.

 <li><p>If <var>userAgent</var> is non-null, then return <var>userAgent</var>,
 <a>isomorphic encoded</a>.

 <li><p>Return the <a>default `<code>User-Agent</code>` value</a>.
</ol>
</div>
```

fetch.bs
```diff
@@ -5986,11 +6001,18 @@ run these steps:
...
+    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a
+    for="header list">does not contain</a> `<code>User-Agent</code>`, then user agents should run
+    the following steps:
+
+    <ol>
+     <li><p>Let <var>userAgent</var> be the <var>httpRequest</var>'s <a for=request>client</a>'s
+     <a>environment default `<code>User-Agent</code>` value</a>.
+
+     <li><p><a for="header list">Append</a> (`<code>User-Agent</code>`, <var>userAgent</var>) to
+     <var>httpRequest</var>'s <a for=request>header list</a>.
```
```suggestion
    <p>If <var>httpRequest</var>'s <a for=request>header list</a> <a
    for="header list">does not contain</a> `<code>User-Agent</code>`, then user agents should:

    <ol>
     <li><p>Let <var>userAgent</var> be <var>httpRequest</var>'s <a for=request>client</a>'s
     <a>environment default `<code>User-Agent</code>` value</a>.

     <li><p><a for="header list">Append</a> (`<code>User-Agent</code>`, <var>userAgent</var>) to
     <var>httpRequest</var>'s <a for=request>header list</a>.
```

annevk (commented):
I had these comments, but you just refactored it further. I'm not sure I like removing the default User-Agent value concept entirely. Is that really what we want?

I now also wonder at what point implementations ask WebDriver for this value as it seems unlikely to happen in het network layer, where the User-Agent value is currently appended.

sadym-chromium:
> I had these comments, but you just refactored it further. I'm not sure I like removing the default User-Agent value concept entirely. Is that really what we want?
> 
> I now also wonder at what point implementations ask WebDriver for this value as it seems unlikely to happen in het network layer, where the User-Agent value is currently appended.

@annevk Ok, I reverted the last change where I  merged the "environment default `<code>User-Agent</code>` value" and "default `<code>User-Agent</code>` value" and applied your proposed changes.

annevk:
So this looks good, but what about the timing of this call? Presumably it happens earlier, but maybe we can decide not to worry about that for now? Hmm.

sadym-chromium:
> I now also wonder at what point implementations ask WebDriver for this value as it seems unlikely to happen in het network layer, where the User-Agent value is currently appended.

@annevk in Chromium, we rely on [`Emulation.setUserAgentOverride`](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setUserAgentOverride) method, which sets the user agent override, which is [read in the renderer](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/loader/frame_loader.cc;drc=48d6f7175422b2c969c14258f9f8d5b196c28d18;l=1496).

annevk:
I think that does mean there are some issues with this setup. In that if you hit a couple of redirects and you call that method repeatedly, the User-Agent header would not change in value, but per this specification as written it ought to change.

To fix that and some race issues we'd have to grab the current User-Agent header before we go in parallel and use it for the entire duration of the fetch. I'm okay with not doing that for now, but I'm curious what other people think about what level of accuracy we need here. Maybe @jgraham is willing to chime in.

sadym-chromium:
> I think that does mean there are some issues with this setup. In that if you hit a couple of redirects and you call that method repeatedly, the User-Agent header would not change in value, but per this specification as written it ought to change.
> 
> To fix that and some race issues we'd have to grab the current User-Agent header before we go in parallel and use it for the entire duration of the fetch. I'm okay with not doing that for now, but I'm curious what other people think about what level of accuracy we need here. Maybe @jgraham is willing to chime in.

The WebDriver BiDi ["WebDriver BiDi emulated User-Agent"](https://www.w3.org/TR/webdriver-bidi/#webdriver-bidi-emulated-user-agent) hook returns the emulated user context based on pre-configured emulations per top-level navigable, so it is not expected to be changed during navigations or redirects. Even though the JS environments will be changed, it would not influence the return result. BiDi uses the environment only to associate the request with the navigable.

annevk:
I see. So you are you saying it's set before navigation starts and can't be changed after?

sadym-chromium:
> I see. So you are you saying it's set before navigation starts and can't be changed after?

Theoretically the user can change the emulation during navigation, but otherwise it is persistent per top-level navigable. I cannot imagine a real-world automation scenario when the user needs to change emulation during navigation and expect persistent results.

jgraham:
You can change it in the middle of a fetch, so I think it's true that redirects could see different user agent headers (you could also do the same thing with request interception). 

In general I don't think people are expected to change this often for existing traversables; the typical use case is "set the user agent when creating a top-level traversable and use it for the entire lifetime of that traversable".

So I think I agree that in principle caching the value at the start of a redirect loop is the right thing to do, but in practice it's likely to be a rare edge case.

annevk:
Since WebDriver interoperability is quite a bit less crucial than web platform interoperability I'll go ahead and merge this. Let's revamp if people end up hitting this in the wild after all.

sadym-chromium:
> Since WebDriver interoperability is quite a bit less crucial than web platform interoperability I'll go ahead and merge this. Let's revamp if people end up hitting this in the wild after all.

Thanks a lot!

Issue #1839, opened 2025-06-30
Compression dictionary support for opaque responses


pmeenan:
### What problem are you trying to solve?

There are some web platform features that are not available when responses are not `CORS-readable`. The protection is in place to make sure private data from responses can't inadvertently be leaked. The features that come to mind are some of the details like content-encoding from the resource timing API and compression dictionary content-encoding.

There are a lot of "public" resources that are embedded in a third-party context where it is difficult, if not impossible to enable CORS on the requests (either anonymous or credentialed). This includes things like images loaded through CSS and scripts that are sprinkled across the web (like analytics, ads, etc) where the content of the response is the same when fetched with and without credentials.

We would like to explore a mechanism for an origin to explicitly mark responses to `no-cors` requests as being readable.

### What solutions exist today?

For the image-in-css case today, there is no way to opt-in to CORS so there is no solution.

For the case of scripts or stylesheets embedded in a third-party context, the solution is to add `crossorigin` attributes to the markup. That is somewhat impractical to apply at web-scale but, more importantly, it comes with side-effects. Depending on the CORS mode used, the requests will either require a separate connection from no-cors requests or they will require a preflight check.

### How would you solve it?

For `no-cors` requests, add a `Sec-Accept-Content-Readability: public` request header to advertise support to the origin for marking responses as readable.

Responses would be considered `CORS-readable` if they contain both of the following response headers:

```
Content-Readability: public
Access-Control-Allow-Origin: *
```

For redirect chains, every response in the redirect chain would need both headers for the final response to be readable (similar to CORS tainting on the request path).

The client-advertisiment/negotiation for cases like compression dictionary encoding where the response will fail if it is not readable.

### Anything else?

Original discussion in the HTML spec: https://github.com/whatwg/html/issues/8143

Chrome platform status: https://chromestatus.com/feature/5509970048843776
Mozilla standards position: https://github.com/mozilla/standards-positions/issues/1240
Webkit standards position: https://github.com/WebKit/standards-positions/issues/505

domenic:
> For `no-cors` requests, add a `Sec-Accept-Content-Readability: public` request header to advertise support to the origin for marking responses as readable.

Is this part necessary?

pmeenan:
>> For no-cors requests, add a Sec-Accept-Content-Readability: public > request header to advertise support to the origin for marking responses as readable.

> Is this part necessary?

Maybe not if we caught it in time with recent additions to the web platform tests for compression dictionary.

[...]

PR #1854, opened 2025-08-25
Add support for compression dictionary transport

### LINKED ISSUE #1739: Add compression dictionary negotiation and decoding to the fetch processing model
### What problem are you trying to solve?

[Compression dictionary transport](https://github.com/WICG/compression-dictionary-transport) provides a mechanism for using fetched resources as a compression dictionary for content-encoding of HTTP responses. The IETF draft for the http-level negotiation and compression is [here](https://datatracker.ietf.org/doc/draft-ietf-httpbis-compression-dictionary/).

Most of the HTTP-level negotiation is defined in the IETF draft but there are a few places where it intersects with the fetch processing model:

* Dictionary cache, associated metadata, partitioning and clearing.
* Matching an existing dictionary against outbound requests in [HTTP-network fetch](https://fetch.spec.whatwg.org/#http-network-fetch).
* CORS-readability requirements for storing dictionaries.
* CORS-readability requirements for decoding dictionary-compressed responses.

### What solutions exist today?

_No response_

### How would you solve it?

Add the necessary processing steps once we are comfortable that they are appropriate.

### Anything else?

Chromium is currently the only browser engine experimenting with compression dictionaries so it might not make sense to fold into the standard quite yet but it is worth having the discussion to make sure that the implementation is something that we can all agree with before it gets too far.

pmeenan:
Add processing steps for handling [HTTP Compression Dictionary Transport](https://datatracker.ietf.org/doc/draft-ietf-httpbis-compression-dictionary/) content encoding and dictionary negotiation (RFC pending publication).

This adds a processing layer between the HTTP cache and network fetch that handles most of the dictionary-based content encoding (including matching dictionaries to outgoing requests).

Additionally, it adds processing above the HTTP cache for storing the dictionaries for future use and defines the "compression-dictionary" initiator and destination (the matching HTML spec update is in-process).

Support for clearing the caches through clear-site-data is in [this PR](https://github.com/w3c/webappsec-clear-site-data/pull/94).

Fix #1739, #1839

   * [Chrome](https://chromestatus.com/feature/5124977788977152)
   * [Firefox](https://github.com/WebKit/standards-positions/issues/160)
   * [Safari](https://github.com/WebKit/standards-positions/issues/160)
   * https://github.com/web-platform-tests/wpt/tree/master/fetch/compression-dictionary
   * PR for removal of "tentative": https://github.com/web-platform-tests/wpt/pull/54532
   * Chromium: https://crbug.com/1413922
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1882979
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=295249
   * Deno (not for CORS changes): https://github.com/denoland/deno/issues/30533
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport

***
<a href="https://whatpr.org/fetch/1854.html" title="Last updated on Jun 10, 2026, 9:38 PM UTC (4d3401a)">Preview</a> | <a href="https://whatpr.org/fetch/1854/30140d0...4d3401a.html" title="Last updated on Jun 10, 2026, 9:38 PM UTC (4d3401a)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -82,9 +82,7 @@ urlPrefix:https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-layered-cooki
         "aliasOf": "RFC9111"
     },
     "HTTP-COMPRESSION-DICTIONARIES": {
-        "authors": ["Patrick Meenan", "Yoav Weiss"],
-        "href": "https://datatracker.ietf.org/doc/draft-ietf-httpbis-compression-dictionary/",
-        "title": "Compression Dictionary Transport"
+        "aliasOf": "RFC9842"
     },
     "HTTP1": {
         "aliasOf": "RFC9112"
--- a/fetch.bs
+++ b/fetch.bs
@@ -6582,8 +6582,14 @@ run these steps:
    <li><p>Let <var>compressionDictionaryCache</var> be the result of
    <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
 
-   <li><p>If <var>compressionDictionaryCache</var> is null or <var>request</var>'s
-   <a for=request>response tainting</a> is "<code>opaque</code>", then return <var>response</var>.
+   <li><p>If <var>compressionDictionaryCache</var> is null, then return <var>response</var>.
+
+   <li><p>Let <var>corpPolicy</var> be the result of <a for="header list">getting</a>
+   `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` from <var>response</var>'s
+   <a for=response>header list</a>.
+
+   <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>" and
+   <var>corpPolicy</var> is not `<code>cross-origin</code>`, then return <var>response</var>.
 
    <li><p>Let <var>expirationTime</var> be the time at which the <var>response</var> becomes stale.
 
@@ -6651,7 +6657,12 @@ given a <a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
  <li><p>If <var>codings</var> is null or does not contain `<code>dcb</code>` or `<code>dcz</code>`,
  then return <var>response</var>.
 
- <li><p>If <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>", then return a
+ <li><p>Let <var>corpPolicy</var> be the result of <a for="header list">getting</a>
+ `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` from <var>response</var>'s
+ <a for=response>header list</a>.
+
+ <li><p>If <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>" and
+ <var>corpPolicy</var> is not `<code>cross-origin</code>`, then return a
  <a>network error</a>.
 
  <li><p>Let <var>availableDictionaryHash</var> be the result of
--- a/fetch.bs
+++ b/fetch.bs
@@ -6638,8 +6638,34 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
  <li><p><a>Combine</a> (`<code>Accept-Encoding</code>`, `<code>dcz</code>`) in <var>request</var>'s <a for=request>header list</a>.
 
- <li><p>Return the result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
- <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+ <li><p>Let <var>response</var> be the result of running <a>HTTP-network fetch</a> given
+ <var>fetchParams</var>, <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+
+ <li><p>Let <var>codings</var> be the result of <a>extracting header list values</a> given
+ `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
+
+ <li><p>If <var>codings</var> is null or does not contain `<code>dcb</code>` or `<code>dcz</code>`, then
+ return <var>response</var>.
+
+ <li><p>If <var>response</var>'s <a for=response>type</a> is "<code>opaque</code>", then return a
+ <a>network error</a>.
+
+ <li><p>Let <var>availableDictionaryHash</var> be the result of <a>getting a structured field value</a>
+ given `<code>Available-Dictionary</code>`, "<code>bytestring</code>", and <var>request</var>'s
+ <a for=request>header list</a>.
+
+ <li><p>Let <var>newBody</var> be a new <a for=/>body</a> whose <a for=body>stream</a> is the result
+ of transforming <var>response</var>'s <a for=response>body</a>'s <a for=body>stream</a> with an
+ algorithm that verifies that the dictionary hash in the stream matches
+ <var>availableDictionaryHash</var> and decodes the rest of the stream with the applicable
+ algorithm as defined in [[!HTTP-COMPRESSION-DICTIONARIES]]. If verification or decoding fails,
+ the transformed stream must error.
+
+ <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>newBody</var>.
+
+ <li><p><a>Delete</a> `<code>Content-Encoding</code>` from <var>response</var>'s <a for=response>header list</a>.
+
+ <li><p>Return <var>response</var>.
 </ol>
 </div>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -2314,14 +2314,9 @@ Unless stated otherwise, it is unset.
 <dfn export for=request id=timing-allow-failed>timing allow failed flag</dfn>. Unless stated
 otherwise, it is unset.
 
-<p>A <a for=/>request</a> has an associated
-<dfn export for=request id=compression-dictionary-disallowed>compression dictionary disallowed flag</dfn>.
-Unless stated otherwise, it is unset.
-
 <p class=note>A <a for=/>request</a>'s <a for=request>URL list</a>, <a for=request>current URL</a>,
 <a for=request>redirect count</a>, <a for=request>response tainting</a>,
-<a for=request>done flag</a>, <a for=request>timing allow failed flag</a>, and
-<a for=request>compression dictionary disallowed flag</a> are used as
+<a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
 <p>A <a for=/>request</a> has an associated
@@ -5022,18 +5017,6 @@ steps:
     </ol>
   </dl>
 
- <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>",
- then:
-
- <ol>
-   <li><p>Let <var>corpPolicy</var> be the result of <a for="header list">getting</a>
-   `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` from <var>response</var>'s
-   <a for=response>header list</a>.
-
-   <li><p>If <var>corpPolicy</var> is not `<code>cross-origin</code>`, then set
-   <var>request</var>'s <a for=request>compression dictionary disallowed flag</a>.
- </ol>
-
  <li><p>If <var>recursive</var> is true, then return <var>response</var>.
 
  <li>
@@ -6577,7 +6560,7 @@ run these steps:
  <var>request</var> and the given realm.
 
  <li>
-  <p>If <var>request</var>'s <a for=request>compression dictionary disallowed flag</a> is not set
+  <p>If <var>request</var>'s <a for=request>response tainting</a> is not "<code>opaque</code>"
   and <var>response</var>'s <a for=response>header list</a> <a for="header list">contains</a>
   `<code>Use-As-Dictionary</code>`:
   <!-- This is defined in [[!RFC9842]] -->
@@ -6669,8 +6652,8 @@ given a <a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
  <li><p>If <var>codings</var> is null or does not contain `<code>dcb</code>` or `<code>dcz</code>`,
  then return <var>response</var>.
 
- <li><p>If <var>request</var>'s <a for=request>compression dictionary disallowed flag</a>
- is set, then return a <a>network error</a>.
+ <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>",
+ then return a <a>network error</a>.
 
  <li><p>Let <var>availableDictionaryHash</var> be the result of
  <a>getting a structured field value</a> given `<code>Available-Dictionary</code>`,
--- a/fetch.bs
+++ b/fetch.bs
@@ -1770,7 +1770,6 @@ processing requirements.
 "<code>audio</code>",
 "<code>beacon</code>",
 "<code>body</code>",
-"<code>compression-dictionary</code>",
 "<code>css</code>",
 "<code>early-hints</code>",
 "<code>embed</code>",
@@ -1978,7 +1977,7 @@ not always relevant and might require different behavior.
   <tr>
    <td>"<code>compression-dictionary</code>"
    <td>"<code>compression-dictionary</code>"
-   <td><code>default-src</code> (no specific directive)
+   <td><code>default-src</code>
    <td>HTML's <code>&lt;link rel=compression-dictionary&gt;</code>
   <tr>
    <td>"<code>download</code>"
@@ -3333,6 +3332,22 @@ or an <a>implementation-defined</a> value.
  <li><p>If <var>key</var> is null, then return null.
 
  <li><p>Return the unique HTTP cache associated with <var>key</var>. [[!HTTP-CACHING]]
+</ol>
+</div>
+
+
+<h3 id=compression-dictionary-cache-partitions>Compression-dictionary cache partitions</h3>
+
+<div algorithm>
+<p>To <dfn>determine the compression-dictionary cache partition</dfn>, given a <a for=/>request</a> <var>request</var>:
+
+<ol>
+ <li><p>Let <var>key</var> be the result of <a for=request>determining the network partition key</a>
+ given <var>request</var>.
+
+ <li><p>If <var>key</var> is null, then return null.
+
+ <li><p>Return the unique compression-dictionary cache associated with <var>key</var>. [[!HTTP-COMPRESSION-DICTIONARIES]]
 </ol>
 </div>
 
@@ -6544,6 +6559,40 @@ run these steps:
 
  <li><p>If <var>isAuthenticationFetch</var> is true, then create an <a>authentication entry</a> for
  <var>request</var> and the given realm.
+
+ <li>
+  <p>If <var>response</var>'s <a for=response>header list</a>
+  <a for="header list">contains</a> `<code>Use-As-Dictionary</code>`, then:
+  <!-- This is defined in [[!HTTP-COMPRESSION-DICTIONARIES]] -->
+
+  <ol>
+   <li><p>Let <var>dictionaryValue</var> be the result of
+   <a for="header list">getting a structured field value</a> given `<code>Use-As-Dictionary</code>`,
+   "<code>dictionary</code>", and <var>response</var>'s <a for=response>header list</a>.
+
+   <li><p>If <var>dictionaryValue</var> is null or <var>dictionaryValue</var>["<code>match</code>"]
+   does not <a for=map>exist</a>, then return <var>response</var>.
+
+   <li><p>Let <var>pattern</var> be the result of creating a URL pattern from
+   <var>dictionaryValue</var>["<code>match</code>"] and <var>request</var>'s
+   <a for=request>current URL</a>.
+
+   <li><p>If <var>pattern</var> is failure or <var>pattern</var> has regexp groups, then return
+   <var>response</var>.
+
+   <li><p>Let <var>compressionDictionaryCache</var> be the result of
+   <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+   <li><p>If <var>compressionDictionaryCache</var> is null or <var>request</var>'s
+   <a for=request>response tainting</a> is "<code>opaque</code>", then return <var>response</var>.
+
+   <li><p>Let <var>expirationTime</var> be the time at which the <var>response</var> becomes stale.
+
+   <li><p>If <var>expirationTime</var> is not in the future, then return <var>response</var>.
+
+   <li><p>Store <var>response</var> in <var>compressionDictionaryCache</var> with its associated
+   <var>dictionaryValue</var> and <var>expirationTime</var>.
+  </ol>
 
  <li><p>Return <var>response</var>. <span class=note>Typically <var>response</var>'s
  <a for=response>body</a>'s <a for=body>stream</a> is still being enqueued to after
--- a/fetch.bs
+++ b/fetch.bs
@@ -6410,8 +6410,9 @@ run these steps:
    <li><p>If <var>httpRequest</var>'s <a for=request>cache mode</a> is
    "<code>only-if-cached</code>", then return a <a>network error</a>.
 
-   <li><p>Let <var>forwardResponse</var> be the result of running <a>HTTP-network fetch</a> given
-   <var>httpFetchParams</var>, <var>includeCredentials</var>, and <var>isNewConnectionFetch</var>.
+   <li><p>Let <var>forwardResponse</var> be the result of running
+   <a>HTTP-network compression-dictionary fetch</a> given <var>httpFetchParams</var>,
+   <var>includeCredentials</var>, and <var>isNewConnectionFetch</var>.
 
    <li><p>If <var>httpRequest</var>'s <a for=request>method</a> is <a>unsafe</a> and
    <var>forwardResponse</var>'s <a for=response>status</a> is in the range 200 to 399, inclusive,
@@ -6600,6 +6601,47 @@ run these steps:
 </ol>
 </div>
 
+<h3 id=http-network-compression-dictionary-fetch>HTTP-network compression-dictionary fetch</h3>
+
+<div algorithm>
+<p>To <dfn>HTTP-network compression-dictionary fetch</dfn>, given a <a for=/>fetch params</a>
+<var>fetchParams</var>, an optional boolean <var>includeCredentials</var> (default false), and an
+optional boolean <var>forceNewConnection</var> (default false), run these steps:
+
+<ol>
+ <li><p>Let <var>request</var> be <var>fetchParams</var>'s <a for="fetch params">request</a>.
+
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is "<code>no-cors</code>", then return the result of
+ running <a>HTTP-network fetch</a> given <var>fetchParams</var>, <var>includeCredentials</var>, and
+ <var>forceNewConnection</var>.
+
+ <li><p>If the user agent is configured to block cookies for <var>request</var>, then return the
+ result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
+ <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+
+ <li><p>Let <var>compressionDictionaryCache</var> be the result of
+ <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+ <li><p>If <var>compressionDictionaryCache</var> is null, then return the result of running
+ <a>HTTP-network fetch</a> given <var>fetchParams</var>, <var>includeCredentials</var>, and
+ <var>forceNewConnection</var>.
+
+ <li><p>Let <var>bestMatch</var> be the result of finding the best matching dictionary in
+ <var>compressionDictionaryCache</var> for <var>request</var> as defined in [[!HTTP-COMPRESSION-DICTIONARIES]].
+
+ <li><p>If <var>bestMatch</var> is null, then return the result of running <a>HTTP-network fetch</a>
+ given <var>fetchParams</var>, <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+
+ <li><p>Add the `<code>Available-Dictionary</code>` and `<code>Dictionary-ID</code>` (if applicable) headers to <var>request</var> using <var>bestMatch</var> as defined in [[!HTTP-COMPRESSION-DICTIONARIES]].
+
+ <li><p><a>Combine</a> (`<code>Accept-Encoding</code>`, `<code>dcb</code>`) in <var>request</var>'s <a for=request>header list</a>.
+
+ <li><p><a>Combine</a> (`<code>Accept-Encoding</code>`, `<code>dcz</code>`) in <var>request</var>'s <a for=request>header list</a>.
+
+ <li><p>Return the result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
+ <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+</ol>
+</div>
 
 <h3 id=http-network-fetch>HTTP-network fetch</h3>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -81,9 +81,6 @@ urlPrefix:https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-layered-cooki
     "HTTP-CACHING": {
         "aliasOf": "RFC9111"
     },
-    "HTTP-COMPRESSION-DICTIONARIES": {
-        "aliasOf": "RFC9842"
-    },
     "HTTP1": {
         "aliasOf": "RFC9112"
     },
@@ -3345,7 +3342,7 @@ or an <a>implementation-defined</a> value.
 
  <li><p>If <var>key</var> is null, then return null.
 
- <li><p>Return the unique compression-dictionary cache associated with <var>key</var>. [[!HTTP-COMPRESSION-DICTIONARIES]]
+ <li><p>Return the unique compression-dictionary cache associated with <var>key</var>. [[!RFC9842]]
 </ol>
 </div>
 
@@ -6561,8 +6558,8 @@ run these steps:
 
  <li>
   <p>If <var>response</var>'s <a for=response>header list</a>
-  <a for="header list">contains</a> `<code>Use-As-Dictionary</code>`, then:
-  <!-- This is defined in [[!HTTP-COMPRESSION-DICTIONARIES]] -->
+  <a for="header list">contains</a> `<code>Use-As-Dictionary</code>`:
+  <!-- This is defined in [[!RFC9842]] -->
 
   <ol>
    <li><p>Let <var>dictionaryValue</var> be the result of
@@ -6596,7 +6593,7 @@ run these steps:
    <li><p>If <var>expirationTime</var> is not in the future, then return <var>response</var>.
 
    <li><p>Store <var>response</var> in <var>compressionDictionaryCache</var> with its associated
-   <var>dictionaryValue</var> and <var>expirationTime</var>.
+   <var>pattern</var>, <var>dictionaryValue</var> and <var>expirationTime</var>.
   </ol>
 
  <li><p>Return <var>response</var>. <span class=note>Typically <var>response</var>'s
@@ -6633,14 +6630,14 @@ given a <a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
 
  <li><p>Let <var>bestMatch</var> be the result of finding the best matching dictionary in
  <var>compressionDictionaryCache</var> for <var>request</var> as defined in
- [[!HTTP-COMPRESSION-DICTIONARIES]].
+ [[!RFC9842]].
 
  <li><p>If <var>bestMatch</var> is null, then return the result of running <a>HTTP-network fetch</a>
  given <var>fetchParams</var>, <var>includeCredentials</var>, and <var>forceNewConnection</var>.
 
  <li><p>Add the `<code>Available-Dictionary</code>` and `<code>Dictionary-ID</code>`
  (if applicable) headers to <var>request</var> using <var>bestMatch</var> as defined in
- [[!HTTP-COMPRESSION-DICTIONARIES]].
+ [[!RFC9842]].
 
  <li><p><a for="header list">Combine</a> (`<code>Accept-Encoding</code>`, `<code>dcb</code>`)
  in <var>request</var>'s <a for=request>header list</a>.
@@ -6673,7 +6670,7 @@ given a <a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
  result of transforming <var>response</var>'s <a for=response>body</a>'s <a for=body>stream</a>
  with an algorithm that verifies that the dictionary hash in the stream matches
  <var>availableDictionaryHash</var> and decodes the rest of the stream with the applicable
- algorithm as defined in [[!HTTP-COMPRESSION-DICTIONARIES]]. If verification or decoding fails,
+ algorithm as defined in [[!RFC9842]]. If verification or decoding fails,
  the transformed stream must error.
 
  <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>newBody</var>.
@@ -6755,7 +6752,7 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
 
     <ul>
      <li><p>Follow the relevant requirements from HTTP. [[!HTTP]] [[!HTTP-CACHING]]
-     [[!HTTP-COMPRESSION-DICTIONARIES]]
+     [[!RFC9842]]
 
      <li>
       <p>If <var>request</var>'s <a for=request>body</a> is non-null, and <var>request</var>'s
--- a/fetch.bs
+++ b/fetch.bs
@@ -66,6 +66,11 @@ urlPrefix:https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-layered-cooki
     url:name-retrieve-cookies;text:retrieve cookies
     url:name-serialize-cookies;text:serialize cookies
     url:name-garbage-collect-cookies;text:garbage collect cookies
+
+urlPrefix:https://urlpattern.spec.whatwg.org/#;type:dfn;spec:urlpattern
+    url:url-pattern-create;text:creating a URL pattern
+    url:url-pattern-has-regexp-groups;text:has regexp groups
+
 </pre>
 
 <pre class=biblio>
@@ -1804,7 +1809,6 @@ is "<code>all</code>" or "<code>none</code>". Unless stated otherwise it is "<co
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-initiator>initiator</dfn>, which is
 the empty string,
-"<code>compression-dictionary</code>",
 "<code>download</code>",
 "<code>imageset</code>",
 "<code>manifest</code>",
@@ -1971,8 +1975,7 @@ not always relevant and might require different behavior.
    <td>HTML's <code>&lt;video></code> element
   <tr>
    <td>"<code>compression-dictionary</code>"
-   <td>"<code>compression-dictionary</code>"
-   <td><code>default-src</code>
+   <td><code>connect-src</code>
    <td>HTML's <code>&lt;link rel=compression-dictionary&gt;</code>
   <tr>
    <td>"<code>download</code>"
@@ -2299,6 +2302,11 @@ Unless stated otherwise, it is zero.
 <dfn export for=request id=concept-request-response-tainting>response tainting</dfn>,
 which is "<code>basic</code>", "<code>cors</code>", or "<code>opaque</code>".
 Unless stated otherwise, it is "<code>basic</code>".
+
+<p>A <a for=/>request</a> has an associated
+<dfn export for=request id=concept-request-corp-tainting>corp tainting</dfn>,
+which is "<code>allowed</code>", or "<code>blocked</code>".
+Unless stated otherwise, it is "<code>allowed</code>".
 
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=no-cache-prevent-cache-control>prevent no-cache cache-control header modification flag</dfn>.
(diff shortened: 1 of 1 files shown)


pmeenan:
The RFC is pending publication so this will have to wait until that happens (should be any day now).  RFC number has been assigned and final edits are complete: https://www.rfc-editor.org/auth48/rfc9842

pmeenan:
RFC has published now so this should be ready to go (just rebased it).

pmeenan:
From the TPAC 2025 discussions, I added support for opaque responses to use compression dictionary when the `Cross-Origin-Resource-Policy` response header is set to `cross-origin`.

fetch.bs
```diff
@@ -81,6 +81,9 @@ urlPrefix:https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-layered-cooki
     "HTTP-CACHING": {
         "aliasOf": "RFC9111"
     },
+    "HTTP-COMPRESSION-DICTIONARIES": {
+        "aliasOf": "RFC9842"
+    },
```
annevk: We can just reference the RFC number directly. No need for an alias.

fetch.bs
```diff
@@ -1798,6 +1801,7 @@ is "<code>all</code>" or "<code>none</code>". Unless stated otherwise it is "<co
 <p>A <a for=/>request</a> has an associated
 <dfn export for=request id=concept-request-initiator>initiator</dfn>, which is
 the empty string,
+"<code>compression-dictionary</code>",
```
annevk: Why do we need a distinct initiator if we also have a distinct destination?

fetch.bs
```diff
@@ -3267,6 +3277,22 @@ or an <a>implementation-defined</a> value.
...
+<div algorithm>
+<p>To <dfn>determine the compression-dictionary cache partition</dfn>, given a <a for=/>request</a> <var>request</var>:
+
+<ol>
+ <li><p>Let <var>key</var> be the result of <a for=request>determining the network partition key</a>
+ given <var>request</var>.
+
+ <li><p>If <var>key</var> is null, then return null.
+
+ <li><p>Return the unique compression-dictionary cache associated with <var>key</var>. [[!HTTP-COMPRESSION-DICTIONARIES]]
```
annevk: Are there PRs against Storage and Clear-Site-Data about clearing this cache? It would be good if those were linked from the commit message.

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
  <li><p>If <var>isAuthenticationFetch</var> is true, then create an <a>authentication entry</a> for
  <var>request</var> and the given realm.
 
+ <li>
+  <p>If <var>response</var>'s <a for=response>header list</a>
+  <a for="header list">contains</a> `<code>Use-As-Dictionary</code>`, then:
```
```suggestion
  <a for="header list">contains</a> `<code>Use-As-Dictionary</code>`:
```

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+
+  <ol>
+   <li><p>Let <var>dictionaryValue</var> be the result of
+   <a for="header list">getting a structured field value</a> given `<code>Use-As-Dictionary</code>`,
+   "<code>dictionary</code>", and <var>response</var>'s <a for=response>header list</a>.
+
+   <li><p>If <var>dictionaryValue</var> is null or <var>dictionaryValue</var>["<code>match</code>"]
+   does not <a for=map>exist</a>, then return <var>response</var>.
+
+   <li><p>Let <var>pattern</var> be the result of creating a URL pattern from
```
annevk: "creating a URL pattern" needs some kind of link to a definition.

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+   <var>dictionaryValue</var>["<code>match</code>"] and <var>request</var>'s
+   <a for=request>current URL</a>.
+
+   <li><p>If <var>pattern</var> is failure or <var>pattern</var> has regexp groups, then return
```
annevk: "has regexp groups" needs a definition.

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+   <var>response</var>.
+
+   <li><p>Let <var>compressionDictionaryCache</var> be the result of
+   <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+   <li><p>If <var>compressionDictionaryCache</var> is null, then return <var>response</var>.
```
annevk, line 6662: Maybe we should do this before creating a URL pattern. This seems less expensive.

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+
+   <li><p>Let <var>corpPolicy</var> be the result of <a for="header list">getting</a>
+   `<a http-header><code>Cross-Origin-Resource-Policy</code></a>` from <var>response</var>'s
+   <a for=response>header list</a>.
+
+   <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>" and
+   <var>corpPolicy</var> is not `<code>cross-origin</code>`, then return <var>response</var>.
```
annevk: This isn't good enough. We want _corpPolicy_ to be a good value across redirects. We should have tests for that as well.

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+
+   <li><p>Let <var>expirationTime</var> be the time at which the <var>response</var> becomes stale.
```
annevk: This needs more links to definitions.

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+
+   <li><p>If <var>expirationTime</var> is not in the future, then return <var>response</var>.
+
+   <li><p>Store <var>response</var> in <var>compressionDictionaryCache</var> with its associated
+   <var>dictionaryValue</var> and <var>expirationTime</var>.
```
annevk: Shouldn't we also store _pattern_ somewhere?

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+<ol>
+ <li><p>Let <var>request</var> be <var>fetchParams</var>'s <a for="fetch params">request</a>.
+
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is "<code>no-cors</code>", then return the
+ result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
+ <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+
+ <li><p>If the user agent is configured to block cookies for <var>request</var>, then return the
+ result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
+ <var>includeCredentials</var>, and <var>forceNewConnection</var>.
```
annevk: I don't understand this. The dictionary is partitioned so how is it different from the HTTP cache?

fetch.bs
```diff
@@ -6385,12 +6412,131 @@ run these steps:
...
+
+ <li><p>Add the `<code>Available-Dictionary</code>` and `<code>Dictionary-ID</code>`
+ (if applicable) headers to <var>request</var> using <var>bestMatch</var> as defined in
+ [[!HTTP-COMPRESSION-DICTIONARIES]].
+
+ <li><p><a for="header list">Combine</a> (`<code>Accept-Encoding</code>`, `<code>dcb</code>`)
+ in <var>request</var>'s <a for=request>header list</a>.
+
+ <li><p><a for="header list">Combine</a> (`<code>Accept-Encoding</code>`, `<code>dcz</code>`)
+ in <var>request</var>'s <a for=request>header list</a>.
```
annevk: Are we doing this even if web developers set these headers? That seems rather sketchy.

fetch.bs
pmeenan: The `Use-As-Dictionary` response header can attach an arbitrary ID to arbitrary requests from the same origin (that match a given pattern) without having to access the underlying response. This allows non-executable responses to attach a tracking ID to an origin that behaves like a cookie without having to execute code (image-only origins for example).

To keep things consistent with dev expectations and CSP, the security and privacy teams requested that they also be treated as cookies and not be allowed in situations where cookies would not have been allowed.

fetch.bs
pmeenan: Done

fetch.bs
pmeenan: Fixed, thanks.

fetch.bs
pmeenan: Added (let me know if the reference to the urlpattern spec should be handled differently - I tried to mimick what was used for resource timing.

fetch.bs
pmeenan: Fixed.

fetch.bs
pmeenan: I don't know if something more generic would make sense but my current plan is:
* Add a `compression dictionary blocked` flag to request that starts out as unset
* In main fetch, each time the response is processed and after `response tainting` has been updated and is currently `opaque`, check the corp header.
* If the header is missing or not set to `cross-origin` then set the `compression dictionary blocked` flag
* Replace the check here with a check of the `compression dictionary blocked` flag

I don't know if the same CORP readability logic could or should be used for anything else and a more generic tainting name should be used.

fetch.bs
pmeenan: Added a link to the `stale response` definition.

fetch.bs
pmeenan: `Accept-Encoding` is a [forbidden request header](https://fetch.spec.whatwg.org/#forbidden-request-header) so developers shouldn't be able to set it.

fetch.bs
pmeenan: Done

fetch.bs
pmeenan: Fixed.

fetch.bs
pmeenan: Fixed.

fetch.bs
pmeenan: I went ahead and added a `compression dictionary disallowed` flag to `request` which is updated in `Main fetch` just after the `response tainting` is evaluated for each step and then the flag is just checked when doing dictionary-specific work (ignoring the `Use-As-Dictionary` header and erroring dictionary-compressed opaque responses.

fetch.bs
pmeenan: There is a PR for clear-site-data now (linked). I looked through the storage spec and couldn't find anything relevant since it's another one of the various "network caches" and isn't directly addressable. Let me know if that's not the case and there's a specific API or area where it crosses with the storage spec.

pmeenan:
Sorry for dragging this out. I think I addressed all of the questions/issues. The main question outstanding right now is if the no-cors path using CORP is "approved" or if there are more steps to getting that integrated before this can land.

I filed it as a [question with the TAG](https://github.com/w3ctag/design-reviews/issues/1203) but I'm happy do do a more formal review somewhere if necessary.

pmeenan:
I switched back to only allowing dictionary use for non-opaque requests. It's cleaner from the browser's perspective and harder to get wrong for sites. This is what the browsers have implemented anyway and what the current WPT's test.

The spec change should be ready to go now and match Chrome's and Mozilla's implementations.

If we want to bring dictionary support to third-party embeds in some way (the main pain point for `no-cors`), we can solve that separately.

fetch.bs
```diff
@@ -6531,12 +6559,121 @@ run these steps:
...
+   does not <a for=map>exist</a>, then return <var>response</var>.
+
+   <li><p>Let <var>compressionDictionaryCache</var> be the result of
+   <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+   <li><p>If <var>compressionDictionaryCache</var> is null, then return <var>response</var>.
+
+   <li><p>Let <var>pattern</var> be the result of
+   <a for=/>creating a URL pattern</a> from <var>dictionaryValue</var>["<code>match</code>"]
+   and <var>request</var>'s <a for=request>current URL</a>.
```
fred-wang: It seems the text for callers of this algorithm is written that way: "creating a URL pattern, given dictionaryValue["match"], request's current URL and an empty map".

Also,  request's current URL is an URL not a string, so we should perform a serialization similar to https://urlpattern.spec.whatwg.org/#other-specs-http (IIUC we can't use that one, because dictionary["match"] is a dictionary value not an HTTP structured field value).

fetch.bs
pmeenan: I removed all of the no-cors stuff and went back to the as-implemented `non-opaque` requirement.

fetch.bs
pmeenan: Thanks. This should be fixed now. I changed it to extract the "item" from match and to serialize the URL. Let me know if you think a different serialization is needed than the one I used.

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+   <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+   <li><p>If <var>compressionDictionaryCache</var> is null, then return <var>response</var>.
+
+   <li><p>Let <var>pattern</var> be the result of
+   <a for=/>creating a URL pattern</a> given the bare item of <var>dictionaryValue</var>["<code>match</code>"],
+   the <a lt="URL serializer">serialization</a> of <var>request</var>'s <a for=request>current URL</a>,
+   and an empty map.
+
+   <li><p>If <var>pattern</var> is failure or <var>pattern</var> <a for=/>has regexp groups</a>,
```
fred-wang, line 6669: "creating a URL pattern" uses algos from https://url.spec.whatwg.org/ that can return failure, but when that happens it actually throw an exception, rather than returning failure?

I wanted to ask about these potential exceptions the other day, but couldn't find an obvious input here that would make the algo throw... Anyway, I find this spec indeed deals with the case when an exception is thrown so I guess we probably want the same here: https://wicg.github.io/connection-allowlists/#abstract-opdef-parse-a-connection-allowlist-header

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+  `<code>Use-As-Dictionary</code>`:
+  <!-- This is defined in [[!RFC9842]] -->
+
+  <ol>
+   <li><p>Let <var>dictionaryValue</var> be the result of
+   <a for="header list">getting a structured field value</a> given `<code>Use-As-Dictionary</code>`,
+   "<code>dictionary</code>", and <var>response</var>'s <a for=response>header list</a>.
+
+   <li><p>If <var>dictionaryValue</var> is null or <var>dictionaryValue</var>["<code>match</code>"]
+   does not <a for=map>exist</a>, then return <var>response</var>.
```
fred-wang, line 6636: What about other parameters?

For example https://www.rfc-editor.org/rfc/rfc9842#name-type says a client should not deal with a `type` value it does not understand, so I guess we should bail out if that happens.

For `id` what happens if it exceeds the 1024 characters limit?

For  `match-dest`, I guess clients would just ignore unknown destinations.

fetch.bs
annevk: It also seems we lack tests for this. We should have somewhat exhaustive tests for error conditions.

fetch.bs
```diff
@@ -1818,6 +1823,7 @@ device to assist defining CSP and Mixed Content. It is not exposed to JavaScript
 <p>A <dfn export>destination type</dfn> is one of:
 the empty string,
 "<code>audio</code>",
+"<code>compression-dictionary</code>",
```
annevk: This should be below audioworklet.

fetch.bs
```diff
@@ -3323,6 +3334,22 @@ or an <a>implementation-defined</a> value.
...
+<div algorithm>
+<p>To <dfn>determine the compression-dictionary cache partition</dfn>, given a <a for=/>request</a> <var>request</var>:
+
+<ol>
+ <li><p>Let <var>key</var> be the result of <a for=request>determining the network partition key</a>
+ given <var>request</var>.
+
+ <li><p>If <var>key</var> is null, then return null.
+
+ <li><p>Return the unique compression-dictionary cache associated with <var>key</var>. [[!RFC9842]]
```
annevk, line 3358: I think we should make compression-dictionary cache a link.

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+   <li><p>If <var>dictionaryValue</var> is null or <var>dictionaryValue</var>["<code>match</code>"]
+   does not <a for=map>exist</a>, then return <var>response</var>.
+
+   <li><p>Let <var>compressionDictionaryCache</var> be the result of
+   <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+   <li><p>If <var>compressionDictionaryCache</var> is null, then return <var>response</var>.
+
+   <li><p>Let <var>pattern</var> be the result of
+   <a for=/>creating a URL pattern</a> given the bare item of <var>dictionaryValue</var>["<code>match</code>"],
```
annevk: What is bare item?

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
 </ol>
 </div>
 
+<h3 id=http-network-compression-dictionary-fetch>HTTP-network compression-dictionary fetch</h3>
+
+<div algorithm>
+<p>To <dfn id=concept-http-network-compression-dictionary-fetch>HTTP-network compression-dictionary fetch</dfn>,
+given a <a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
+<var>includeCredentials</var> (default false), and an optional boolean <var>forceNewConnection</var>
+(default false), run these steps:
```
```suggestion
(default false):
```

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+ <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+
+ <li><p>Let <var>compressionDictionaryCache</var> be the result of
+ <a>determining the compression-dictionary cache partition</a> given <var>request</var>.
+
+ <li><p>If <var>compressionDictionaryCache</var> is null, then return the result of running
+ <a>HTTP-network fetch</a> given <var>fetchParams</var>, <var>includeCredentials</var>, and
+ <var>forceNewConnection</var>.
+
+ <li><p>Let <var>bestMatch</var> be the result of finding the best matching dictionary in
```
annevk: Can we add a link here as well? (In the way we refer to other RFC terms.)

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+ <var>compressionDictionaryCache</var> for <var>request</var> as defined in
+ [[!RFC9842]].
+
+ <li><p>If <var>bestMatch</var> is null, then return the result of running <a>HTTP-network fetch</a>
+ given <var>fetchParams</var>, <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+
+ <li><p>Add the `<code>Available-Dictionary</code>` and `<code>Dictionary-ID</code>`
+ (if applicable) headers to <var>request</var> using <var>bestMatch</var> as defined in
```
annevk: We have append/set algorithms for this. We should use them.

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+ `<code>Content-Encoding</code>` and <var>response</var>'s <a for=response>header list</a>.
+
+ <li><p>If <var>codings</var> is null or does not contain `<code>dcb</code>` or `<code>dcz</code>`,
+ then return <var>response</var>.
+
+ <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>",
+ then return a <a>network error</a>.
+
+ <li><p>Let <var>availableDictionaryItem</var> be the result of
+ <a>getting a structured field value</a> given `<code>Available-Dictionary</code>`,
```
annevk: add the for="header list" attribute to "getting a ..."

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+given a <a for=/>fetch params</a> <var>fetchParams</var>, an optional boolean
+<var>includeCredentials</var> (default false), and an optional boolean <var>forceNewConnection</var>
+(default false), run these steps:
+
+<ol>
+ <li><p>Let <var>request</var> be <var>fetchParams</var>'s <a for="fetch params">request</a>.
+
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is "<code>no-cors</code>", then return the
+ result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
+ <var>includeCredentials</var>, and <var>forceNewConnection</var>.
```
annevk: We have this return 5 times. I think we could restructure this with a lambda that initializes a variable. This would a) reduce repetition b) scope a number of variables to where they are needed c) hopefully add some clarity.

You would do something like:

> Let x be the result of running these steps:
> 1. ...

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+   <var>pattern</var>, <var>dictionaryValue</var> and <var>expirationTime</var>.
+  </ol>
+
  <li><p>Return <var>response</var>. <span class=note>Typically <var>response</var>'s
  <a for=response>body</a>'s <a for=body>stream</a> is still being enqueued to after
  returning.</span>
 </ol>
 </div>
 
+<h3 id=http-network-compression-dictionary-fetch>HTTP-network compression-dictionary fetch</h3>
```
annevk, line 6687: I wonder if here and below we also want a hyphen between network and compression. cc @sideshowbarker

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+ <li><p>If <var>availableDictionaryItem</var> is null, then return a <a>network error</a>.
+
+ <li><p>Let <var>availableDictionaryHash</var> be the bare item of <var>availableDictionaryItem</var>.
+
+ <li><p>Let <var>newBody</var> be a new <a for=/>body</a> whose <a for=body>stream</a> is the
+ result of transforming <var>response</var>'s <a for=response>body</a>'s <a for=body>stream</a>
+ with an algorithm that verifies that the dictionary hash in the stream matches
+ <var>availableDictionaryHash</var> and decodes the rest of the stream with the applicable
+ algorithm as defined in [[!RFC9842]]. If verification or decoding fails,
+ the transformed stream must error.
```
annevk: This must seems out-of-place.

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+
+ <li><p>Set <var>response</var>'s <a for=response>body</a> to <var>newBody</var>.
+
+ <li><p><a>Delete</a> `<code>Content-Encoding</code>` from <var>response</var>'s
```
annevk: needs a for attribute

fetch.bs
```diff
@@ -1967,6 +1973,10 @@ not always relevant and might require different behavior.
    <td>"<code>video</code>"
    <td><code>media-src</code>
    <td>HTML's <code>&lt;video></code> element
+  <tr>
+   <td>"<code>compression-dictionary</code>"
+   <td><code>connect-src</code>
+   <td>HTML's <code>&lt;link rel=compression-dictionary&gt;</code>
```
```suggestion
   <td>HTML's <code>&lt;link rel=compression-dictionary></code>
```

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+   <li><p>If <var>pattern</var> is failure or <var>pattern</var> <a for=/>has regexp groups</a>,
+   then return <var>response</var>.
+
+   <li><p>Let <var>expirationTime</var> be the time at which the <var>response</var> becomes
+   a <a>stale response</a>.
+
+   <li><p>If <var>expirationTime</var> is not in the future, then return <var>response</var>.
+
+   <li><p>Store <var>response</var> in <var>compressionDictionaryCache</var> with its associated
+   <var>pattern</var>, <var>dictionaryValue</var> and <var>expirationTime</var>.
```
annevk: Oxford comma

fetch.bs
```diff
@@ -6531,12 +6559,126 @@ run these steps:
...
+ <li><p>If <var>request</var>'s <a for=request>response tainting</a> is "<code>opaque</code>",
+ then return a <a>network error</a>.
+
+ <li><p>Let <var>availableDictionaryItem</var> be the result of
+ <a>getting a structured field value</a> given `<code>Available-Dictionary</code>`,
+ "<code>item</code>", and <var>request</var>'s <a for=request>header list</a>.
+
+ <li><p>If <var>availableDictionaryItem</var> is null, then return a <a>network error</a>.
+
+ <li><p>Let <var>availableDictionaryHash</var> be the bare item of <var>availableDictionaryItem</var>.
```
annevk: This could maybe be inlined, though again we need more clarity on bare item.

annevk:
I vaguely remember we discussed the token name incorrectly containing a hyphen, but I can't find it. Do you remember where that is? I suspect it's not worth correcting at this stage, but we should call it out in the specification is some kind of error we regret in the name of web compatibility so that we don't make it again for a new value.

fetch.bs
pmeenan: This mirrors the section above on HTTP cache partitions where it does "Return the unique HTTP cache associated with...".  The RFC itself doesn't go into detail about the cache itself and how it operates. Is that something that needs to be fully defined somewhere?

fetch.bs
pmeenan: FWIW, I was copying how HTTP-network fetch is handled which has a space in the string but hyphens in the ID (`<h3 id=http-network-fetch>HTTP-network fetch</h3>`). Happy to go either way though.

fetch.bs
pmeenan: Added details on how to handle invalid values for the other parameters. I don't explicitly call out that unknown keys should be ignored but I can add an explicit step to delete unknown keys if you think it would help.

fetch.bs
pmeenan: Thanks. I added similar language that it should return response if the URL Pattern creation throws an exception. I'll add WPT tests for that and the rest of the edge cases that aren't currently covered now (will take a few days to work their way through).

fetch.bs
pmeenan: Added a link to the structured field parsing reference of bare item.

pmeenan:
> I vaguely remember we discussed the token name incorrectly containing a hyphen, but I can't find it. Do you remember where that is? I suspect it's not worth correcting at this stage, but we should call it out in the specification is some kind of error we regret in the name of web compatibility so that we don't make it again for a new value.

@annevk the link relation discussion was [here](https://github.com/whatwg/html/issues/10162) which is where the token came from AFAIK.

fetch.bs
pmeenan: I just landed a (hopefully comprehensive) set of WPT tests to cover all of the edge cases I could think of: https://github.com/web-platform-tests/wpt/pull/60164

* Invalid and out-of-scope `match` properties (syntax failures and  cross-origin patterns).
* Invalid/unsupported dictionary `type` tokens.
* Comprehensive `match-dest` parsing (unknown destinations,  matching/non-matching destinations, and explicit wildcards).
* Dictionary `id` maximum length validation (exactly 1024 characters vs  exceeding 1025 characters).
* Robustness against unknown additional dictionary parameters.
* Rejection of entirely malformed structured headers  (`Use-As-Dictionary: ?0`) and non-cacheable responses (`max-age=0`).
* Opaque response tainting resulting from cross-origin redirects under  `no-cors` mode correctly ignoring registration.
* Evaluation of `Available-Dictionary` on matching redirect targets across  redirect chains.
* Precedence evaluation when matching overlapping pattern scopes.
* Decoding failures resulting from dictionary hash mismatches throwing  clean network errors for both `dcb` and `dcz`.

fetch.bs
fred-wang: Yeah, I was not sure what is the common way to handle this, however https://httpwg.org/specs/rfc9651.html#preserving-extensibility mentions this:

> Specifications that use Dictionaries can also allow for forward compatibility by requiring that the presence of -- as well as value and type associated with -- unknown keys be ignored. Subsequent specifications can then add additional keys, specifying constraints on them as appropriate.

So maybe the spec should be explicit about ignoring unknown keys.

Incidentally, https://httpwg.org/specs/rfc9651.html#error-handling says ignoring the entire field is the default behavior when field-specific constraints are violated so it's indeed good the spec is now explicit for id, type, etc that we only ignore the corresponding key.

(note: probably rfc9651 should also have similar wording)

fetch.bs
```diff
@@ -6585,12 +6620,148 @@ run these steps:
...
+ <li><p>If <var>availableDictionaryItem</var> is null, then return a <a>network error</a>.
+
+ <li><p>Let <var>availableDictionaryHash</var> be the <a>bare item</a> of <var>availableDictionaryItem</var>.
+
+ <li><p>Let <var>newBody</var> be a new <a for=/>body</a> whose <a for=body>stream</a> is the
+ result of transforming <var>response</var>'s <a for=response>body</a>'s <a for=body>stream</a>
+ with an algorithm that verifies that the dictionary hash in the stream matches
+ <var>availableDictionaryHash</var> and decodes the rest of the stream with the applicable
+ algorithm as defined in [[!RFC9842]]. If verification or decoding fails,
+ error the transformed stream.
```
fred-wang, line 6759: I'm not sure what's implied by this "error the transformed stream". Shouldn't we return a network error if decoding/verification fails? If not what do we expect for newBody in that case?

Also, about the hash verification, I don't see a lot from RFC 9842... Looking for "hash" or "available-dictionary", I find these paragraphs:
- https://www.rfc-editor.org/info/rfc9842/#name-available-dictionary
- https://www.rfc-editor.org/info/rfc9842/#section-2.3
- https://www.rfc-editor.org/info/rfc9842/#name-changing-content
- https://www.rfc-editor.org/info/rfc9842/#name-negotiating-the-content-enc

and in particular "The dictionary is validated using an SHA-256 hash of the content to make sure that the client and server are both using the same dictionary." from which one can infer the server is expected to send back the same hash if it has found the same on its side.

So can we actually just add a previous step that checks `availableDictionaryHash` matches the `Available-Dictionary` in request's header at step 9 and return a network error otherwise. Or Am I missing something?

pmeenan: `availableDictionaryHash` and `Available-Dictionary` are both generated by the client and attributes of the request. The issue we need to protect against is a server responding with a `dcb` or `dcz` stream that was compressed with a dictionary other than the one requested (has happened to more than a few people when deploying because of incorrectly-configured `Vary` response headers).

I'm happy to change it to be a network error of some kind if there's a sensible way to plumb that. Where it gets a bit complicated is that it's a problem with the stream, not necessarily the HTTP-level response container and is closer to being a corrupt payload (like sending a brotli payload with `Content-Encoding: zstd`). It also won't show up until the body starts being processed/read.

That feels like it's a problem at the stream level rather than the network level but I'm happy to plumb it however it best fits into the spec.

fred-wang: Thank you, I see! So this is referring to the remaining instances of "hash" (that I disregarded yesterday because I thought they were irrelevant):

https://www.rfc-editor.org/info/rfc9842/#name-dictionary-compressed-brotl
https://www.rfc-editor.org/info/rfc9842/#name-dictionary-compressed-zstan

> The header consists of a fixed 4-byte sequence and a 32-byte hash of the external dictionary that was used.

> A "Dictionary-Compressed Zstandard" stream is a binary stream that starts with a 40-byte fixed header

So what do you think of this minor clarifications:

> that first verifies that the dictionary hash in the stream's header matches availableDictionaryHash and decodes the rest of the stream with the applicable algorithm as defined in §4. Dictionary-Compressed Brotli and §5. Dictionary-Compressed Zstandard of [!RFC9842]].

So now I agree this is more an error at the stream level. Still I don't understand the implication of "error the transformed stream" (probably because I'm not familiar with the spec terminologies). What will be newBody after such an error?

fetch.bs
```diff
@@ -6585,12 +6620,148 @@ run these steps:
...
+
+   <li><p>If <var>dictionaryValue</var>["<code>id</code>"] <a for=map>exists</a> and its
+   <a>bare item</a>'s <a for=string>length</a> is greater than 1024, then <a for=map>remove</a>
+   <var>dictionaryValue</var>["<code>id</code>"].
+
+   <li>
+    <p>If <var>dictionaryValue</var>["<code>match-dest</code>"] <a for=map>exists</a>:
+
+    <ol>
+     <li><p>Let <var>matchDestList</var> be <var>dictionaryValue</var>["<code>match-dest</code>"][0].
```
fred-wang, line 6650: What does the `[0]`  mean here?

fetch.bs
fred-wang: Thank you so much for these tests! In general they look good but I have some suggestions/comments/questions:

> Dictionary with unknown match-dest is not used for fetch() API

So this is aligned with your PR because we return the response if matchDestList is empty (so similar to "Dictionary registration with empty match-dest list acts as wildcard'"). But can you please also add a similar test that does not hit the empty case:

```
compression_dictionary_promise_test(async (t) => {
  const match_dest = encodeURIComponent('("asdf", "")');
  ...
```

As I read your PR, "asdf" would be removed but the dictionary registration would still succeed, right? (similar to "Dictionary registration with matching fetch destination")

> Dictionary registration with 1024 character dictionary ID

For completeness, can we also have a test that checks the character range. That would also exercise parsing/serialization of the dquotes and backslash is properly done as per rfc965:

```
+compression_dictionary_promise_test(async (t) => {
+  // https://www.rfc-editor.org/info/rfc9651/#name-parsing-a-string
+  // double quotes and backslash are escaped in headers per RFC9651.
+  let dictionaryIDsuffix = " !#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+  let dictionaryID = `\\"\\\\${dictionaryIDsuffix}`;
+  const dict = await (await fetch(`${kRegisterDictionaryPath}?id=${encodeURIComponent(dictionaryID)}`)).text();
+  // Wait until `available-dictionary` header is available.
+  assert_equals(
+      await waitUntilAvailableDictionaryHeader(t, {}),
+      kDefaultDictionaryHashBase64);
+  assert_equals(await checkHeader('dictionary-id', {}), `"${dictionaryID}"`);
+}, `Dictionary registration with dictionary ID (valid characters and backslash escaping)`);
```

> Dictionary with 1025 character dictionary ID is not registered

This test does not seem aligned with your PR? The ID would be ignored, so that should just behave like the "Simple dictionary registration and unregistration" case?

Also we can have a test for invalid character range (like '€') but in that case https://httpwg.org/specs/rfc9651.html#error-handling says the entire field should be ignored, so we would indeed really check that there is no available-dictionary header here (as opposed to a successful registration without ID).

Note that per https://httpwg.org/specs/1.html#error-handling the spec author can decide whether they want to ignore the entire field or define other handling (here remove the id from the dictionary).

> Overlapping match patterns prioritize the more specific dictionary

Would be nice to test a bit more the priority from https://www.rfc-editor.org/info/rfc9842/#section-2.2.3

At least I think
- A Use-As-Dictionary with "match-dest" has priority over a Use-As-Dictionary without "match-dest" (with a longer "match" length even).
- Two Use-As-Dictionaries with the same match's length and matchDest's emptiness but different fetch time (the most recent wins).

Would be nice to test for the "best match" algorithms that involve more than 2 dictionaries too.

fred-wang: Correction:

- Actually "Dictionary registration with empty match-dest list acts as wildcard" and "Dictionary with unknown match-dest is not used for fetch() API" are contradictory. Since the unknown match-dest is removed they should behave the same. But one expect a dictionary to be registered without a match-dest requirement while the other expect the dictionary to be ignored.

- IIUC "New dictionary registration overrides the existing one" is already exercising the case I mentioned "Two Use-As-Dictionaries with the same match's length and matchDest's emptiness but different fetch time (the most recent wins)." so I guess the case missing is just priority of non-empty match-dest over empty match-dest.

fetch.bs
```diff
@@ -8527,7 +8699,7 @@ dictionary RequestInit {
   any window; // can only be set to null
 };
 
-enum RequestDestination { "", "audio", "audioworklet", "document", "embed", "font", "frame", "iframe", "image", "json", "manifest", "object", "paintworklet", "report", "script", "sharedworker", "style", "text", "track", "video", "worker", "xslt" };
+enum RequestDestination { "", "audio", "audioworklet", "compression-dictionary", "document", "embed", "font", "frame", "iframe", "image", "json", "manifest", "object", "paintworklet", "report", "script", "sharedworker", "style", "text", "track", "video", "worker", "xslt" };
```
fred-wang, line 8747: If you are adding a new destination, that might affect the internal priority calculated by step 15 of https://fetch.spec.whatwg.org/#fetching ; currently that's completely implementation-defined so don't really need to say anything in the spec and that's probably off topic for this PR... But anyway I'm curious for WebKit and Firefox to know how Chromium is calculating fetch priority for the "compression-dictionary" destination? (naively, I would suspect it is low priority).

pmeenan: Chrome defaults it to an idle-level priority request (and I believe it tries to wait until after the document onload has fired but I need to check to verify).

fetch.bs
pmeenan: I'll see if I can find a better way to word it but it is coming from the [tuple definition](https://infra.spec.whatwg.org/#tuple) from the structured field [parsing of an item](https://httpwg.org/specs/rfc9651.html#parse-item) where the individual items of the parsed dictionary are tuples.

fetch.bs
pmeenan: Thanks, I'll update the tests (may take a week to make it through the plumbing). 

As far as priority goes, path specificity is the only factor in prioritization (with freshness being a tie-breaker). `match-dest` is just a filter for what candidates can be considered.

For the unknown types, I'll see if something is broken with the current spec language and update it but the intent is:
- Default to match-everything (by defaulting to an empty list)
- If a `match-dest` are specified and it is not an empty list, ONLY match if the current dest is in the list.

For unsupported dests, this likely means we don't want to register them at all since there is nothing they could match and we don't want them to be treated like a wildcard. This might need some additional logic in the registration flow to remember the initial state of the `match-dest` list (empty or not) before filtering for known dest's and then drop the registration if the list is empty but it wasn't initially empty.

We want to avoid the case of a new dest being introduced and dictionaries that are targeting that dest accidentally behaving like a wildcard (and maybe taking priority over all other dictionaries for browsers that don't support the new dest).

fetch.bs
pmeenan: I'll see if I can find a better way to plumb the error that isn't hand-wavy (FWIW, I'm not all that well versed in writing specs so thanks for pushing on these unclear cases).

fetch.bs
fred-wang: > As far as priority goes, path specificity is the only factor in prioritization (with freshness being a tie-breaker). match-dest is just a filter for what candidates can be considered.

But https://www.rfc-editor.org/info/rfc9842/#name-multiple-matching-dictionar actually does use match-dest non-emptiness as the 1st factor for prioritization, with match length being second and freshness being third:

> 1. For clients that support request destinations, a dictionary that specifies and matches a "match-dest" takes precedence over a match that does not use a destination
> 2. Given equivalent destination precedence [...]
> 3. Given equivalent destination and match length precedence [...]

pmeenan: Yikes - thanks for pointing that out - Chrome's implementation probably needs to be fixed then (sorry, been working on it for so many years at this point that I'm forgetting parts).  I'll fix the tests to make sure they test the behavior (and fix Chrome's implementation).

pmeenan: FYI, I updated the PR to fail registration when `id` is invalid to match the test. Generally we want to ignore unknown fields (or unknown values in fields where we expect for future extensibility of values) but known fields with invalid values we should fail the registration.

fetch.bs
fred-wang: OK, I see thanks

fetch.bs
```diff
@@ -6585,12 +6620,148 @@ run these steps:
...
+   <a>bare item</a>'s <a for=string>length</a> is greater than 1024, then return
+   <var>response</var>.
+
+   <li>
+    <p>If <var>dictionaryValue</var>["<code>match-dest</code>"] <a for=map>exists</a>:
+
+    <ol>
+     <li><p>Let <var>matchDestList</var> be <var>dictionaryValue</var>["<code>match-dest</code>"][0].
+
+     <li><p>For each <var>dest</var> of <var>matchDestList</var>: if <var>dest</var>'s <a>bare item</a>
```
fred-wang, line 6652: So per what you said earlier, condition this and the next item on whether matchDestList is nonempty?

fetch.bs
fred-wang: > FYI, I updated the PR to fail registration when id is invalid to match the test.
> For unsupported dests, this likely means we don't want to register them at all since there is nothing they could match and we don't want them to be treated like a wildcard.

OK I think these address my remaining concerns regarding mismatch between spec and tests. For match-dest, I believe the match-dest change is still pending, but probably you can just do an early check for the emptiness of match-dest rather than remembering the initial state (see my other comment above).

fetch.bs
fred-wang: @pmeenan: By idle-level priority, are you referring to https://github.com/whatwg/html/pull/11620/changes#r3385960016 i.e. when the client decides to perform the fetch ? By "fetch priority" I was talking about https://web.dev/articles/fetch-priority?hl=en#effects which is destination-dependent and will affect how the server prioritizes requests (for example via [RFC9218](https://www.rfc-editor.org/rfc/rfc9218.html)). Since you are introducing a new destination, we need to decide how to handle it. I guess compression-dictionary destination would default to "VeryLow" / "Lowest" (using web.dev article's terminology). And we also need to decide a non-auto fetchpriority affects that default value, if that's desired.

fetch.bs
```diff
@@ -1970,6 +1983,10 @@ not always relevant and might require different behavior.
    <td>"<code>video</code>"
    <td><code>media-src</code>
    <td>HTML's <code>&lt;video></code> element
+  <tr>
+   <td>"<code>compression-dictionary</code>"
+   <td><code>connect-src</code>
```
fred-wang, line 1988: We need to add tests to check that the connect-src directive is taken into account when fetching with a compression-dictionary destination.

fred-wang: Done at https://github.com/web-platform-tests/wpt/pull/61309

fetch.bs
fred-wang: Even if it's implementation-defined, it seems there is some motivation to have very low priority for "compression-dictionary" destinations and that seems to be what Chromium implements. What about having some non-normative note suggesting that?

(Note: I was checking if it's possible to write a test for fetchpriority, it is mentioned on https://html.spec.whatwg.org/multipage/semantics.html#processing-link-headers:link-processing-options-3 but does not seem to be exposed by https://fetch.spec.whatwg.org/#request ; so one really need to check the behavior, which is implementation-defined...)

fetch.bs
fred-wang: > The issue we need to protect against is a server responding with a dcb or dcz stream that was compressed with a dictionary other than the one requested (has happened to more than a few people when deploying because of incorrectly-configured Vary response headers).

I a few questions regarding this incorrectly-configured Vary response headers. I was re-reading https://datatracker.ietf.org/doc/html/rfc9842#section-6.2 and the Vary header is only required when serving cacheable compressed reponse (I assume for other content-encoding than dcb/dcz too but that's probably defined in other spec).

If I understand the issue you refer to is when the client first retrieves a resource with or without a compression dictionary for it, and then retrieves it again after a better matching dictionary become available or after the initial dictionary is removed from the cache. **Do we have tests for these cases?**

For these cases, the available-dictionary header would change and if the server properly setting the Vary header and if the client properly takes them into account, things should just work as expected...

That being said, searching for "vary", the only match is this test:
https://github.com/web-platform-tests/wpt/blob/92ce110bdfaafc84acc1e37b4bbae549f112b8aa/fetch/compression-dictionary/dictionary-registration.tentative.https.html#L62
and it says the opposite: browsers cache should not vary based on those headers!
I understand why the hash protection becomes even more relevant. **Is this behavior specified anywhere?**. https://chromium-review.googlesource.com/c/chromium/src/+/6634910 does not provide much context.

(incidentally, I submitted https://chromium-review.googlesource.com/c/chromium/src/+/8163980 to better align the server with RFC9842 while still testing this case)

pmeenan: Cache implementations where it comes to dictionary-compressed assets can be "complicated". ACTUALLY varying on the `Available-Dictionary` is important for middle-boxes to make sure they serve the correct dictionary-compressed asset to clients that actually have the dictionary that was used to compress the resource. For something like delta-encoded script libraries, the "current" version might be compressed against the last N different previous versions so that a delta can be sent relative to whatever a given client has from when they last visited (particularly for apps that update their scripts frequently).

The browser's private cache is orders-of-magnitude more complicated:

- If the browser varies based on `Available-Dictionary` and a response's `Use-As-Dictionary` header includes itself in the `match` pattern (common for delta-updating scripts), then a cached visit to the same page that uses the same library will now have an `Available-Dictionary` of the hash of the current version of the code, not matching whatever it sent the first time, resulting in a cache miss even though the resource hasn't changed. It turns a cacheable resource into a 1-time cache miss, every time.

- If the browser STORES the response in cache in dictionary-compressed form, the cached version now depends on the dictionary used at fetch time being available to decode the response. If that dictionary was also a dictionary-compressed response then it also depends on the dictionary that was used for it's original request, creating a very long, fragile dependency chain. This can be solved a few different ways at implementation time (keep the dictionary dependency but make sure the dictionary is stored in it's whole form or cache a decoded version of the response). I'm not sure we want to force any specific implementation decisions on how clients handle the dependency chain issue internally (though it is probably worth making sure it is documented for their awareness).

simoneonofri:
Hi all, thanks for your patience, [following up on the TAG discussion](https://github.com/w3ctag/design-reviews/issues/1203) and the discussion in SING.

After checking this PR, I think the issue's immediate concern is no longer present in that patch:
- no-cors -specific handling was removed
- no-cors requests fall back before dictionary negotiation;
- registration is limited to non-opaque responses in secure contexts
- an opaque dictionary-compressed response produces a network error

And this is in line with [RFC 9842 Section 9.3.2](https://www.rfc-editor.org/rfc/rfc9842.html#section-9.3.2) boundary.

If opaque-response support returns in a separate proposal, however, `CORP: cross-origin` would not be an equivalent signal, as it permits a no-cors response to pass CORP but does not make it CORS-readable.

fetch.bs
```diff
@@ -6585,12 +6620,152 @@ run these steps:
...
+
+ <li>
+  <p>Let <var>fallback</var> be the following steps:
+
+  <ol>
+   <li><p>Return the result of running <a>HTTP-network fetch</a> given <var>fetchParams</var>,
+   <var>includeCredentials</var>, and <var>forceNewConnection</var>.
+  </ol>
+
+ <li><p>If <var>request</var>'s <a for=request>mode</a> is "<code>no-cors</code>", then return the
```
fred-wang, line 6706: compressed-large-resources-001 gives an example of compressed resources using img/script/link tags (as opposed to other tests that rely on the fetch API) and when we tried to implement this condition in WebKit, the test was failing unless we use an explicit crossorigin attribute to avoid triggering a no-cors request.

This indeed does not seem aligned with what Chromium is doing here:
https://source.chromium.org/chromium/chromium/src/+/main:services/network/cors/cors_url_loader.cc;l=366

I understand you mean to still allow fetching compression dictionaries for same origin requests too?

See also https://chromium-review.googlesource.com/c/chromium/src/+/8353624

PR #1856, opened 2025-09-13
Add note that `signal` is always initialized

TimvdLippe:
The getter steps [1] specify to return `this.signal`. Its signal is nullable [2], but it is always initialized in the constructor and when cloning.

[1]: https://fetch.spec.whatwg.org/#dom-request-signal
[2]: https://fetch.spec.whatwg.org/#request-signal

***
<a href="https://whatpr.org/fetch/1856.html" title="Last updated on Sep 18, 2025, 1:54 PM UTC (f275bd3)">Preview</a> | <a href="https://whatpr.org/fetch/1856/e28f9a5...f275bd3.html" title="Last updated on Sep 18, 2025, 1:54 PM UTC (f275bd3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -8254,7 +8254,7 @@ interface Request {
   readonly attribute boolean keepalive;
   readonly attribute boolean isReloadNavigation;
   readonly attribute boolean isHistoryNavigation;
-  readonly attribute AbortSignal signal;
+  readonly attribute AbortSignal? signal;
   readonly attribute RequestDuplex duplex;
 
   [NewObject] Request clone();
--- a/fetch.bs
+++ b/fetch.bs
@@ -8748,7 +8748,7 @@ constructor steps are:
  <li><p>Let <var>signals</var> be « <var>signal</var> » if <var>signal</var> is non-null; otherwise
  « ».
 
- <li><p>Set <a>this</a>'s <a for=Request>signal</a> to the result of
+ <li id="signal-initialized-in-constructor"><p>Set <a>this</a>'s <a for=Request>signal</a> to the result of
  <a>creating a dependent abort signal</a> from <var>signals</var>, using {{AbortSignal}} and
  <a>this</a>'s <a>relevant realm</a>.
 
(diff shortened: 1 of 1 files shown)


TimvdLippe:
This seems editorial to me, but I am actually not sure. Do let me know if this does require test coverage and all if it turns out it is normative.

annevk:
I don't think this is correct. While it does start out as null, it is always initialized. By the time the getter returns it will always get an `AbortSignal` object.

TimvdLippe:
Hm, I am implementing it in Servo atm and it does return null sometimes. I haven't been able to figure out why though.

Should we maybe update the steps to assert it is not null and then return this.signal? That makes the invariant explicit

annevk:
I think IDL is that assert. And in a future world the Web IDL specification should have such a generic assert in its getter wrapper. (That what the specification returns matches what it needs to return, or at least matches that post any Infra-to-IDL-to-JS-type conversion.)

TimvdLippe:
Will investigate why Servo is running into the assert and let you know if it is an implementation issue or not.

TimvdLippe:
Seems like I missed implementing the parts of `cloning` the request. Now it works correctly and we always return a non-null value. Should we add a note about this or do you think that's too much?

annevk:
A note for the getter? I suppose that's reasonable.

fetch.bs
```diff
@@ -8919,6 +8919,10 @@ set; otherwise false.
 <p>The <dfn attribute for=Request><code>signal</code></dfn> getter steps are to return <a>this</a>'s
 <a for="Request">signal</a>.
 
+<p class="note"><a>this</a>'s <a for="Request">signal</a> is always initialized in the
+<a href="#ref-for-request-signal②">constructor</a> and when
+<a href="#ref-for-request-signal⑤">cloning</a>.
```
TimvdLippe: Hm, I wanted to be fancy and point to the exact assignment that assures this. However, the numbers already don't match, as this note also refers to signal. Is there a less brittle way to do this or should we link to the algorithm as a whole and that's sufficient?

fetch.bs
```diff
@@ -8919,6 +8919,10 @@ set; otherwise false.
 <p>The <dfn attribute for=Request><code>signal</code></dfn> getter steps are to return <a>this</a>'s
 <a for="Request">signal</a>.
 
+<p class="note"><a>this</a>'s <a for="Request">signal</a> is always initialized in the
+<a href="#ref-for-request-signal②">constructor</a> and when
+<a href="#ref-for-request-signal⑤>cloning</a>.
```
annevk: Let's capitalize "This" (or reword). We also can't use those backreferences unfortunately as they are not stable. Although maybe we could add these as ID to those places to make them stable. (Although it might well be that this ends up renumbering others in a weird way, not sure.)

TimvdLippe: Added identifiers to the list items and updated links. They all work now.

fetch.bs
```diff
@@ -8933,7 +8937,7 @@ set; otherwise false.
  <li><p>Let <var>clonedRequest</var> be the result of <a lt=clone for=request>cloning</a>
  <a>this</a>'s <a for=Request>request</a>.
 
- <li><p><a for=/>Assert</a>: <a>this</a>'s <a for=Request>signal</a> is non-null.
+ <li id="signal-initialized-when-cloning"><p><a for=/>Assert</a>: <a>this</a>'s <a for=Request>signal</a> is non-null.
```
annevk: I think we should add this ID to the next step, where it actually happens.

PR #1858, opened 2025-09-28
Throw `SecurityError` on non-trustworthy URLs in `fetchLater`

TimvdLippe:
Per https://github.com/web-platform-tests/wpt/blob/c009ca2fa7d311b1a257222aa0f9fd7706ffa4d6/fetch/fetch-later/basic.https.window.js#L54-L59 it should not throw a `TypeError`, but a `SecurityError` instead.

***
<a href="https://whatpr.org/fetch/1858.html" title="Last updated on Sep 28, 2025, 8:54 AM UTC (e869c75)">Preview</a> | <a href="https://whatpr.org/fetch/1858/5a015d1...e869c75.html" title="Last updated on Sep 28, 2025, 8:54 AM UTC (e869c75)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -9440,7 +9440,7 @@ method steps are:
  <a>HTTP(S) scheme</a>, then throw a {{TypeError}}.
 
  <li><p>If <var>request</var>'s <a for=request>URL</a> is not a <a>potentially trustworthy URL</a>,
- then throw a {{TypeError}}.
+ then throw a {{SecurityError}}.
 
  <li>
   <p>If <var>request</var>'s <a for=request>body</a> is not null, and <var>request</var>'s


annevk:
cc @noamr

annevk:
I wonder if the test is wrong here. I think we decided that we should align with `fetch()` on throwing `TypeError` everywhere.

noamr:
> I wonder if the test is wrong here. I think we decided that we should align with `fetch()` on throwing `TypeError` everywhere.

Yea that's what I recall as well.
We should fix the test and chromium implementation 
Cc @mingyc

TimvdLippe:
As part of the Servo PR I can update the test. Shall I?

annevk:
Yeah please. After that this PR can be closed. Thank you!

noamr:
> As part of the Servo PR I can update the test. Shall I?

Thanks for opening these issues as part of Servo, it helps makes the specs & general interop bettter!

PR #1859, opened 2025-09-28
Expose `fetchLater` only in a SecureContext

TimvdLippe:
Tested per https://github.com/web-platform-tests/wpt/blob/c009ca2fa7d311b1a257222aa0f9fd7706ffa4d6/fetch/fetch-later/non-secure.window.js

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -9269,7 +9269,7 @@ interface FetchLaterResult {
 };
 
 partial interface Window {
-  [NewObject] FetchLaterResult fetchLater(RequestInfo input, optional DeferredRequestInit init = {});
+  [NewObject, SecureContext] FetchLaterResult fetchLater(RequestInfo input, optional DeferredRequestInit init = {});
 };
 </pre>
 


TimvdLippe:
https://github.com/web-platform-tests/wpt/pull/55123 updates the IDL test to run in a secure context to check for the existence of `fetchLater`. Currently Chromium browsers fail these tests: https://wpt.fyi/results/fetch/api/idlharness.any.html?label=experimental&label=master&aligned

annevk:
cc @noamr

noamr:
LGTM
CC @mingyc

annevk:
This is a normative change and therefore would normally require the PR template. I'm going to waive that requirement given that Chromium is supportive and is the only implementation so far (out of Chromium, Gecko, and WebKit) and we already have test coverage. It should also be non-controversial as this API is partially permission-gated and permissions shouldn't work without HTTPS.

annevk:
Thanks @TimvdLippe!

noamr:
> This is a normative change and therefore would normally require the PR template. I'm going to waive that requirement given that Chromium is supportive and is the only implementation so far (out of Chromium, Gecko, and WebKit) and we already have test coverage. It should also be non-controversial as this API is partially permission-gated and permissions shouldn't work without HTTPS.

Also it was gated by a security error anyway when used outside of a potentially trustworthy context. So this mainly reduces the error surface rather than changes functionality

PR #1862, opened 2025-10-02
Add a textStream() method to the Body mixin

### LINKED ISSUE #1861: Convenience method to stream a response as text
### What problem are you trying to solve?

`response.body` is a stream of `Uint8Array` but often you want to deal with text. `response.text()` decodes the response body as text and returns it, but all at once, not streaming.

### What solutions exist today?

```js
let response = await fetch('/');
let decoder = new TextDecoderStream();
response.body.pipeThrough(decoder);
```

### How would you solve it?

A `textStream()` helper that gives you a stream of decoded text chunks instead.

### Anything else?

This is mainly motivated by https://github.com/whatwg/html/issues/11669, where a `ReadableStream` will needed. If it's as easy to get a string stream as a byte stream, then we could probably make those APIs only accept strings. (My current thinking was to bake text decoding into them to reduce boilerplate, but @noamr suggested a new method instead.)

foolip:
`textStream()` is a convenience method that pipes a body through a utf-8 `TextDecoderStream`,
as it's a common use case for e.g. HTML streaming or any other text streaming.

Fixes https://github.com/whatwg/fetch/issues/1861.

   * Chromium
   * WebKit (see below)
   * https://github.com/web-platform-tests/wpt/pull/60356
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests)     
   * Chromium: [Implemented](https://issues.chromium.org/issues/514448226)
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2044628
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=316191
   * Deno (not for CORS changes): https://github.com/denoland/deno/issues/34792
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: this is done by chromium release process.

***
<a href="https://whatpr.org/fetch/1862.html" title="Last updated on Jun 4, 2026, 3:06 PM UTC (6036a4c)">Preview</a> | <a href="https://whatpr.org/fetch/1862/301650c...6036a4c.html" title="Last updated on Jun 4, 2026, 3:06 PM UTC (6036a4c)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -8503,8 +8503,8 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>UTF-8 decode</a>
  <a>relevant realm</a>.
 
  <li>
-  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with <a
-  for=/>UTF-8</a>.
+  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with
+  <a for=/>UTF-8</a>.
 
   <p class=note>This is done regardless of the presence or the value of a
   `<code>Content-Type</code>` header and regardless of the presence or the value of a
--- a/fetch.bs
+++ b/fetch.bs
@@ -8487,7 +8487,13 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>UTF-8 decode</a>
 
  <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object.
 
- <li><p>Return the result of piping <var>stream</var> through <var>decoder</var>.
+ <li>
+  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with <a for=/>UTF-8</a>.
+  <p class=note>This is done regardless of the presence or the value of a
+  `<code>Content-Type</code>` header and regardless of the presence or the value of a
+  `<code>charset</code>` parameter.</p>
+
+ <li><p>Return the result of <var>stream</var>, <a>piped through</a> <var>decoder</var>.
 </ol>
 </div>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -8302,8 +8302,8 @@ due course.
  <dt><code><var>requestOrResponse</var> . <a method for=Body>text</a>()</code>
  <dd><p>Returns a promise fulfilled with <var>requestOrResponse</var>'s body as string.
 
- <dt><code><var>requestOrResponse</var> . <a method for=Body>textStream</a>()</code>
- <dd><p>Returns a {{ReadableStream}} with <var>requestOrResponse</var>'s body as string chunks.
+ <dt><code><var>requestOrResponse</var> . <a method for=Body>textStream</a>()</code> <dd><p>Returns
+ a {{ReadableStream}} with <var>requestOrResponse</var>'s body as string chunks.
 </dl>
 
 <hr>
@@ -8481,15 +8481,30 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>UTF-8 decode</a>
 <p>The <dfn method for=Body><code>textStream()</code></dfn> method steps are:</p>
 
 <ol>
- <li><p>If <span>this</span>'s <a for=Body>body</a> is null, throw a {{TypeError}}.
+ <li><p>If <span>this</span> is <a for=Body>unusable</a>, then throw a {{TypeError}}.
+
+ <li>
+  <p>If <span>this</span>'s <a for=Body>body</a> is null, then:
+
+  <ol>
+   <li><p>Let <var>emptyStream</var> be a new {{ReadableStream}} in <a>this</a>'s <a>relevant
+   realm</a>,
+
+   <li><p><a for="ReadableStream">Set up</a> <var>emptyStream</var>.
+
+   <li><p><a for="ReadableStream">Close</a> <var>emptyStream</var>.
+
+   <li><p>Return <var>emptyStream</var>.
+  </ol>
 
  <li><p>Let <var>stream</var> be <a>this</a>'s <a for=Body>body</a>'s <a for=body>stream</a>.
 
- <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object.
+ <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object in <a>this</a>'s <a>relevant
+ realm</a>.
 
  <li>
-  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with <a for=/>UTF-8</a>.
-  <p class=note>This is done regardless of the presence or the value of a
+  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with <a
+  for=/>UTF-8</a>. <p class=note>This is done regardless of the presence or the value of a
   `<code>Content-Type</code>` header and regardless of the presence or the value of a
   `<code>charset</code>` parameter.</p>
 
--- a/fetch.bs
+++ b/fetch.bs
@@ -8259,6 +8259,7 @@ interface mixin Body {
   [NewObject] Promise&lt;FormData> formData();
   [NewObject] Promise&lt;any> json();
   [NewObject] Promise&lt;USVString> text();
+  [NewObject] ReadableStream textStream();
 };</pre>
 
 <p class=note>Formats you would not want a network layer to be dependent upon, such as
@@ -8300,6 +8301,9 @@ due course.
 
  <dt><code><var>requestOrResponse</var> . <a method for=Body>text</a>()</code>
  <dd><p>Returns a promise fulfilled with <var>requestOrResponse</var>'s body as string.
+
+ <dt><code><var>requestOrResponse</var> . <a method for=Body>textStream</a>()</code>
+ <dd><p>Returns a {{ReadableStream}} with <var>requestOrResponse</var>'s body as string chunks.
 </dl>
 
 <hr>
@@ -8471,6 +8475,20 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from 
 <div algorithm>
 <p>The <dfn method for=Body><code>text()</code></dfn> method steps are to return the result
 of running <a for=Body>consume body</a> with <a>this</a> and <a>UTF-8 decode</a>.
+</div>
+
+<div algorithm>
+<p>The <dfn method for=Body><code>textStream()</code></dfn> method steps are:</p>
+
+<ol>
+ <li><p>If <span>this</span>'s <a for=Body>body</a> is null, throw a {{TypeError}}.
+
+ <li><p>Let <var>stream</var> be <a>this</a>'s <a for=Body>body</a>'s <a for=body>stream</a>.
+
+ <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object.
+
+ <li><p>Return the result of piping <var>stream</var> through <var>decoder</var>.
+</ol>
 </div>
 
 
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -8246,6 +8250,20 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
 
+<div algorithm>
+<p>The <dfn method for=Body><code>textStream()</code></dfn> method steps are:</p>
+
+<ol>
+ <li><p>If <span>this</span>'s <a for=Body>body</a> is null, throw a {{TypeError}}.
+
+ <li><p>Let <var>stream</var> be <a>this</a>'s <a for=Body>body</a>'s <a for=body>stream</a>.
+
+ <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object.
```
foolip: This and the next step would need improving if there's interest in this method.

The constructor steps for `TextDecoderStream` aren't implicitly run by just saying "new object", but what I'm after is an UTF-8 decoder with the default settings.

fetch.bs
```diff
@@ -8246,6 +8250,20 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
+
+ <li><p>Return the result of piping <var>stream</var> through <var>decoder</var>.
```
foolip: Here "pipe through" would need to link to something. I'm not sure if there are other APIs that internally do something similar that I could copy?

fetch.bs
annevk: @MattiasBuelens or @ricea might have thoughts, but ideally we don't use public API for internal algorithms.

fetch.bs
```diff
@@ -8246,6 +8250,20 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
 of running <a for=Body>consume body</a> with <a>this</a> and <a>UTF-8 decode</a>.
 </div>
 
+<div algorithm>
+<p>The <dfn method for=Body><code>textStream()</code></dfn> method steps are:</p>
```
annevk, line 8481: I think for the null case it'd be nicer to return an empty stream, similar to what `text()` does.

Another thing that's not entirely clear to me is whether piping ends up making the underlying stream unusable so that you can't invoke `text()` after `textStream()` for instance. (That's a property we want.)

foolip: Ah, I didn't realize what `text()` does if `this.body` is null. I agree an empty stream would be better, code that doesn't consider this case would likely just work.

I thought this should behave like when you try to use both `blob()` and `bytes()`, or indeed call one of those twice. That throws a TypeError in the first step of https://fetch.spec.whatwg.org/#concept-body-consume-body.

Since it's not possible to call `text()` twice, should it really be possible to call `text()` after `textStream()`?

annevk: It should not be possible.

foolip: Good, that should be easy to achieve, even though the current spec text doesn't make clear what happens.

fetch.bs
MattiasBuelens: We should probably expose an algorithm to create a `TextDecoderStream` from other specifications, [like we do for Streams](https://streams.spec.whatwg.org/#other-specs).

fetch.bs
MattiasBuelens: You should be able to use the [pipe through](https://streams.spec.whatwg.org/#readablestream-pipe-through) algorithm exposed for other specifications. `[=piping through=]` should work. 😉

noamr:
See corresponding FileAPI PR: https://github.com/w3c/FileAPI/pull/221

fetch.bs
noamr: How deep does this need to go? Seems like the internal algorithms in the encoding spec operate on a `TextDecoderStream` object rather than on an internal representation which can make this into quite a big yak shaving for a small ergonomic API addition.

MattiasBuelens: We can still use `TextDecoderStream`, that's not the problem. The question is rather: what is the "correct" wording to create it and call [its specified constructor](https://encoding.spec.whatwg.org/#dom-textdecoderstream)?

Web IDL seems to assume that constructor operations are only intended to be called by the JavaScript binding, not by other specifications. That's why we came up with the ["create + setup" pattern for `ReadableStream`](https://streams.spec.whatwg.org/#readablestream-set-up), since those only use exported *algorithms* instead of *operations*.

I agree that it would be cumbersome to create both an "algorithm" and "constructor operation" version of every constructor, just to satisfy both use cases. If there's already an agreed-upon way for specifications to refer to and call a constructor operation, I'd love to know. 🙂

noamr: OK, splitting the constructor itself into a web-exposed constructor and internal setup steps seem reasonable enough.

noamr: See https://github.com/whatwg/encoding/pull/364

reillyeon:
If an HTTP response includes a `Content-Type` header with `charset` to something other than `utf-8` should this method use that when decoding the byte stream to a text stream? It seems like in some other places we explicitly say that `charset` is ignored and the data is decoded as UTF-8 so we probably need something similar here to disambiguate.

noamr:
> If an HTTP response includes a `Content-Type` header with `charset` to something other than `utf-8` should this method use that when decoding the byte stream to a text stream? It seems like in some other places we explicitly say that `charset` is ignored and the data is decoded as UTF-8 so we probably need something similar here to disambiguate.

Yes, this is discussed on the issue (#1861)

reillyeon:
Ack, I just don't see any mention of it in the text added by the PR, while other methods like `text()` and `formData()` are explicit about it.

annevk:
A note similar to what we have for `formData()` would probably be good.

fetch.bs
noamr: Done. `textStream` now returns an empty stream for a null body.

fetch.bs

fetch.bs

noamr:
> A note similar to what we have for `formData()` would probably be good.

Done

noamr (commented):
Using the new "set up" algo and also resolved some other comments.

fetch.bs
```diff
@@ -8300,6 +8301,9 @@ due course.
 
  <dt><code><var>requestOrResponse</var> . <a method for=Body>text</a>()</code>
  <dd><p>Returns a promise fulfilled with <var>requestOrResponse</var>'s body as string.
+
+ <dt><code><var>requestOrResponse</var> . <a method for=Body>textStream</a>()</code> <dd><p>Returns
```
annevk: Is this another specfmt bug?

fetch.bs
```diff
@@ -8473,6 +8477,41 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
 </div>
 
+<div algorithm>
+<p>The <dfn method for=Body><code>textStream()</code></dfn> method steps are:</p>
+
+<ol>
+ <li><p>If <span>this</span> is <a for=Body>unusable</a>, then throw a {{TypeError}}.
+
+ <li>
+  <p>If <span>this</span>'s <a for=Body>body</a> is null, then:
```
```suggestion
  <p>If <span>this</span>'s <a for=Body>body</a> is null:
```

fetch.bs
```diff
@@ -8473,6 +8477,41 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
+
+  <ol>
+   <li><p>Let <var>emptyStream</var> be a new {{ReadableStream}} in <a>this</a>'s <a>relevant
+   realm</a>,
```
annevk: No wrapping inside phrasing elements. And there's also a comma that should be period here...

fetch.bs
```diff
@@ -8473,6 +8477,41 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
+
+   <li><p><a for="ReadableStream">Close</a> <var>emptyStream</var>.
+
+   <li><p>Return <var>emptyStream</var>.
+  </ol>
+
+ <li><p>Let <var>stream</var> be <a>this</a>'s <a for=Body>body</a>'s <a for=body>stream</a>.
+
+ <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object in <a>this</a>'s <a>relevant
+ realm</a>.
```
annevk: No wrapping inside phrasing elements.

fetch.bs
```diff
@@ -8473,6 +8477,41 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
+
+ <li>
+  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with <a
+  for=/>UTF-8</a>. <p class=note>This is done regardless of the presence or the value of a
```
annevk: specfmt?

fetch.bs
noamr: Yes. I double checked the formatting and *then* ran `specfmt`, should do it in reverse order next time. Anyway, fixed

annevk: I'd recommend committing, then running specfmt and reviewing the diff.

fetch.bs
```diff
@@ -8473,6 +8477,43 @@ of running <a for=Body>consume body</a> with <a>this</a> and <a>parse JSON from
...
+  </ol>
+
+ <li><p>Let <var>stream</var> be <a>this</a>'s <a for=Body>body</a>'s <a for=body>stream</a>.
+
+ <li><p>Let <var>decoder</var> be a new {{TextDecoderStream}} object in <a>this</a>'s
+ <a>relevant realm</a>.
+
+ <li>
+  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with <a
+  for=/>UTF-8</a>.
```
annevk: no newline inside phrasing

fetch.bs
```diff
@@ -8300,6 +8301,9 @@ due course.
 
  <dt><code><var>requestOrResponse</var> . <a method for=Body>text</a>()</code>
  <dd><p>Returns a promise fulfilled with <var>requestOrResponse</var>'s body as string.
+
+ <dt><code><var>requestOrResponse</var> . <a method for=Body>textStream</a>()</code>
+ <dd><p>Returns a {{ReadableStream}} with <var>requestOrResponse</var>'s body as string chunks.
```
```suggestion
 <dd><p>Returns a {{ReadableStream}} object with <var>requestOrResponse</var>'s body as string chunks.
```

annevk (approved):
The _emptyStream_ thing is both correct and annoying.

fetch.bs
noamr:
```suggestion
  <p><a data-lt="set up a text decoder stream">Set up</a> <var>decoder</var> with
  <a for=/>UTF-8</a>.
```

noamr:
I've ticked all the checkboxes, apart from implementer support. Do we need a whole "Standards position" for this @annevk @zcorpan ?

annevk:
You can consider WebKit supportive. I left one nit on the test coverage. Thanks by the way for making those tentative PRs easily reviewable on WPT, that's really nice and maybe we should just enshrine that somewhere so that everyone starts doing that.

Once the tests are done feel free to land them and the specification changes, but please be careful to follow the commit message guidelines (ask me if you're unsure).

noamr:
> You can consider WebKit supportive. I left one nit on the test coverage. Thanks by the way for making those tentative PRs easily reviewable on WPT, that's really nice and maybe we should just enshrine that somewhere so that everyone starts doing that.

I thought it was the standard way of doing this? It's at least on the WHATWG PR template

annevk:
It's supposed to be, yes. But people don't always do it well and then you have to dig through Mozilla or Chromium changelogs to find what you want.

noamr:
> Once the tests are done feel free to land them and the specification changes, but please be careful to follow the commit message guidelines (ask me if you're unsure).

I don't have permissions to land the spec changes.

PR #1867, opened 2025-10-14
Handle `Unencoded-Digest` assertions.

mikewest:
Signature-based Integrity [1] relies on user agents properly handling `Unencoded-Digest` headers [2] which deliver a server's assertions about the integrity of a given response's body.

This patch extracts the relevant algorithms from [1], spelling out the processing model for the header, and verifying response integrity at the end of Main Fetch, alongside SRI's existing check.

This is one step of the Signature-based Integrity upstreaming work detailed in [3].

[1]: https://wicg.github.io/signature-based-sri/
[2]: https://httpwg.org/http-extensions/draft-ietf-httpbis-unencoded-digest.html
[3]: https://github.com/WICG/signature-based-sri/issues/49

- [X] At least two implementers are interested (and none opposed):
   * Chromium [shipped support](https://chromestatus.com/feature/5032324620877824) for this functionality in M141.
   * WebKit expressed support in https://github.com/WebKit/standards-positions/issues/434
   * Mozilla's opinions have been solicited in https://github.com/mozilla/standards-positions/issues/1139
   * https://wpt.fyi/results/subresource-integrity/unencoded-digest/tentative?label=experimental&label=master&aligned
   * Chromium: …
   * Gecko: …
   * WebKit: …
- [X] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use.

***
<a href="https://whatpr.org/fetch/1867.html" title="Last updated on Nov 19, 2025, 8:22 AM UTC (307d393)">Preview</a> | <a href="https://whatpr.org/fetch/1867/0e72db8...307d393.html" title="Last updated on Nov 19, 2025, 8:22 AM UTC (307d393)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -4508,19 +4508,19 @@ represent a digest of the response produced via the specified algorithm. [[!UNEN
 <a>byte sequence</a> <var>bytes</var> and a <a for=/>header list</a> <var>list</var>:
 
 <ol>
- <li><p>Let <var>header</var> be the result of
+ <li><p>Let <var>value</var> be the result of
  <a for="header list" lt="get a structured field value">getting</a> the
  `<a http-header><code>Unencoded-Digest</code></a>` header as a "<code>dictionary</code>" from
  <var>list</var>.
 
- <li><p>If <var>header</var> is null, then return <b>verified</b>.
+ <li><p>If <var>value</var> is null, then return <b>verified</b>.
 
  <li>
-  <p><a for="list">For each</a> <var>alg</var> → <var>digest</var> of <var>header</var>:
+  <p><a for="list">For each</a> <var>algorithm</var> → <var>digest</var> of <var>value</var>:
 
   <ol>
    <li>
-    <p>Switch on <var>alg</var>:
+    <p>Switch on <var>algorithm</var>:
 
     <dl class=switch>
      <dt>"<code>sha-256</code>"
@@ -4529,7 +4529,7 @@ represent a digest of the response produced via the specified algorithm. [[!UNEN
        <li><p>Let <var>body digest</var> be the result of executing the SHA-256 algorithm on
        <var>bytes</var>. [[!FIPS-180-4]]
 
-       <li><p>If <var>body digest</var> matches <var>digest</var>, <a for="iteration">continue</a>.
+       <li><p>If <var>body digest</var> is <var>digest</var>, <a for="iteration">continue</a>.
       </ol>
 
      <dt>"<code>sha-512</code>"
@@ -4538,7 +4538,7 @@ represent a digest of the response produced via the specified algorithm. [[!UNEN
        <li><p>Let <var>body digest</var> be the result of executing the SHA-512 algorithm on
        <var>bytes</var>. [[!FIPS-180-4]]
 
-       <li><p>If <var>body digest</var> matches <var>digest</var>, <a for="iteration">continue</a>.
+       <li><p>If <var>body digest</var> is <var>digest</var>, <a for="iteration">continue</a>.
       </ol>
 
      <dt><b>Otherwise</b>


mikewest:
Hey folks! I'd like to start upstreaming signature-based integrity checks; this seemed like the simplest self-contained place to start. Assuming folks are on board, I'll file implementation and MDN bugs, and put up more PRs against SRI and CSP.

cc @annevk @mozfreddyb

mikewest:
Friendly ping. It'd be nice to solidify things prior to TPAC. :)

mikewest:
Friendlier ping? Are there concerns with beginning to upstream this work?

fetch.bs
```diff
@@ -4483,6 +4494,72 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+assertions about the integrity of a response's content. It is a <a>structured header</a> whose value
+must be a <a data-lt="structured field dictionary">dictionary</a> whose keys specify hashing
+algorithms, and whose values are <a data-lt="structured field byte sequence">byte sequences</a>
+represent a digest of the response produced via the specified algorithm. [[!UNENCODED-DIGEST]]
+
+
+<div algorithm>
+<p>To <dfn export>verify `<code>Unencoded-Digest</code>` assertions</dfn>, given a
+<a>byte sequence</a> <var>bytes</var> and a <a for=/>header list</a> <var>list</var>, run these
+steps:
```
```suggestion
<p>To <dfn export>verify `<code>Unencoded-Digest</code>` assertions</dfn>, given a
<a>byte sequence</a> <var>bytes</var> and a <a for=/>header list</a> <var>list</var>:
```

fetch.bs
```diff
@@ -4483,6 +4494,72 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+
+<ol>
+ <li><p>Let <var>header</var> be the result of
+ <a for="header list" lt="get a structured field value">getting</a> the
+ `<a http-header><code>Unencoded-Digest</code></a>` header as a "<code>dictionary</code>" from
+ <var>list</var>.
+
+ <li><p>If <var>header</var> is null, then return <b>verified</b>.
```
annevk: I think we should restructure this to return a boolean. `<b>return value</b>` is not part of Infra and it's not necessarily that much clearer. I suppose we could return an enum instead, but we don't really have enums with only two values. (Maybe we should and maybe we should refactor everything away from booleans after mostly refactoring away from flags, but that would require a bit of investigation as to what that would look like.)

fetch.bs
```diff
@@ -4483,6 +4494,72 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+represent a digest of the response produced via the specified algorithm. [[!UNENCODED-DIGEST]]
+
+
+<div algorithm>
+<p>To <dfn export>verify `<code>Unencoded-Digest</code>` assertions</dfn>, given a
+<a>byte sequence</a> <var>bytes</var> and a <a for=/>header list</a> <var>list</var>, run these
+steps:
+
+<ol>
+ <li><p>Let <var>header</var> be the result of
```
annevk: Shouldn't this be called _value_?

fetch.bs
```diff
@@ -4483,6 +4494,72 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+ <a for="header list" lt="get a structured field value">getting</a> the
+ `<a http-header><code>Unencoded-Digest</code></a>` header as a "<code>dictionary</code>" from
+ <var>list</var>.
+
+ <li><p>If <var>header</var> is null, then return <b>verified</b>.
+
+ <li>
+  <p><a for="list">For each</a> <var>alg</var> → <var>digest</var> of <var>header</var>:
```
annevk: algorithm*

fetch.bs
```diff
@@ -4483,6 +4494,72 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+    <p>Switch on <var>alg</var>:
+
+    <dl class=switch>
+     <dt>"<code>sha-256</code>"
+     <dd>
+      <ol>
+       <li><p>Let <var>body digest</var> be the result of executing the SHA-256 algorithm on
+       <var>bytes</var>. [[!FIPS-180-4]]
+
+       <li><p>If <var>body digest</var> matches <var>digest</var>, <a for="iteration">continue</a>.
```
annevk: bodyDigest

What does "matches" mean here? Can we say "is"?

Also, perhaps we should pass algorithm and bytes to SRI so it can have the dependencies on SHA? Perhaps this entire algorithm can be in SRI and we only have the processing model hooks in fetch? Hmm.

fetch.bs
```diff
@@ -5022,7 +5099,9 @@ steps:
   <p class=note>This standardizes the error handling for servers that violate HTTP.
 
  <li>
-  <p>If <var>request</var>'s <a for=request>integrity metadata</a> is not the empty string, then:
+  <p>If <var>request</var>'s <a for=request>integrity metadata</a> is not the empty string, or if
+  <var>internalResponse</var>'s <a for="response">header list</a> <a for="header list">contains</a>
+  `<a http-header><code>Unencoded-Digest</code></a>`, then:
```
annevk, line 5103: This means we enforce it for opaque responses. Is that what we want? What about requests that didn't care about integrity enforcement?

fetch.bs
mikewest: Booleans require reading the algorithm name carefully, so I find it clearer to name the output in some way. We do this already in at least one place in Fetch (https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check), and I'd prefer that we do that more often.

I could put up a PR against Infra that attempted to generalize this pattern (define some set of values (**Success** and **Failure**, **Allowed** and **Blocked**?) that can be returned from algorithms, and use them in the algorithm linked above?) if you'd be interested in that. We might find more values over time, but the things I can easily remember from CSP and SRI would fall into one of those categories...

fetch.bs
mikewest: "is" is accurate.

I thought about folding this into SRI as well. That might make more sense. My feeling is that this header is more closely tied to the fetching mechanism, as it's a server's declaration of the intended response body's content, but I suppose that's true for the signature mechanism as well that would come in a subsequent PR, and that probably doesn't fit into Fetch terribly cleanly. Hrm.

@mozfreddyb might have an opinion here?

fetch.bs
mikewest: The way this is currently defined would apply the check to responses regardless of a page's integrity assertions. It's based on a server's assertions, not on integrity metadata associated with the request. That layering makes the SRI checks quite straightforward (similar conceptually to CSP's layering on top of SRI), as the algorithms can assume that the `unencoded-digest` header will be enforced, so they only need to concern themselves with signature verification.

This layering also gives servers the same ability as clients to make binding assertions about responses. Philosophically, I find that appealing.

fetch.bs
annevk: That would improve it, though I worry what happens when we suddenly need more than two values. I guess we switch to "success", "failure", "report", or some such at that point?

mikewest: No reason we couldn't define multiple common patterns. The two I noted above (allow/block from https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check, and failure from all over URL and a number of HTML algorithms) are ones I could easily find examples for. The latter in particular seems like it would benefit from putting something in Infra.

Filed https://github.com/whatwg/infra/issues/686 to think about it some more.

fetch.bs
```diff
@@ -4483,6 +4494,71 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+
+  <ol>
+   <li>
+    <p>Switch on <var>algorithm</var>:
+
+    <dl class=switch>
+     <dt>"<code>sha-256</code>"
+     <dd>
+      <ol>
+       <li><p>Let <var>body digest</var> be the result of executing the SHA-256 algorithm on
```
annevk, line 4529: bodyDigest*

fetch.bs
```diff
@@ -4483,6 +4494,71 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+     <dt>"<code>sha-512</code>"
+     <dd>
+      <ol>
+       <li><p>Let <var>body digest</var> be the result of executing the SHA-512 algorithm on
+       <var>bytes</var>. [[!FIPS-180-4]]
+
+       <li><p>If <var>body digest</var> is <var>digest</var>, <a for="iteration">continue</a>.
+      </ol>
+
+     <dt><b>Otherwise</b>
```
annevk, line 4544: No need for `<b>`

fetch.bs
```diff
@@ -4483,6 +4494,71 @@ indicates the request’s purpose is to fetch a resource that is anticipated to
...
+  </ol>
+
+ <li><p>Return <b>verified</b>.
+</ol>
+
+<p class="note">This algorithm requires all valid digests delivered via
+`<a http-header><code>Unencoded-Digest</code></a>` to match the response’s decoded body, while
+ignoring unknown algorithms. Since the server controls both the body and the headers, it seems
+unnecessary to allow the flexibility of allowing the asserted digests to match more than one
+resource (as we do in client-initiated checks via [[SRI]], which need to support servers'
```
annevk, line 4558: References go at the end of a paragraph. If you want something inline you have to use `<cite>` and such.

fetch.bs
```diff
@@ -5022,7 +5098,9 @@ steps:
   <p class=note>This standardizes the error handling for servers that violate HTTP.
 
  <li>
-  <p>If <var>request</var>'s <a for=request>integrity metadata</a> is not the empty string, then:
+  <p>If <var>request</var>'s <a for=request>integrity metadata</a> is not the empty string, or if
```
annevk, line 5101:
```suggestion
  <p>If <var>request</var>'s <a for=request>integrity metadata</a> is not the empty string or
```

fetch.bs
```diff
@@ -5022,7 +5098,9 @@ steps:
...
+  <var>internalResponse</var>'s <a for="response">header list</a> <a for="header list">contains</a>
+  `<a http-header><code>Unencoded-Digest</code></a>`, then:
```
annevk, line 5103:
```suggestion
  `<a http-header><code>Unencoded-Digest</code></a>`:
```

fetch.bs
annevk: But we don't want SRI to work for opaque responses so there will be some mismatches in the overall model. Is the idea that SRI also ends up poking at the header list?

I'm not a big fan of us poking at the internal header list as it means this is yet another header that cannot be hidden from web content processes.

mozfreddyb:
I noticed that `Unencoded-Digest` is also supported in trailers, as per the draft. That doesn't seem like it would align with the proposed integration here?

PR #1876, opened 2025-11-09
Spec 'text directive user activation' boolean

domfarolino:
This PR complements https://github.com/whatwg/html/pull/11895, in service of standardizing https://wicg.github.io/scroll-to-text-fragment/.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1876.html" title="Last updated on Dec 30, 2025, 5:44 PM UTC (5a07747)">Preview</a> | <a href="https://whatpr.org/fetch/1876/0e72db8...5a07747.html" title="Last updated on Dec 30, 2025, 5:44 PM UTC (5a07747)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2234,6 +2234,10 @@ Unless stated otherwise, it is unset.
 <p>A <a for=/>request</a> has an associated boolean <dfn export for=request>user-activation</dfn>.
 Unless stated otherwise, it is false.
 
+<p>A <a for=/>request</a> has an associated boolean
+<dfn export for=request>text directive user activation</dfn>.
+Unless stated otherwise, it is false.
+
 <p class=note>This is for exclusive use by HTML's navigate algorithm. [[!HTML]]
 
 <p>A <a for=/>request</a> has an associated boolean <dfn export for=request>render-blocking</dfn>.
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -2277,6 +2277,10 @@ otherwise, it is unset.
 <a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
+<p>A <a for=/>request</a> has an associated boolean
+<dfn export for=request id=text-directive-user-activation>text directive user activation</dfn>.
```
annevk: No need for an explicit ID here. For new `<dfn>` elements we let Bikeshed generate them (it will be different so you'll need to adjust the HTML side).

fetch.bs
```diff
@@ -2277,6 +2277,10 @@ otherwise, it is unset.
 <a for=request>done flag</a>, and <a for=request>timing allow failed flag</a> are used as
 bookkeeping details by the <a for=/>fetch</a> algorithm.
 
+<p>A <a for=/>request</a> has an associated boolean
+<dfn export for=request>text directive user activation</dfn>.
+Unless stated otherwise, it is false.
```
annevk: We should group this with "user-activation". I'm not entirely sure anymore why that has a hyphen in it either. Hopefully @sideshowbarker remembers.

And do we really need both? Seems unfortunate.

sideshowbarker: I don't remember, but regardless I would imagine "user-activation" should only have a hyphen in cases where it's used as an adjective. If it's used as a noun, we shouldn't need any hyphen.

domfarolino: > And do we really need both? Seems unfortunate.

Good question. I don't know off the top of my head, I will investigate.

domfarolino: I think https://github.com/WICG/scroll-to-text-fragment/blob/main/redirects.md spells out why we cannot use user activation for this. It's because user activation does not span across client-side redirects, whereas text directive user activation must. The doc describes it more.

PR #1914, opened 2026-03-13
Fix Bikeshed API endpoint in Makefile

yoavweiss:
The `api.csswg.org/bikeshed/` endpoint is no longer reliable. Migrate to `www.w3.org/publications/spec-generator/` with the required `type=bikeshed-spec` and `output=html` parameters.

See https://github.com/w3c/spec-prod/issues/227#issuecomment-4049021313

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,15 @@ SHELL=/bin/bash -o pipefail
 .PHONY: local remote deploy
 
 remote: fetch.bs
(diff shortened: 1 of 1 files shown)


yoavweiss:
https://github.com/whatwg/spec-factory/pull/83 fixes this at the source

annevk:
#1915

PR #1922, opened 2026-04-14
Define multipart/form-data

annevk:
Based on Andreu's work at https://github.com/andreubotella/multipart-form-data with mostly editorial changes and some corrections.

WPT coverage at fetch/api/response/response-form-data.html

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/59216
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1922.html" title="Last updated on Sep 12, 2026, 8:58 AM UTC (2bab7b5)">Preview</a> | <a href="https://whatpr.org/fetch/1922/394d20d...2bab7b5.html" title="Last updated on Sep 12, 2026, 8:58 AM UTC (2bab7b5)">Diff</a>


fetch.bs
```diff
@@ -9802,6 +9779,542 @@ that RFC's normative processing requirements to be compatible with deployed cont
...
+
+ <li><p>it is composed by bytes in the ranges 0x30 to 0x39, 0x41 to 0x5A, or 0x61 to 0x7A,
+ inclusive (<a>ASCII alphanumeric</a>), or which are 0x27 ('), 0x2D (-) or 0x5F (_).
+</ul>
+
+<div algorithm>
+<p>To <dfn export for="multipart/form-data/boundary">generate</dfn> a
+<a for="multipart/form-data" lt="boundary"><code>multipart/form-data</code> boundary</a>, return an
+<a>implementation-defined</a> <a for=/>byte sequence</a> which fulfills the conditions for
+boundaries, such that part of it is randomly generated, with a minimum entropy of 95 bits.
```
andreubotella, line 9802: This was based on my analysis which IIRC concluded that Chromium/WebKit's boundary had exactly 95 bits of entropy, and that Gecko's was probably close to but less than 96 bits. But I see that at least Gecko has changed its boundary string generation now, and I'm not sure where it stands.

Do we want to spec something here that aligns with current implementations, or do we want to change this to some round number that makes sense (e.g. 32 * 3 = 96 bits would be easy to implement), and make implementations match?

fetch.bs
```diff
@@ -9802,6 +9779,542 @@ that RFC's normative processing requirements to be compatible with deployed cont
...
+   <var>position</var>. (Do nothing with those bytes.)
+
+   <li>
+    <p>If <var>position</var> points to a sequence of bytes starting with 0x2D 0x2D
+    (`<code>--</code>`):
+
+    <ol>
+     <li><p>If <var>position</var> + 2 points to the end of <var>input</var>, or
+     <var>position</var> + 2 points to a sequence of bytes starting with 0x0D 0x0A (CR LF), then
+     return <var>entryList</var>.
```
youennf, line 10114: According RFC, there could be some spaces between `--` and CRLF, not sure what UAs are doing though.
Also, I am not sure what happens with UAs if there are non whitespace characters after `--`. Maybe entryList is returned instead of a failure.

fetch.bs
annevk: I think for now matching implementations makes sense (Gecko indeed seems to have a far higher number) and then separately pursue an improvement on top.

fetch.bs
annevk: Thanks, this uncovered some more issues.

PR #1930, opened 2026-05-30
Update :protocol to webtransport-h3 for WebTransport over HTTP/3

jan-ivar:
[WebTransport over HTTP/3 draft-15](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-15#name-creating-a-new-session:~:text=be%20set%20to-,webtransport%2Dh3,-.) renamed the extended-CONNECT `:protocol` pseudo-header value from `webtransport` to `webtransport-h3`. [WebTransport over HTTP/2](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http2-14#name-creating-a-new-session:~:text=to-,webtransport) keeps `webtransport`, making it depend on _connection_.

   * Firefox: implementing draft-15 in [bug 2033974](https://bugzil.la/2033974)
   * Chromium: …
   * WebKit: …
- [ ] Tested with web-platform-tests: WebTransport WPT exercises the H3 handshake; the draft-15 `:protocol` value is covered by the server update in [bug 2033974](https://bugzil.la/2033974).
- [ ] MDN issue / update: N/A (not author-observable).
- [x] The top of this comment includes a clear commit message.
   * …
   * …

***
<a href="https://whatpr.org/fetch/1930.html" title="Last updated on Jun 26, 2026, 12:18 PM UTC (eb06c09)">Preview</a> | <a href="https://whatpr.org/fetch/1930/301650c...eb06c09.html" title="Last updated on Jun 26, 2026, 12:18 PM UTC (eb06c09)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -126,6 +126,12 @@ urlPrefix:https://www.rfc-editor.org/rfc/rfc6454;type:dfn;spec:RFC6454
         "publisher": "IETF",
         "title": "WebTransport over HTTP/3"
     },
+    "WEBTRANSPORT-HTTP2": {
+        "authors": ["E. Kinnear"],
+        "href": "https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http2",
+        "publisher": "IETF",
+        "title": "WebTransport over HTTP/2"
+    },
     "SVCB": {
         "aliasOf": "RFC9460"
     }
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -6656,6 +6656,13 @@ optional boolean <var>forceNewConnection</var> (default false), run these steps:
    to the <a for=/>coarsened shared current time</a> given <var>fetchParams</var>'s
    <a for="fetch params">cross-origin isolated capability</a>.
 
+   <li>
+    <p>If <var>request</var>'s <a for=request>mode</a> is "<code>webtransport</code>",
+    <var>connection</var> is an HTTP/3 connection, and the "<code>:protocol</code>"
+    pseudo-header associated with <var>request</var>'s <a for=request>method</a> is
+    "<code>webtransport</code>", then set the "<code>:protocol</code>" pseudo-header to
```
annevk: Why is the pseudo-header associated with request's method? I think you just mean associated with request? What other values besides "webtransport" can it have at this point and given the H3 precondition?

jan-ivar: It is set to "`webtransport`" [in the WT constructor](https://w3c.github.io/webtransport/#pseudo-header) so that is the only possible value rn and is valid for H2. But H3 is not a precondition at that point, hence the need for code here. I'll add a note to clarify this.

annevk: I think if we reach this point and the value has to be "webtransport" because of the other values, we should assert it instead. This will require some nested steps, but that seems okay.

jan-ivar: You're right I used conditionals to clarify preconditions here since they're in a different spec (which might someday change).

If we're OK with cross-spec asserts I can change it to use those (for the method and pseudo header?)

annevk (approved):
@jan-ivar sorry for the delay. Could you confirm you are okay with my nits?

One thing I somewhat dislike is that if we ever add HTTP/4 it'll go back to "`webtransport`", but not much that can be done about that. Hopefully we'll stick with 3 versions for the foreseeable future and eventually get the equivalent of C++ enum classes in standards.

annevk:
@jan-ivar heya, did you see my reply above?

PR #1931, opened 2026-06-05
Add TAO destination check for navigation redirect chains

yoavweiss:
As part of the [WebPerfWG discussion](https://w3c.github.io/web-performance/meetings/2026/2026-06-04/index.html) on [re-exposing redirectCount for opted-in cross-origin redirects](https://github.com/whatwg/html/pull/12513), we discussed the fact that TAO currently only opts-in "backwards" in terms of origins, and doesn't constitute an opt-in to the destination origin, which is the appropriate opt-in for navigations (contrary to subresources).

This PR adds the mechanisms that can enable such "forward" opt-in, and can enable a response to a redirect chain to know that the redirect chain opted in to expose the redirects to its origin.

Once this lands, we'd use the relevant algorithms in the related HTML PR.

   * Chromium: https://github.com/whatwg/fetch/pull/1931#issuecomment-4660492429
   * WebKit: https://github.com/whatwg/fetch/pull/1931#pullrequestreview-4560631006
   * https://github.com/web-platform-tests/wpt/pull/60823
   * Chromium: https://issues.chromium.org/issues/521861828
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2049166
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=316647
   * Deno (not for CORS changes): N/A as it doesn't do navigations.
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/841

***
<a href="https://whatpr.org/fetch/1931.html" title="Last updated on Jun 24, 2026, 9:51 AM UTC (48b86d4)">Preview</a> | <a href="https://whatpr.org/fetch/1931/301650c...48b86d4.html" title="Last updated on Jun 24, 2026, 9:51 AM UTC (48b86d4)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -2308,8 +2308,8 @@ Unless stated otherwise, it is unset.
 otherwise, it is unset.
 
 <p>A <a for=/>request</a> has an associated
-<dfn for=request>navigation timing allow check set</dfn> (a <a for=/>list</a> of zero or
-more <a for=/>strings</a>). Unless stated otherwise, it is « "<code>*</code>" ».
+<dfn for=request>navigation timing allow check set</dfn> (an <a for=/>ordered set</a> of zero
+or more <a for=/>strings</a>). Unless stated otherwise, it is « "<code>*</code>" ».
 
 <p class=note>A <a for=/>request</a>'s <a for=request>URL list</a>, <a for=request>current URL</a>,
 <a for=request>redirect count</a>, <a for=request>response tainting</a>,
@@ -2614,8 +2614,8 @@ the response of a redirect has to be set if it was set for previous responses in
 this is also tracked internally using the request's <a for=request>timing allow failed flag</a>.
 
 <p>A <a for=/>response</a> has an associated
-<dfn export for=response>navigation timing allow check set</dfn> (a <a for=/>list</a> of zero or
-more <a for=/>strings</a>). Unless stated otherwise, it is « "<code>*</code>" ».
+<dfn export for=response>navigation timing allow check set</dfn> (an <a for=/>ordered set</a> of
+zero or more <a for=/>strings</a>). Unless stated otherwise, it is « "<code>*</code>" ».
 
 <p class=note>This is used by <a for=/>responses</a> to <a>navigation requests</a> so that
 callers can determine, once the navigation's destination origin is known, whether that origin is
@@ -5119,7 +5119,7 @@ steps:
 
  <li><p>If <var>request</var> is a <a>navigation request</a>, then set
  <var>internalResponse</var>'s <a for=response>navigation timing allow check set</a> to a
- <a for=list>clone</a> of <var>request</var>'s
+ <a for=set>clone</a> of <var>request</var>'s
  <a for=request>navigation timing allow check set</a>.
 
  <li><p>If <var>request</var>'s <a for=request>timing allow failed flag</a> is unset, then set
@@ -5903,7 +5903,7 @@ these steps:
 
   <ol>
    <li><p>If <var>request</var> is a <a>navigation request</a>, then
-   <a>update a request's navigation timing allow check set</a> given <var>request</var> and
+   <a>restrict a request's navigation timing allow check set</a> given <var>request</var> and
    <var>internalResponse</var>.
 
    <li>
@@ -7334,7 +7334,7 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
 </div>
 
 <div algorithm>
-<p>To <dfn>update a request's navigation timing allow check set</dfn>, given a
+<p>To <dfn>restrict a request's navigation timing allow check set</dfn>, given a
 <a for=/>request</a> <var>request</var> and a <a for=/>response</a> <var>response</var>, run these
 steps:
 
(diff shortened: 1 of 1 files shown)


noamr:
Chrome team is supportive.

noamr (commented):
Wouldn't it be simpler to keep a list of TAO value sets and iterate over them when computing the "navigation TAO check"?

yoavweiss:
> Wouldn't it be simpler to keep a list of TAO value sets and iterate over them when computing the "navigation TAO check"?

@annevk made a similar comment earlier on.

What we can do is something like:
* We need the end response of the chain to have a list of lists of all the TAO values each response seen. We can get that by adding the a list of lists of past values to the redirect request and then move it to the response. 
* Navigation TAO check would verify that the TAO values of each response contain "*" or destination origin, otherwise it'd fail.

I don't know that it's simpler (and it seems like we'd be copying more values around between redirect responses), but happy to move to that model if that's simpler/easier

noamr:
> > Wouldn't it be simpler to keep a list of TAO value sets and iterate over them when computing the "navigation TAO check"?
> 
> @annevk made a similar comment earlier on.
> 
> What we can do is something like:
> 
> * We need the end response of the chain to have a list of lists of all the TAO values each response seen. We can get that by adding the a list of lists of past values to the redirect request and then move it to the response.
> * Navigation TAO check would verify that the TAO values of each response contain "*" or destination origin, otherwise it'd fail.
> 
> I don't know that it's simpler (and it seems like we'd be copying more values around between redirect responses), but happy to move to that model if that's simpler/easier

It's simpler to read  (IMO) and potentially cleaner impl wise as there is no set manipulation as you go along and it's not a dedup check for each redirect.

noamr (approved):
LGTM (though needs the checkboxes etc)

yoavweiss:
@annevk - Can you take a look? Do I need to file a WebKit position on this, or will you be able to state support here?

fetch.bs
```diff
@@ -7311,6 +7338,49 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
 </ol>
 </div>
 
+<div algorithm>
+<p>To <dfn>append to a request's navigation timing allow check list</dfn>, given a
+<a for=/>request</a> <var>request</var> and a <a for=/>response</a> <var>response</var>, run these
+steps:
```
```suggestion
<a for=/>request</a> <var>request</var> and a <a for=/>response</a> <var>response</var>:
```

fetch.bs
```diff
@@ -7311,6 +7338,49 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+
+ <li><p><a for=list>Append</a> <var>taoValues</var> to <var>request</var>'s
+ <a for=request>navigation timing allow check list</a>.
+</ol>
+</div>
+
+<div algorithm>
+<p>To perform a <dfn export id=concept-navigation-tao-check>navigation TAO check</dfn> for a
+<a for=/>response</a> <var>response</var> and an <a for=/>origin</a>
+<var>destinationOrigin</var>, run these steps:
```
```suggestion
<var>destinationOrigin</var>:
```

fetch.bs
```diff
@@ -2307,10 +2307,19 @@ Unless stated otherwise, it is unset.
 <dfn export for=request id=timing-allow-failed>timing allow failed flag</dfn>. Unless stated
 otherwise, it is unset.
 
+<p>A <a for=/>request</a> has an associated
+<dfn for=request>navigation timing allow check list</dfn> (a <a for=/>list</a> of
```
annevk: instead of "check list" maybe "values list" would be nicer?

fetch.bs
```diff
@@ -7311,6 +7338,49 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+  <a for=response>navigation timing allow check list</a>:
+
+  <ol>
+   <li><p>If <var>taoValues</var> <a for=list>contains</a> "<code>*</code>", then
+   <a for=iteration>continue</a>.
+
+   <li><p>If <var>taoValues</var> <a for=list>contains</a> <var>destinationOrigin</var>,
+   <a lt="ASCII serialization of an origin">serialized</a>, then <a for=iteration>continue</a>.
+
+   <li><p>Return failure.
```
```suggestion
   <li><p>Return <a for=/>failure</a>.
```

fetch.bs
```diff
@@ -7311,6 +7338,49 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+  </ol>
+
+ <li><p>Return success.
```
```suggestion
 <li><p>Return <a for=/>success</a>.
```

fetch.bs
```diff
@@ -7311,6 +7338,49 @@ agent's <a>CORS-preflight cache</a> for which there is a <a>cache entry match</a
...
+
+ <li><p>If <var>taoValues</var> is null, then set <var>taoValues</var> to « ».
+
+ <li><p><a for=list>Append</a> <var>taoValues</var> to <var>request</var>'s
+ <a for=request>navigation timing allow check list</a>.
+</ol>
+</div>
+
+<div algorithm>
+<p>To perform a <dfn export id=concept-navigation-tao-check>navigation TAO check</dfn> for a
```
annevk: Let's not add new novel IDs when we don't have to.

annevk (approved):
This design seemed reasonable to @achristensen07 and I so you can consider WebKit supportive.

fetch.bs
yoavweiss: done!

fetch.bs
yoavweiss: done

fetch.bs
yoavweiss: done

fetch.bs
yoavweiss: Done

fetch.bs
yoavweiss: done

fetch.bs
yoavweiss: removed

fetch.bs
yoavweiss: This required adding a link disambiguation, to point at infra, rather than WebDriver.

Bas-moz:
This looks good from the Mozilla side as well.

PR #1937, opened 2026-06-23
Hide range header values on no-cors requests

### LINKED ISSUE #1936: Hide range values from no-cors cross-origin range requests
https://github.com/whatwg/fetch/pull/560 allowed range requests from media elements to pass through the service worker. However, you can see the range header in the service worker, and this can be used to learn the byte size of the media resource, even if it's cross-origin, by forcing a request for the final chunk of the media.

Media is pretty leaky already, as it'll give you the duration of the resource, but it seems like we shouldn't make the byte length observable.

We could do one of the following:

- Hide the range header value on these requests.
- Lie that the range header value is `0-`.

The advantage of the latter is that code that detects whether the request is a range request will still work… maybe.

The rest of the rules we have in the spec, like removing the range header if the request is modified, will remain.

jakearchibald:
Fixed #1936

The rest of the form is TODO.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1937.html" title="Last updated on Jun 24, 2026, 8:39 AM UTC (20d2ac7)">Preview</a> | <a href="https://whatpr.org/fetch/1937/45097d9...20d2ac7.html" title="Last updated on Jun 24, 2026, 8:39 AM UTC (20d2ac7)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -622,9 +622,17 @@ efficient representation, as long as they also support an associated data struct
 <dfn export for="header list" id=concept-header-list-hidden-range-value>hidden range value</dfn>
 (null or a <a for=/>header value</a>). It is initially null.
 
-<p class=note>This is for cases where a "<code>no-cors</code>" <a for=/>request</a> needs to include a
-`<code>Range</code>` header, but it can not be exposed via {{Headers}}. See <a for=request>add a range
-header</a>.
+<div class=note>
+ <p>This is for cases where a "<code>no-cors</code>" <a for=/>request</a> needs to include a
+ `<code>Range</code>` header, but it can not be exposed via {{Headers}}. It can be set by privileged
+ APIs, and will be preserved if its associated request object is copied, but will be removed if the
+ request is modified by unprivileged APIs.
+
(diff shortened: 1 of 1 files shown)


fetch.bs
```diff
@@ -8032,8 +8056,7 @@ method steps are to <a for=Headers>append</a> (<var>name</var>, <var>value</var>
   <p class=note>Passing a dummy <a>header value</a> ought not to have any negative repercussions.
 
  <li><p>If <a>this</a>'s <a for=Headers>guard</a> is "<code>request-no-cors</code>", <var>name</var>
- is not a <a>no-CORS-safelisted request-header name</a>, and <var>name</var> is not a
- <a>privileged no-CORS request-header name</a>, then return.
+ is not a <a>no-CORS-safelisted request-header name</a>, then return.
```
jakearchibald, line 8059: This is a behaviour/capability change. Previously you could remove the range from a range request. Now you cannot.

I could make this work, but it seems weird that you can remove a header that doesn't appear to be there (would we fake `has` too?).

You can remove the range header by modifying some other header - even if you set it back to its original value.

Fwiw this wouldn't be an issue if we returned a fake value. Although that may have other issues.

PR #1943, opened 2026-07-21
Add getStructured/setStructured to Headers

jasnell:
Adds new APIs to the Headers class for getting/setting structured header fields.

Structured fields are defined in RFC 8941. Newer HTTP header definitions build on it. Fetch's handling of all header values as strings works but loses some of the utility. This commit adds new `getStructured/setStructured` APIS to `Headers` for getting/setting header field values as structured fields. The existing `get`/`set`/`append`/etc are left untouched. Header iteration is left untouched. It remains possible to get all fields as strings.

There are currently ~36 standard headers that use structured header fields:

#### Dictionary

| Header              | Spec                                                               | Description                                                        |
| ------------------- | ------------------------------------------------------------------ | ------------------------------------------------------------------ |
| `Priority`            | [RFC 9218](https://www.rfc-editor.org/rfc/rfc9218)                  | HTTP response prioritization (urgency, incremental delivery)       |
| `Signature`           | [RFC 9421](https://www.rfc-editor.org/rfc/rfc9421)                  | HTTP message signatures                                            |
| `Signature-Input`     | [RFC 9421](https://www.rfc-editor.org/rfc/rfc9421)                  | Metadata for message signatures (covered components, key ID, etc.) |
| `Accept-Signature`    | [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1) | Requests that recipient apply a signature                          |
| `Content-Digest`      | [RFC 9530](https://www.rfc-editor.org/rfc/rfc9530)                  | Integrity digest over HTTP message content                         |
| `Repr-Digest`         | [RFC 9530](https://www.rfc-editor.org/rfc/rfc9530)                  | Integrity digest over HTTP representation                          |
| `Want-Content-Digest` | [RFC 9530](https://www.rfc-editor.org/rfc/rfc9530)                  | Requests `Content-Digest` with algorithm preferences                 |
| `Want-Repr-Digest`    | [RFC 9530](https://www.rfc-editor.org/rfc/rfc9530)                  | Requests `Repr-Digest` with algorithm preferences                    |
| `CDN-Cache-Control`   | [RFC 9213](https://www.rfc-editor.org/rfc/rfc9213)                  | Targeted cache directives for CDN caches                           |
| `Use-As-Dictionary`   | [RFC 9842](https://www.rfc-editor.org/rfc/rfc9842)                  | Marks a response as a compression dictionary                       |

#### List

| Header                   | Spec                                                | Description                                         |
| ------------------------ | --------------------------------------------------- | --------------------------------------------------- |
| `Cache-Status`             | [RFC 9211](https://www.rfc-editor.org/rfc/rfc9211)   | Per-cache handling report (`hit`, `fwd`, `ttl`, etc.)     |
| `Proxy-Status`             | [RFC 9209](https://www.rfc-editor.org/rfc/rfc9209)   | Per-intermediary handling report with error details |
| `Accept-CH`                | [RFC 8942](https://www.rfc-editor.org/rfc/rfc8942)   | Advertises server support for Client Hints          |
| `Client-Cert-Chain`        | [RFC 9440](https://www.rfc-editor.org/rfc/rfc9440)   | Client certificate chain from TLS-terminating proxy |
| `Accept-Query`             | [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008) | Accepted media types for HTTP QUERY body            |
| `Cache-Groups`             | [RFC 9875](https://www.rfc-editor.org/rfc/rfc9875)   | Associates cached responses with named groups       |
| `Cache-Group-Invalidation` | [RFC 9875](https://www.rfc-editor.org/rfc/rfc9875)   | Invalidates all responses in named cache groups     |

#### Item

| Header                                   | Spec                                                                                              | Description                                              |
| ---------------------------------------- | ------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
| `Client-Cert`                              | [RFC 9440](https://www.rfc-editor.org/rfc/rfc9440)                                                 | End-entity client certificate (Byte Sequence)            |
| `Capsule-Protocol`                         | [RFC 9297](https://www.rfc-editor.org/rfc/rfc9297)                                                 | Enables the Capsule Protocol on an HTTP stream (Boolean) |
| `Deprecation`                              | [RFC 9745](https://www.rfc-editor.org/rfc/rfc9745)                                                 | Signals resource deprecation (Date)                      |
| `Available-Dictionary`                     | [RFC 9842](https://www.rfc-editor.org/rfc/rfc9842)                                                 | Client has a compression dictionary available            |
| `Dictionary-ID`                            | [RFC 9842](https://www.rfc-editor.org/rfc/rfc9842)                                                 | Assigns a stable ID to a compression dictionary response |
| `Concealed-Auth-Export`                    | [RFC 9729](https://www.rfc-editor.org/rfc/rfc9729)                                                 | Exported keying material for concealed HTTP auth         |
| `Cross-Origin-Embedder-Policy`             | [HTML Standard](https://html.spec.whatwg.org/multipage/origin.html#coep)                           | Controls cross-origin resource loading policy            |
| `Cross-Origin-Embedder-Policy-Report-Only` | [HTML Standard](https://html.spec.whatwg.org/multipage/origin.html#coep)                           | COEP in report-only mode                                 |
| `Cross-Origin-Opener-Policy`               | [HTML Standard](https://html.spec.whatwg.org/multipage/browsers.html#cross-origin-opener-policies) | Controls browsing context group sharing                  |
| `Cross-Origin-Opener-Policy-Report-Only`   | [HTML Standard](https://html.spec.whatwg.org/multipage/browsers.html#cross-origin-opener-policies) | COOP in report-only mode                                 |
| `Origin-Agent-Cluster`                     | [HTML Standard](https://html.spec.whatwg.org/multipage/origin.html#origin-agent-cluster)           | Requests origin-keyed agent cluster (Boolean)            |
| `Sec-Fetch-Dest`                           | [Fetch Metadata](https://w3c.github.io/webappsec-fetch-metadata/)                                  | Request destination type (Token)                         |
| `Sec-Fetch-Mode`                           | [Fetch Metadata](https://w3c.github.io/webappsec-fetch-metadata/)                                  | Request mode (Token)                                     |
| `Sec-Fetch-Site`                           | [Fetch Metadata](https://w3c.github.io/webappsec-fetch-metadata/)                                  | Request-vs-target origin relationship (Token)            |
| `Sec-Fetch-User`                           | [Fetch Metadata](https://w3c.github.io/webappsec-fetch-metadata/)                                  | User activation (Boolean)                                |
| `Sec-Purpose`                              | [Fetch Standard](https://fetch.spec.whatwg.org/)                                                   | Request purpose, e.g. `prefetch` (Token)                   |

#### Reading the `Priority` header (Dictionary)

```js
// Priority: u=0, i
const p = response.headers.getStructured("Priority", "dictionary");
const urgency = p?.get("u")?.value ?? 3;     // 0
const incremental = p?.get("i")?.value ?? false; // true
```

Compared with strings:

```js
const raw = response.headers.get("Priority"); // "u=0, i"
// ... now what? Split on comma? Parse key=value? Handle quoting?
// Every app rolls its own parser and gets edge cases wrong.
```

#### Reading Cache-Status (List)

```js
// Cache-Status: ReverseProxy;hit, CDN;fwd=miss;stored;ttl=3600
const cs = response.headers.getStructured("Cache-Status", "list");
for (const entry of cs) {
  console.log(entry.value);                    // "ReverseProxy", "CDN"
  console.log(entry.params.get("hit"));        // true, undefined
  console.log(entry.params.get("fwd"));        // undefined, "miss"
  console.log(entry.params.get("ttl"));        // undefined, 3600
}
```

#### Reading Content-Digest (Dictionary with Byte Sequences)

```js
// Content-Digest: sha-256=:X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=:
const digest = response.headers.getStructured("Content-Digest", "dictionary");
const hash = digest?.get("sha-256")?.value;    // Uint8Array
```

#### Reading Sec-Purpose (Item)

```js
// Sec-Purpose: prefetch
const purpose = request.headers.getStructured("Sec-Purpose", "item");
if (purpose?.value === "prefetch") {
  // serve a lighter response
}
```

#### Writing Priority (Dictionary — plain object form)

```js
// Sets: Priority: u=0, i
request.headers.setStructured("Priority", "dictionary", {
  u: { value: 0 },
  i: { value: true }
});
```

#### Writing Cache-Status (List with parameters)

```js
// Sets: Cache-Status: MyProxy;hit;ttl=7200
response.headers.setStructured("Cache-Status", "list", [
  { value: "MyProxy", params: { hit: true, ttl: 7200 } }
]);
```

#### Writing Content-Digest (Dictionary with Byte Sequence)

```js
const body = await response.arrayBuffer();
const hash = new Uint8Array(await crypto.subtle.digest("SHA-256", body));

// Sets: Content-Digest: sha-256=:base64encodedhash=:
response.headers.setStructured("Content-Digest", "dictionary", {
  "sha-256": { value: hash }
});
```

#### Graceful fallback when unsupported

```js
// getStructured returns null if the header is absent, malformed,
// or if the implementation doesn't support structured field parsing.
const priority = request.headers.getStructured("Priority", "dictionary");
const urgency = priority?.get("u")?.value ?? 3;  // always works, defaults to 3
```

#### Token vs String serialization

```js
// Strings matching token syntax serialize as unquoted tokens:
headers.setStructured("Example", "item", { value: "foo" });
// Sets: Example: foo

// Strings that don't match token syntax serialize as quoted strings:
headers.setStructured("Example", "item", { value: "hello world" });
// Sets: Example: "hello world"
```

#### All input forms for dictionaries and parameters (HeadersInit pattern)

```js
// Plain object (most ergonomic)
headers.setStructured("Priority", "dictionary", {
  u: { value: 3 },
  i: { value: true }
});

// Map (preserves insertion order explicitly)
headers.setStructured("Priority", "dictionary", new Map([
  ["u", { value: 3 }],
  ["i", { value: true }]
]));

// Sequence of pairs
headers.setStructured("Priority", "dictionary", [
  ["u", { value: 3 }],
  ["i", { value: true }]
]);

// Parameters accept the same forms:
headers.setStructured("Cache-Status", "list", [
  { value: "cdn", params: { hit: true, ttl: 3600 } },       // plain object
  { value: "origin", params: new Map([["fwd", "miss"]]) },   // Map
  { value: "edge", params: [["stored", true]] }              // sequence
]);
```

---

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fetch/1943.html" title="Last updated on Aug 7, 2026, 6:02 PM UTC (0b6bae3)">Preview</a> | <a href="https://whatpr.org/fetch/1943/586cd2a...0b6bae3.html" title="Last updated on Aug 7, 2026, 6:02 PM UTC (0b6bae3)">Diff</a>

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -8057,14 +8057,15 @@ if (purpose?.value === "prefetch") { /* ... */ }  // direct string comparison
   whether the number is an integer), booleans serialize as structured field
   booleans, {{Uint8Array}} serializes as a byte sequence, and strings
   serialize as tokens if they conform to <a>structured field token</a> syntax
-  or as quoted structured field strings otherwise. Item objects can include an
-  <!--grammar-check-override--><code>sfType</code> property to override the auto-detected structured field
-  type &mdash; for example, <code>{ value: "foo", sfType: "string" }</code>
-  forces a quoted string even when the value matches token syntax. Supported
-  <code>sfType</code> values are "<code>boolean</code>",
-  "<code>integer</code>", "<code>decimal</code>", "<code>date</code>",
-  "<code>string</code>", "<code>token</code>",
-  "<code>displaystring</code>", and "<code>bytesequence</code>".
+  or as quoted structured field strings otherwise. To override the
+  auto-detected structured field type, set the <code>sfType</code> property
+  on the item object &mdash; for example,
+  <code>{ value: "foo", sfType: "string" }</code> forces a quoted string even
+  when the value matches token syntax. Supported <code>sfType</code> values
+  are "<code>boolean</code>", "<code>integer</code>",
+  "<code>decimal</code>", "<code>date</code>", "<code>string</code>",
+  "<code>token</code>", "<code>displaystring</code>", and
+  "<code>bytesequence</code>".
 
   <div class=example id=example-setstructured>
    <pre><code class=lang-javascript>
--- a/fetch.bs
+++ b/fetch.bs
@@ -8057,7 +8057,7 @@ if (purpose?.value === "prefetch") { /* ... */ }  // direct string comparison
   whether the number is an integer), booleans serialize as structured field
   booleans, {{Uint8Array}} serializes as a byte sequence, and strings
   serialize as tokens if they conform to <a>structured field token</a> syntax
-  or as quoted structured field strings otherwise. Item objects may include an
+  or as quoted structured field strings otherwise. Item objects can include an
   <code>sfType</code> property to override the auto-detected structured field
   type &mdash; for example, <code>{ value: "foo", sfType: "string" }</code>
   forces a quoted string even when the value matches token syntax. Supported
--- a/fetch.bs
+++ b/fetch.bs
@@ -7991,16 +7991,18 @@ new Headers(meta2);
  <dd><p>Replaces the value of the first header whose name is <var>name</var> with <var>value</var>
  and removes any remaining headers whose name is <var>name</var>.
 
- <dt><code><var>result</var> = <var>headers</var> . <a method for=Headers lt=getStructured()>getStructured</a>(<var>name</var>, <var>type</var>)</code>
+ <dt><code><var>headers</var> . <a method for=Headers lt=getStructured()>getStructured</a>(<var>name</var>, <var>type</var>)</code>
  <dd>
   <p>Parses the value of the header named <var>name</var> as a structured field
   ([[RFC9651]]) of the given <var>type</var> and returns the result as a
   JavaScript value.
 
-  <p>Returns null if the header is not present or its value fails to parse as
-  the specified structured field type.
+  <p>Returns null if the header is not present, or if its value does not
+  strictly conform to the specified structured field type as defined in
+  [[RFC9651]]. No partial or lenient parsing is performed; any
+  deviation from the grammar causes an immediate null return.
 
-  <p>The <var>type</var> must be one of "<code>item</code>",
+  <p>The <var>type</var> argument is one of "<code>item</code>",
   "<code>list</code>", or "<code>dictionary</code>".
 
   <p>Structured field values are converted to JavaScript as follows:
@@ -8015,9 +8017,13 @@ new Headers(meta2);
   <p class=note>Tokens are represented as plain strings, following the same
   pattern as {{DOMString}}, {{USVString}}, and {{ByteString}} &mdash; they are a
   constrained profile of string rather than a separate type. When serializing
-  via {{Headers/setStructured()}}, strings that conform to
-  <a>structured field token</a> syntax are serialized as tokens; all other
-  strings are serialized as quoted structured field strings.
+  via {{Headers/setStructured()}}, strings are serialized as tokens or quoted
+  strings by default (auto-detected from syntax). To override the type, set
+  <code>sfType</code> on the item object (e.g.,
+  <code>{ value: "foo", sfType: "string" }</code> forces a quoted string).
+  The <code>sfType</code> property can also be used to produce
+  <a>structured field date</a> and <a>structured field display string</a>
+  values that cannot be expressed through auto-detection.
 
   <div class=example id=example-getstructured>
    <pre><code class=lang-javascript>
@@ -8047,11 +8053,18 @@ if (purpose?.value === "prefetch") { /* ... */ }  // direct string comparison
   <p>Throws a {{TypeError}} if <var>value</var> cannot be serialized as the
   specified structured field type or if the headers are immutable.
 
-  <p>For serialization, numbers serialize as integers or decimals (determined
-  by whether the number is an integer), booleans serialize as structured field
+  <p>By default, numbers serialize as integers or decimals (determined by
+  whether the number is an integer), booleans serialize as structured field
   booleans, {{Uint8Array}} serializes as a byte sequence, and strings
   serialize as tokens if they conform to <a>structured field token</a> syntax
-  or as quoted structured field strings otherwise.
+  or as quoted structured field strings otherwise. Item objects may include an
+  <code>sfType</code> property to override the auto-detected structured field
+  type &mdash; for example, <code>{ value: "foo", sfType: "string" }</code>
+  forces a quoted string even when the value matches token syntax. Supported
+  <code>sfType</code> values are "<code>boolean</code>",
+  "<code>integer</code>", "<code>decimal</code>", "<code>date</code>",
+  "<code>string</code>", "<code>token</code>",
+  "<code>displaystring</code>", and "<code>bytesequence</code>".
 
   <div class=example id=example-setstructured>
    <pre><code class=lang-javascript>
@@ -8075,6 +8088,21 @@ headers.setStructured("Priority", "dictionary", new Map([
   ["u", { value: 0 }],
   ["i", { value: true }]
 ]));
+
+// sfType overrides auto-detection of the structured field type:
+// Force a quoted string even though "foo" matches token syntax
+headers.setStructured("Example", "item", { value: "foo", sfType: "string" });
+// Result: Example: "foo"
+
+// Without sfType, "foo" would serialize as an unquoted token:
+headers.setStructured("Example", "item", { value: "foo" });
+// Result: Example: foo
+
+// Produce a Date value (not possible without sfType)
+headers.setStructured("Deprecation", "item", {
+  value: 1659578233, sfType: "date"
+});
+// Result: Deprecation: @1659578233
 </code></pre>
   </div>
 
@@ -8288,7 +8316,7 @@ given a structured field item <var>sfItem</var>:
 </ol>
 
 <p class=note>The returned object has the shape
-<code>{ value: <var>bareItem</var>, params: Map }</code>.
+<code>{ value, params: Map }</code>.
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


panva:
@jasnell are you building up towards [RFC 9421: HTTP Message Signatures](https://www.rfc-editor.org/info/rfc9421/) support? I'd be happy to help.

jasnell:
@panva ... That's definitely one of the items on the agenda, yes.

mnot:
Would it be useful to have a spec of a canonical mapping of SF to JSON?

reschke:
... like the one used in the test suite?

jasnell:
> Would it be useful to have a spec of a canonical mapping of SF to JSON?

I would think so, yes

reschke:
Any reason why this needs to be baked in into `headers`? Woun't something based on strings be simpler?

jasnell:
"Something based on strings" is just what Headers already provides. Using an additional parser/serializer is obviously possible but adds an additional dependency which is what this is aiming to eliminate.

fetch.bs
```diff
@@ -7970,6 +7991,93 @@ new Headers(meta2);
...
+  lists are returned as objects with <code>items</code> and <code>params</code>
+  properties. Lists are returned as arrays. Dictionaries and parameters are
+  returned as {{Map}} objects.
+
+  <p class=note>Tokens are represented as plain strings, following the same
+  pattern as {{DOMString}}, {{USVString}}, and {{ByteString}} &mdash; they are a
+  constrained profile of string rather than a separate type. When serializing
+  via {{Headers/setStructured()}}, strings that conform to
+  <a>structured field token</a> syntax are serialized as tokens; all other
+  strings are serialized as quoted structured field strings.
```
martinthomson: This string->token implicit conversion is not going to work for some uses of the API.  Is it possible to override the type?  For instance, by adding a value as `{value: "v", sfType: "string"}` or something like that?  Otherwise, there are cases where a SF-defined value will expect a string and get a token that it will choke one.

martinthomson:
What is the thinking about this sort of thing?

```js
const cc = response.headers.getStructured("Cache-Control", "dictionary");
```

Presumably, this just runs the string through the identified SF parser, which might work out fine (or not), caveat emptor and all that jazz.

fetch.bs
```diff
@@ -8075,6 +8183,587 @@ from a {{Headers}} object (<var>headers</var>), run these steps:
...
+  <ol>
+   <li><p>If <var>jsValue</var> is NaN, +&infin;, or &minus;&infin;,
+   then <a>throw</a> a {{TypeError}}.
+
+   <li><p>If <var>jsValue</var> is an
+   <a href="https://tc39.es/ecma262/#sec-isinteger">integral number</a>,
+   return a <a>structured field integer</a> with <var>jsValue</var>'s value.
+
+   <li><p>Otherwise, return a <a>structured field decimal</a> with
+   <var>jsValue</var>'s value.
```
martinthomson: What do you do with values outside of the range that SF requires support for?  Would 6e110 be presented with all 110 digits?

fetch.bs
```diff
@@ -8174,6 +8863,77 @@ method steps are:
...
+</ol>
+</div>
+
+<p class=note>Serializing structured fields is entirely optional.
+Implementations that do not support structured field serialization are
+fully compliant with this specification. The minimum conformance
+requirement is that {{Headers/setStructured()}} does not throw for valid
+inputs &mdash; an implementation that silently ignores the call is compliant.
+Whether to actually serialize and set the header is an implementation
+decision.
```
martinthomson: This seems like it could be an interoperability nightmare.  I think that you need to support SF serialization if you add this.

jasnell: I've been going back and forth on this, to be honest. But, you're right.

fetch.bs
jasnell: No, in that case the correct response would be to return null. The fallback would be to just use the existing get method.

jasnell:
> Presumably, this just runs the string through the identified SF parser, which might work out fine (or not), caveat emptor and all that jazz

Yes. If it can be parsed as the specified type, then it will be. Otherwise null is returned.

fetch.bs
martinthomson: Is that because 6e110 doesn't fit in a double?  I'm thinking of those cases where the range of permissible values in a JS Number is outside the range of permissible SF integers.  What then?  Translate, and let the application enjoy the (possible) surprise (possible because there is nothing preventing an implementation from accepting more than the minimal range).

You need this to be explicit in the algorithm.

reschke:
> "Something based on strings" is just what Headers already provides. Using an additional parser/serializer is obviously possible but adds an additional dependency which is what this is aiming to eliminate.

Yes. What I was trying to say is that the API does not necessarily have to be attached to the `headers` object. It could be independant, but yes, it should be in FETCH.

fetch.bs
jasnell: Ok, updated....

```js
// Without sfType — auto-detects as token (unquoted)
headers.setStructured("Example", "item", { value: "foo" });
// → Example: foo

// With sfType — forces quoted string
headers.setStructured("Example", "item", { value: "foo", sfType: "string" });
// → Example: "foo"

// Token validation — throws if value doesn't match token syntax
headers.setStructured("Example", "item", { value: "hello world", sfType: "token" });
// → TypeError (spaces not allowed in tokens)

// Deprecation header (RFC 9745) — defined as an SF Date
headers.setStructured("Deprecation", "item", {
  value: 1659578233,
  sfType: "date"
});
// → Deprecation: @1659578233

// Without sfType, the same number becomes an integer:
headers.setStructured("Deprecation", "item", { value: 1659578233 });
// → Deprecation: 1659578233   (wrong — missing @ prefix)

headers.setStructured("Example", "item", {
  value: "Ünïcödé",
  sfType: "displaystring"
});
// → Example: %"...percent-encoded UTF-8..."

// Without sfType, this would try to serialize as a regular SF string,
// which only permits printable ASCII — serialization would fail.

// Force decimal even for an integer value
headers.setStructured("Example", "item", { value: 42, sfType: "decimal" });
// → Example: 42.0

// Force integer — throws if not integral
headers.setStructured("Example", "item", { value: 3.14, sfType: "integer" });
// → TypeError

// Dictionary with mixed sfType overrides
headers.setStructured("My-Header", "dictionary", {
  name: { value: "cdn", sfType: "string" },   // quoted: "cdn"
  kind: { value: "cdn" },                      // auto: unquoted token cdn
  ts:   { value: 1700000000, sfType: "date" }  // date: @1700000000
});
// → My-Header: name="cdn", kind=cdn, ts=@1700000000

// List with sfType
headers.setStructured("My-List", "list", [
  { value: "keep-alive", sfType: "string" },  // "keep-alive"
  { value: "upgrade", sfType: "token" }        // upgrade
]);
// → My-List: "keep-alive", upgrade

// Integer out of range — throws
headers.setStructured("X", "item", { value: 1e16 });
// → TypeError (exceeds 15-digit SF integer limit)

// Decimal with too-large integer component — throws
headers.setStructured("X", "item", { value: 9999999999999.5 });
// → TypeError (integer component exceeds 12 digits)

// sfType: "date" with non-integer — throws
headers.setStructured("X", "item", { value: 3.14, sfType: "date" });
// → TypeError (dates must be integral)

// sfType only applies to item objects, not parameter values.
// Parameters always use auto-detection:
headers.setStructured("Cache-Status", "list", [
  { value: "cdn", sfType: "token", params: { hit: true, ttl: 3600 } }
  //                                         ^^^ bare values, no sfType
]);
// → Cache-Status: cdn;hit;ttl=3600
```

fetch.bs
jasnell: Updated to be more explicit

fetch.bs
jasnell: Updated to make it required.

jasnell:
> Yes. What I was trying to say is that the API does not necessarily have to be attached to the headers object. It could be independant, but yes, it should be in FETCH.

Hmm... you're not wrong but I think having it on `Headers` is the most ergonomic for the typical use cases.

PR #1955, opened 2026-08-30
Make a request's reserved client always be an environment

shannonbooth:
For: https://github.com/whatwg/html/pull/12837

### DIFF
--- a/fetch.bs
+++ b/fetch.bs
@@ -1710,12 +1710,10 @@ this field to be set to null due to certain redirects.
 
 <p>A <a for=/>request</a> has an associated
 <dfn id=concept-request-reserved-client export for=request>reserved client</dfn>
-(null, an <a for=/>environment</a>, or an
-<a>environment settings object</a>). Unless stated otherwise it is null.
+(null or an <a for=/>environment</a>). Unless stated otherwise it is null.
 
 <p class=note>This is only used by <a>navigation requests</a> and worker requests, but not service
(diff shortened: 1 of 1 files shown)


annevk (approved):

annevk:
Just to make sure I understand correctly, we need this because while the top-level fetch for the worker will have an environment settings object (from the creator), the subsequent fetches need to use the worker's environment, which isn't yet an environment settings object as we need the full module graph?

shannonbooth:
> Just to make sure I understand correctly, we need this because while the top-level fetch for the worker will have an environment settings object (from the creator), the subsequent fetches need to use the worker's environment, which isn't yet an environment settings object as we need the full module graph?

Not quite, descendant fetches also use the creators ESO as the fetch client. Reserved client is here is for building the Worker ESO. And worker ESO is needed for the module map/parse realm.

I think Domenic's comment here: https://github.com/whatwg/html/issues/11340#issuecomment-2921215917 explains the intent of these changes much better than I could, it's pretty much exactly the direction I followed. Using environment instead of ESO here breaks the circular dependency of an ESO being needed for both the input of a worker fetch and the output of a worker fetch.

So this is really about fixing the top level fetch, the module script case is handled after the top level request is available, at which point the ESO can be built, and then wire that ESO up through to "fetch a worklet/module worker script graph"  (but with a customisation that doesn't need to actually fetch, conceptually kind of like service workers do, since the top level fetch already has bytes). I think this part ties into the question here: https://github.com/whatwg/html/pull/12837#discussion_r3888690910

